PR #38, opened 2025-05-05
FEAT: Entra ID support with `mssql-auth` inside package

bewithgaurav:


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -246,6 +246,15 @@ void LoadDriverOrThrowException() {
         ThrowStdException("Failed to load driver");
     }
 
+    // Preload mssql-auth.dll from the same path if available
+    std::wstring authDllDir = std::wstring(ddbcModulePath) + L"\\libs\\win\\mssql-auth.dll";
+    HMODULE hAuthModule = LoadLibraryW(authDllDir.c_str());
+    if (hAuthModule) {
+        LOG("Authentication library loaded successfully from - {}", authDllDir.c_str());
```
Ramudaykumar, line 254: This apporach lookes good to me, even we are not needed to make any change to parent msodbcsql library.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -246,6 +246,15 @@ void LoadDriverOrThrowException() {
...
+    } else {
+        LOG("Note: Authentication library not found at - {}. This is OK if not using Windows Authentication.", authDllDir.c_str());
```
Ramudaykumar: Here we should put msg more accurate, "This Ok if you are not doing Entra Id Auth." As SQL Auth will also work withoiut MSSQL-AUth.

Ramudaykumar (approved):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -246,6 +246,15 @@ void LoadDriverOrThrowException() {
         ThrowStdException("Failed to load driver");
     }
 
+    // Preload mssql-auth.dll from the same path if available
+    std::wstring authDllDir = std::wstring(ddbcModulePath) + L"\\libs\\win\\mssql-auth.dll";
+    HMODULE hAuthModule = LoadLibraryW(authDllDir.c_str());
```
Ramudaykumar, line 252: If we can make it conditional load, only when Entra ID Auth is required. It will be even better.

Ramudaykumar (approved):

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: sure, added this as an improvement point as TODO inside comment since we would need to test it further extensively

codewatchzen (approved):

PR #45, opened 2025-05-13
FEAT: Access Token Login

jahnvi480:


mssql_python/connection.py
```diff
@@ -76,23 +77,34 @@ def _construct_connection_string(self, connection_str: str, **kwargs) -> str:
...
-            elif key.lower() == "encrypt":
-                key = "Encrypt"
-            elif key.lower() == "trust_server_certificate":
-                key = "TrustServerCertificate"
-            else:
-                continue
-            conn_str += f"{key}={value};"
+
+        # Check if access token authentication is being used
+        if "attrs_before" in kwargs:
```
bewithgaurav: probably not needed since attrs_before is a registered param now

mssql_python/connection.py
```diff
@@ -115,9 +127,40 @@ def _initializer(self) -> None:
         self._allocate_environment_handle()
         self._set_environment_attributes()
         self._allocate_connection_handle()
-        self._set_connection_attributes()
+        if self._attrs_before != {}:
+            self._apply_attrs_before()  # Apply pre-connection attributes
```
bewithgaurav, line 126: lets just strictly handle only access token attr (1256) as of now

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -692,18 +692,58 @@ SQLRETURN SQLSetEnvAttr_wrap(SqlHandlePtr EnvHandle, SQLINTEGER Attribute, intpt
...
+        int intValue = ValuePtr.cast<int>();
+        value = reinterpret_cast<SQLPOINTER>(intValue);
+        length = SQL_IS_INTEGER;  // Integer values don't require a length
+    } else if (py::isinstance<py::str>(ValuePtr)) {
+        // Handle Unicode string values
+        static std::wstring unicodeValueBuffer;
+        unicodeValueBuffer = ValuePtr.cast<std::wstring>();
+        value = const_cast<SQLWCHAR*>(unicodeValueBuffer.c_str());
+        length = SQL_NTS;  // Indicates null-terminated string
+    } else if (py::isinstance<py::bytes>(ValuePtr) || py::isinstance<py::bytearray>(ValuePtr)) {
```
bewithgaurav, line 719: we should handle only cases for autocommit and access token attrs in this since others are unclear to us atm

mssql_python/connection.py
```diff
@@ -76,23 +77,34 @@ def _construct_connection_string(self, connection_str: str, **kwargs) -> str:
...
-            conn_str += f"{key}={value};"
+
+        # Check if access token authentication is being used
+        if "attrs_before" in kwargs:
+            # Skip adding Uid and Pwd for access token authentication
+            if ENABLE_LOGGING:
+                logger.info("Using access token authentication. Skipping Uid and Pwd.")
+        else:
+            # Add additional key-value pairs to the connection string
+            for key, value in kwargs.items():
```
bewithgaurav: pls add/modify tests in this - construction of conn strings should be consistent with/without access token
there's a test called `test_construct_connection_string` - can add attrs before as a param there possibly

bewithgaurav (changes requested):

mssql_python/db_connection.py
```diff
@@ -6,7 +6,7 @@
 from mssql_python.connection import Connection
 
 
-def connect(connection_str: str, autocommit: bool = True, **kwargs) -> Connection:
+def connect(connection_str: str, autocommit: bool = True, attrs_before: dict = {}, **kwargs) -> Connection:
```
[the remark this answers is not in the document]
bewithgaurav: This is actually a good practice, python does call by reference in params which causes a lot of inconsistencies sometimes

bewithgaurav (approved):

PR #48, opened 2025-05-15
FEAT: Support for Windows ARM64 and added x86 binaries

bewithgaurav:
Support for Windows ARM64 and added x86 for future support


sumitmsft (approved):
Added files to support ARM64 architecture.

PR #51, opened 2025-05-19
REFACTOR: modularize ddbc_bindings.cpp

gargsaumya:
ADO Task ID [AB#37534](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/37534)

This pull request introduces a new header file, `ddbc_bindings.h`, for managing SQL database interactions in the `mssql_python` module. The file provides function pointer typedefs for SQL APIs, extern declarations for these function pointers, and utility classes and functions for logging, exception handling, driver loading, and SQL handle management.

### Summary  
Refactored the native layer to clearly separate binding logic and ODBC utility functions. This decouples the tightly integrated components from ddbc_bindings.cpp. This has been done to improve modularity, reusability, and maintainability of the native C++ layer to support cross-platform development.
We have also introduced core utilities such as SqlHandle and  DriverLoader to promote reuse and reduce redundancy.

### Checklist  
- [x] **Tests Passed** (if applicable)  : All pytests passed
- [x] **Docs Updated** (if necessary)  : Not needed

### Testing Performed  
 The changes were verified against all the pytests on Windows. (current support)

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -9,22 +9,18 @@
 #include <cstdint>
 #include <iomanip>  // std::setw, std::setfill
 #include <iostream>
-#include <string>
 #include <utility>  // std::forward
 
 // Replace std::filesystem usage with Windows-specific headers
 #include <shlwapi.h>
 #pragma comment(lib, "shlwapi.lib")
 
+#include "ddbc_bindings.h"
 #include <pybind11/chrono.h>
 #include <pybind11/complex.h>
 #include <pybind11/functional.h>
 #include <pybind11/pytypes.h>  // Add this line for datetime support
 #include <pybind11/stl.h>
-#include <windows.h>  // windows.h needs to be included before sql.h
-#include <sql.h>
-#include <sqlext.h>
-
 
 namespace py = pybind11;
 using namespace pybind11::literals;
@@ -115,59 +111,6 @@ struct ErrorInfo {
     std::wstring ddbcErrorMsg;
 };
 
-
-//-------------------------------------------------------------------------------------------------
-// Function pointer typedefs
-//-------------------------------------------------------------------------------------------------
-
-// Handle APIs
-typedef SQLRETURN (*SQLAllocHandleFunc)(SQLSMALLINT, SQLHANDLE, SQLHANDLE*);
-typedef SQLRETURN (*SQLSetEnvAttrFunc)(SQLHANDLE, SQLINTEGER, SQLPOINTER, SQLINTEGER);
-typedef SQLRETURN (*SQLSetConnectAttrFunc)(SQLHDBC, SQLINTEGER, SQLPOINTER, SQLINTEGER);
-typedef SQLRETURN (*SQLSetStmtAttrFunc)(SQLHSTMT, SQLINTEGER, SQLPOINTER, SQLINTEGER);
-typedef SQLRETURN (*SQLGetConnectAttrFunc)(SQLHDBC, SQLINTEGER, SQLPOINTER, SQLINTEGER,
-                                           SQLINTEGER*);
-
-// Connection and Execution APIs
-typedef SQLRETURN (*SQLDriverConnectFunc)(SQLHANDLE, SQLHWND, SQLWCHAR*, SQLSMALLINT, SQLWCHAR*,
-                                          SQLSMALLINT, SQLSMALLINT*, SQLUSMALLINT);
-typedef SQLRETURN (*SQLExecDirectFunc)(SQLHANDLE, SQLWCHAR*, SQLINTEGER);
-typedef SQLRETURN (*SQLPrepareFunc)(SQLHANDLE, SQLWCHAR*, SQLINTEGER);
-typedef SQLRETURN (*SQLBindParameterFunc)(SQLHANDLE, SQLUSMALLINT, SQLSMALLINT, SQLSMALLINT,
-                                          SQLSMALLINT, SQLULEN, SQLSMALLINT, SQLPOINTER, SQLLEN,
-                                          SQLLEN*);
-typedef SQLRETURN (*SQLExecuteFunc)(SQLHANDLE);
-typedef SQLRETURN (*SQLRowCountFunc)(SQLHSTMT, SQLLEN*);
-typedef SQLRETURN (*SQLSetDescFieldFunc)(SQLHDESC, SQLSMALLINT, SQLSMALLINT, SQLPOINTER, SQLINTEGER);
-typedef SQLRETURN (*SQLGetStmtAttrFunc)(SQLHSTMT, SQLINTEGER, SQLPOINTER, SQLINTEGER, SQLINTEGER*);
-
-// Data retrieval APIs
-typedef SQLRETURN (*SQLFetchFunc)(SQLHANDLE);
-typedef SQLRETURN (*SQLFetchScrollFunc)(SQLHANDLE, SQLSMALLINT, SQLLEN);
-typedef SQLRETURN (*SQLGetDataFunc)(SQLHANDLE, SQLUSMALLINT, SQLSMALLINT, SQLPOINTER, SQLLEN,
-                                    SQLLEN*);
-typedef SQLRETURN (*SQLNumResultColsFunc)(SQLHSTMT, SQLSMALLINT*);
-typedef SQLRETURN (*SQLBindColFunc)(SQLHSTMT, SQLUSMALLINT, SQLSMALLINT, SQLPOINTER, SQLLEN,
-                                    SQLLEN*);
-typedef SQLRETURN (*SQLDescribeColFunc)(SQLHSTMT, SQLUSMALLINT, SQLWCHAR*, SQLSMALLINT,
-                                        SQLSMALLINT*, SQLSMALLINT*, SQLULEN*, SQLSMALLINT*,
-                                        SQLSMALLINT*);
-typedef SQLRETURN (*SQLMoreResultsFunc)(SQLHSTMT);
-typedef SQLRETURN (*SQLColAttributeFunc)(SQLHSTMT, SQLUSMALLINT, SQLUSMALLINT, SQLPOINTER,
-                                         SQLSMALLINT, SQLSMALLINT*, SQLPOINTER);
-
-// Transaction APIs
-typedef SQLRETURN (*SQLEndTranFunc)(SQLSMALLINT, SQLHANDLE, SQLSMALLINT);
-
-// Disconnect/free APIs
-typedef SQLRETURN (*SQLFreeHandleFunc)(SQLSMALLINT, SQLHANDLE);
-typedef SQLRETURN (*SQLDisconnectFunc)(SQLHDBC);
-typedef SQLRETURN (*SQLFreeStmtFunc)(SQLHSTMT, SQLUSMALLINT);
-
-// Diagnostic APIs
-typedef SQLRETURN (*SQLGetDiagRecFunc)(SQLSMALLINT, SQLHANDLE, SQLSMALLINT, SQLWCHAR*, SQLINTEGER*,
-                                       SQLWCHAR*, SQLSMALLINT, SQLSMALLINT*);
-
 //-------------------------------------------------------------------------------------------------
 // Function pointer initialization
 //-------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mssql_python/pybind/ddbc_bindings.h
```diff
@@ -0,0 +1,140 @@
...
+void LOG(const std::string& formatString, Args&&... args);
+
+// -- Exception helper --
+void ThrowStdException(const std::string& message);
+
+// -- Driver loader --
+std::wstring LoadDriverOrThrowException();
+
+// -- Singleton wrapper --
+class DriverLoader {
```
sumitmsft, line 130: While the code is generally well-structured, comments explaining why the DriverLoader is a singleton or why SqlHandle's destructor is a no-op (which is already well-commented regarding Python's GC) would be beneficial for future maintainers. The existing comment on SqlHandle destructor is excellent.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -702,11 +470,214 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
...
+    if (!success) {
+        LOG("Failed to load required function pointers from driver - {}", dllDirStr);
+        ThrowStdException("Failed to load required function pointers from driver");
+    }
+    LOG("Successfully loaded function pointers from driver");
+    
+    return dllDir;
+}
+
+// DriverLoader definition
```
sumitmsft, line 623: While the code is generally well-structured, comments explaining why the DriverLoader is a singleton or why SqlHandle's destructor is a no-op (which is already well-commented regarding Python's GC) would be beneficial for future maintainers. The existing comment on SqlHandle destructor is excellent (Lines 643-647).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -702,11 +470,214 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
...
+    std::string module_file = module_path.cast<std::string>();
+    
+    char path[MAX_PATH];
+    strncpy_s(path, MAX_PATH, module_file.c_str(), module_file.length());
+    PathRemoveFileSpecA(path);
+    return std::string(path);
+}
+
+// Helper to load the driver
+// TODO: We don't need to do explicit linking using LoadLibrary. We can just use implicit
```
sumitmsft, line 502: The TODO for using implicit linking (// TODO: We don't need to do explicit linking using LoadLibrary. We can just use implicit linking to load this DLL. It will simplify the code a lot.) is an important one. If this is planned for a follow-up, it's fine. Otherwise, it might be worth addressing as part of this refactor if the scope allows, as it would simplify the driver loading logic significantly.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -702,11 +470,214 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
...
+SQLHANDLE SqlHandle::get() const {
+    return _handle;
+}
+
+SQLSMALLINT SqlHandle::type() const {
+    return _type;
+}
+
+void SqlHandle::free() {
+    if (_handle && SQLFreeHandle_ptr) {
```
sumitmsft, line 657: Are there any potential issues if SQLFreeHandle_ptr is null when SqlHandle::free() is called? The code checks if (_handle && SQLFreeHandle_ptr), which is good.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
```
sumitmsft, line 3: The current implementation for loading the ODBC driver and mssql-auth.dll is Windows-specific. This is what I understand from the code, however, If this module is intended to be cross-platform, this is a significant area that needs to be addressed or clarified. Please clarify in your comments in the header of the file that this implementation is Windows-centric.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -9,22 +9,18 @@
 #include <cstdint>
 #include <iomanip>  // std::setw, std::setfill
 #include <iostream>
-#include <string>
 #include <utility>  // std::forward
 
 // Replace std::filesystem usage with Windows-specific headers
```
sumitmsft, line 14: The current implementation for loading the ODBC driver and mssql-auth.dll is Windows-specific. This is what I understand from the code, however, If this module is intended to be cross-platform, this is a significant area that needs to be addressed or clarified. Please clarify in your comments in the header of the file that this implementation is Windows-centric.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -702,11 +470,214 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
 
 }  // namespace
 
+// TODO: Revisit GIL considerations if we're using python's logger
```
sumitmsft, line 473: This part of the code is C++, not Python. However, the interaction with Python (e.g., logging via Python's logging module, GIL considerations mentioned in comments) is relevant.

The comment // TODO: Revisit GIL considerations if we're using python's logger is important. If Python's logging is called from C++ without proper GIL management, it can lead to deadlocks or crashes. Ensure the GIL is acquired before calling Python logging functions and released afterward if the C++ code might be running without the GIL held (e.g., in a separate thread).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -702,11 +470,214 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
...
+        archDir = L"arm64";
+    } else {
+        archDir = L"x86";
+    }
+    dllDir += archDir;
+    std::wstring mssqlauthDllPath = dllDir + L"\\mssql-auth.dll";
+    dllDir += L"\\msodbcsql18.dll";
+
+    // Preload mssql-auth.dll from the same path if available
+    // TODO: Only load mssql-auth.dll if using Entra ID Authentication modes (Active Directory modes)
```
sumitmsft: The comment // TODO: Only load mssql-auth.dll if using Entra ID Authentication modes (Active Directory modes) is still present. If this logic can be implemented, it would prevent unnecessarily trying to load this DLL. Is there any specific reason why we want to load mssql-auth.dll only when auth mode is EntraId?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -740,7 +712,8 @@ SQLRETURN SQLSetConnectAttr_wrap(SqlHandlePtr ConnectionHandle, SQLINTEGER Attri
                                  py::object ValuePtr) {
     LOG("Set SQL Connection Attribute");
     if (!SQLSetConnectAttr_ptr) {
-        LoadDriverOrThrowException();
+        LOG("Function pointer not initialized. Loading the driver.");
+        DriverLoader::getInstance().loadDriver();  // Load the driver
     }
 
     // Print the type of ValuePtr and attribute value - helpful for debugging
```
sumitmsft, line 718: ## Putting a comment here as a placeholder, even though this comment is not related to the changes made in this PR
@gargsaumya @jahnvi480 

Ref: Line Number 766/739:
        // Handle byte or bytearray values (like access tokens)
        // Store in static buffer to ensure memory remains valid during connection
        static std::vector<std::string> bytesBuffers;
        bytesBuffers.push_back(ValuePtr.cast<std::string>());
        value = const_cast<char*>(bytesBuffers.back().c_str());


Issue: The bytesBuffers vector is declared static. This means it persists for the lifetime of the module. Each time SQLSetConnectAttr_wrap is called with a byte string (e.g., for an access token), a new string is added to this vector. The vector never shrinks, and these strings are never removed. This is a memory leak. The amount of leaked memory will grow with each relevant call to set a connection attribute.

This buffer should not be static, or its contents need a proper lifecycle management strategy (e.g., clear it when the connection is closed, or use non-static buffers if their lifetime can be shorter).

This may lead to unbounded memory growth over time if connection attributes are set repeatedly. Let's keep an eye on this and create a workitem to explore this problem further.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -702,11 +470,214 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
...
+}
+
+void DriverLoader::loadDriver() {
+    if (!m_driverLoaded) {
+        LoadDriverOrThrowException();
+        m_driverLoaded = true;
+    }
+}
+
+// SqlHandle definition
```
sumitmsft, line 638: ### Let's Explore this as well, please create a work item to track this issue and put relevant comments in the code

The C++ SqlHandle object itself is managed by std::shared_ptr, so its memory won't leak. However, the underlying ODBC handles (e.g., HENV, HDBC, HSTMT) stored in _handle are only released when the Python-exposed free() method is explicitly called. The C++ destructor SqlHandle::~SqlHandle() is a no-op by design (lines 643-647 in .cpp).

What I foresee here is that if the Python user of this library does not diligently call the free() method on these handle objects, the underlying ODBC resources will leak. This is a common pattern for managing native resources in Python extensions but relies heavily on correct usage from the Python side.

**My two cents and open for further discussions**
Clearly and prominently document that users of the library must call the free() method on SqlHandle objects to prevent ODBC resource leaks. Consider adding Python-side context managers (__enter__, __exit__) for these handle objects.

Can we do something to handle this more gracefully or are fully dependent on Python to call free()?

sumitmsft (changes requested):
The refactoring is a positive step towards better code organization and maintainability. The introduction of the DriverLoader singleton and the move of declarations to a header file are good improvements. The main areas to focus on for further improvement and to ensure robustness are cross-platform compatibility (if applicable), GIL handling for Python callbacks, and thorough testing. 

I have left a few comments, let's discuss and resolve them.

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: This was a suggestion which we had discussed in our team call, but this is a bit redundant since conditional loading of mssql-auth would be a micro task for the driver and if not loaded in some scenarios could cause issues. I would suggest that we remove this comment altogether.

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: yes, its there as a failsafe mechanism in case the driver APIs are not properly loaded. if not, then calling freehandle would result in violations/issues.

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: I agree, this approach can be discussed:
@gargsaumya - lets get some more clarity on this approach since this will make things very neat.. If this is a high effort task we can put this in for later iteration. lets create a work item anyways.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2156,7 +2144,7 @@ PYBIND11_MODULE(ddbc_bindings, m) {
     
     try {
         // Try loading the ODBC driver when the module is imported
-        LoadDriverOrThrowException();
+        DriverLoader::getInstance().loadDriver();  // Load the driver
```
bewithgaurav, line 2147: Please add a log message here to indicate trying to load Driver

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: Agreed, below is a brief description on why this was done:
This was a workaround approach when we were dealing with having driver logic both in Python and cpp. 
The free() method is called by mssql-python logic since python driver logic was made responsible for memory lifecycle, and the method was not exposed to users - it was strictly a part of driver code.
This will definitely get taken care of once we port our connection and cursor classes since we will use inbuilt destructors for the memory management once C++ takes care of it, and free() will be deprecated.

bewithgaurav (changes requested):
Added some context and minor suggestions in places wherever needed.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: The note is present in Line-4.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: We can take up a separate workitem  (Task #37537) to revisit GIL considerations if we're using python's logger. We can also discuss to have a logging utility in C++ itself and proceed further.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Created a workitem in ADO for this. (#37538)

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: This will be deprecated in upcoming PRs, as the Connection management will happen in C++.

sumitmsft (approved):
All looks good. PR ready for merge.

bewithgaurav (approved):

PR #52, opened 2025-05-19
FEAT: adding connection class structure in C++

gargsaumya:
ADO Task ID [[AB#34962](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34962)]
This pull request introduces a new `Connection` class for managing MSSQL database connections in the `mssql_python` module. The class provides methods for connecting, disconnecting, handling transactions, and managing autocommit settings. Below are the key changes:

### Implementation of the `Connection` class:

* **Header file (`connection.h`)**:
  - Added the `Connection` class definition, including methods for connecting (`connect`), disconnecting (`close`), transaction management (`commit`, `rollback`, `end_transaction`), and autocommit settings (`set_autocommit`, `get_autocommit`).
  - Introduced member variables such as `_conn_str` for the connection string, `_env_handle` and `_dbc_handle` for SQL handles, and `_autocommit` to track autocommit mode.

* **Source file (`connection.cpp`)**:
  - Added the `Connection` class constructor, destructor, and method stubs with logging statements for key operations like connecting, disconnecting, committing, rolling back, and managing autocommit.

### Checklist  
- [x] **Tests Passed** (if applicable)  : All pytests passed successfully.
- [ ] **Docs Updated** (if necessary) Will be added in upcoming PRs once the Connection management is fully integrated in c++.

### DIFF
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -1,17 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-// Implements the Connection class declared in connection.h.
-// This class wraps low-level ODBC operations like connect/disconnect,
-// transaction control, and autocommit configuration.
-
 #include "connection.h"
 #include <iostream>
 
+//-------------------------------------------------------------------------------------------------
+// Implements the Connection class declared in connection.h.
+// This class wraps low-level ODBC operations like connect/disconnect,
+// transaction control, and autocommit configuration.
+//-------------------------------------------------------------------------------------------------
 Connection::Connection(const std::wstring& conn_str) : _conn_str(conn_str) {}
 
 Connection::~Connection() {
-    close();
+    LOG("Connection destructor called");
+    close();    // Ensure the connection is closed when the object is destroyed.
 }
 
 SQLRETURN Connection::connect() {
@@ -39,11 +41,11 @@ SQLRETURN Connection::end_transaction(SQLSMALLINT completion_type) {
 }
 
 SQLRETURN Connection::set_autocommit(bool enable) {
-    LOG("Setting autocommit mode C++");
+    LOG("Setting autocommit mode");
     // to be added
 }
 
 bool Connection::get_autocommit() const {
-    LOG("Getting autocommit mode C++");
+    LOG("Getting autocommit mode");
     // to be added
 }
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+// INFO|TODO - Note that is file is Windows specific right now. Making it arch agnostic will be
+//             taken up in future
+
 #include "connection.h"
 #include <iostream>
 
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#include "connection.h"
+#include <iostream>
+
+Connection::Connection(const std::wstring& conn_str) : _conn_str(conn_str) {}
+
+Connection::~Connection() {
+    close();
+}
+
+SQLRETURN Connection::connect() {
+    LOG("Connecting to MSSQL");
+    // to be added 
+}
+
+SQLRETURN Connection::close() {
+    LOG("Disconnect from MSSQL");
+    // to be added
+}
+
+SQLRETURN Connection::commit() {
+    LOG("Committing transaction");
+    // to be added
+}
+
+SQLRETURN Connection::rollback() {
+    LOG("Rolling back transaction");
+    // to be added
+}
+
+SQLRETURN Connection::end_transaction(SQLSMALLINT completion_type) {
+    // to be added
+}
+
+SQLRETURN Connection::set_autocommit(bool enable) {
+    LOG("Setting autocommit mode C++");
+    // to be added
+}
+
+bool Connection::get_autocommit() const {
+    LOG("Getting autocommit mode C++");
+    // to be added
+}
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -1,6 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+// Implements the Connection class declared in connection.h.
+// This class wraps low-level ODBC operations like connect/disconnect,
+// transaction control, and autocommit configuration.
+
 #include "connection.h"
 #include <iostream>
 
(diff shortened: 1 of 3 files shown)


mssql_python/pybind/connection/connection.cpp
```diff
@@ -0,0 +1,54 @@
...
+    LOG("Committing transaction");
+    // to be added
+}
+
+SQLRETURN Connection::rollback() {
+    LOG("Rolling back transaction");
+    // to be added
+}
+
+SQLRETURN Connection::end_transaction(SQLSMALLINT completion_type) {
```
sumitmsft, line 42: Although this is just a structure of the connection class, I would like to just make sure that we evaluate the following for future discussion:
For me, the end_transaction() method overlaps with the functionality of commit() and rollback(). Typically, these two methods are sufficient for transaction management in most database systems. Please comment is the end_transaction() has a different purpose.

mssql_python/pybind/connection/connection.h
```diff
@@ -0,0 +1,51 @@
...
+    bool get_autocommit() const;
+
+private:
+
+    std::wstring _conn_str;     // Connection string
+    SqlHandlePtr _env_handle;   // Environment handle
+    SqlHandlePtr _dbc_handle;   // Connection handle
+
+    bool _autocommit = false;
+    std::shared_ptr<Connection> _conn;
```
sumitmsft, line 49: The _conn member is a std::shared_ptr<Connection> allocated within the same Connection class. Just make sure you validate and verify that this doesn't introduce a risk of circular references and memory leaks if _conn is ever assigned or initialized with a std::shared_ptr pointing to the current object (this). Let's just be mindful of it.

sumitmsft (approved):
This PR is just structure of connection class, more comprehensive review will be required to be done once we have the implementation.

mssql_python/pybind/connection/connection.cpp
gargsaumya: The end_transaction() method is only for commit() and rollback() and has no other purpose. The implementation for both will be handled in next PR.

mssql_python/pybind/connection/connection.h
gargsaumya: This will be needed in pooling. It is not being used in connection class anywhere until pooling is introduced, will remove it for now.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -0,0 +1,54 @@
...
+    // to be added
+}
+
+SQLRETURN Connection::set_autocommit(bool enable) {
+    LOG("Setting autocommit mode");
+    // to be added
+}
+
+bool Connection::get_autocommit() const {
```
bewithgaurav, line 51: The DB API specs in python has autocommit as an attribute which we set - [ref](https://peps.python.org/pep-0249/#autocommit), also in future setautocommit() will be used for setting the attribute - [ref](https://peps.python.org/pep-0249/#footnotes) (last point of footnotes)
wanted to check on which cases will we be using this

mssql_python/pybind/connection/connection.cpp
```diff
@@ -0,0 +1,54 @@
...
+SQLRETURN Connection::rollback() {
+    LOG("Rolling back transaction");
+    // to be added
+}
+
+SQLRETURN Connection::end_transaction(SQLSMALLINT completion_type) {
+    // to be added
+}
+
+SQLRETURN Connection::set_autocommit(bool enable) {
```
bewithgaurav, line 46: setautocommit() instead of set_autocommit() as per DB API specs 2.0

mssql_python/pybind/connection/connection.h
```diff
@@ -0,0 +1,51 @@
...
+    SQLRETURN commit();
+
+    // Rollback the current transaction.
+    SQLRETURN rollback();
+
+    // End the transaction with the specified completion type.
+    SQLRETURN end_transaction(SQLSMALLINT completion_type);
+
+    // Enable or disable autocommit mode.
+    SQLRETURN set_autocommit(bool value);
```
bewithgaurav, line 37: same

mssql_python/pybind/connection/connection.cpp
bewithgaurav: Adding a comment in same lines as sumit just to understand, will end_transaction be an end user facing function or an internal method?

bewithgaurav (changes requested):

mssql_python/pybind/connection/connection.cpp
```diff
@@ -0,0 +1,54 @@
...
+// transaction control, and autocommit configuration.
+//-------------------------------------------------------------------------------------------------
+Connection::Connection(const std::wstring& conn_str) : _conn_str(conn_str) {}
+
+Connection::~Connection() {
+    LOG("Connection destructor called");
+    close();    // Ensure the connection is closed when the object is destroyed.
+}
+
+SQLRETURN Connection::connect() {
```
bewithgaurav, line 22: the `connect()` method will actually be a part of global module (can check `db_connection.py`) which is responsible for creating a new Connection class object. 
I'm assuming this might be needed to call DDBCSQLConnect - can we rename this to something else which will be internal and not public? you can check `connection.py` for reference where we are using a custom private method called `_initializer()` which calls `_connect_to_db()` - if that feels clear we can try taking the same approach as well, i.e.
- `_initializer()` code can be a part of constructor and will contain allocation of handles and calling SQLConnect via `_connect_to_db()` 
Suggesting this change since the function names are same and will contradict the usage

mssql_python/pybind/connection/connection.cpp
gargsaumya: The Connection class in C++ is internal and so are the APIs. The APIs in the python class will remain DB API compliant as those will be user facing.

mssql_python/pybind/connection/connection.cpp
gargsaumya: This is mapped with current function in python class that returns the current autocommit mode of the connection.
Can you explain what's concerning?

mssql_python/pybind/connection/connection.h
gargsaumya: Addressed above.

mssql_python/pybind/connection/connection.cpp
gargsaumya: This is an internal API which will not be used by the user. This is going to be the DDBCSQLConnect call, I'm not sure how this will contradict the usage.

mssql_python/pybind/connection/connection.cpp
gargsaumya: These are all internal functions.

mssql_python/pybind/connection/connection.cpp
bewithgaurav: I think we don't need a python class if the connection class is completely transferred in C++, I think I might have a different idea on the overall approach, lets discuss on call today

bewithgaurav (approved):
As discussed, we have finalized on the below approach:
- Connection is encapsulated in two classes:
- ddbc_bindings will have a _Connection class and Python will be using that class and wrap it into its own connection class
- this is being done since we want to implement connection pooling in cpp rn, eventually we can think of doing a complete migration, and eliminate python layer altogether maybe
- also, this is being done keeping in mind Cursor will also follow the same changes
Approving this under the above context

PR #53, opened 2025-05-20
FEAT: Adding options for BCP with comparison bcp.exe

jahnvi480:
#### For feature requests
FEAT: This pull request introduces a new data model for managing BCP (Bulk Copy Program) options in the `mssql_python` package. The changes add two `@dataclass` structures, `ColumnFormat` and `BCPOptions`, to encapsulate the configuration details for BCP operations in a structured and type-safe manner.

### Summary  
* `mssql_python/bcp_options.py`:  
  - Added the `ColumnFormat` dataclass to define column-specific configurations such as `prefix_len`, `data_len`, and terminators (`field_terminator` and `row_terminator`).  
  - Added the `BCPOptions` dataclass to represent BCP operation settings, including options for direction (`in` or `out`), file paths (`data_file`, `error_file`, etc.), batch size, error handling, and column configurations.

### DIFF
--- a/mssql_python/bcp_options.py
+++ b/mssql_python/bcp_options.py
@@ -1,0 +1,29 @@
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+@dataclass
+class ColumnFormat:
+    prefix_len: int # Option: (format_file) or (prefix_len, data_len)
+    data_len: int # Option: (format_file) or (prefix_len, data_len)
+    field_terminator: Optional[str] = None # Option: (-t)
+    row_terminator: Optional[str] = None # Option: (-r)
+    server_col: int = 1 # Option: (format_file) or (server_col)
+    file_col: int = 1 # Option: (format_file) or (file_col)
+
+@dataclass
+class BCPOptions:
+    direction: str  # 'in' or 'out' Option: (-i or -o)
+    data_file: str # Option: (positional argument)
+    error_file: Optional[str] = None # Option: (-e)
+    format_file: Optional[str] = None # Option: (-f)
+    write_format_file: Optional[str] = None # Option: (-x)
+    batch_size: Optional[int] = None # Option: (-b)
+    max_errors: Optional[int] = None # Option: (-m)
+    first_row: Optional[int] = None # Option: (-F)
+    last_row: Optional[int] = None # Option: (-L)
+    code_page: Optional[str] = None # Option: (-C)
+    keep_identity: bool = False # Option: (-E)
+    keep_nulls: bool = False # Option: (-k)
+    hints: Optional[str] = None # Option: (-h)
+    bulk_mode: str = "native"  # native, char, unicode Option: (-n, -c, -w)
+    columns: Optional[List[ColumnFormat]] = field(default_factory=list) # Option: (format_file) or (columns)


mssql_python/bcp_options.py
```diff
@@ -0,0 +1,58 @@
...
+        first_row (Optional[int]): The first row to process. Option: (-F).
+        last_row (Optional[int]): The last row to process. Option: (-L).
+        code_page (Optional[str]): The code page. Option: (-C).
+        keep_identity (bool): Keep identity values. Option: (-E).
+        keep_nulls (bool): Keep null values. Option: (-k).
+        hints (Optional[str]): Additional hints. Option: (-h).
+        bulk_mode (str): Bulk mode ('native', 'char', 'unicode'). Option: (-n, -c, -w).
+        columns (List[ColumnFormat]): Column formats. Option: (format_file) or (columns).
+    """
+    direction: Literal["in", "out"]
```
sumitmsft, line 74: Applicable on line# 44 and 57:

Type hints like Literal are primarily for static analysis (e.g., by type checkers like MyPy). They don't automatically enforce the allowed values at runtime. This means that if someone were to pass an invalid value for direction or bulk_mode (e.g., "invalid"), the code would not raise an error immediately. This could lead to unexpected behavior or bugs later in the program's execution.

The assert statements would check if the value of direction is in the list of allowed values (["in", "out"]) and if the value of bulk_mode is in the list of allowed values (["native", "char", "unicode"]). If the assertion fails (i.e., the value is invalid), an AssertionError is raised with a descriptive error message that includes the invalid value. This makes it easier to debug the problem.

Per my research.. __post_init__ can be used in dataclasses that's automatically called after the dataclass is initialized. It's a good place to perform validation. Please explore this.

mssql_python/bcp_options.py
```diff
@@ -0,0 +1,58 @@
...
+        field_terminator (Optional[str]): Option: (-t). The field terminator string.
+        row_terminator (Optional[str]): Option: (-r). The row terminator string.
+        server_col (int): Option: (format_file) or (server_col). The server column number.
+        file_col (int): Option: (format_file) or (file_col). The file column number.
+    """
+    prefix_len: int
+    data_len: int
+    field_terminator: Optional[str] = None
+    row_terminator: Optional[str] = None
+    server_col: int = 1
```
sumitmsft, line 30: **Applicable to Line# 20 and 21**

In the ColumnFormat dataclass, both server_col and file_col are initialized with a default value of 1. Are these default values always appropriate? Meaning, is there any other value that could be more appropriate than 1? 
server_col: Represents the column number in the SQL Server table.
file_col: Represents the column number in the data file.

- Can we add validation to ensure that server_col and file_col are within a reasonable range (e.g., greater than 0)?
- Add a clear documentation (docstrings or comments) explaining the purpose of server_col and file_col, and why the default value of 1. Also, make sure we add this in public documentation as well.

mssql_python/bcp_options.py
```diff
@@ -0,0 +1,58 @@
...
+    Represents the format of a column in a bulk copy operation.
+    Attributes:
+        prefix_len (int): Option: (format_file) or (prefix_len, data_len).
+        data_len (int): Option: (format_file) or (prefix_len, data_len).
+        field_terminator (Optional[str]): Option: (-t). The field terminator string.
+        row_terminator (Optional[str]): Option: (-r). The row terminator string.
+        server_col (int): Option: (format_file) or (server_col). The server column number.
+        file_col (int): Option: (format_file) or (file_col). The file column number.
+    """
+    prefix_len: int
```
sumitmsft, line 26: **Line# 16 and 17**

The ColumnFormat dataclass includes the attributes prefix_len and data_len, which represent lengths. Lengths, by definition, cannot be negative. However, the current code doesn't prevent a user from accidentally setting these values to negative numbers. 

If prefix_len or data_len are negative, it could lead to:

Unexpected Behavior: The BCP operation might behave in unpredictable ways, potentially corrupting data or causing errors.
Logic Errors: Calculations or operations that rely on these lengths might produce incorrect results.
Security Vulnerabilities: In some cases, negative lengths could potentially be exploited to cause buffer overflows or other security issues (although this is less likely in Python due to its memory management).

Add runtime validation to the ColumnFormat dataclass to ensure that prefix_len and data_len are non-negative (i.e., greater than or equal to 0).

mssql_python/bcp_options.py
```diff
@@ -0,0 +1,58 @@
...
+    write_format_file: Optional[str] = None
+    batch_size: Optional[int] = None
+    max_errors: Optional[int] = None
+    first_row: Optional[int] = None
+    last_row: Optional[int] = None
+    code_page: Optional[str] = None
+    keep_identity: bool = False
+    keep_nulls: bool = False
+    hints: Optional[str] = None
+    bulk_mode: Literal["native", "char", "unicode"] = "native"
```
sumitmsft, line 87: bulk_mode defaults to "native". Plese document why this default was chosen.

mssql_python/bcp_options.py
```diff
@@ -0,0 +1,58 @@
+from dataclasses import dataclass, field
+from typing import List, Optional, Literal
+
+@dataclass
+class ColumnFormat:
```
sumitmsft, line 6: **Lines 5-22 (ColumnFormat)** 

The current implementation of the ColumnFormat dataclasses lacks explicit validation of the input data. While type hints provide some level of static type checking, they don't guarantee that the values passed to the dataclasses are semantically valid or within acceptable ranges at runtime. This means that invalid data could potentially be passed to these dataclasses, leading to unexpected behavior, errors, or even security vulnerabilities.

While the attributes have type hints (e.g., int, Optional[str]), there's nothing preventing a user from passing a negative value for prefix_len or data_len, or an empty string for field_terminator if that's not allowed.

mssql_python/bcp_options.py
```diff
@@ -0,0 +1,58 @@
...
+        file_col (int): Option: (format_file) or (file_col). The file column number.
+    """
+    prefix_len: int
+    data_len: int
+    field_terminator: Optional[str] = None
+    row_terminator: Optional[str] = None
+    server_col: int = 1
+    file_col: int = 1
+
+@dataclass
```
sumitmsft, line 52: **Lines 23-58 (BCPOptions)** 

The current implementation of the BCPOptions dataclasses lacks explicit validation of the input data. While type hints provide some level of static type checking, they don't guarantee that the values passed to the dataclasses are semantically valid or within acceptable ranges at runtime. This means that invalid data could potentially be passed to these dataclasses, leading to unexpected behavior, errors, or even security vulnerabilities.

Similarly, a user could pass an invalid value for direction (something other than "in" or "out"), a negative value for batch_size or max_errors, or an empty string for data_file.

sumitmsft (approved):
Although this PR has just added the options for the BCP, I have added a few comments\best practices to be followed when you implement the logic. Make sure you adhere to all the validations I have suggested along with others wherever possible.

bewithgaurav (approved):

sumitmsft (approved):
All requested changes have been merged. PR approved.

bewithgaurav (approved):
Re-approved since the pending comments and changes are resolved now.

PR #54, opened 2025-05-20
FEAT: Implementation of C++ Connection Class API

gargsaumya:
ADO Task ID [[AB#37539](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/37539)]
This pull request introduces a new `Connection` class and supporting infrastructure for managing ODBC database connections in the `mssql_python` project. Key changes include the implementation of the `Connection` class, the addition of function pointer typedefs for ODBC APIs, and the creation of utility classes for driver management and handle lifecycle management.

### New `Connection` Class Implementation:
* [`mssql_python/pybind/connection/connection.cpp`](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR1-R110): Implements the `Connection` class, which provides methods for connecting to a database, managing transactions, and configuring autocommit. It uses RAII for resource management via `SqlHandle`.
* [`mssql_python/pybind/connection/connection.h`](diffhunk://#diff-c74277ea1f36fff8f6bbe9df3fd722f7d270f3d6a3069ac7a747fe6d4235cf13R1-R47): Declares the `Connection` class, including methods for connection management, transaction handling, and autocommit configuration.

### ODBC Function Pointer Typedefs and Utilities:
* [`mssql_python/pybind/ddbc_bindings.h`](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635R1-R158): Adds typedefs for ODBC API function pointers, logging utilities, and exception helpers. Introduces the `DriverLoader` singleton for centralized driver loading and the `SqlHandle` class for RAII-based handle management.

### Checklist  
- [x] **Tests Passed** (if applicable)  : All pytests are passing.
- [ ] **Docs Updated** (if necessary) : Will be done in upcoming PR once the feature is completely implemented.

### DIFF
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -7,131 +7,15 @@
 #include "connection.h"
 #include <iostream>
 
+SqlHandlePtr Connection::_envHandle = nullptr;
 //-------------------------------------------------------------------------------------------------
 // Implements the Connection class declared in connection.h.
 // This class wraps low-level ODBC operations like connect/disconnect,
 // transaction control, and autocommit configuration.
 //-------------------------------------------------------------------------------------------------
 Connection::Connection(const std::wstring& conn_str, bool autocommit)
-    : _conn_str(conn_str) , _autocommit(autocommit) {}
-
-Connection::~Connection() {
-    close();    // Ensure the connection is closed when the object is destroyed.
-}
-
-SQLRETURN Connection::connect() {
-    allocDbcHandle();
-    return connectToDb();
-}
-
-// Allocates DBC handle
-void Connection::allocDbcHandle() {
-    SQLHANDLE dbc = nullptr;
-    LOG("Allocate SQL Connection Handle");
-    SQLRETURN ret = SQLAllocHandle_ptr(SQL_HANDLE_DBC, getSharedEnvHandle()->get(), &dbc);
-    if (!SQL_SUCCEEDED(ret)) {
-        throw std::runtime_error("Failed to allocate connection handle");
-    }
-    _dbc_handle = std::make_shared<SqlHandle>(SQL_HANDLE_DBC, dbc);
-}
-
-// Connects to the database
-SQLRETURN Connection::connectToDb() {
-    LOG("Connecting to database");
-    SQLRETURN ret = SQLDriverConnect_ptr(_dbc_handle->get(), nullptr,
-                                         (SQLWCHAR*)_conn_str.c_str(), SQL_NTS,
-                                         nullptr, 0, nullptr, SQL_DRIVER_NOPROMPT);
-    if (!SQL_SUCCEEDED(ret)) {
-        throw std::runtime_error("Failed to connect to database");
-    }
-    LOG("Connected to database successfully");
-    return ret;
-}
-
-SQLRETURN Connection::close() {
-    if (!_dbc_handle) {
-        LOG("No connection handle to close");
-        return SQL_SUCCESS;
-    }
-    LOG("Disconnect from MSSQL");
-    if (!SQLDisconnect_ptr) {
-        LOG("Function pointer not initialized. Loading the driver.");
-        DriverLoader::getInstance().loadDriver();
-    }
-
-    SQLRETURN ret = SQLDisconnect_ptr(_dbc_handle->get());
-    _dbc_handle.reset();
-    return ret;
-}
-
-SQLRETURN Connection::commit() {
-    if (!_dbc_handle) {
-        throw std::runtime_error("Connection handle not allocated");
-    }
-    LOG("Committing transaction");
-    SQLRETURN ret = SQLEndTran_ptr(SQL_HANDLE_DBC, _dbc_handle->get(), SQL_COMMIT);
-    if (!SQL_SUCCEEDED(ret)) {
-        throw std::runtime_error("Failed to commit transaction");
-    }
-    return ret;
-}
-
-SQLRETURN Connection::rollback() {
-    if (!_dbc_handle) {
-        throw std::runtime_error("Connection handle not allocated");
-    }
-    LOG("Rolling back transaction");
-    SQLRETURN ret = SQLEndTran_ptr(SQL_HANDLE_DBC, _dbc_handle->get(), SQL_ROLLBACK);
-    if (!SQL_SUCCEEDED(ret)) {
-        throw std::runtime_error("Failed to rollback transaction");
-    }
-    return ret;
-}
-
-SQLRETURN Connection::setAutocommit(bool enable) {
-    if (!_dbc_handle) {
-        throw std::runtime_error("Connection handle not allocated");
-    }
-    SQLINTEGER value = enable ? SQL_AUTOCOMMIT_ON : SQL_AUTOCOMMIT_OFF;
-    LOG("Set SQL Connection Attribute");
-    SQLRETURN ret = SQLSetConnectAttr_ptr(_dbc_handle->get(), SQL_ATTR_AUTOCOMMIT, (SQLPOINTER)value, 0);
-    if (!SQL_SUCCEEDED(ret)) {
-        throw std::runtime_error("Failed to set autocommit mode.");
-    }
-    _autocommit = enable;
-    return ret;
-}
-
-bool Connection::getAutocommit() const {
-    if (!_dbc_handle) {
-        throw std::runtime_error("Connection handle not allocated");
-    }
-    LOG("Get SQL Connection Attribute");
-    SQLINTEGER value;
-    SQLINTEGER string_length;
-    SQLGetConnectAttr_ptr(_dbc_handle->get(), SQL_ATTR_AUTOCOMMIT, &value, sizeof(value), &string_length);
-    
-    return value == SQL_AUTOCOMMIT_ON;
-}
-
-SqlHandlePtr Connection::allocStatementHandle() {
-    if (!_dbc_handle) {
-        throw std::runtime_error("Connection handle not allocated");
-    }
-    LOG("Allocating statement handle");
-    SQLHANDLE stmt = nullptr;
-    SQLRETURN ret = SQLAllocHandle_ptr(SQL_HANDLE_STMT, _dbc_handle->get(), &stmt);
-    if (!SQL_SUCCEEDED(ret)) {
-        throw std::runtime_error("Failed to allocate statement handle");
-    }
-    return std::make_shared<SqlHandle>(SQL_HANDLE_STMT, stmt);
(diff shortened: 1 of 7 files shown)


mssql_python/pybind/connection/connection.cpp
```diff
@@ -12,39 +12,143 @@
 // This class wraps low-level ODBC operations like connect/disconnect,
 // transaction control, and autocommit configuration.
 //-------------------------------------------------------------------------------------------------
-Connection::Connection(const std::wstring& conn_str) : _conn_str(conn_str) {}
+Connection::Connection(const std::wstring& conn_str, bool autocommit)
+    : _conn_str(conn_str) , _autocommit(autocommit) {}
 
 Connection::~Connection() {
-    LOG("Connection destructor called");
     close();    // Ensure the connection is closed when the object is destroyed.
```
sumitmsft: Can you please validate if after close(), we should reset the handles (_env_handle and _dbc_handle) to nullptr to avoid dangling shared pointers?

mssql_python/pybind/connection/connection.cpp
```diff
@@ -12,39 +12,143 @@
...
 }
 
 SQLRETURN Connection::close() {
+    if (!_dbc_handle) {
+        LOG("No connection handle to close");
+        return SQL_SUCCESS;
+    }
     LOG("Disconnect from MSSQL");
-    // to be added
+    if (!SQLDisconnect_ptr) {
```
sumitmsft: If `DriverLoader::getInstance().loadDriver()` is called concurrently from multiple threads within the `call_once` or from `Connection::close(),` the `DriverLoader` itself needs to be thread-safe to prevent race conditions while loading the driver and initializing function pointers. This depends on the implementation of `DriverLoader`.
 
Because `DriverLoader::getInstance().loadDriver()` can be invoked from multiple code paths (lines 57-59 in `close()` and lines 137-139 in `getSharedEnvHandle`), and because `DriverLoader` is a singleton managing shared state (the function pointers), its `loadDriver()` method must be internally thread-safe. If it isn't, concurrent calls could lead to corrupted function pointers, inconsistent state, or crashes. 
The `std::call_once` in `getSharedEnvHandle` only protects the call within its lambda from other calls to that same lambda, not from calls to `loadDriver()` originating from other parts of the code like `Connection::close()`.

Let's validate and see if we are covered on this issue.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -12,39 +12,143 @@
...
+SqlHandlePtr Connection::getSharedEnvHandle() {
+    static std::once_flag flag;
+    static SqlHandlePtr env_handle;
+
+    std::call_once(flag, []() {
+        LOG("Allocating environment handle");
+        SQLHANDLE env = nullptr;
+        if (!SQLAllocHandle_ptr) {
+            LOG("Function pointers not initialized, loading driver");
+            DriverLoader::getInstance().loadDriver();
```
sumitmsft: **This comment is exactly similar to the one put in against line number 57-59**

If `DriverLoader::getInstance().loadDriver()` is called concurrently from multiple threads within the `call_once` or from `Connection::close(),` the `DriverLoader` itself needs to be thread-safe to prevent race conditions while loading the driver and initializing function pointers. This depends on the implementation of `DriverLoader`.
 
Because `DriverLoader::getInstance().loadDriver()` can be invoked from multiple code paths (lines 57-59 in `close()` and lines 137-139 in `getSharedEnvHandle`), and because `DriverLoader` is a singleton managing shared state (the function pointers), its `loadDriver()` method must be internally thread-safe. If it isn't, concurrent calls could lead to corrupted function pointers, inconsistent state, or crashes. 
The `std::call_once` in `getSharedEnvHandle` only protects the call within its lambda from other calls to that same lambda, not from calls to `loadDriver()` originating from other parts of the code like `Connection::close()`.

Let's validate and see if we are covered on this issue.

mssql_python/pybind/connection/connection.cpp
sumitmsft: **Review for entire code in lines 57-59**

This logic seems a bit unusual in `close()`. If `SQLDisconnect_ptr` is null, it means the driver wasn't loaded successfully initially (or was somehow unloaded). Attempting to load it again just to call `SQLDisconnect_ptr` might be problematic if the handle `_dbc_handle->get()` is invalid or belongs to a driver instance that is no longer active. 

Typically, if the function pointers aren't loaded, critical operations like connect would have failed earlier. If `_dbc_handle `exists, it implies a successful connection at some point, which should mean the driver was loaded. This might warrant a second look to ensure robustness under unusual driver states. Usually, you'd expect all necessary function pointers to be loaded when the Connection object is successfully established or when `getSharedEnvHandle` is first called.

While the code is trying to be safe by reloading the driver, doing it so late in the `close()` method, for a connection that was supposedly active, might be papering over an earlier, more fundamental problem, or it might not interact correctly with the existing connection handle if the driver state is genuinely compromised.

It's generally better to ensure the driver is loaded correctly once, at the beginning, and that it stays correctly loaded. If it somehow unloads or its function pointers become null, that usually indicates a more severe issue that a simple reload in close() might not cleanly resolve for the existing `_dbc_handle`.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -12,39 +12,143 @@
 // This class wraps low-level ODBC operations like connect/disconnect,
 // transaction control, and autocommit configuration.
 //-------------------------------------------------------------------------------------------------
-Connection::Connection(const std::wstring& conn_str) : _conn_str(conn_str) {}
+Connection::Connection(const std::wstring& conn_str, bool autocommit)
```
sumitmsft, line 19: The `autocommit` flag is passed in the constructor. Going forward we will have other connection attributes which are needed to be configurable (e.g., timeouts, network packet size etc.), a more extensible configuration mechanism (like properties struct or map) could be considered for the Connection constructor or dedicated setter methods. This makes the code cleaner, easier to use correctly, and much simpler to update when new features are added.

bewithgaurav: I think that `autocommit` can be a public attribute for this class and should be accessible and editable from python itself. Porting to cpp later on would also require the same, since end user should be able to directly set the attribute. Agreeing with Sumit on other part, they can be passed in Connection using a dict (maybe can be called `attr_dict`)

mssql_python/pybind/connection/connection.cpp
bewithgaurav: Agreed, the ODBC function ptrs being null signifies some inconsistencies from previous run, and the env handle must've gotten abandoned in that case. Reloading driver again to access the connection/env handle again would cause access violations. It will cause inconsistencies in case of multiple threads as mentioned above. I believe it might just be better if we exit the code in this case after logging and raising an exception since this means something bad has happened already.

mssql_python/pybind/connection/connection.h
```diff
@@ -10,12 +10,12 @@
...
 
 // Represents a single ODBC database connection.
-// Manages its own environment and connection handles.
+// Manages connection handles.
 // Note: This class does NOT implement pooling logic directly.
 
 class Connection {
 public:
-    Connection(const std::wstring& conn_str);
+    Connection(const std::wstring& conn_str, bool autocommit = false);
```
bewithgaurav, line 16: same as above comment on autocommit

mssql_python/pybind/connection/connection.h
```diff
@@ -31,18 +31,23 @@ class Connection {
     SQLRETURN rollback();
 
     // Enable or disable autocommit mode.
-    SQLRETURN set_autocommit(bool value);
+    SQLRETURN setAutocommit(bool value);
 
     //  Check whether autocommit is enabled.
-    bool get_autocommit() const;
+    bool getAutocommit() const;
```
bewithgaurav, line 35: if we make autocommit as non private, we wouldn't be needing getAutocommit

bewithgaurav (changes requested):

mssql_python/pybind/connection/connection.cpp
gargsaumya: Right, it is already a public attribute right now and all the actions on it are being maintained by C++, python class redirects it to C++.
I agree with using a structured config (or dict) for other connection attributes which will be needed to be configurable in future. For now, I’ll proceed with the current approach to keep things consistent and unblock current task, but we can plan a follow-up change to maybe use a dict.

mssql_python/pybind/connection/connection.cpp
gargsaumya: Ack.

mssql_python/pybind/connection/connection.h
gargsaumya: If we do that, we'll just return the cached attribute. Whereas, this function is used to call SQLGetConnectAttr(SQL_ATTR_AUTOCOMMIT), when you want guaranteed accuracy of the current state directly from the driver.

This is more robust and is needed because the autocommit mode might have been changed externally, e.g.,:
1. By direct SQLSetConnectAttr calls outside your control (maybe by other layers).
2. By executing SQL statements that implicitly change connection state.

mssql_python/pybind/connection/connection.cpp
gargsaumya: Ack.

mssql_python/pybind/connection/connection.cpp
gargsaumya: Ack.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -7,44 +7,124 @@
...
+    : _connStr(conn_str) , _autocommit(autocommit) {
+    if (!_envHandle) {
+        LOG("Allocating environment handle");
+        SQLHANDLE env = nullptr;
+        if (!SQLAllocHandle_ptr) {
```
bewithgaurav, line 24: we missed removing this as well as per [previous comment](https://github.com/microsoft/mssql-python/pull/54#discussion_r2109271876)
sorry - might've missed adding a comment here last time

bewithgaurav: Adding more context since I see that `loadDriver()` is guarded by `call_once`, I would still suggest removing this since we already have a `success` variable which already indicates proper loading of ptrs

sumitmsft (approved):

bewithgaurav (approved):
Create a task for the memory issue https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37606

PR #55, opened 2025-05-20
FEAT: Integrate python <->C++ flow with C++ connection class

gargsaumya:
This pull request makes the database connection and cursor handling in the `mssql_python` module via new C++ `Connection` class and simplifying the Python-side implementation. The changes include replacing low-level connection management with a higher-level abstraction, removing redundant methods, and delegating key operations to the new C++ backend. Below are the most important changes grouped by theme:

### Refactoring Python Connection Management
* Replaced manual handle allocation (`henv`, `hdbc`) and initialization in `mssql_python/connection.py` with a new `Connection` class from the C++ backend. The `_initializer`, `_allocate_environment_handle`, `_allocate_connection_handle`, and related methods were removed. 
* Simplified transaction methods (`commit`, `rollback`) and connection lifecycle methods (`close`) by delegating their logic to the C++ `Connection` class.
* Updated autocommit handling to use the `set_autocommit` and `get_autocommit` methods from the C++ `Connection` class. 
### Refactoring Python Cursor Management
* Removed checks for closed connections in `mssql_python/cursor.py` and delegated statement handle allocation to the C++ `Connection` class. 
* Simplified cursor cleanup by removing explicit handle freeing logic. 

### Integration with Python Bindings
* Updated `ddbc_bindings.cpp` to include the new `Connection` class, enabling its use in the Python layer. 

### Checklist  
- [ ] **Tests Passed** (if applicable) : Pytests need to be fixed with connection logic in C++.

### DIFF
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -53,14 +53,20 @@ class Connection:
         preparing it for further operations such as connecting to the 
         database, executing queries, etc.
         """
-        self.henv = None
-        self.hdbc = None
         self.connection_str = self._construct_connection_string(
             connection_str, **kwargs
         )
-        self._attrs_before = attrs_before
-        self._autocommit = autocommit  # Initialize _autocommit before calling _initializer
-        self._initializer()
+        # self._attrs_before = attrs_before
+        # self._autocommit = autocommit
+        # if self._attrs_before != {}:
+        #     self._apply_attrs_before()  # Apply pre-connection attributes
+        # if self._autocommit:
+        #     self.setautocommit(autocommit)
+        print("Connection string:", self.connection_str)
+        self._conn = ddbc_bindings.Connection(self.connection_str)
+        self._conn.connect()
+        print("Connection established")
+        self._autocommit = autocommit
         self.setautocommit(autocommit)
 
     def _construct_connection_string(self, connection_str: str = "", **kwargs) -> str:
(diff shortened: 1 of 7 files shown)


mssql_python/pybind/connection/connection.cpp
```diff
@@ -15,12 +18,18 @@
 Connection::Connection(const std::wstring& conn_str, bool autocommit)
     : _conn_str(conn_str) , _autocommit(autocommit) {}
 
-Connection::~Connection() {
-    close();    // Ensure the connection is closed when the object is destroyed.
-}
+Connection::~Connection() {}
```
sumitmsft: Consider reinstating the call to `close()` or an equivalent cleanup logic within this destructor. If the Python Connection object is garbage collected without an explicit `close()` call, the current empty destructor will result in the `SQLDisconnect_ptr` function (from the `Connection::close() `method) not being called. This could lead to database server sessions not being gracefully terminated, potentially consuming server resources until a timeout. While the `SqlHandle` destructor will free the client-side handle, explicitly disconnecting from the server is crucial for proper resource management. Ref. https://github.com/microsoft/mssql-python/pull/55/files#r2107519526

mssql_python/pybind/connection/connection.cpp
```diff
@@ -60,7 +69,7 @@ SQLRETURN Connection::close() {
     }
 
     SQLRETURN ret = SQLDisconnect_ptr(_dbc_handle->get());
```
sumitmsft: The call to SQLDisconnect_ptr is essential for proper server-side resource cleanup. The Connection destructor should ensure this (or an equivalent) is called if the connection is still active upon destruction.

mssql_python/connection.py
```diff
@@ -340,9 +150,6 @@ def cursor(self) -> Cursor:
             DatabaseError: If there is an error while creating the cursor.
             InterfaceError: If there is an error related to the database interface.
         """
-        if self._is_closed():
```
sumitmsft, line 343: Previously, before creating a Cursor, the Python code explicitly checked if self._is_closed():. If the connection was closed, it would raise a Python Exception. Now, this check is gone. The expectation is that if you try to create a cursor (which involves self.connection._conn.alloc_statement_handle() calling into C++) on a closed connection, the C++ layer (or the underlying ODBC driver) will detect this. This detection should then result in an error that propagates back to Python, ideally as a specific DatabaseError or InterfaceError (rather than a generic C++ runtime error). 
You need to test what happens if you call connection.cursor() after connection.close() has been called. Does it raise an appropriate and informative Python exception? If this test was already performed then we are good, otherwise we need to put that verification.

mssql_python/connection.py
```diff
@@ -357,17 +164,8 @@ def commit(self) -> None:
         Raises:
             DatabaseError: If there is an error while committing the transaction.
         """
-        if self._is_closed():
```
sumitmsft, line 360: Similar to cursor(), the commit() method used to check if self._is_closed():. Now, it directly calls self._conn.commit(), which is a method on the C++ Connection object. If the connection is already closed at the C++ level (i.e., _dbc_handle is invalid or disconnected), the C++ commit() method (in connection.cpp) should ideally recognize this state. It might try to call SQLEndTran_ptr, which would likely fail on a closed handle. This failure should then be translated into a Python exception. 
What I would recommend it to test calling connection.commit() after connection.close(). Does it fail gracefully with a clear Python exception (e.g., "Cannot commit on a closed connection")?

mssql_python/connection.py
```diff
@@ -382,17 +180,8 @@ def rollback(self) -> None:
         Raises:
             DatabaseError: If there is an error while rolling back the transaction.
         """
-        if self._is_closed():
```
sumitmsft, line 385: The reasoning is identical to the commit() method. The Python-level _is_closed() check is removed, relying on the C++ layer to handle attempts to rollback on a closed connection and to surface an appropriate error to Python. 

Test calling connection.rollback() after connection.close(). Does it produce a clear and correct Python exception?

mssql_python/connection.py
```diff
@@ -53,14 +53,13 @@ def __init__(self, connection_str: str = "", autocommit: bool = False, attrs_bef
...
         self.connection_str = self._construct_connection_string(
             connection_str, **kwargs
         )
-        self._attrs_before = attrs_before
-        self._autocommit = autocommit  # Initialize _autocommit before calling _initializer
-        self._initializer()
+        self._attrs_before = attrs_before or {}
+        self._conn = ddbc_bindings.Connection(self.connection_str, autocommit)
+        self._conn.connect(self._attrs_before)
+        self._autocommit = autocommit
```
sumitmsft: **Minor observation regarding the state of self._autocommit**

This line sets the Python instance variable `self._autocommit` based on the `autocommit` parameter. Immediately after, `self.setautocommit(autocommit)` is called, which itself updates `self._autocommit` again after ensuring the C++ object's `autocommit` state is set via `self._conn.set_autocommit(value)`.

While the `autocommit` property getter (`self._conn.get_autocommit()`) correctly fetches the state from the C++ object, and `setautocommit()` ensures final consistency, there's a very brief moment where `self._autocommit` (set here on L2) might not perfectly reflect the actual effective `autocommit` state of the underlying C++ Connection object if its constructor or `connect()` method had any complex internal logic that modified `autocommit` before `self.setautocommit()` is called.

This is likely a very low potential bug and probably has no practical impact in the current flow because `self.setautocommit()` immediately follows and synchronizes everything. However, for strict state consistency, one might consider if `self._autocommit` should only be authoritatively set after the C++ state is confirmed (e.g., solely within `setautocommit` or by fetching from C++ after initial setup if needed).

No immediate action required unless there's a known C++ side-effect I'm missing, but worth a quick thought for future refactoring or if issues around `autocommit` state were to arise. The current approach with the subsequent `self.setautocommit()` call is robust in practice.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -127,6 +132,54 @@ SqlHandlePtr Connection::allocStatementHandle() {
...
+
+    SQLPOINTER ptr = nullptr;
+    SQLINTEGER length = 0;
+
+    if (py::isinstance<py::int_>(value)) {
+        int intValue = value.cast<int>();
+        ptr = reinterpret_cast<SQLPOINTER>(static_cast<uintptr_t>(intValue));
+        length = SQL_IS_INTEGER;
+    } else if (py::isinstance<py::bytes>(value) || py::isinstance<py::bytearray>(value)) {
+        static std::vector<std::string> buffers;
```
sumitmsft, line 153: **Critical Concern: Memory Leak and Security Risk with Static Token Buffer**

The use of `static std::vector<std::string> buffers;` for handling byte string attributes (like access tokens) presents two significant issues:

**Memory Leak:** The buffers vector is static, meaning it's shared across all calls to `setAttribute` and persists for the application's lifetime. Each time this block is executed, a new token string is added via `emplace_back` but never removed. This will cause the vector to grow indefinitely, leading to a memory leak.

**Security Risk (Sensitive Data Exposure):** Access tokens are highly sensitive. Storing them in a long-lived, ever-growing static vector means these tokens remain in application memory far longer than necessary. This significantly increases the risk of token exposure if the application's memory is compromised (e.g., via other vulnerabilities, core dumps, or direct memory inspection).
Ideally, sensitive data like tokens should have their memory securely zeroed out as soon as they are no longer needed for the immediate ODBC operation. This is not happening here.

Possibly you may have to make come out with some solutions to take care of both issues. 

I find these two issues as high priority, as it impacts both stability (memory leak) and security (token exposure). We can discuss in detail if you wish.

mssql_python/connection.py
```diff
@@ -53,14 +53,13 @@ def __init__(self, connection_str: str = "", autocommit: bool = False, attrs_bef
...
-        self.henv = None
-        self.hdbc = None
         self.connection_str = self._construct_connection_string(
             connection_str, **kwargs
         )
-        self._attrs_before = attrs_before
-        self._autocommit = autocommit  # Initialize _autocommit before calling _initializer
-        self._initializer()
+        self._attrs_before = attrs_before or {}
+        self._conn = ddbc_bindings.Connection(self.connection_str, autocommit)
```
bewithgaurav, line 60: My comments for managing autocommit is in #54 . If looks good, this would also require changes

mssql_python/pybind/connection/connection.cpp
bewithgaurav: Usage of static vectors was done to ensure that ODBC is able to access the token at any time using the pointers - [changes ref](https://sqlclientdrivers.visualstudio.com/mssql-python/_git/mssql-python/pullRequest/5476#1744109444)
However I agree on the comments made by sumit, I believe theekshna had suggested a different approach which we weren't able to integrate at that time. 
@gargsaumya - please checkout [this comment](https://sqlclientdrivers.visualstudio.com/mssql-python/_git/mssql-python/pullRequest/5476#1744111469) from ADO, this might be a better way of doing things

bewithgaurav (changes requested):

mssql_python/pybind/connection/connection.cpp
gargsaumya: This is the same function/bug from Access token support that we identified in earlier PRs. As discussed earlier, since this issue is not a part of this PR, we'll be taking this up on priority in separate PR so that it does not affect our progress for release.

mssql_python/connection.py
gargsaumya: Addressed it in mentioned PR.

mssql_python/pybind/connection/connection.cpp
gargsaumya: This is an outdated file, the changes have been rebased with PR-54

mssql_python/pybind/connection/connection.cpp
gargsaumya: This is already being handled in destructor. (the file was outdated)

mssql_python/connection.py
gargsaumya: These scenarios are now handled in the C++ layer, so the corresponding checks are no longer needed in Python. However, we plan to introduce proper exception classes in C++ in the future, aligned with the DB-API specification, similar to what we previously implemented in Python. For now, the C++ layer raises generic RuntimeError exceptions.

mssql_python/connection.py
gargsaumya: We are safely handling these scenarios now in the C++ layer, so the corresponding checks are no longer needed in Python. However, we plan to introduce proper exception classes in C++ in the future, aligned with the DB-API specification, similar to what we previously implemented in Python. For now, the C++ layer raises generic RuntimeError exceptions.

mssql_python/connection.py
gargsaumya: addressed in above comments.

mssql_python/connection.py
gargsaumya: This Python variable is no longer needed, as its functionality is now handled in the C++ layer. It has now been removed.

mssql_python/pybind/connection/connection.cpp
gargsaumya: Created a WI to address this: https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37606

PR #57, opened 2025-05-22
FEAT: Adding BCP APIs for python

jahnvi480:
This pull request introduces a new `BCPClient` class in the `mssql_python/bcp_main.py` file to handle bulk copy operations using the BCP (Bulk Copy Program) utility. The class provides methods for initializing and executing BCP operations, with robust validation and configuration options.

### New Feature: `BCPClient` class for bulk copy operations
* Added a `BCPClient` class to encapsulate bulk copy operations (`in` and `out`) using the BCP utility. This includes methods for initializing BCP operations and executing them with configurable options.
* Introduced `_validate_options` to validate or provide default `BCPOptions`, ensuring proper configuration for the BCP operation.
* Implemented `run_bcp` to execute the BCP operation, supporting features like batch size, error handling, row range, code page, identity and null handling, hints, and custom column formats or format files.


mssql_python/bcp_main.py
```diff
@@ -0,0 +1,98 @@
...
+                options.direction must be 'in' or 'out'.
+        """
+        current_options = self._validate_options(options)
+
+        # BCPOptions.__post_init__ ensures data_file and error_file are set for 'in'/'out'
+        self.wrapper.bcp_initialize_operation(table, current_options.data_file, current_options.error_file, current_options.direction)
+
+        # Set BCP control options
+        if current_options.batch_size is not None:
+            self.wrapper.bcp_control("BCPBATCH", current_options.batch_size)
```
sumitmsft: **Constants for BCP Controls**

Instead of using string literals like "BCPBATCH", "BCPMAXERRS", etc., directly in bcp_control calls (lines 54-72), consider defining these as constants. This can prevent typos and make the code easier to understand and maintain. For example:
```
BCP_BATCH_SIZE = "BCPBATCH"
# ...
self.wrapper.bcp_control(BCP_BATCH_SIZE, current_options.batch_size)

```

mssql_python/bcp_main.py
```diff
@@ -0,0 +1,98 @@
...
+        if current_options.max_errors is not None:
+            self.wrapper.bcp_control("BCPMAXERRS", current_options.max_errors)
+        if current_options.first_row is not None:
+            self.wrapper.bcp_control("BCPFIRST", current_options.first_row)
+        if current_options.last_row is not None:
+            self.wrapper.bcp_control("BCPLAST", current_options.last_row)
+        if current_options.code_page is not None:
+            self.wrapper.bcp_control("BCPFILECP", current_options.code_page)
+        if hasattr(current_options, 'keep_identity') and current_options.keep_identity:
```
sumitmsft: **keep_identity and keep_nulls on BCPOptions (lines 64-66):**

`keep_identity` and `keep_nulls` are considered standard configuration options for `BCPOptions`. It might be more robust if `BCPOptions.__init__ `always initializes them, perhaps to a default like `False `or `None`. 

This would mean any `BCPOptions` instance is guaranteed to have these attributes. The client code could then directly access their values (e.g., `if current_options.keep_identity:`), making the interaction clearer and relying on a more consistent object structure.

What we can do:
- Ensure the `BCPOptions` class constructor (`__init__`) always initializes these attributes, even if it's to a default value like `False `or `None`. 

Example:
```
class BCPOptions:
    def __init__(self, ..., keep_identity=False, keep_nulls=False, ...):
        # ...
        self.keep_identity = keep_identity
        self.keep_nulls = keep_nulls
        # ...
```

- This way any instance of `BCPOptions` is guaranteed to have `keep_identity` and `keep_nulls`. The `BCPClient` can then directly use:
```
# In BCPClient.run_bcp
if current_options.keep_identity:
    self.wrapper.bcp_control("BCPKEEPIDENTITY", 1)
if current_options.keep_nulls:
    self.wrapper.bcp_control("BCPKEEPNULLS", 1)
```

This way the code will be cleaner, expresses the intent more directly (we're checking the value of the flag, assuming its existence), and relies on a clearer contract from `BCPOptions`. If `None `is a meaningful state (e.g., "not specified, let BCP use its default"), the check could be `if current_options.keep_identity is not None:`.

mssql_python/bcp_main.py
```diff
@@ -0,0 +1,98 @@
...
+                    col_name=getattr(col, 'col_name', None), # Assuming None if not present
+                    server_col=col.server_col,
+                    file_col=col.file_col
+                )
+        
+        self.wrapper.exec_bcp()
+        self.wrapper.finish()
+        
+        if hasattr(self.wrapper, 'close'):
+            self.wrapper.close()
```
sumitmsft: The `self.wrapper `object (which is an instance of `BCPWrapper` from `ddbc_bindings`) might not always have a `close `method. If `close `might not be there, how is resource cleanup guaranteed in those cases?

If a wrapper instance should have been closed but didn't have a `close `method (and the `hasattr` check prevented an error), resources might be leaked.

IMO, the `BCPWrapper` class should ideally always provide a `close()` method as part of its defined interface, especially if it manages underlying native resources. If there are scenarios where `close()` does nothing (e.g., if resources were already released or none were acquired), the method can simply be a no-op (i.e., `def close(self): pass`).

The `BCPClient` can then unconditionally call `self.wrapper.close()` (perhaps in a `finally `block to ensure it's called even if errors occur earlier).

```
# In BCPClient.run_bcp
try:
    # ... bcp operations ...
    self.wrapper.exec_bcp()
finally:
    self.wrapper.finish() # Assuming finish is always present
    self.wrapper.close()  # Now confidently callable
```

mssql_python/bcp_main.py
```diff
@@ -0,0 +1,98 @@
...
+                    data_len=col.data_len,
+                    terminator=col.field_terminator, 
+                    server_col_type=getattr(col, 'server_col_type', 0), # Assuming 0 if not present
+                    col_name=getattr(col, 'col_name', None), # Assuming None if not present
+                    server_col=col.server_col,
+                    file_col=col.file_col
+                )
+        
+        self.wrapper.exec_bcp()
+        self.wrapper.finish()
```
sumitmsft: The proper management of underlying BCP handles and resources by BCPWrapper is critical. The calls to self.wrapper.finish() (line 95) and the conditional self.wrapper.close() (line 98, though the hasattr check) are key to this.

It's crucial that these methods reliably release all resources acquired by the BCP operation to prevent leaks, especially under error conditions.

Consider explicitly using a `try...finally` block around the core BCP execution logic (`self.wrapper.bcp_initialize_operation(...) `through `self.wrapper.exec_bcp()`) to guarantee that `self.wrapper.finish()` and `self.wrapper.close()` (or its equivalent unconditional call) are executed regardless of whether the operations within the try block succeed or raise an exception.

For example:

```
# In BCPClient.run_bcp method
try:
    # ... (initialization and option setting) ...
    self.wrapper.bcp_initialize_operation(...)
    # ... (other bcp_control calls) ...
    self.wrapper.exec_bcp()
finally:
    if self.wrapper: # Ensure wrapper exists
        self.wrapper.finish()
        # Ideally, self.wrapper.close() would be unconditionally callable here
        # based on previous discussions about its interface.
        # If hasattr must remain for now:
        if hasattr(self.wrapper, 'close'):
            self.wrapper.close()
```

This pattern would significantly improve the robustness of resource cleanup, ensuring that BCP resources are consistently released even when exceptions occur during the data transfer phase.

mssql_python/bcp_main.py
```diff
@@ -0,0 +1,98 @@
...
+            self.wrapper.read_format_file(current_options.format_file)
+        elif current_options.columns:
+            self.wrapper.define_columns(len(current_options.columns))
+            for i, col in enumerate(current_options.columns):
+                self.wrapper.define_column_format(
+                    col_num=i + 1, 
+                    prefix_len=col.prefix_len,
+                    data_len=col.data_len,
+                    terminator=col.field_terminator, 
+                    server_col_type=getattr(col, 'server_col_type', 0), # Assuming 0 if not present
```
sumitmsft: For `server_col_type = getattr(col_format, "server_col_type", 0)`: It's important to ensure that 0 is a genuinely valid, safe, and meaningful default value for `server_col_type `when passed to the underlying `self.wrapper.bcp_colfmt()` method.

Does the `BCPWrapper` (and the native BCP API it interfaces with) interpret 0 for `server_col_type` in a well-defined way that doesn't lead to errors or unexpected type conversions?
If 0 signifies "let BCP infer the type" or some other specific default behavior, this understanding is crucial.

It would be beneficial to:
1. Double-check the BCP documentation or BCPWrapper implementation to confirm that 0 is indeed a correct and safe default for `server_col_type`.
2. Consider adding a comment in the code or in the ColumnFormat class documentation explaining the meaning or implication of `server_col_type` being 0 (or not being provided, thus defaulting to 0).

mssql_python/bcp_main.py
```diff
@@ -0,0 +1,98 @@
...
+        self.wrapper = BCPWrapper(connection)
+
+    def _validate_options(self, options: Optional[BCPOptions]) -> BCPOptions:
+        """
+        Validates the provided BCP options or provides a default.
+        Ensures options are suitable for 'in' or 'out' operations.
+        """
+        if options is None:
+            # Default options for 'in' operation
+            return BCPOptions(direction="in", data_file="default_data.bcp", error_file="default_error.err", bulk_mode="native")
```
sumitmsft: If a user runs bcp without providing options, the tool should not execute successfully. Instead, it should display a usage message that outlines the correct syntax and available options. BCP expects at least the direction (in, out, queryout, or format) and the target (table or query) to be specified.
- If options are missing, BCP should fail gracefully by printing the help/usage instructions or perhaps with some error message. Perhaps something like ValueError
- It shouldn't cause a crash.
- Even for missing essential options like data_file or table there must be a ValueError

The BCPOptions class itself could be designed to handle defaults more robustly, perhaps having its __init__ method require essential parameters like direction and table etc.

mssql_python/bcp_options.py
```diff
@@ -71,51 +72,75 @@ class BCPOptions:
...
+    code_page: Optional[Union[int, str]] = None
+    hints: Optional[str] = None
+    columns: Optional[List[ColumnFormat]] = field(default_factory=list)
     keep_identity: bool = False
     keep_nulls: bool = False
-    hints: Optional[str] = None
-    bulk_mode: Literal["native", "char", "unicode"] = "native"
-    columns: List[ColumnFormat] = field(default_factory=list)
+
+    ALLOWED_DIRECTIONS = {"in", "out"}
```
bewithgaurav, line 91: can be placed in constants.py

mssql_python/bcp_options.py
```diff
@@ -71,51 +72,75 @@ class BCPOptions:
         columns (List[ColumnFormat]): Column formats.
     """
 
-    direction: Literal["in", "out"]
+    direction: str
     data_file: str  # data_file is mandatory for 'in' and 'out'
     error_file: Optional[str] = None
     format_file: Optional[str] = None
     # write_format_file is removed as 'format' direction is not actively supported
+    bulk_mode: Optional[str] = "native"  # Default to 'native' mode
```
bewithgaurav, line 80: can we defined the allowed modes as a list (similar to ALLOWED DIRECTIONS) and use it similarly to filter unwanted modes? this as well can be placed in constants.py

mssql_python/bcp_options.py
```diff
@@ -71,51 +72,75 @@ class BCPOptions:
...
             and self.last_row is not None
-            and self.last_row < self.first_row
+            and self.first_row > self.last_row
         ):
-            raise ValueError("last_row must be greater than or equal to first_row.")
-        if self.code_page is not None and not self.code_page:
-            raise ValueError("code_page, if provided, must not be an empty string.")
-        if self.hints is not None and not self.hints:
-            raise ValueError("hints, if provided, must not be an empty string.")
-        if self.bulk_mode not in ["native", "char", "unicode"]:
```
bewithgaurav, line 120: are these checks still ensured? if not, can re-add them and check with constants if the values are filtered well or not.

mssql_python/bcp_main.py
```diff
@@ -0,0 +1,146 @@
...
+                        terminator_wstr=col_format_obj.field_terminator,
+                        col_name=col_format_obj.col_name,
+                        server_col=col_format_obj.server_col,
+                        file_col=col_format_obj.file_col,
+                    )
+
+            self.wrapper.exec_bcp()
+
+        finally:
+            if self.wrapper:
```
codewatchzen, line 144: Just a suggestion: can cleanup ever fail in finally block? If yes, considering logging exception.

mssql_python/bcp_main.py
codewatchzen, line 1: Have you considered adding unit test cases for bcp_options scenarios?

PR #58, opened 2025-05-23
FEAT: MacOS Support

bewithgaurav:
This pull request introduces platform-specific adjustments to the `mssql_python` package, adds a new `Connection` class for macOS, and refines module loading logic for compatibility across different operating systems. The most significant changes include the addition of macOS-specific connection handling, updates to dynamic module loading, and removal of unused constants.

### Platform-specific adjustments:

* [`mssql_python/db_connection.py`](diffhunk://#diff-dc442a0da15dacf049752b6c1aa47a48f9e84fb6965215dd3b2457d9cc19de82R6-R10): Added conditional imports to load the appropriate `Connection` class based on the operating system (`Windows` or macOS).
* [`mssql_python/ddbc_bindings.py`](diffhunk://#diff-65fc4dc854e915ab51bf5ef8703891df3f52f15fb8ab3e152d3c81ac19864525L9-R36): Updated dynamic module loading logic to determine the correct file extension (`.pyd` for Windows, `.so` for macOS/Linux) and added fallback handling for missing modules.

### New macOS connection handling:

* [`mssql_python/connection_mac.py`](diffhunk://#diff-fc969de9bfd18c6c71b4703f7abe8d6a7c82a4a09fef92d97c34088b6ae51474R1-R330): Introduced a new `Connection` class tailored for macOS, providing methods for database connection management, transaction handling, and cursor creation.

### Code cleanup:

* [`mssql_python/constants.py`](diffhunk://#diff-e6d80f1000af6fd5afca05f435b11fd82df7f5c3e75ecf5763f85d3aacdbe758L119-L120): Removed unused constants `SQL_IS_POINTER` and `SQL_COPT_SS_ACCESS_TOKEN` from the `ConstantsDDBC` class.

### Refactored imports:

* [`mssql_python/__init__.py`](diffhunk://#diff-d95f3a67986de29f30453416b1b4c34e6a43207e9a33e2b1b80ef0c378b0a538L39-R39): Simplified imports by combining `connect` and `Connection` into a single line.

### ADO Task ID
- [AB#34977](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34977)


mssql_python/connection_mac.py
bewithgaurav, line 1: created a new file for mac connection since the smart pointer update has not yet been propagated to ddbc bindings in mac. will be porting this change in the next iteration.

mssql_python/pybind/ddbc_bindings_mac.cpp
bewithgaurav, line 1: created a new bindings file for mac since the smart pointer changes have not yet been ported to MacOS changes.

tests/test_005_exceptions.py
```diff
@@ -126,5 +126,5 @@ def test_foreign_key_constraint_error(cursor, db_connection):
 
 def test_connection_error(db_connection):
     with pytest.raises(OperationalError) as excinfo:
-        Connection("InvalidConnectionString")
+        connect("InvalidConnectionString")
```
bewithgaurav, line 136: using a connect() method instead of calling Connection class directly

mssql_python/cursor_mac.py
bewithgaurav, line 1: created a new file for mac cursor since the smart pointer update has not yet been propagated to ddbc bindings in mac. will be porting this change in the next iteration.

Task - https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37606

mssql_python/pybind/ddbc_bindings_mac.cpp
```diff
@@ -0,0 +1,2416 @@
...
+#include <iomanip>  // std::setw, std::setfill
+#include <iostream>
+#include <string>
+#include <utility>  // std::forward
+
+// Platform-specific headers
+#include <sql.h>
+#include <sqlext.h>
+
+#ifdef _WIN32
```
bewithgaurav, line 19: since this thing started off as a common file for windows and mac - I have not removed the ifdef WIN32/APPLE portions of the file. Kept this as it is since we might merge this to commonize ddbc_bindings across platforms. Parking this into another task - [link](https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37626)

mssql_python/pybind/ddbc_bindings_mac.cpp
```diff
@@ -0,0 +1,2416 @@
...
+    // Ensure the buffer is null-terminated
+    LOG("Connection string buffer size - {}", connStrBuffer.size());
+    connStrPtr = connStrBuffer.data();
+    LOG("Connection string buffer created");
+#else
+    connStrPtr = const_cast<SQLWCHAR*>(ConnectionString.c_str());
+#endif
+
+    // Log the actual connection string, not the pointer value
+    LOG("Connection string - {}", ConnectionString);
```
sumitmsft: I've noticed a potential security concern. The entire connection string is logged using the `LOG` macro:
```cpp
1104|     // Log the actual connection string, not the pointer value
1105|     LOG("Connection string - {}", ConnectionString);
```
If the `ConnectionString` variable includes sensitive information such as passwords or access tokens, these will be written to the Python logger output when debug logging is enabled for the `mssql_python.logging_config` logger. This could inadvertently expose credentials in log files or console outputs, posing a security risk.

Can we modify the logging statement to mask or omit sensitive parts of the connection string before logging. For example, common parameters like `Password`, `PWD`, `UID`, or tokens could be replaced with a placeholder like `********`.

sumitmsft: Perhaps, keep it at P0 and create a backlog item for this and fix it immediately after June 2nd?

mssql_python/pybind/ddbc_bindings_mac.cpp
```diff
@@ -0,0 +1,2416 @@
...
+    }
+    std::cerr << "DEBUG: Could not extract directory from path: " << module_file << std::endl;
+    return module_file;
+#endif
+}
+
+namespace {
+
+// TODO: Revisit GIL considerations if we're using python's logger
+template <typename... Args>
```
sumitmsft, line 309: As I understand this code base, the `LOG` macro calls Python's logging system, which needs the Python Global Interpreter Lock (GIL). If our C++ code holds its own lock (like a `std::mutex`) and then calls `LOG`, it will wait for the GIL. If another Python thread already has the GIL and tries to get that **same C++ lock**, both threads can get stuck waiting for each other – a deadlock. would happen. I think The `TODO` on Line 311 ("Revisit GIL considerations") points to this. 
The simplest way to avoid this is to make sure the `LOG` macro isn't called when C++ code is holding its own locks that Python might also need. This is important for stability, especially in multi-threaded apps.

Can you please verify if we have backlog item for this TODO? If not, let's have it and fix it in subsequent releases.

mssql_python/pybind/ddbc_bindings_mac.cpp
```diff
@@ -0,0 +1,2416 @@
...
+                LOG("Received numeric parameter: precision - {}, scale- {}, sign - {}, value - {}",
+                    decimalParam.precision, decimalParam.scale, decimalParam.sign,
+                    decimalParam.val);
+                SQL_NUMERIC_STRUCT* decimalPtr =
+                    AllocateParamBuffer<SQL_NUMERIC_STRUCT>(paramBuffers);
+                decimalPtr->precision = decimalParam.precision;
+                decimalPtr->scale = decimalParam.scale;
+                decimalPtr->sign = decimalParam.sign;
+                // Convert the integer decimalParam.val to char array
+                std:memset(static_cast<void*>(decimalPtr->val), 0, sizeof(decimalPtr->val));
```
sumitmsft, line 790: @bewithgaurav I've found a potential issue in how we handle `NUMERIC` or `DECIMAL` types when sending them as parameters to the database. 

**This will be applicable across all platforms, so please check this out on Windows code base as well (ddbc_bindings.cpp) as well.** Cc. @gargsaumya  

1.  When Python sends a `Decimal` number, we have a C++ structure `NumericData` where the actual numeric value is stored as a `std::uint64_t val` (e.g., for `123.45`, `val` becomes `12345`).
2.  This `uint64_t val` is then copied byte-for-byte using `memcpy` into the `val` field of the ODBC `SQL_NUMERIC_STRUCT`.
   
*   A `std::uint64_t` can only hold numbers up to about 19-20 decimal digits.
*   SQL `NUMERIC` or `DECIMAL` types can be much larger (e.g., `NUMERIC(38, 2)` can have 38 digits).
*   If Python sends a `Decimal` whose scaled integer value is too big to fit into our `uint64_t val`, the value will be **truncated (cut short) or mangled** when it's converted to `uint64_t`.
*   The `memcpy` then copies these incorrect, truncated bytes into the `SQL_NUMERIC_STRUCT` that ODBC uses.

- If you try to send a large `Decimal` value (e.g., one with more than ~19 digits for its scaled integer part) from Python to the database, the database will receive an incorrect, smaller, or garbled value.

- It does *not* appear to affect fetching `NUMERIC`/`DECIMAL` data from the database, which seems to handle full precision correctly by converting to a string first.

**Example:**
If you try to insert `Decimal('12345678901234567890123')` (23 digits) into a `NUMERIC(25,0)` column:
1.  This number is too big for a `uint64_t`.
2.  It gets truncated/changed when put into `NumericData::val`.
3.  The database receives this wrong, smaller value.

**Suggestion:**
To fix this, the `NumericData::val` (or a similar mechanism) needs to be able to hold the full precision of Python's `Decimal` as a byte array that matches the ODBC `SQL_NUMERIC_STRUCT::val` format, rather than being limited by `uint64_t`. The Python side would need to serialize its `Decimal` into this byte array format.

This looks like a significant bug for users working with high-precision decimal or numeric data.

sumitmsft: We can potentially fix after Public Preview. Please create a backlog workitem for this bug.

mssql_python/pybind/mac_fix.cpp
```diff
@@ -0,0 +1,111 @@
...
+        memcpy(&utf16Bytes[i * kUcsLength], &sqlwStr[i], kUcsLength);
+    }
+    
+    // Convert UTF-16LE to std::wstring (UTF-32 on macOS)
+    try {
+        // Use C++11 codecvt to convert between UTF-16LE and wstring
+        std::wstring_convert<std::codecvt_utf8_utf16<wchar_t, 0x10ffff, std::little_endian>> converter;
+        return converter.from_bytes(reinterpret_cast<const char*>(utf16Bytes.data()), 
+                                   reinterpret_cast<const char*>(utf16Bytes.data() + utf16Bytes.size()));
+    } catch (const std::exception& e) {
```
sumitmsft, line 51: We have a function `SQLWCHARToWString` to convert strings between how ODBC gives them to us and how C++ (on macOS) likes them.

The code first tries a good, reliable method (`std::codecvt`) for these conversions. If that method fails for some reason, it "falls back" to a simpler, more basic conversion.

This simpler fallback conversion might not handle all special characters correctly (perhaps, emoji for example) , which could lead to:
- Incorrect characters (like '?' instead of the real character).
- Subtle data loss or changes in strings.

It would be helpful to log a warning message whenever this fallback conversion is used. For example: "Warning: Using fallback string conversion on macOS. Character data might be inexact."

If users report problems with funny characters or garbled text on macOS, these warnings in the logs would point us to where it might be happening. It helps us understand if the main, reliable conversion method (`codecvt`) is failing unexpectedly.

This is just about adding a log message to make troubleshooting easier if string issues pop up on macOS.

mssql_python/pybind/mac_fix.cpp
```diff
@@ -0,0 +1,111 @@
...
+        std::wstring_convert<std::codecvt_utf8_utf16<wchar_t, 0x10ffff, std::little_endian>> converter;
+        std::string utf16Bytes = converter.to_bytes(str);
+        
+        // Convert the bytes to SQLWCHAR array
+        std::vector<SQLWCHAR> result(utf16Bytes.size() / kUcsLength + 1, 0);  // +1 for null terminator
+        for (size_t i = 0; i < utf16Bytes.size() / kUcsLength; ++i) {
+            memcpy(&result[i], &utf16Bytes[i * kUcsLength], kUcsLength);
+        }
+        return result;
+    } catch (const std::exception& e) {
```
sumitmsft, line 77: Here also I want to add the log a warning message whenever this fallback conversion is used (similar to my previous comment)

mssql_python/pybind/mac_fix.cpp
```diff
@@ -0,0 +1,111 @@
...
+        utf16Bytes.push_back(bytes[0]);
+        utf16Bytes.push_back(bytes[1]);
+        i++;
+    }
+    
+    try {
+        std::wstring_convert<std::codecvt_utf8_utf16<char16_t, 0x10ffff, std::little_endian>> converter;
+        return converter.to_bytes(reinterpret_cast<const char16_t*>(utf16Bytes.data()), 
+                                 reinterpret_cast<const char16_t*>(utf16Bytes.data() + utf16Bytes.size()));
+    } catch (const std::exception& e) {
```
sumitmsft, line 108: Here also I want to add the log a warning message whenever this fallback conversion is used (similar to my previous comment)

mssql_python/connection_mac.py
```diff
@@ -0,0 +1,330 @@
...
+            elif key.lower() == "encrypt":
+                key = "Encrypt"
+            elif key.lower() == "trust_server_certificate":
+                key = "TrustServerCertificate"
+            else:
+                continue
+            conn_str += f"{key}={value};"
+
+        if ENABLE_LOGGING:
+            logger.info("Final connection string: %s", conn_str)
```
sumitmsft: This looks to be  a major security vulnerability related to how connection strings are logged. If `ENABLE_LOGGING` is set to true, the code logs the entire connection string, including any sensitive credentials like passwords or access tokens. 

Let's identify sensitive keys like passwords, tokens, or secrets in a connection string. Then mask sensitive values and then replace the actual values of these sensitive keys with a placeholder like `****` or `<masked>` or `<REDACTED>` before logging.

This is a high-priority issue and needs to be addressed immediately to prevent accidental credential exposure.

mssql_python/cursor_mac.py
```diff
@@ -0,0 +1,736 @@
...
+        Raises:
+            Error: If the previous call to execute did not produce any result set.
+        """
+        self._check_closed()  # Check if the cursor is closed
+
+        # Fetch all remaining rows
+        rows = []
+        ret = ddbc_bindings.DDBCSQLFetchAll(self.hstmt.value, rows)
+        check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt.value, ret)
+        if ret != ddbc_sql_const.SQL_NO_DATA.value:
```
sumitmsft, line 713: This issue catches my eye all of a sudden and it seems there is a logic problem here. Please correct me if I am wrong. 

The `fetchall()` would fetch all (remaining) rows of the query result.  

When fetching data from the database:
- If there's data, give me the data.
- If there's no data left, tell me "no data."
- If there's an error, tell me about the error.

**What this code line *actually* does:**
1. If Data Was Fetched Successfully (ret is SQL_SUCCESS). In this case, ret is not SQL_NO_DATA. So, the condition ret != SQL_NO_DATA is TRUE. The code executes return [].
2. If an Error Occurred During Fetch (ret is SQL_ERROR). In this case, ret is not SQL_NO_DATA. So, the condition ret != SQL_NO_DATA is TRUE. The code executes return [].
3. If No Data Was Found (ret is SQL_NO_DATA). In this case, the condition ret != SQL_NO_DATA is FALSE. The return [] statement is skipped.

A function fetching data should generally return the data if successful, raise an error if an error occurs, and return an empty list (or None) only if SQL_NO_DATA is received (meaning no data was available to fetch). This line does the opposite for success and error cases. It only proceeds to potentially return actual data (or handle the no-data case correctly) if the initial status was SQL_NO_DATA.

This needs to be fixed so it returns data when there's data, and properly reports errors. Let me know if this doesn't make sense.

sumitmsft: Please check for this issue in Windows cursor class as well

mssql_python/connection_mac.py
```diff
@@ -0,0 +1,330 @@
...
+                ctypes.pointer(self.hdbc), ctypes.c_void_p
+            ).value,  # SQL output handle pointer
+        )
+        check_error(ddbc_sql_const.SQL_HANDLE_DBC.value, self.hdbc.value, ret)
+
+    def _set_connection_attributes(self):
+        """
+        Set the connection attributes before connecting.
+        """
+        if self.autocommit:
```
imasud00, line 156: could there be a race condition that somesome calls this method and we are connecting but has not connected yet.  overall I think there are three states idle, connecting and connected.  Unless in Python that may not possible becasue GIL lock.

mssql_python/cursor_mac.py
bewithgaurav: I have made a fix to remove redundant line of code in the above section.
For the cases, actually we have covered all aspects - below is the explanation:

1. `SQL_SUCCESS` never comes out of DDBCSQLFetchAll - Inside the CPP implementation. We return/break the while loop only when we get `SQL_ERROR`/`SQL_NO_DATA`. I have removed the part where it contains the if condition (since it never gets called)
2.  A `SQL_ERROR` would go to check_error immediately and make an exception
3. If no data is found, still we will be returning a blank list anyways []

Made the change which cleans up the code, please resolve if all looks okay.

mssql_python/pybind/ddbc_bindings_mac.cpp
bewithgaurav: Acknowledged as a bug, we will plan for this in our upcoming release cycles.
Work item link - https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37630

mssql_python/pybind/ddbc_bindings_mac.cpp
bewithgaurav: Acknowledged. Created a work item in the backlog - https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37631

gargsaumya: We already created a work item for this : [Task 37537 Revisit GIL considerations if we're using python's logger or Add a separate logging utility in C++.](https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37537)

gargsaumya:
Though this is high LOC PR, but majorly contains the codebase already reviewed and running for Windows OS, there were no major concerns except the security concern highlighted by Sumit, hence approving the PR.

mssql_python/pybind/ddbc_bindings_mac.cpp
bewithgaurav: Fixed in the new push

mssql_python/pybind/mac_fix.cpp
bewithgaurav: Fixed in the new push

mssql_python/pybind/mac_fix.cpp
bewithgaurav: Done

mssql_python/pybind/mac_fix.cpp
bewithgaurav: Done

mssql_python/connection_mac.py
bewithgaurav: Fixed in the new push

gargsaumya (approved):
Though this is high LOC PR, but majorly contains the codebase already reviewed and running for Windows OS, there were no major concerns except the security concern highlighted by Sumit, hence approving the PR.

mssql_python/connection_mac.py
bewithgaurav: Acknowledged, this is definitely in our upcoming research points and we've created a task and referenced this comment to track and check upon this as p0 in our upcoming release cycle, Task:
https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/37537
Resolving the comment to unblock the public preview. We will start a separate discussion once we start working on this task.

PR #62, opened 2025-05-29
FEAT: Python support for Pooling

gargsaumya:
This pull request introduces connection pooling functionality to the `mssql_python` library. The changes include adding a new module to manage pooling configuration, modifying existing connection logic to support pooling, and extending the C++ bindings to handle the new pooling parameter.

### Connection pooling feature:

* [`mssql_python/pool_config.py`](diffhunk://#diff-de1c56e9ee3454304c984c5590b878a0548851323c72dcb3a8d2f14ccc46532cR1-R27): Added a new module to manage connection pooling, including a global `_pooling_enabled` flag and an `enable_pooling` function to configure pooling parameters.

### Integration of pooling into Python connection logic:

* [`mssql_python/connection.py`](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90L60-R61): Updated the `Connection` class to include the `_pooling_enabled` flag when creating a connection, ensuring pooling is respected in subsequent operations.
* [`mssql_python/__init__.py`](diffhunk://#diff-d95f3a67986de29f30453416b1b4c34e6a43207e9a33e2b1b80ef0c378b0a538R41): Imported the `enable_pooling` function from the new `pool_config` module to make it accessible at the package level.

### Extension of C++ bindings for pooling:

* `mssql_python/pybind/connection/connection.cpp` and `connection.h`: Modified the `Connection` constructor to accept a new `use_pooling` parameter and store it as `_usePool`. [[1]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecL19-R20) [[2]](diffhunk://#diff-c74277ea1f36fff8f6bbe9df3fd722f7d270f3d6a3069ac7a747fe6d4235cf13L16-R16)
* [`mssql_python/pybind/ddbc_bindings.cpp`](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1923-R1925): Updated the Pybind11 bindings for the `Connection` class to include the `use_pooling` parameter, allowing Python code to pass pooling configuration to the underlying C++ implementation.

### Checklist  
- [x] **Tests Passed** (if applicable) : All pytests are passing


mssql_python/pool_config.py
```diff
@@ -0,0 +1,27 @@
+"""
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+This module provides a way to create a new connection object to interact with the database.
+"""
+from mssql_python import ddbc_bindings
+
+# Global state: to track if pooling is enabled
+_pooling_enabled = False
```
sumitmsft: While using a global variable is functional for the current scope, for future maintainability, testability, and potential extensibility of the pooling feature, it might be beneficial to consider encapsulating the pooling state and its related configuration within a dedicated class (e.g., a `PoolingManager`).

I can think of three major benefits, we can discuss if required:

1. A class would group pooling-related data (like `is_enabled`, `max_size`, `idle_timeout`) and the logic to manage it (e.g., an `enable()` method) together. This improves organization and makes it clearer how the pooling state is managed.
2. It's generally easier to test components in isolation when they don't rely on mutable global state. An instance of a `PoolingManager` could be created or mocked per test.
3. If the library ever needs to support more advanced pooling configurations (e.g., multiple named pools, or dynamically changing settings beyond a one-time enable), a class-based approach provides a more flexible foundation.

This is just a suggestion for future-proofing and adhering to common design patterns that tend to scale well. The current implementation is clear for its immediate purpose.

If this change is feasible right now (for PuPr), I'd recommend to do it otherwise we can park it in backlog.

mssql_python/pool_config.py
```diff
@@ -0,0 +1,27 @@
...
+
+def enable_pooling(max_size=100, idle_timeout=600):
```
sumitmsft: **Looking at the `enable_pooling` function (lines 11-27)**

The current logic, specifically the check `if _pooling_enabled: return` on line 20, means that pooling parameters (`max_size`, `idle_timeout`) are effectively configured only by the first call to this function. Subsequent calls will return without applying any new parameters.

This behavior implies a one-time setup for connection pooling.

Could you clarify if this one-time configuration is the intended design?

If this is intended as a one-time setup: The current logic is fine. It might be beneficial to make this explicit in the function's docstring to manage user expectations. For example, something like: "Enables connection pooling. This function configures pooling parameters on its first successful invocation. Subsequent calls will have no effect."

If re-configuration is a desired feature (allowing users to change `max_size` or `idle_timeout` after initial enablement), the logic at line 20 would need to be adjusted, and the underlying (future) C++ binding `ddbc_bindings.enable_pooling` would also need to support updates to these parameters on an already active pool.

Understanding the intended behavior here will help ensure the feature aligns with user expectations and future development.

sumitmsft: Theoretically, as I understand if multiple threads call enable_pooling() around the same time before pooling is first enabled, there's a chance they might both try to initialize the C++ pooling mechanism (line 26: ddbc_bindings.enable_pooling()) concurrently.

Concern:

If the C++ enable_pooling function isn't designed to handle simultaneous setups from multiple threads, it could lead to issues. The Python _pooling_enabled flag (lines 20 & 27) helps, but there's a small window where multiple threads could pass the check before the flag is set.

Recommendation:

When implementing the C++ ddbc_bindings.enable_pooling(), please ensure it's safe to be called by multiple threads at once (e.g., it's idempotent or uses internal locks). Alternatively, a threading.Lock in the Python enable_pooling function could guard the initialization call.

sumitmsft (commented):
I have added a few comments to keep a note of. Please see if you can incorporate them when you make changes to ddbc_bindings.cpp.

mssql_python/pool_config.py
gargsaumya: Thanks for the suggestion @sumitmsft . 
For the current scope, we've intentionally kept the pooling configuration as a global, one-time setup, consistent with how it's typically handled in other drivers such as pyodbc. 
However, we can consider enhancing the API in future releases to support per-pool configuration and dynamic reconfiguration, as the codebase and user needs mature.

mssql_python/pool_config.py
gargsaumya: Ack.

mssql_python/pooling.py
```diff
@@ -0,0 +1,29 @@
...
+class PoolingManager:
+    _enabled = False
+    _lock = threading.Lock()
+    _config = {
+        "max_size": 100,
+        "idle_timeout": 600
+    }
+
+    @classmethod
+    def enable(cls, max_size=100, idle_timeout=600):
```
bewithgaurav, line 14: improvement suggestion: can define 100 & 600 under a new class in constants.py (something like ConnectionPoolConstants and name them MAX_POOL_SIZE and IDLE_TIMEOUT) and use it across the module

mssql_python/__init__.py
```diff
@@ -53,3 +53,18 @@
 apilevel = "2.0"
 paramstyle = "qmark"
 threadsafety = 1
+
+from .pooling import PoolingManager
+def pooling(max_size=100, idle_timeout=600):
```
bewithgaurav, line 58: same

mssql_python/pooling.py
```diff
@@ -0,0 +1,29 @@
+# mssql_python/pooling.py
+from mssql_python import ddbc_bindings
+import threading
+
+class PoolingManager:
+    _enabled = False
+    _lock = threading.Lock()
+    _config = {
+        "max_size": 100,
```
bewithgaurav, line 9: same

bewithgaurav (approved):
Approved with some minor suggestions for refactoring, can be added as a TODO for later use.

PR #64, opened 2025-05-29
FEAT: C++ support for pooling

gargsaumya:
This pull request introduces connection pooling to improve database connection management in the `mssql_python` project. Key changes include the addition of a `ConnectionPool` class, updates to the `Connection` class to support pooling, and modifications to the Python bindings to expose pooling functionality.

### Connection Pooling Implementation:

* **New `ConnectionPool` and `ConnectionPoolManager` classes**: Added `ConnectionPool` for managing reusable database connections and `ConnectionPoolManager` as a singleton to manage multiple pools. These classes handle connection acquisition, release, and idle timeout logic. (`mssql_python/pybind/connection/connection_pool.cpp` - [[1]](diffhunk://#diff-35a698bc3405cc30c1bf656066bd0575302239b5e39083a8160a6202026ca360R1-R85) `mssql_python/pybind/connection/connection_pool.h` - [[2]](diffhunk://#diff-396cb5eb8f38803cad4ec84bad73aa55c78a528275be99cb1f710aace9593f5eR1-R64)

### Updates to `Connection` Class:

* **Support for connection pooling**: Refactored `Connection` class to include attributes and methods for pooling, such as `_lastUsed`, `isAlive()`, and `reset()`. Added `ConnectionHandle` as a wrapper for pooled connections. (`mssql_python/pybind/connection/connection.cpp` - [[1]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecL19-R21) [[2]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR194-R288) [[3]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR96) [[4]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR106) [[5]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR139) `mssql_python/pybind/connection/connection.h` - [[6]](diffhunk://#diff-c74277ea1f36fff8f6bbe9df3fd722f7d270f3d6a3069ac7a747fe6d4235cf13R36-R39) [[7]](diffhunk://#diff-c74277ea1f36fff8f6bbe9df3fd722f7d270f3d6a3069ac7a747fe6d4235cf13L47-R72)

### Python Bindings:

* **Exposed pooling functionality**: Updated Python bindings to replace the `Connection` class with `ConnectionHandle` and added a global `enable_pooling` function for configuring pooling parameters. (`mssql_python/pybind/ddbc_bindings.cpp` - [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1880-R1883) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1922-R1935)

### Build System:

* **Updated build configuration**: Modified `CMakeLists.txt` to include the new `connection_pool.cpp` file in the build process. (`mssql_python/pybind/CMakeLists.txt` - [mssql_python/pybind/CMakeLists.txtL93-R93](diffhunk://#diff-dbb5892fbbb28149d1639664797cf3adb48ced28ec11aba95f2e2b338ca46badL93-R93))

### Header File Updates:

* **Included new headers**: Added `connection_pool.h` to relevant files for access to pooling functionality. (`mssql_python/pybind/connection/connection.cpp` - [[1]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR8) `mssql_python/pybind/ddbc_bindings.cpp` - [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R8)
### Checklist  
- [ ] **Tests Passed** (if applicable)


mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -0,0 +1,85 @@
...
+            }
+            return conn;
+        } else {
+            conn->disconnect();
+            --_current_size;
+        }
+    }
+    if (_current_size < _max_size) {
+        auto conn = std::make_shared<Connection>(connStr, true);
+        conn->connect(attrs_before);
```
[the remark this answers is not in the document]
bewithgaurav: Wanted to double check on this.. do we not need to increment current size var here?

mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -0,0 +1,91 @@
...
+        if (idle_time > _idle_timeout_secs) {
+            conn->disconnect();
+            return true;
+        }
+        return false;
+    }), _pool.end());
+    size_t pruned = before - _pool.size();
+    _current_size = (_current_size >= pruned) ? (_current_size - pruned) : 0;
+
+    while (!_pool.empty()) {
```
bewithgaurav: Since the acquire and release logic is very intricate, could you please put detailed comments? This will help understand the code better..

mssql_python/pybind/connection/connection_pool.h
```diff
@@ -0,0 +1,64 @@
...
+    // Acquires a connection from the pool or creates a new one if under limit
+    std::shared_ptr<Connection> acquire(const std::wstring& connStr, const py::dict& attrs_before = py::dict());
+
+    // Returns a connection to the pool for reuse
+    void release(std::shared_ptr<Connection> conn);
+
+private:
+    size_t _max_size;       // Maximum number of connections allowed
+    int _idle_timeout_secs; // Idle time before connections are considered stale
+    size_t _current_size = 0;
```
bewithgaurav, line 30: I have a suggestion to eliminate _current_size altogether, I believe since we are using deque, we shouldn't need any other var to store size since _pool.size() at any point of time should return same as _current_size, and the onus to update it will be not ours.

bewithgaurav (changes requested):

sumitmsft (approved):

mssql_python/pybind/connection/connection_pool.h
gargsaumya: If we remove _current_size and rely only on _pool.size(), our acquire() logic will be incorrect. Because once a connection is checked out from the pool, it's not in _pool anymore, so _pool.size() would be lower and our logic would over-allocate new connections, violating the max limit. We are not looking at just idle connections but also the in-use connections through this, that's why we need it.

bewithgaurav: got it.. resolving this, thanks

mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -0,0 +1,92 @@
...
+    : _max_size(max_size),  _idle_timeout_secs(idle_timeout_secs), _current_size(0) {}
+
+std::shared_ptr<Connection> ConnectionPool::acquire(const std::wstring& connStr, const py::dict& attrs_before) {
+    std::lock_guard<std::mutex> lock(_mutex);
+    auto now = std::chrono::steady_clock::now();
+    size_t before = _pool.size();
+    _pool.erase(std::remove_if(_pool.begin(), _pool.end(), [&](const std::shared_ptr<Connection>& conn) {
+        auto idle_time = std::chrono::duration_cast<std::chrono::seconds>(now - conn->lastUsed()).count();
+        if (idle_time > _idle_timeout_secs) {
+            conn->disconnect();
```
sumitmsft: You’re calling `conn->disconnect()` (and indirectly accessing `conn->lastUsed()`) inside the predicate of `std::remove_if` while holding `_mutex`. If `disconnect()` ever throws (or otherwise fails), the exception will escape the predicate and that mutex will never be unlocked—leading to a deadlock for any other thread trying to acquire or release connections.

I recommend splitting cleanup into two phases:
1. Under the lock, scan the pool and collect all stale `shared_ptr<Connection>` handles into a temporary list—but do not call `disconnect()` yet.
2. Unlock the mutex, then iterate that list and call disconnect() on each.

Alternatively, you could wrap the call to `disconnect()` in a try/catch inside the predicate, swallow or log errors, and return true to prune—but the two-phase approach ensures that disconnect logic runs outside the critical section. This will prevent any thrown exception from escaping while `_mutex` is held, and avoid blocking other threads for the duration of potentially slow I/O in `disconnect()`.

We can keep this in backlog and fix it later.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -185,4 +190,99 @@ void Connection::applyAttrsBefore(const py::dict& attrs) {
...
+
+void Connection::updateLastUsed() {
+    _lastUsed = std::chrono::steady_clock::now();
+}
+
+std::chrono::steady_clock::time_point Connection::lastUsed() const {
+    return _lastUsed;
+}
+
+ConnectionHandle::ConnectionHandle(const std::wstring& connStr, bool usePool, const py::dict& attrsBefore)
```
sumitmsft, line 240: Currently `ConnectionHandle` relies on callers to invoke `close()` explicitly. If a user neglects to call `close()`, then:

- Pooled connections are never returned to the pool (leaking active handles).
- Non-pooled connections are never disconnect()-ed.

I'd suggest implement a destructor that enforces RAII and always returns or disconnects the underlying connection.

```
// Add to ConnectionHandle in connection.cpp
ConnectionHandle::~ConnectionHandle() {
    if (_conn) {
        // this will either return it to the pool or disconnect it
        close();
    }
}
```

With this change, even if close() is forgotten, the ~ConnectionHandle() will ensure no connection is lost. 

Let me know if doesn't make sense.

mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+    if (_current_size < _max_size) {
+        auto conn = std::make_shared<Connection>(connStr, true);
+        conn->connect(attrs_before);
+        ++_current_size;
+        return conn;
+    } else {
+        LOG("Cannot acquire connection: pool size limit reached");
+        return nullptr;
```
sumitmsft: Right now, when the pool has reached its maximum size, you do:

```
LOG("Cannot acquire connection: pool size limit reached");
return nullptr;
```
Returning `nullptr` silently hands off the burden of null‐checking to every caller of `acquire()`. If a caller forgets to guard against a null shared_ptr, it "might" lead to a crash or use‐after‐null.

I'd suggest throwing an exception with a clear message (instead of silently returning `nullptr`) or block/wait on a condition_variable. 

1. Throw an exception instead of returning `nullptr`, so the error is explicit and cannot be ignored:
```
if (_current_size >= _max_size) {
    throw std::runtime_error("ConnectionPool::acquire: pool size limit reached");
}
```
OR

2. Block until a connection is released, using a `std::condition_variable` to wait rather than fail fast. That would look roughly like:
```
std::unique_lock<std::mutex> lock(_mutex);
_cond.wait(lock, [&]{ return _pool.size() > 0 || _current_size < _max_size; });
// then proceed to pop or create a connection

```
Either approach makes error handling more robust—either by surfacing an immediate exception or by naturally throttling threads to available pool capacity.

mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -0,0 +1,92 @@
...
+//             taken up in future.
+
+#include "connection_pool.h"
+#include <iostream>
+#include <exception>
+
+ConnectionPool::ConnectionPool(size_t max_size, int idle_timeout_secs)
+    : _max_size(max_size),  _idle_timeout_secs(idle_timeout_secs), _current_size(0) {}
+
+std::shared_ptr<Connection> ConnectionPool::acquire(const std::wstring& connStr, const py::dict& attrs_before) {
```
sumitmsft, line 14: In `ConnectionPool::acquire`, you prune idle connections using `std::remove_if` while holding `_mutex`, then call `conn->disconnect()` inside that predicate. Disconnecting an ODBC handle can be expensive, which:
- Holds the lock for the duration of the disconnect I/O, degrading concurrency under high contention
- Risks unlocking never occurring if `disconnect()` throws, leading to deadlocks

If we refactor the cleanup into two distinct phases, then the disconnect operations occur outside the critical section - minimizing lock hold time and preventing any exception from escaping while `_mutex` is held.

- Under the lock, scan `_pool`, move all stale `shared_ptr<Connection>` to a temporary `std::vector`, remove them from `_pool`, and adjust `_current_size`.
- After unlocking, iterate the temporary list and call `disconnect()` on each handle (wrapped in a try/catch if desired).

We can put this issue under backlog and fix it later.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -16,8 +17,8 @@ SqlHandlePtr Connection::_envHandle = nullptr;
 // This class wraps low-level ODBC operations like connect/disconnect,
 // transaction control, and autocommit configuration.
 //-------------------------------------------------------------------------------------------------
-Connection::Connection(const std::wstring& conn_str, bool autocommit, bool use_pooling)
-    : _connStr(conn_str) , _autocommit(autocommit), _usePool(use_pooling) {
+Connection::Connection(const std::wstring& conn_str, bool use_pool)
+    : _connStr(conn_str), _autocommit(false), _fromPool(use_pool) {
     if (!_envHandle) {
```
sumitmsft: The one-time initialization of the static `_envHandle` is guarded only by `if (!_envHandle)` and is not thread-safe. If two threads enter the constructor concurrently, both may see `_envHandle == nullptr` and allocate separate environment handles, violating the requirement for a single shared handle and potentially leaking resources.

Can we plan to use a dedicated mutex to ensure the initialization block runs exactly once, even under concurrent construction? 
This would guarantee that only one thread performs the allocation and attribute setup, eliminating the race and ensuring correct, single-initialize semantics.

This can be put into backlog if turns out complex.

mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -0,0 +1,92 @@
...
+        return false;
+    }), _pool.end());
+    size_t pruned = before - _pool.size();
+    _current_size = (_current_size >= pruned) ? (_current_size - pruned) : 0;
+
+    while (!_pool.empty()) {
+        auto conn = _pool.front();
+        _pool.pop_front();
+        if (conn->isAlive()) {
+            if (!conn->reset()) {
```
sumitmsft: Inside the loop you have:

```
if (!conn->reset()) {
    continue;
}
return conn;
```

When `reset()` fails, you simply skip that handle without calling `disconnect()` or adjusting `_current_size`. That abandoned handle remains live, and `_current_size` stays inflated—eventually starving the pool of valid connections.

I think if on `reset()` failure we explicitly clean up that handle and decrement the pool count will ensure failed or dead connections are removed and resources freed, keeping `_current_size` accurate and preventing depletion of healthy connections.

sumitmsft (changes requested):
I have a few comments. Some of them can be put into backlog for future improvements. A couple of them can be thought to get fixed right away.

mssql_python/pybind/connection/connection.cpp
gargsaumya: Makes sense, we 're handling it for conn object but we should handle it here too. Ack.

mssql_python/pybind/connection/connection.cpp
gargsaumya: Ack.

mssql_python/pybind/connection/connection_pool.cpp
gargsaumya: Ack.

mssql_python/pybind/connection/connection_pool.cpp
gargsaumya: Ack.

mssql_python/pybind/connection/connection_pool.cpp
gargsaumya: Ack.

mssql_python/pybind/connection/connection.cpp
sumitmsft: Hi @gargsaumya Could you please validate if the issue is fixed from race condition pov? Seems the static _envHandle is still vulnerable to concurrent initialization by multiple threads. What I meant was, we can use a dedicated mutex or one-time initialization primitive so only one thread can enter the initialization block. This will make sure we avoid race. Please validate this.

mssql_python/pybind/connection/connection_pool.cpp
sumitmsft: On this, there is no indication that failed connections are cleaned up or that _current_size is decremented when reset() fails. Could you please validate this? I fear, that if reset() fails, the pool will leak handles and become depleted over time.

gargsaumya: @sumitmsft I think you may have reviewed an older version of this file. In the current implementation, I've already handled the reset failure case as per your previous suggestion: failed connections are added to a disconnect list, _current_size is decremented, and cleanup happens outside the lock. Could you please take another look and confirm?

sumitmsft: My bad! Yes you're correct. This issue was fixed.

mssql_python/pybind/connection/connection.cpp
gargsaumya: Thanks. I've updated the _envHandle initialization by moving it into a dedicated getEnvHandle() function that uses a function-local static variable. This pattern ensures that the environment handle is created only once and in a thread-safe manner.

PR #75, opened 2025-06-11
FIX: Row Object support

jahnvi480:
[AB#37746](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/37746)
This pull request introduces several enhancements to the database interaction logic and improves the testing framework for SQL data types. The most significant changes include adding support for named tuples in `fetchone` results, improving error handling in the C++ bindings, and refactoring test cases for better cleanup and assertion practices.

### Enhancements to `fetchone` functionality:
* [`mssql_python/cursor.py`](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L651-R689): Modified the `fetchone` method to return a named tuple when valid field names are available, enabling attribute-based access to query results. Falls back to a regular tuple if named tuple creation fails.
* [`mssql_python/cursor.py`](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R9): Added `collections` import to support named tuple creation.

### Improvements to C++ bindings:
* [`mssql_python/pybind/ddbc_bindings.cpp`](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1810-R1852): Updated `FetchOne_wrap` to handle named tuples on the Python side, added error handling for column count retrieval, and improved logging for debugging.

### Refactoring and cleanup in test cases:
* [`tests/test_004_cursor.py`](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69R370-R372): Refactored multiple test cases to use the `drop_table_if_exists` helper function for cleanup, ensuring tables are dropped before and after tests. [[1]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69R370-R372) [[2]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L393-R400) [[3]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L476-R495)
* [`tests/test_004_cursor.py`](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L383-R390): Replaced direct tuple comparisons in assertions with the `compare_row_value` helper function for improved readability and flexibility. [[1]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L383-R390) [[2]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L437-R460) [[3]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L466-R477)

### Debugging and logging enhancements:
* [`main.py`](diffhunk://#diff-b10564ab7d2c520cdd0243874879fb0a782862c3c902ab535faabe57d5a505e1L6-R32): Added debug statements to print the type and content of rows, and safely handle cases where `rows` is `None`.

### DIFF
--- a/mssql_python/row.py
+++ b/mssql_python/row.py
@@ -18,7 +18,12 @@ class Row:
             cursor_description: The cursor description containing column metadata
         """
         self._values = values
-        self.cursor_description = cursor_description
+        
+        # TODO: ADO task - Optimize memory usage by sharing column map across rows
+        # Instead of storing the full cursor_description in each Row object:
+        # 1. Build the column map once at the cursor level after setting description
+        # 2. Pass only this map to each Row instance
+        # 3. Remove cursor_description from Row objects entirely
         
         # Create mapping of column names to indices
         self._column_map = {}
@@ -43,6 +48,8 @@ class Row:
         """
         if isinstance(other, list):
             return self._values == other
+        elif isinstance(other, Row):
+            return self._values == other._values
         return super().__eq__(other)
     
     def __len__(self):
--- a/tests/test_004_cursor.py
+++ b/tests/test_004_cursor.py
@@ -1145,6 +1145,101 @@ def test_numeric_precision_scale_negative_exponent(cursor, db_connection):
         cursor.execute("DROP TABLE pytest_numeric_test")
         db_connection.commit()
 
+def test_row_attribute_access(cursor, db_connection):
+    """Test accessing row values by column name as attributes"""
+    try:
+        # Create test table with multiple columns
+        cursor.execute("""
+            CREATE TABLE pytest_row_attr_test (
+                id INT PRIMARY KEY,
+                name VARCHAR(50),
+                email VARCHAR(100),
+                age INT
+            )
+        """)
+        db_connection.commit()
+        
+        # Insert test data
+        cursor.execute("""
+            INSERT INTO pytest_row_attr_test (id, name, email, age)
+            VALUES (1, 'John Doe', 'john@example.com', 30)
+        """)
+        db_connection.commit()
+        
+        # Test attribute access
+        cursor.execute("SELECT * FROM pytest_row_attr_test")
+        row = cursor.fetchone()
+        
+        # Access by attribute
+        assert row.id == 1, "Failed to access 'id' by attribute"
+        assert row.name == 'John Doe', "Failed to access 'name' by attribute"
+        assert row.email == 'john@example.com', "Failed to access 'email' by attribute"
+        assert row.age == 30, "Failed to access 'age' by attribute"
+        
+        # Compare attribute access with index access
+        assert row.id == row[0], "Attribute access for 'id' doesn't match index access"
+        assert row.name == row[1], "Attribute access for 'name' doesn't match index access"
+        assert row.email == row[2], "Attribute access for 'email' doesn't match index access"
+        assert row.age == row[3], "Attribute access for 'age' doesn't match index access"
+        
+        # Test attribute that doesn't exist
+        with pytest.raises(AttributeError):
+            value = row.nonexistent_column
+            
+    except Exception as e:
+        pytest.fail(f"Row attribute access test failed: {e}")
+    finally:
+        cursor.execute("DROP TABLE pytest_row_attr_test")
+        db_connection.commit()
+
+def test_row_comparison_with_list(cursor, db_connection):
+    """Test comparing Row objects with lists (__eq__ method)"""
+    try:
+        # Create test table
+        cursor.execute("CREATE TABLE pytest_row_comparison_test (col1 INT, col2 VARCHAR(20), col3 FLOAT)")
+        db_connection.commit()
+        
+        # Insert test data
+        cursor.execute("INSERT INTO pytest_row_comparison_test VALUES (10, 'test_string', 3.14)")
+        db_connection.commit()
+        
+        # Test fetchone comparison with list
+        cursor.execute("SELECT * FROM pytest_row_comparison_test")
+        row = cursor.fetchone()
+        assert row == [10, 'test_string', 3.14], "Row did not compare equal to matching list"
+        assert row != [10, 'different', 3.14], "Row compared equal to non-matching list"
+        
+        # Test full row equality
+        cursor.execute("SELECT * FROM pytest_row_comparison_test")
+        row1 = cursor.fetchone()
+        cursor.execute("SELECT * FROM pytest_row_comparison_test")
+        row2 = cursor.fetchone()
+        assert row1 == row2, "Identical rows should be equal"
+        
+        # Insert different data
+        cursor.execute("INSERT INTO pytest_row_comparison_test VALUES (20, 'other_string', 2.71)")
+        db_connection.commit()
+        
+        # Test different rows are not equal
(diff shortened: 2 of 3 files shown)


mssql_python/cursor.py
```diff
@@ -648,63 +649,117 @@ def fetchone(self) -> Union[None, tuple]:
         Fetch the next row of a query result set.
```
sumitmsft, line 650: I noticed that while the fetch methods (`fetchone`, `fetchmany`, `fetchall`) now support returning named tuples for valid column names (which is great for usability), the docstrings could be improved to clarify this new behavior.

**Suggestion:**  
Could you please update the docstrings for these methods to clearly describe:
- That they return a named tuple if valid field names are available, otherwise they fall back to a regular tuple.
- What "valid field names" means (i.e., names must be valid Python identifiers).
- How users can access row data in each case (e.g., by attribute or by index).

This will help users understand what to expect from these methods and make the API easier to use.

mssql_python/cursor.py
```diff
@@ -648,63 +649,117 @@ def fetchone(self) -> Union[None, tuple]:
...
+        # Get field names from the description attribute
+        field_names = [desc[0] for desc in self.description]
+        
+        # Check if field names are valid for namedtuple
+        valid_fieldnames = all(isinstance(name, str) and name.isidentifier() for name in field_names)
+        
+        if valid_fieldnames:
+            # Create a namedtuple on the Python side
+            try:
+                RowRecord = collections.namedtuple('RowRecord', field_names, rename=True)
```
sumitmsft: Currently, the named tuple class (`RowRecord`) is created every time `fetchone`, `fetchmany`, or `fetchall` is called, even if the schema (column names) hasn’t changed. For queries that fetch many rows from the same result set, this could introduce unnecessary overhead.

**Suggestion:** 
Cache the named tuple class on the cursor instance keyed by the current schema (e.g., the column names from `self.description`). You could set/reset a private attribute (like `_row_namedtuple_class`) when a new result set is prepared, and reuse it for subsequent rows of the same query.

This will avoid repeated creation of identical named tuple classes and improve performance, especially in loops or bulk fetches.

mssql_python/cursor.py
```diff
@@ -648,63 +649,117 @@ def fetchone(self) -> Union[None, tuple]:
...
+                result = RowRecord(*row_list)
+                return result
+            except (TypeError, ValueError) as e:
+                # Fall back to a regular tuple for any error in namedtuple creation
```
sumitmsft: If named tuple creation fails (due to invalid field names or an exception), the code silently falls back to returning a regular tuple. This can make it harder to debug issues if users expect attribute access but get a tuple instead.

Suggestion:
Consider adding a debug-level log message (or at least a comment) in the except block where the fallback occurs. This will help with diagnosing issues during development and understanding why named tuple creation failed.

sumitmsft (changes requested):
The PR follows good practices, is secure, and robustly upgrades the user-facing API. With minor refactoring for efficiency and logging, it would be even better. No memory leaks, deadlocks, or security issues identified. Logic is correct, and changes are well-tested.

bewithgaurav (changes requested):
As discussed, it would be cleaner and more future-proof to define a dedicated Row class instead of relying on namedtuple generation at runtime. By having a separate Row class, we can:

**Encapsulate row behavior:** A custom class allows us to centralize logic for attribute and index access, custom formatting, and future enhancements (like dict-like access, type conversions, or richer error handling).
**Improve maintainability:** Making changes or adding features to row handling (e.g., supporting column name mapping, case-insensitivity, or value transformations) becomes much easier in a single class, without having to manage the quirks of namedtuple limitations.
**Enable richer row features:** A dedicated class can provide context-aware methods, like .as_dict(), or even hooks for validation, which would be awkward with namedtuples.

Could we refactor this so that fetchone, fetchmany, and fetchall return instances of a Row class, instead of namedtuples or plain tuples? This would keep the codebase cleaner and also pave the way for any future features related to row handling.

tests/test_004_cursor.py
```diff
@@ -1155,3 +1155,4 @@ def test_close(db_connection):
         pytest.fail(f"Cursor close test failed: {e}")
     finally:
         cursor = db_connection.cursor()
+
```
bewithgaurav, line 1253: Add tests for these scenarios:
- Accessing rows using column_name as attr (testing `__getattr__`)
- comparing with lists? (for changes in `__eq__`) - it might already be there but double check

mssql_python/row.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, values, cursor_description):
+        """
+        Initialize a Row object with values and cursor description.
+        
+        Args:
+            values: List of values for this row
+            cursor_description: The cursor description containing column metadata
+        """
+        self._values = values
+        self.cursor_description = cursor_description
```
bewithgaurav: I see here that each Row currently stores its own copy of the full description, which in turn increases memory usage. Instead, we need to build a single column name-to-index map after setting cursor.description, store it on the cursor, and pass this map to each Row. This way, all rows share the same metadata reference, greatly reducing per-row memory consumption for large result sets.
This is not a p0, you can probably entirely skip `cursor_description` in Row class for now, and add an ADO task and a TODO comment since this PR is already of a sufficient size.

bewithgaurav (approved):

mssql_python/row.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Initialize a Row object with values and cursor description.
+        
+        Args:
+            values: List of values for this row
+            cursor_description: The cursor description containing column metadata
+        """
+        self._values = values
+        
+        # TODO: ADO task - Optimize memory usage by sharing column map across rows
```
sumitmsft, line 22: Please create a workitem to track this TODO item.

mssql_python/row.py
```diff
@@ -0,0 +1,65 @@
+class Row:
```
sumitmsft, line 1: In the below code we are using dictionary which is dynamically created and allows you to add attributes to an object at runtime. However, for classes where the attributes are known in advance and won't change, this dynamic dictionary can be inefficient in terms of memory usage.

`__slots__` is a class-level attribute that you can define to explicitly declare the attributes that an instance of the class will have. By defining `__slots__`, you tell Python not to create the `__dict__` for instances of that class. Instead, Python allocates space for the attributes listed in `__slots__`.  This will further help in reduced memory footprint: This is the primary benefit. This will be helpful especially when you're creating a large number of instances, which is possible considering many users can invoke the instance of the same class concurrently. 

    ```python
    class Row:
        __slots__ = ('_values', '_column_map')
        # ... rest of the class definition ...
    ```
Using `__slots__` can provide a noticeable memory saving, which is crucial for a Python driver handling database results.

You can put this in backlog, create a workitem and enhance the code when you work on the below TODO item.

PR #77, opened 2025-06-11
FIX: validate numeric data for range (upper and lower bound)

gargsaumya:
This pull request enhances the `BindParameters` function in `mssql_python/pybind/ddbc_bindings.cpp` by introducing range validation for various parameter types to ensure data integrity and prevent out-of-range errors. Below are the key changes grouped by parameter type:

### Integer Parameter Validations

* **Signed 16-bit integer (`short`)**: Added range validation to ensure the value is within the limits of a signed 16-bit integer. Throws an exception if the value is out of range.
* **Unsigned 16-bit integer (`unsigned short`)**: Added validation to check that the value does not exceed the maximum limit of an unsigned 16-bit integer.
* **Signed 64-bit integer (`int64_t`)**: Introduced range validation to confirm the value is within the bounds of a signed 64-bit integer.
* **Unsigned 64-bit integer (`uint64_t`)**: Added a check to ensure the value does not exceed the maximum limit for an unsigned 64-bit integer.

### Date Parameter Validation

* **SQL Server Date Range**: Implemented validation for `year` attributes in date parameters to ensure they fall within the valid SQL Server date range (1753-9999).

### Checklist  
- [ ] **Tests Passed** (if applicable)  : All pytests passed.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -265,6 +265,11 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                 if (!py::isinstance<py::int_>(param)) {
                     ThrowStdException(MakeParamMismatchErrorStr(paramInfo.paramCType, paramIndex));
                 }
+                int value = param.cast<int>();
+                // Range validation for signed 16-bit integer
+                if (value < std::numeric_limits<short>::min() || value > std::numeric_limits<short>::max()) {
+                    ThrowStdException("Signed short integer parameter out of range at paramIndex " + std::to_string(paramIndex));
+                }
                 dataPtr =
                     static_cast<void*>(AllocateParamBuffer<int>(paramBuffers, param.cast<int>()));
                 break;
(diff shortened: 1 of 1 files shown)


sumitmsft (approved):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -265,6 +265,11 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                 if (!py::isinstance<py::int_>(param)) {
                     ThrowStdException(MakeParamMismatchErrorStr(paramInfo.paramCType, paramIndex));
                 }
+                int value = param.cast<int>();
+                // Range validation for signed 16-bit integer
+                if (value < std::numeric_limits<short>::min() || value > std::numeric_limits<short>::max()) {
```
sumitmsft, line 270: In the current implementation, range-checking logic for various integer types and date values is repeated multiple times with only minor variations (type and message). We can factor out this repeated logic into a reusable template helper function. Not necessary but could be done. Good to have, not must have.

gargsaumya: sure, we'll add a separate reusable function for this. Will take it up in a new PR later, resolving this for now as it is not a blocker.

bewithgaurav (approved):

PR #108, opened 2025-06-25
FEAT: Bulk Copy Wrapper 

jahnvi480:
This pull request introduces support for Bulk Copy Program (BCP) functionality in the `mssql_python` project. The changes include adding a new `BCPWrapper` class, integrating BCP APIs, and modifying existing classes to support BCP operations. Below is a summary of the most important changes:

### Integration of BCP Functionality

* **New `BCPWrapper` Class**: Added `BCPWrapper` in `mssql_python/pybind/bcp/bcp_wrapper.{cpp,h}` to encapsulate BCP operations such as initialization, control, column formatting, and execution. This class interfaces with ODBC BCP APIs and handles diagnostic messages for error reporting.
* **BCP API Pointers**: Declared and initialized function pointers for ODBC BCP APIs (`BCPInitW`, `BCPControlW`, `BCPReadFmtW`, etc.) in `ddbc_bindings.cpp`. These are used by `BCPWrapper` to interact with BCP operations.

### Modifications to Support BCP

* **CMake Configuration**: Updated `CMakeLists.txt` to include the new `bcp_wrapper.cpp` file and its corresponding header directory. This ensures the new functionality is compiled and linked.

* **Connection Class Enhancements**: Added methods to retrieve the database connection handle (`getDbcHandle`) and the `Connection` object (`getConnection`) in `connection.h`. These are required by `BCPWrapper` to access the underlying ODBC connection handle.

* **BCP Attribute Setting**: Modified `applyAttrsBefore` in `connection.cpp` to support setting the `SQL_COPT_SS_BCP` attribute before establishing a connection. This enables bulk copy mode for connections.

### General Updates

* **Header Inclusion**: Added `bcp_wrapper.h` to `ddbc_bindings.cpp` to ensure the new `BCPWrapper` class and BCP-related functionality are accessible.

These changes collectively add robust support for BCP operations, enabling efficient bulk data transfer in the `mssql_python` library.

### ADO
* [USER STORY 34946](https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/34946)

### DIFF
--- a/mssql_python/pybind/bcp/bcp_wrapper.cpp
+++ b/mssql_python/pybind/bcp/bcp_wrapper.cpp
@@ -1,0 +1,361 @@
+#include "bcp_wrapper.h" // Includes ddbc_bindings.h (and thus sql.h, sqlext.h, <string>, <memory>) and connection.h
+
+// Pybind11 headers (needed for py::cast and potentially std::optional if used with pybind types)
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h> 
+#include <pybind11/pytypes.h> // Ensure this is included for py::bytes if not transitively
+
+// Standard C++ headers (not covered by ddbc_bindings.h)
+#include <vector>
+#include <stdexcept>
+#include <unordered_map>
+#include <string> // For std::to_string, std::string
+#include <iostream> // Added for std::cout
+
+namespace py = pybind11; // Alias for pybind11 namespace
+
+// Helper to manage BCP properties for bcp_control
+enum class BCPCtrlPropType { INT, WSTRING };
+struct BCPCtrlPropertyInfo {
+    INT option_code;
+    BCPCtrlPropType type;
+};
+
+// Map property names to their ODBC codes and types
+const std::unordered_map<std::wstring, BCPCtrlPropertyInfo> bcp_control_properties = {
+    {L"BCPMAXERRS",     {BCPMAXERRS,        BCPCtrlPropType::INT}},
+    {L"BCPBATCH",       {BCPBATCH,          BCPCtrlPropType::INT}},
+    {L"BCPKEEPNULLS",   {BCPKEEPNULLS,      BCPCtrlPropType::INT}},
+    {L"BCPKEEPIDENTITY",{BCPKEEPIDENTITY,   BCPCtrlPropType::INT}},
+    {L"BCPHINTS",       {BCPHINTS,          BCPCtrlPropType::WSTRING}},
+    {L"BCPFILECP",      {BCPFILECP,         BCPCtrlPropType::INT}},
+    // Example if you were to add global terminators via bcp_control:
+    // {L"BCPFIELDTERM",   {BCPFIELDTERM_Constant, BCPCtrlPropType::BYTES}}, // Assuming BCPFIELDTERM_Constant is the ODBC int
+    // {L"BCPROWTERM",     {BCPROWTERM_Constant,   BCPCtrlPropType::BYTES}}, // Assuming BCPROWTERM_Constant is the ODBC int
+};
+
+// Helper for bcp_init direction string
+INT get_bcp_direction_code(const std::wstring& direction_str) {
+    if (direction_str == L"in") return DB_IN;
+    if (direction_str == L"out" || direction_str == L"queryout") return DB_OUT;
+    throw std::runtime_error("Invalid BCP direction string: " + py::cast(direction_str).cast<std::string>());
+}
+
+// Helper function (can be a static private method or a lambda in C++11 and later)
+// to retrieve ODBC diagnostic messages and populate ErrorInfo.
+// This uses SQLGetDiagRec_ptr directly, which is loaded by DriverLoader.
+static ErrorInfo get_odbc_diagnostics_for_handle(SQLSMALLINT handle_type, SQLHANDLE handle) {
+    ErrorInfo error_info;
+    error_info.sqlState = L""; // Initialize
+    error_info.ddbcErrorMsg = L""; // Initialize
+
+    if (!SQLGetDiagRec_ptr) {
+        LOG("get_odbc_diagnostics_for_handle: SQLGetDiagRec_ptr is null.");
+        error_info.ddbcErrorMsg = L"SQLGetDiagRec_ptr not loaded. Cannot retrieve diagnostics.";
+        return error_info;
+    }
+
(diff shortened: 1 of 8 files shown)


mssql_python/pybind/bcp/bcp_wrapper.cpp
```diff
@@ -0,0 +1,361 @@
...
+            LOG("BCPWrapper Error: Invalid HDBC from Connection object.");
+            throw std::runtime_error("BCPWrapper: Invalid HDBC from Connection object.");
+        }
+    } catch (const std::runtime_error& e) {
+        // Re-throw with more context or just let it propagate
+        throw std::runtime_error(std::string("BCPWrapper Constructor: Failed to get valid HDBC - ") + e.what());
+    }
+}
+
+BCPWrapper::~BCPWrapper() {
```
Ramudaykumar, line 135: Destructor currently only logs. If possible, ensure that resources are reliably released (calling close() or finish() if not already done), possibly via RAII.

mssql_python/pybind/bcp/bcp_wrapper.cpp
```diff
@@ -0,0 +1,361 @@
+#include "bcp_wrapper.h" // Includes ddbc_bindings.h (and thus sql.h, sqlext.h, <string>, <memory>) and connection.h
```
Ramudaykumar: Comment Style Suggestions

- Start with a brief summary of what a block does, especially if it’s not immediately obvious.
- Use inline comments for tricky lines, and block comments for larger logic sections.
- If you’re handling edge cases (like empty strings, or non-ASCII data), call it out explicitly.
- For all conversions between types (especially between Python/C++ or encodings), note assumptions and possible gotchas for future readers.

mssql_python/pybind/bcp/bcp_wrapper.cpp
```diff
@@ -0,0 +1,361 @@
...
+        ErrorInfo diag = get_odbc_diagnostics_for_handle(SQL_HANDLE_DBC, _hdbc);
+        msg += " ODBC Diag: SQLState: " + py::cast(diag.sqlState).cast<std::string>() + ", Message: " + py::cast(diag.ddbcErrorMsg).cast<std::string>();
+        LOG(msg);
+        throw std::runtime_error(msg);
+    }
+    return ret;
+}
+
+SQLRETURN BCPWrapper::bcp_control(const std::wstring& property_name, const std::wstring& value) {
+    if (!_bcp_initialized || _bcp_finished) {
```
Ramudaykumar, line 254: a failed property lookup or type mismatch only logs and returns SQL_ERROR, does not throw. For consistency, consider always throwing on logic errors (as you do for the int version).

mssql_python/pybind/bcp/bcp_wrapper.cpp
```diff
@@ -0,0 +1,361 @@
...
+    if (it == bcp_control_properties.end() || it->second.type != BCPCtrlPropType::WSTRING) {
+        LOG("BCPWrapper Error: bcp_control(wstring) - property '" + py::cast(property_name).cast<std::string>() + "' not found or type mismatch.");
+        return SQL_ERROR; 
+    }
+    
+    LOG("BCPWrapper: Calling bcp_controlW for property '" + py::cast(property_name).cast<std::string>() + "' with wstring value '" + py::cast(value).cast<std::string>() + "'.");
+    
+    std::string value_utf8 = py::cast(value).cast<std::string>();
+    // Fix the BCPHINTS issue: use the wide string directly rather than converting to narrow
+    SQLRETURN ret = BCPControlW_ptr(_hdbc, it->second.option_code, (LPVOID)value_utf8.c_str());
```
Ramudaykumar, line 276: If ODBC expects UTF-16/SQLWCHAR* for a property like BCPHINTS, passing a UTF-8 narrow string pointer is incorrect, especially for non-ASCII data.
You should convert std::wstring to a buffer of SQLWCHAR (UTF-16), and pass that pointer instead.

std::vector<SQLWCHAR> wide_value(value.c_str(), value.c_str() + value.size() + 1);
SQLRETURN ret = BCPControlW_ptr(_hdbc, it->second.option_code, (LPVOID)wide_value.data());

The way wide-string values are currently handled for BCP control is not robust for non-ASCII input and will cause bugs in some scenarios.

mssql_python/pybind/bcp/bcp_wrapper.cpp
```diff
@@ -0,0 +1,361 @@
...
+    // Example if you were to add global terminators via bcp_control:
+    // {L"BCPFIELDTERM",   {BCPFIELDTERM_Constant, BCPCtrlPropType::BYTES}}, // Assuming BCPFIELDTERM_Constant is the ODBC int
+    // {L"BCPROWTERM",     {BCPROWTERM_Constant,   BCPCtrlPropType::BYTES}}, // Assuming BCPROWTERM_Constant is the ODBC int
+};
+
+// Helper for bcp_init direction string
+INT get_bcp_direction_code(const std::wstring& direction_str) {
+    if (direction_str == L"in") return DB_IN;
+    if (direction_str == L"out" || direction_str == L"queryout") return DB_OUT;
+    throw std::runtime_error("Invalid BCP direction string: " + py::cast(direction_str).cast<std::string>());
```
Ramudaykumar, line 45: Make the string matching case-insensitive for better robustness.

If user passes "IN" instead of "in" (capitalization), it will throw.

Ramudaykumar (changes requested):
There is no explicit thread safety; if the same BCPWrapper object is used from multiple threads, state flags and ODBC calls could race.

Some methods (e.g., bcp_control int overload) throw exceptions on error, others (e.g., bcp_control wstring overload, read_format_file) just return SQL_ERROR. For a pybind11-exposed wrapper, it's better to throw exceptions for all error states, so Python users get errors, not silent failures.

Recommendation: 
Add a note/documentation about thread-safety, or add a mutex if needed.
Always throw std::runtime_error on unrecoverable errors.

mssql_python/pybind/bcp/bcp_wrapper.cpp
jahnvi480: ODBC does not accept wide string value for BCPHINT

mssql_python/pybind/bcp/bcp_wrapper.cpp
jahnvi480: I'm converting the direction to lower in python.

PR #123, opened 2025-07-04
FIX: Ensure cursors are freed before connection (segfault) with added tests

bewithgaurav:
### ADO Work Item Reference  

> [AB#37863](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/37863)
### Summary   
  
This pull request introduces several updates to enhance connection management, improve cursor handling, and update test cases for better coverage and consistency. The most important changes include the implementation of weak references for cursor tracking, additional safeguards for connection cleanup, and modifications to test cases to align with schema conventions.

### Connection Management Enhancements:
* Replaced `ctypes` with `weakref` and introduced `WeakSet` for managing cursors to prevent memory leaks and ensure automatic cleanup of unused cursors. (`mssql_python/connection.py`) [[1]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90L8-R8) [[2]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R57-R69)
* Added logic to ensure all cursors are closed before closing the connection and made `close()` idempotent to handle multiple calls safely. (`mssql_python/connection.py`)

### Cursor Handling Improvements:
* Updated the `cursor()` method to raise an exception if called on a closed connection and track active cursors using `WeakSet`. (`mssql_python/connection.py`)

### Test Case Updates:
* Added new tests to validate cursor cleanup behavior, weak reference handling, cleanup order, and connection idempotency. (`tests/test_003_connection.py`)
* Updated stored procedure-related tests to use the `dbo` schema prefix for better compatibility and clarity. (`tests/test_004_cursor.py`) [[1]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L873-R879) [[2]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L895-R895) [[3]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L907-R907) [[4]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L918-R918)
* Modified table creation and cleanup logic in multiple test cases to use temporary tables (`#pytest_*`) for isolation and reduced side effects. (`tests/test_004_cursor.py`, `tests/test_005_exceptions.py`) [[1]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L1153-R1153) [[2]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L1164-R1170) [[3]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L1192-R1231) [[4]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L1240-R1240) [[5]](diffhunk://#diff-282f76cf72519247beed3fe56aa378a339af5ecbdef32fb468a3470969d15b69L88-R128)

### DIFF
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -4,6 +4,11 @@ Licensed under the MIT license.
 This module defines the Connection class, which is used to manage a connection to a database.
 The class provides methods to establish a connection, create cursors, commit transactions, 
 roll back transactions, and close the connection.
+Resource Management:
+- All cursors created from this connection are tracked internally.
+- When close() is called on the connection, all open cursors are automatically closed.
+- Do not use any cursor after the connection is closed; doing so will raise an exception.
+- Cursors are also cleaned up automatically when no longer referenced, to prevent memory leaks.
 """
 import weakref
 from mssql_python.cursor import Cursor
@@ -12,6 +17,7 @@ from mssql_python.constants import ConstantsDDBC as ddbc_sql_const
 from mssql_python.helpers import add_driver_to_connection_str, check_error
 from mssql_python import ddbc_bindings
 from mssql_python.pooling import PoolingManager
+from mssql_python.exceptions import DatabaseError, InterfaceError
 
 logger = get_logger()
 
@@ -64,6 +70,7 @@ class Connection:
         # It is a set that holds weak references to its elements.
         # When an object is only weakly referenced, it can be garbage collected even if it's still in the set.
         # It prevents memory leaks by ensuring that cursors are cleaned up when no longer in use without requiring explicit deletion.
+        # TODO: Think and implement scenarios for multi-threaded access to cursors
         self._cursors = weakref.WeakSet()
 
         # Auto-enable pooling if user never called
@@ -162,8 +169,9 @@ class Connection:
         """
         """Return a new Cursor object using the connection."""
         if self._closed:
-            raise Exception("Cannot create cursor on closed connection")
-        
+            # raise InterfaceError
+            raise InterfaceError("Cannot create cursor on closed connection")
+
         cursor = Cursor(self)
         self._cursors.add(cursor)  # Track the cursor
         return cursor
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -5,7 +5,7 @@ This module defines the Connection class, which is used to manage a connection t
 The class provides methods to establish a connection, create cursors, commit transactions, 
 roll back transactions, and close the connection.
 """
-import ctypes
+import weakref
 from mssql_python.cursor import Cursor
 from mssql_python.logging_config import get_logger, ENABLE_LOGGING
 from mssql_python.constants import ConstantsDDBC as ddbc_sql_const
@@ -54,10 +54,19 @@ class Connection:
         preparing it for further operations such as connecting to the 
         database, executing queries, etc.
         """
+        print("[SEGDEBUGGING - PYTHON - PYTHON] Connection.__init__ called")
         self.connection_str = self._construct_connection_string(
             connection_str, **kwargs
         )
         self._attrs_before = attrs_before or {}
+        self._closed = False
+        
+        # Using WeakSet which automatically removes cursors when they are no longer in use
+        # It is a set that holds weak references to its elements.
+        # When an object is only weakly referenced, it can be garbage collected even if it's still in the set.
+        # It prevents memory leaks by ensuring that cursors are cleaned up when no longer in use without requiring explicit deletion.
+        self._cursors = weakref.WeakSet()
+
         # Auto-enable pooling if user never called
         if not PoolingManager.is_initialized():
             PoolingManager.enable()
@@ -152,7 +161,13 @@ class Connection:
             DatabaseError: If there is an error while creating the cursor.
             InterfaceError: If there is an error related to the database interface.
         """
-        return Cursor(self)
+        """Return a new Cursor object using the connection."""
+        if self._closed:
+            raise Exception("Cannot create cursor on closed connection")
+        
+        cursor = Cursor(self)
+        self._cursors.add(cursor)  # Track the cursor
+        return cursor
 
     def commit(self) -> None:
         """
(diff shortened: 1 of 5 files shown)


tests/test_005_exceptions.py
```diff
@@ -85,46 +85,47 @@ def test_table_not_found_error(cursor):
...
+        cursor.execute("CREATE TABLE #pytest_test_truncation (id INT, name NVARCHAR(5))")
+        cursor.execute("INSERT INTO #pytest_test_truncation (id, name) VALUES (?, ?)", [1, 'TooLongName'])
     except (ProgrammingError, DataError) as excinfo:
         # DataError is raised on Windows but ProgrammingError on MacOS
         # Included catching both ProgrammingError and DataError in this test
         # TODO: Make this test platform independent
         assert "String or binary data would be truncated" in str(excinfo)
     finally:
-        drop_table_if_exists(cursor, "pytest_test_truncation")
+        drop_table_if_exists(cursor, "#pytest_test_truncation")
```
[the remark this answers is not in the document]
bewithgaurav: made changes in `drop_table_if_exists`

tests/test_004_cursor.py
```diff
@@ -892,7 +892,7 @@ def test_create_stored_procedure(cursor, db_connection):
 def test_execute_stored_procedure_with_parameters(cursor):
     """Test executing stored procedure with parameters"""
     try:
-        cursor.execute("{CALL GetEmployeeProjects(?)}", [1])
+        cursor.execute("{CALL dbo.GetEmployeeProjects(?)}", [1])
```
bewithgaurav, line 895: Using `dbo` schema to avoid issues with Azure SQL with Azure AD/Entra ID authentication. It can misinterpret email-format usernames (e.g., user@domain.com) as schema names. Added a comment in exception tests.

mssql_python/connection.py
```diff
@@ -201,6 +215,18 @@ def close(self) -> None:
...
-        self._conn.close()
-        if ENABLE_LOGGING:
-            logger.info("Connection closed successfully.")
+        if self._closed:
+            return
+        
+        # Close all cursors first
+        if hasattr(self, '_cursors'):
+            for cursor in list(self._cursors):
+                if not cursor.closed:
```
gargsaumya: Consider wrapping the cursor.close() call in a try/except block when closing all tracked cursors in the close() method. This way, even if one cursor raises an exception during close, the rest of the cursors and the connection itself will still be properly cleaned up, reducing the risk of leaked resources.

mssql_python/connection.py
```diff
@@ -201,6 +215,18 @@ def close(self) -> None:
...
+                    cursor.close()
+
```
gargsaumya: We can also clear the _cursors set after closing all cursors, ensuring that any internal weak references are released.

mssql_python/connection.py
```diff
@@ -58,6 +58,14 @@ def __init__(self, connection_str: str = "", autocommit: bool = False, attrs_bef
...
         )
         self._attrs_before = attrs_before or {}
+        self._closed = False
+        
+        # Using WeakSet which automatically removes cursors when they are no longer in use
+        # It is a set that holds weak references to its elements.
+        # When an object is only weakly referenced, it can be garbage collected even if it's still in the set.
+        # It prevents memory leaks by ensuring that cursors are cleaned up when no longer in use without requiring explicit deletion.
+        self._cursors = weakref.WeakSet()
+
```
sumitmsft, line 75: One thing to consider for future improvements: if this library is used in multithreaded contexts, access to `self._cursors` (the WeakSet used for cursor tracking) may not be thread-safe. To prevent potential race conditions during cursor creation and cleanup, it would be good to add documentation or a `threading.Lock` (or similar synchronization mechanism) to guard access to `self._cursors`. This would help ensure robust behavior in concurrent scenarios - especially when async operations would be added as a feature to the driver.

mssql_python/connection.py
```diff
@@ -152,7 +160,13 @@ def cursor(self) -> Cursor:
             DatabaseError: If there is an error while creating the cursor.
             InterfaceError: If there is an error related to the database interface.
         """
-        return Cursor(self)
+        """Return a new Cursor object using the connection."""
+        if self._closed:
+            raise Exception("Cannot create cursor on closed connection")
```
sumitmsft: Can we be more specific to exception type—such as `InterfaceError` nstead of a generic Exception? This will allow client code to catch and handle errors more precisely.

mssql_python/connection.py
```diff
@@ -5,7 +5,7 @@
 The class provides methods to establish a connection, create cursors, commit transactions, 
 roll back transactions, and close the connection.
 """
```
sumitmsft, line 12: Please add the following doc string for Connection Class:


Resource Management:
        - All cursors created from this connection are tracked internally.
        - When `close()` is called on the connection, all open cursors are automatically closed.
        - Do not use any cursor after the connection is closed; doing so will raise an exception.
        - Cursors are also cleaned up automatically when no longer referenced, to prevent memory leaks.


Similarly, please add the following doc string on `cursor class` to be consistent with this change:


Resource Management:
        - Cursors are tracked by their parent connection.
        - Closing the connection will automatically close all open cursors.
        - Do not use a cursor after it is closed, or after its parent connection is closed.
        - Use `close()` to release resources held by the cursor as soon as it is no longer needed.

This is to make sure the devs and contributors understand this important change\logic when they read through the code.

sumitmsft:
_**Additional Comment:**_ 

**New Test Case**
I my opinion we should also add a test for cursor creation/cleanup in a multi-threaded context if this is a use case (I have left https://github.com/microsoft/mssql-python/pull/123#discussion_r2184813012 comment already. This is important because code that works fine in a single-threaded situation can sometimes have bugs or “race conditions” when used by several threads at once. This test helps us to make sure our code is safe and works right, even if users use threads. Multi-threading can be induced by customers themselves, and it will be an onus on us to take care of it internally.

sumitmsft (changes requested):
Left a few comments please fix them

mssql_python/connection.py
bewithgaurav: Yes, will be adding those - added a TODO  and a new ADO task

mssql_python/connection.py
bewithgaurav: Added.

mssql_python/connection.py
bewithgaurav: Done

gargsaumya (approved):

jahnvi480 (approved):

sumitmsft (approved):

PR #129, opened 2025-07-09
FEAT: Azure Interactive support for Mac and Linux

jahnvi480:
### ADO Work Item Reference  

> [AB#34946](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34946)
### Summary   
This pull request introduces significant enhancements to the handling of connection strings in the `mssql_python` package, focusing on improved support for Active Directory Interactive Authentication (AAD) and more robust handling of attributes passed to the database driver. The changes ensure better compatibility across platforms and allow the seamless merging of attributes required by the driver.

### Enhancements to Connection String Handling:
* Updated the `_construct_connection_string` method in `mssql_python/connection.py` to support returning a tuple containing the connection string and a dictionary of attributes (`attrs_before`) when needed. This allows for more flexible handling of attributes required by the database driver. [[1]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90L63-R74) [[2]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90L93-R112) [[3]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R135-R136)
* Modified the `add_driver_to_connection_str` function in `mssql_python/helpers.py` to return either a string or a tuple of the connection string and `attrs_before`. This change ensures that attributes required by the driver are preserved and passed correctly. [[1]](diffhunk://#diff-7e002427036075c912554bb92d9e19681ea477e7c2a0eec56d284934d9125343L25-R39) [[2]](diffhunk://#diff-7e002427036075c912554bb92d9e19681ea477e7c2a0eec56d284934d9125343R58-L61)

### Support for Active Directory Interactive Authentication:
* Enhanced the `add_driver_name_to_app_parameter` function in `mssql_python/helpers.py` to handle AAD Interactive Authentication. On non-Windows platforms, it now uses the `azure-identity` library to fetch a token and includes it in the connection attributes. This ensures compatibility with Azure SQL databases requiring AAD authentication.

These updates improve the flexibility and platform compatibility of the `mssql_python` package, making it more robust for a variety of use cases.

### DIFF
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -114,15 +114,15 @@ class Connection:
         for key, value in kwargs.items():
             if key.lower() == "host" or key.lower() == "server":
                 key = "Server"
-            elif key.lower() == "user" or key.lower() == "uid" and attrs_before is not None:
+            elif key.lower() == "user" or key.lower() == "uid":
                 key = "Uid"
-            elif key.lower() == "password" or key.lower() == "pwd" and attrs_before is not None:
+            elif key.lower() == "password" or key.lower() == "pwd":
                 key = "Pwd"
             elif key.lower() == "database":
                 key = "Database"
-            elif key.lower() == "encrypt" and attrs_before is not None:
+            elif key.lower() == "encrypt":
                 key = "Encrypt"
-            elif key.lower() == "trust_server_certificate" and attrs_before is not None:
+            elif key.lower() == "trust_server_certificate":
                 key = "TrustServerCertificate"
             else:
                 continue


mssql_python/row.py
```diff
@@ -62,4 +62,8 @@ def __iter__(self):
     
     def __repr__(self):
         """Return a string representation of the row"""
-        return f"Row{tuple(self._values)}"
\ No newline at end of file
+        return f"Row{tuple(self._values)}"
+    
+    def __str__(self):
+        """Return a string representation of the row"""
+        return f"Row({', '.join(map(str, self._values))})"
```
bewithgaurav, line 69: we wouldn't be needing Row(...) in the string representations, for end users the output should look like a tuple (e.g., (1, 2, 3)).

I suggest updating both the methods as follows:

```
def __str__(self):
    return str(tuple(self._values))

def __repr__(self):
    return repr(tuple(self._values))
```
this makes sure that printing or representing a Row instance makes it behave just like a tuple, which is the expected behaviour, also please add tests for this inside test cursor to test fetch functions results.

mssql_python/helpers.py
```diff
@@ -80,38 +86,75 @@ def check_error(handle_type, handle, ret):
...
     modified_parameters = []
+    has_aad_interactive = False
 
     # Iterate through the key-value pairs
     for param in parameters:
+        param = param.strip()
+        if not param:
+            continue
+            
+        if param.lower().startswith("authentication="):
```
bewithgaurav, line 114: The auth logic is currently spread across multiple helpers and is resulting in cascading function changes and unstable signatures.
I recommend refactoring all authentication-related code into a dedicated mssql_python/auth.py module. This will centralize and simplify authentication logic and help maintain single responsibility in other modules (like helpers and connection).

mssql_python/row.py
bewithgaurav: I believe these changes should be a separate PR of its own

bewithgaurav (changes requested):
need some major structural changes for auth related code and create another PR for row.py changes.

PR #138, opened 2025-07-15
FEAT: Optimize executemany() performance

gargsaumya:
### ADO Work Item Reference  

> AB#37935
### Summary   
  
This pull request introduces significant updates to the `executemany` method in the `mssql_python` library to enable column-wise parameter binding for batched SQL execution. It also includes enhancements to the underlying C++ bindings to support this functionality. The most important changes involve implementing column-wise binding logic, adding helper functions for parameter buffer allocation, and exposing a new `SQLExecuteMany` method for executing batched statements.

### Enhancements to Python `executemany` method:
* [`mssql_python/cursor.py`](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R623-R736): Refactored the `executemany` method to use column-wise parameter binding for improved performance and scalability. Added `_transpose_rowwise_to_columnwise` helper function to convert row-wise parameters into column-wise format.

### Updates to C++ bindings:
* [`mssql_python/pybind/ddbc_bindings.cpp`](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R945-R1108): Added `BindParameterArray` function to handle column-wise parameter binding, supporting multiple data types such as `SQL_C_LONG`, `SQL_C_DOUBLE`, and `SQL_C_WCHAR`.
* [`mssql_python/pybind/ddbc_bindings.cpp`](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R945-R1108): Introduced `SQLExecuteMany_wrap` function to enable batched execution of SQL statements with multiple parameter sets.
* [`mssql_python/pybind/ddbc_bindings.cpp`](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R2288): Exposed the new `SQLExecuteMany` method to Python via `PYBIND11_MODULE`.

### Helper functions for buffer allocation:
* [`mssql_python/pybind/ddbc_bindings.cpp`](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R188-R196): Added `AllocateParamBufferArray` template function to allocate buffers for column-wise parameter binding.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -18,6 +18,8 @@ from mssql_python.helpers import check_error
 from mssql_python.logging_config import get_logger, ENABLE_LOGGING
 from mssql_python import ddbc_bindings
 from .row import Row
+from typing import Sequence, Any
+
 
 logger = get_logger()
 
@@ -539,7 +541,7 @@ class Cursor:
             # Add more mappings as needed
         }
         return sql_to_python_type.get(sql_type, str)
-
+    
     def execute(
         self,
         operation: str,
@@ -617,38 +619,121 @@ class Cursor:
 
         # Initialize description after execution
         self._initialize_description()
+    
+
+    # def executemany(self, operation: str, seq_of_parameters: list) -> None:
+    #     self._check_closed()
+    #     self._reset_cursor()
+
+    #     if not seq_of_parameters:
+    #         return
+
+    #     # Transpose to column-major format
+    #     columns = list(zip(*seq_of_parameters))  # Each column: tuple of values
+    #     sample_params = seq_of_parameters[0]
+    #     param_info = ddbc_bindings.ParamInfo
+
+    #     parameters_type = []
+    #     for i, sample_val in enumerate(sample_params):
+    #         paraminfo = self._create_parameter_types_list(sample_val, param_info, sample_params, i)
+
+    #         # Fix: Adjust string column sizes based on actual max length across all rows
+    #         if isinstance(sample_val, str):
+    #             max_len = max(
+    #                 (len(v) for v in columns[i] if isinstance(v, str)),
+    #                 default=1  # fallback if all values are None
+    #             )
+    #             paraminfo.columnSize = max_len
+
+    #         parameters_type.append(paraminfo)
+
+    #     # Now execute with adjusted parameter types
+    #     ret = ddbc_bindings.SQLExecuteMany(
+    #         self.hstmt, operation, columns, parameters_type, len(seq_of_parameters)
+    #     )
+    #     check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
+
+    #     self.rowcount = ddbc_bindings.DDBCSQLRowCount(self.hstmt)
+    #     self._initialize_description()
+
+    def _transpose_rowwise_to_columnwise(self, seq_of_parameters: list) -> list:
+        """
+        Convert list of rows (row-wise) into list of columns (column-wise),
+        for array binding via ODBC.
+        
+        Example:
+            Input: [(1, "a"), (2, "b")]
+            Output: [[1, 2], ["a", "b"]]
+        """
+        if not seq_of_parameters:
+            return []
+
+        num_params = len(seq_of_parameters[0])
+        columnwise = [[] for _ in range(num_params)]
+        for row in seq_of_parameters:
+            if len(row) != num_params:
+                raise ValueError("Inconsistent parameter row size in executemany()")
+            for i, val in enumerate(row):
+                columnwise[i].append(val)
+        return columnwise
 
     def executemany(self, operation: str, seq_of_parameters: list) -> None:
         """
         Prepare a database operation and execute it against all parameter sequences.
-
-        Args:
-            operation: SQL query or command.
-            seq_of_parameters: Sequence of sequences or mappings of parameters.
-
-        Raises:
-            Error: If the operation fails.
+        This version uses column-wise parameter binding and a single batched SQLExecute().
         """
-        self._check_closed()  # Check if the cursor is closed
-
+        self._check_closed()
         self._reset_cursor()
 
-        first_execution = True
-        total_rowcount = 0
-        for parameters in seq_of_parameters:
-            parameters = list(parameters)
-            if ENABLE_LOGGING:
-                logger.info("Executing query with parameters: %s", parameters)
-            prepare_stmt = first_execution
-            first_execution = False
-            self.execute(
-                operation, parameters, use_prepare=prepare_stmt, reset_cursor=False
-            )
-            if self.rowcount != -1:
-                total_rowcount += self.rowcount
-            else:
-                total_rowcount = -1
-        self.rowcount = total_rowcount
+        if not seq_of_parameters:
+            self.rowcount = 0
+            return
+
+        # # Infer types from the first row
+        # first_row = list(seq_of_parameters[0])
+        # param_info = ddbc_bindings.ParamInfo
+        # parameters_type = [
+        #     self._create_parameter_types_list(param, param_info, first_row, i)
+        #     for i, param in enumerate(first_row)
+        # ]
+        param_info = ddbc_bindings.ParamInfo
+        param_count = len(seq_of_parameters[0])
+        parameters_type = []
+
+        for col_index in range(param_count):
+            # Use the longest string (or most precise value) in that column for inference
+            column = [row[col_index] for row in seq_of_parameters]
+            sample_value = column[0]
+
+            # For strings, pick the value with max len
+            if isinstance(sample_value, str):
+                sample_value = max(column, key=lambda s: len(str(s)) if s is not None else 0)
+
+            # For decimals, use the one with highest precision
+            elif isinstance(sample_value, decimal.Decimal):
+                sample_value = max(column, key=lambda d: len(d.as_tuple().digits) if d is not None else 0)
+
+            param = sample_value
+            dummy_row = list(seq_of_parameters[0])  # to pass for `_get_numeric_data()` mutation
+            parameters_type.append(self._create_parameter_types_list(param, param_info, dummy_row, col_index))
+
+
+        # Transpose to column-wise format for array binding
+        columnwise_params = self._transpose_rowwise_to_columnwise(seq_of_parameters)
+
+        # Execute batched statement
+        ret = ddbc_bindings.SQLExecuteMany(
+            self.hstmt,
+            operation,
+            columnwise_params,
+            parameters_type,
+            len(seq_of_parameters)
+        )
+        check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
+
+        self.rowcount = len(seq_of_parameters)
+        self.last_executed_stmt = operation
+        self._initialize_description()
 
     def fetchone(self) -> Union[None, Row]:
         """
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -185,6 +185,15 @@ ParamType* AllocateParamBuffer(std::vector<std::shared_ptr<void>>& paramBuffers,
     return static_cast<ParamType*>(paramBuffers.back().get());
 }
 
+template <typename ParamType>
+ParamType* AllocateParamBufferArray(std::vector<std::shared_ptr<void>>& paramBuffers,
+                                    size_t count) {
+    std::shared_ptr<ParamType> buffer(new ParamType[count], std::default_delete<ParamType[]>());
+    ParamType* raw = buffer.get();
+    paramBuffers.push_back(buffer);
+    return raw;
+}
+
 std::string DescribeChar(unsigned char ch) {
     if (ch >= 32 && ch <= 126) {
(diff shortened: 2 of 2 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -933,6 +942,170 @@
...
+
+                    std::wstring wstr = columnValues[i].cast<std::wstring>();
+                    if (wstr.length() > info.columnSize) {
+                        std::string offending = WideToUTF8(wstr);
+                        ThrowStdException("String too long at param " + std::to_string(paramIndex) +
+                                          ", value: " + offending +
+                                          ", len: " + std::to_string(wstr.length()) +
+                                          " > columnSize: " + std::to_string(info.columnSize));
+                    }
+                    std::memcpy(wcharArray + i * (info.columnSize + 1), wstr.c_str(), (wstr.length() + 1) * sizeof(SQLWCHAR));
```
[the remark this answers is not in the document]
gargsaumya: Bounds are explicitly checked above to ensure wstr.length() <= info.columnSize.wcharArray has been pre-allocated with (info.columnSize + 1) * paramSetSize elements.
Also, std::wstring::c_str() is guaranteed non-null, and no overlap exists between src and dest. so, this memcpy is safe and intentional.

mssql_python/cursor.py
```diff
@@ -618,37 +617,65 @@ def execute(
         # Initialize description after execution
         self._initialize_description()
 
+    def _transpose_rowwise_to_columnwise(self, seq_of_parameters: list) -> list:
```
jahnvi480, line 662: Can we use numpy library to do this, It will be much more faster than using 2 loops

```python
    if not seq_of_parameters:
        return []
    arr = np.array(seq_of_parameters)
    return arr.T.tolist()
```

But for this we need to include numpy in setup.py while installing our driver.

mssql_python/cursor.py
```diff
@@ -618,37 +617,65 @@ def execute(
...
+        for row in seq_of_parameters:
+            if len(row) != num_params:
+                raise ValueError("Inconsistent parameter row size in executemany()")
+            for i, val in enumerate(row):
+                columnwise[i].append(val)
+        return columnwise
+
     def executemany(self, operation: str, seq_of_parameters: list) -> None:
         """
         Prepare a database operation and execute it against all parameter sequences.
```
jahnvi480, line 683: Keep the initial comments for function same as other functions by including the args passed and what is it returning or raising.

jahnvi480 (dismissed):

bewithgaurav (approved):

mssql_python/cursor.py
gargsaumya: Thanks for the suggestion!
NumPy is indeed good for handling large, numeric arrays, but in this case, the performance didn't improve upon testing locally. For now, keeping this as is since it will require changes in setup.py as well but we can revisit this in future as improvement.

bewithgaurav (approved):

mssql_python/cursor.py
jahnvi480: Let's add TODO for this

jahnvi480: Or you can make the changes since we need to change setup.py for authentication as well. So, we can do both of them together.

gargsaumya: Since NumPy didn't show any perf benefit here, I believe we can skip adding a TODO for now.
We can always revisit this while testing on other platforms to see any improvement due to this (have noted that).
We can also discuss this with the team.

jahnvi480 (dismissed):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -933,6 +942,173 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
     }
 }
 
+SQLRETURN BindParameterArray(SQLHANDLE hStmt,
```
sumitmsft, line 945: There could be a resource management issues in `BindParameterArray` and that could be a memory leak under certain failure conditions. Could you please validate it once? This could be an edge case but worth exploring.

Inside the `for` loop (line 964), we call `AllocateParamBufferArray` to create the buffers. 
If something goes wrong halfway through the loop (like an exception), the buffers we created in the previous loops never get cleaned up. 
Since they weren't successfully bound, the ODBC won't handle them either, so the memory just leaks.

Can we track all the buffers we allocate? Ans wrap around loop in a `try...catch` block? If we catch an exception, the `catch` block can just delete everything we've allocated so far before re-throwing.

mssql_python/cursor.py
```diff
@@ -631,37 +630,71 @@ def execute(
...
-        self.rowcount = total_rowcount
+        if not seq_of_parameters:
+            self.rowcount = 0
+            return
+
+        param_info = ddbc_bindings.ParamInfo
+        param_count = len(seq_of_parameters[0])
+        parameters_type = []
+
+        for col_index in range(param_count):
```
sumitmsft, line 703: In the previous implementation, there was `logger.info` logging inside the for-loop, which was helpful for tracking individual executions and diagnosing issues. Could we consider adding parameter-level or batch-level logging in the new batched approach? This would help maintain visibility for batch diagnostics and troubleshooting.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -933,6 +942,173 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
+                SQLWCHAR* wcharArray = AllocateParamBufferArray<SQLWCHAR>(paramBuffers, paramSetSize * (info.columnSize + 1));
+                strLenOrIndArray = AllocateParamBufferArray<SQLLEN>(paramBuffers, paramSetSize);
+                for (size_t i = 0; i < paramSetSize; ++i) {
+                    if (columnValues[i].is_none()) {
+                        strLenOrIndArray[i] = SQL_NULL_DATA;
+                        std::memset(wcharArray + i * (info.columnSize + 1), 0, (info.columnSize + 1) * sizeof(SQLWCHAR));
+                        continue;
+                    }
+                    std::wstring wstr = columnValues[i].cast<std::wstring>();
+                    if (wstr.length() > info.columnSize) {
```
sumitmsft: Can you please confirm that the `columnSize` used in this check is derived from the database metadata (e.g., the column definition from the SQL Server table) and not hardcoded or directly from user input?  If it's hardcoded or user-provided, it could be a potential vulnerability. 

Using a hardcoded or user-supplied columnSize for string length checks can introduce serious security risks. It may lead to buffer overflows, data truncation, or even allow malicious users to bypass schema constraints defined in the database. Instead of relying on assumptions or manual inputs, it’s always safer and more reliable to use the actual column size from the database metadata.

mssql_python/cursor.py
```diff
@@ -631,37 +630,71 @@ def execute(
         # Initialize description after execution
         self._initialize_description()
 
+    def _transpose_rowwise_to_columnwise(self, seq_of_parameters: list) -> list:
```
sumitmsft, line 662: I do not see test cases for this function. Do we have unit tests that specifically verify the behavior of this function with various input scenarios (e.g., different data types, empty lists, lists of different lengths)? I would suggest it to be added.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -933,6 +942,173 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
+        const py::list& columnValues = columnwise_params[paramIndex].cast<py::list>();
+        const ParamInfo& info = paramInfos[paramIndex];
+        if (columnValues.size() != paramSetSize) {
+            ThrowStdException("Column " + std::to_string(paramIndex) + " has mismatched size.");
+        }
+        void* dataPtr = nullptr;
+        SQLLEN* strLenOrIndArray = nullptr;
+        SQLLEN bufferLength = 0;
+
+        switch (info.paramCType) {
```
sumitmsft: Could you please verify that *all* supported data types are handled correctly and that appropriate conversions are performed between Python types and the corresponding SQL Server data types? It would be beneficial to add more unit tests that specifically cover a wide range of data types, including edge cases and boundary conditions, to ensure no data loss or unexpected behavior occurs. 
Consider including tests for `NULL` values as well.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -933,6 +942,173 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
+                             const py::list& columnwise_params,
+                             const std::vector<ParamInfo>& paramInfos,
+                             size_t paramSetSize,
+                             std::vector<std::shared_ptr<void>>& paramBuffers) {
+    LOG("Starting column-wise parameter array binding. paramSetSize: {}, paramCount: {}", paramSetSize, columnwise_params.size());
+    for (int paramIndex = 0; paramIndex < columnwise_params.size(); ++paramIndex) {
+        const py::list& columnValues = columnwise_params[paramIndex].cast<py::list>();
+        const ParamInfo& info = paramInfos[paramIndex];
+        if (columnValues.size() != paramSetSize) {
+            ThrowStdException("Column " + std::to_string(paramIndex) + " has mismatched size.");
```
sumitmsft: For better error handling and debugging, can we consider using more specific exception types (e.g., `invalid_argument`, `runtime_error`, custom exception classes) that accurately reflect the nature of the error? 

This would allow calling code to catch specific exceptions and handle them appropriately, providing more context about the error's cause.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -933,6 +942,173 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
+                for (size_t i = 0; i < paramSetSize; ++i) {
+                    if (columnValues[i].is_none()) {
+                        strLenOrIndArray[i] = SQL_NULL_DATA;
+                        std::memset(wcharArray + i * (info.columnSize + 1), 0, (info.columnSize + 1) * sizeof(SQLWCHAR));
+                        continue;
+                    }
+                    std::wstring wstr = columnValues[i].cast<std::wstring>();
+                    if (wstr.length() > info.columnSize) {
+                        std::string offending = WideToUTF8(wstr);
+                        ThrowStdException("String too long at param " + std::to_string(paramIndex) +
```
sumitmsft: "String too long at param ..." is informative, which is good for debugging. However, make sure that the message does *not* inadvertently expose any sensitive data, such as user-specific information, internal paths, or any other details that could pose a security risk if exposed. We can redact the info if necessary to protect sensitive information while still providing useful context for developers.

For example:
Instead of: "String too long at param `user_password` for user `sumitmsft`"

We can raise: "String too long for parameter at index X".

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Currently, for parameter binding, we dynamically infer columnSize on the Python side by inspecting the actual values in the input sequence. So, we do not hardcode sizes, nor do we rely on user-supplied values. This approach is reasonably safe and avoids silent truncation in most cases.
But also, later we can integrate SQLDescribeParam() for parameter binding especially for execute(), to align better with schema-defined column types. We can add this enhancement in a future update.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Ack.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: We are supporting most data types now; here is the list:

Numeric types: SQL_C_LONG, SQL_C_SHORT, SQL_C_DOUBLE, SQL_C_FLOAT, SQL_C_TINYINT, SQL_C_UTINYINT, SQL_C_SBIGINT, SQL_C_SLONG, SQL_C_UBIGINT, SQL_C_ULONG, SQL_C_USHORT
Character types: SQL_C_WCHAR, SQL_C_CHAR
Binary type: SQL_C_BINARY
Boolean: SQL_C_BIT
Temporal types: SQL_C_TYPE_DATE, SQL_C_TYPE_TIME, SQL_C_TYPE_TIMESTAMP
Numeric/Decimal: SQL_C_NUMERIC with explicit precision, scale, sign, and value encoding
NULL handling: Each case includes logic to check for None and set SQL_NULL_DATA and zero/empty buffer values appropriately.

And we have tested locally for all supported types. But we'll add the formal unit tests for this in the next update, so that we do not block this release.

mssql_python/cursor.py
gargsaumya: Though tested locally for all supported types, we'll add the formal unit tests for this in the next update (unblocking this release).

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: This will be covered in following task: https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/34964

jahnvi480 (approved):

PR #157, opened 2025-07-29
FIX: Making type objects and constructor compatible with pyodbc

jahnvi480:

> [AB#38059](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38059)

### Summary   
This pull request refactors type objects in `mssql_python/type.py` to inherit from their respective Python built-in types, simplifies constructors, and updates related tests to align with the new implementation. Additionally, it modifies timestamp and time handling functions for better compatibility and enhances the `Binary` function to handle multiple input types.

### Refactoring of type objects:
- [`mssql_python/type.py`](diffhunk://#diff-57fcb124a0b5958b9034fbd0c3ee1bc65e4140e1671e642d5dcb5aabd60d55a7L12-R55): Changed type classes (`STRING`, `BINARY`, `NUMBER`, `DATETIME`, `ROWID`) to inherit from Python built-in types (`str`, `bytearray`, `float`, `datetime.datetime`, `int`) and replaced `__init__` methods with `__new__` methods for direct instantiation.

### Updates to utility functions:
- [`mssql_python/type.py`](diffhunk://#diff-57fcb124a0b5958b9034fbd0c3ee1bc65e4140e1671e642d5dcb5aabd60d55a7L93-R113): Modified `TimeFromTicks` to use `time.localtime` instead of `time.gmtime` and updated `TimestampFromTicks` to remove the UTC timezone. Enhanced `Binary` to handle both `str` and `bytes` inputs, with fallback for other types by converting to string first.

### Updates to tests:
- [`tests/test_002_types.py`](diffhunk://#diff-15437630102d01c37a02763e0080246da102ccedaeea931d7c433470ff0fb009L6-R19): Updated tests for type objects (`STRING`, `BINARY`, `NUMBER`, `DATETIME`, `ROWID`) to validate against their respective Python built-in types instead of custom attributes.
- [`tests/test_002_types.py`](diffhunk://#diff-15437630102d01c37a02763e0080246da102ccedaeea931d7c433470ff0fb009L47-R58): Adjusted `test_time_from_ticks` and `test_timestamp_from_ticks` to reflect changes in `TimeFromTicks` and `TimestampFromTicks` behavior. Improved `test_binary_constructor` to test compatibility with both `bytes` and `bytearray`.

### DIFF
--- a/mssql_python/type.py
+++ b/mssql_python/type.py
@@ -9,50 +9,50 @@ import time
 
 
 # Type Objects
-class STRING:
+class STRING(str):
     """
     This type object is used to describe columns in a database that are string-based (e.g. CHAR).
     """
 
-    def __init__(self) -> None:
-        self.type = "STRING"
+    def __new__(cls):
+        return str.__new__(cls, "")
 
 
-class BINARY:
+class BINARY(bytearray):
     """
     This type object is used to describe (long)
     binary columns in a database (e.g. LONG, RAW, BLOBs).
     """
 
-    def __init__(self) -> None:
-        self.type = "BINARY"
+    def __new__(cls):
+        return bytearray.__new__(cls)
 
 
-class NUMBER:
+class NUMBER(float):
     """
     This type object is used to describe numeric columns in a database.
     """
 
-    def __init__(self) -> None:
-        self.type = "NUMBER"
+    def __new__(cls):
+        return float.__new__(cls, 0.0)
 
 
-class DATETIME:
+class DATETIME(datetime.datetime):
     """
     This type object is used to describe date/time columns in a database.
     """
 
-    def __init__(self) -> None:
-        self.type = "DATETIME"
+    def __new__(cls, year: int = 1, month: int = 1, day: int = 1):
+        return datetime.datetime.__new__(cls, year, month, day)
 
 
-class ROWID:
+class ROWID(int):
     """
-    This type object is used to describe the “Row ID” column in a database.
+    This type object is used to describe the "Row ID" column in a database.
     """
 
-    def __init__(self) -> None:
-        self.type = "ROWID"
+    def __new__(cls):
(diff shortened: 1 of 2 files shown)


bewithgaurav (approved):

mssql_python/type.py
```diff
@@ -90,18 +90,24 @@ def TimeFromTicks(ticks: int) -> datetime.time:
...
+    Converts a string or bytes to bytes using UTF-8 encoding.
     """
-    return bytes(string, "utf-8")
+    if isinstance(value, bytes):
+        return value
+    elif isinstance(value, str):
+        return bytes(value, "utf-8")
+    else:
+        # Handle other types by converting to string first
+        return bytes(str(value), "utf-8")
```
sumitmsft: Currently, the `Binary(value)` function converts any unexpected input type to a string and then to bytes. This implicit conversion can hide bugs and make debugging harder if an unsupported type is accidentally passed. 

Instead of converting unexpected types to string, raise a `TypeError` for unsupported types. This will ensure that only valid types (`str` or `bytes`) are accepted, and provide clear error messages for developer mistakes.

raise TypeError(f"Cannot convert type {type(value)} to bytes")

This change will improve type safety, make errors more explicit, and help catch bugs earlier in development.

Perhaps, if we are doing what PyODBC is doing then we need to triage it and discuss in detail.

sumitmsft (changes requested):
Left one comment on how to handle unexpected type in Binary func,

mssql_python/type.py
jahnvi480: According to DBAPI2.0 only string is allowed... but since pyodbc supports bytes... I have added that option too. With that I'm adding bytearray. Apart from these datatypes it will raise an error.

sumitmsft (approved):

Issue #23, opened 2025-03-14
Context managers


eavanvalkenburg:
I would love to see correct managers on both the connection, the cursor but also on a transaction, where the exit commits a transaction

LarnuUK:
Context Managers on the connection, and that they didn't dispose of the connection, when used with pyodbc was something that bugged me. It would be great to see context managers added to connection added to mssql-python and that they *dispose* of the connection, like the change that was made to [Psycopg 3](https://www.psycopg.org/psycopg3/docs/basic/usage.html#connection-context).

sumitmsft:
Hello @eavanvalkenburg - Thanks for reaching out to us. We wanted to let you know that we have started the development of Context Manager feature now. Soon, we will be releasing it. Please stay tuned! 

Appreciate your early feedback. 

@LarnuUK - I think we are on the same page regarding your concerns :)

sumitmsft:
This is merged and will be released in v0.10.0.

PR #160, opened 2025-08-04
FEAT: Context manager support for connection and cursor class

jahnvi480:

> [AB#38074](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38074)

> GitHub Issue: #23 

### Summary   
This pull request introduces context manager support for the `Connection` and `Cursor` classes in the `mssql_python` module, aligning their behavior with `pyodbc`. It also adds extensive tests to ensure correct functionality, covering various scenarios such as normal exits, exceptions, nested transactions, and manual commit/rollback. The key changes are grouped below:

### Context Manager Support
* Added `__enter__` and `__exit__` methods to the `Connection` class to enable usage with the `with` statement. The connection commits transactions on normal exit and rolls back on exceptions, without closing the connection. 
* Added `__enter__` and `__exit__` methods to the `Cursor` class to allow usage with the `with` statement. The cursor commits transactions on normal exit but does not close itself, aligning with `pyodbc` behavior.

### Logging Enhancements
* Enhanced the `Connection.close` method to log when uncommitted changes are rolled back before closing.

### Test Coverage
* Added new tests in `tests/test_003_connection.py` to verify context manager behavior for connections, including:
  - Committing transactions on normal exit.
  - Rolling back transactions on exceptions.
  - Handling autocommit mode.
  - Ensuring connections remain open after context exit.
  - Supporting nested context managers.
  - Testing manual commit/rollback within a context manager. 
* Added a test for using `contextlib.closing` with connections to ensure proper closure after context exit.
* Updated `tests/test_004_cursor.py` to include `contextlib.closing` for cursor tests.


LarnuUK:
Why, out of interest @jahnvi480 , should the context managers *not* close the connection/cursor? I am aware that you state that this aligns with `pyodbc` but is that the *right* choice? The behaviour, honestly, feels more like a "gotcha", as often it would be expected that the connection/cursor *would* be disposed of when the context manage is closed. Look, for example, at [[Psycopg 3](https://www.psycopg.org/psycopg3/docs/basic/usage.html#connection-context)](https://www.psycopg.org/psycopg3/docs/basic/usage.html#connection-context) which changed their default behaviour to close to the connection.

mssql_python/connection.py
```diff
@@ -278,6 +293,66 @@ def close(self) -> None:
...
+        Args:
+            exc_type: The exception type if an exception occurred, None otherwise
+            exc_val: The exception value if an exception occurred, None otherwise  
+            exc_tb: The exception traceback if an exception occurred, None otherwise
+            
+        Note:
+            This method does not return True, so exceptions are not suppressed
+            and will propagate normally.
+        """
+        try:
```
dlevy-msft-sql: Do we need to verify that the connection is still open before this? This code should run even if there is an exception in the connection manager so it feels like a disconnected connection should get us here too.

mssql_python/cursor.py
```diff
@@ -789,6 +789,46 @@ def nextset(self) -> Union[bool, None]:
...
+        
+        Args:
+            exc_type: Exception type (if any)
+            exc_val: Exception value (if any) 
+            exc_tb: Exception traceback (if any)
+        """
+        try:
+            # Only commit if autocommit is False, following pyodbc behavior
+            if not self.connection.autocommit:
+                self.connection.commit()
```
sumitmsft: This commits even when exc_type is not None. We need to handle it well.

mssql_python/cursor.py
```diff
@@ -789,6 +789,46 @@ def nextset(self) -> Union[bool, None]:
...
+        - If autocommit is False, commit the transaction
+        - The cursor is NOT closed (unlike connection context manager)
+        - Returns None to propagate any exceptions
+        
+        Args:
+            exc_type: Exception type (if any)
+            exc_val: Exception value (if any) 
+            exc_tb: Exception traceback (if any)
+        """
+        try:
```
sumitmsft: Add exception handling similar to the connection.. We need to handle rollback on exceptions

mssql_python/connection.py
```diff
@@ -278,6 +293,66 @@ def close(self) -> None:
...
+        Example:
+            with connect(connection_string) as conn:
+                cursor = conn.cursor()
+                cursor.execute("INSERT INTO table VALUES (?)", [value])
+                # Transaction will be committed automatically when exiting
+        """
+        log('info', "Entering connection context manager.")
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
```
bewithgaurav: we probably don't need to have any exception overriding w.r.t. logic around `exc_type`, python can take care of it itself I guess
if I understand this correctly, the method can be as simple as:
```
if not self._closed:
    self.close()
```

roganjoshp:
I have to strongly agree with @LarnuUK here. In a past job, I was so convinced by the context-manager behaviour of Python that I persuaded my whole team to do a code review and change all of their connections to use a context manager.

These idle connections linger for no purpose at all that I can see. The server will clear them (over long periods of time), but the more software that's running over the DB, the more they start to accumulate. Depending on the RDBMS you're using, it's when you hit ~500 that things will go south. In this case, my decision ended up crashing our central Redshift cluster and brought down our services for more than 10 clients and launched a _major_ issue. 

As mentioned previously; `psycopg3` changed the default behaviour of `__exit__` to also cover closing the connection, which `psycopg2` did not do. Although I know that the traditional DB API suggested that this behaviour exists, I am yet to understand _any_ utility that it adds. I can only see this as a flaw in the design of the context manager. Please reconsider following the `pyodbc` design and instead follow the example of `psycopg3`.

sumitmsft:
Thanks @LarnuUK @roganjoshp for your invaluable feedback and comments. This PR is still under review and our team was still investigating these points in detail. Having said that, we will close connection/cursors automatically upon __exit__() as default behavior of this driver.

tests/test_003_connection.py
```diff
@@ -16,12 +16,16 @@
 - test_connection_string_with_attrs_before: Check if the connection string is constructed correctly with attrs_before.
 - test_connection_string_with_odbc_param: Check if the connection string is constructed correctly with ODBC parameters.
 - test_rollback_on_close: Test that rollback occurs on connection close if autocommit is False.
+- test_context_manager_commit: Test that context manager commits transaction on normal exit.
+- test_context_manager_autocommit_mode: Test context manager behavior with autocommit enabled.
+- test_context_manager_connection_closes: Test that context manager closes the connection.
```
sumitmsft, line 21: Probably add one more test:

Insert with autocommit True, then connection.close() should not trigger rollback or error.

tests/test_004_cursor.py
```diff
@@ -1313,6 +1314,291 @@ def test_row_column_mapping(cursor, db_connection):
         cursor.execute("DROP TABLE #pytest_row_test")
         db_connection.commit()
 
+def test_cursor_context_manager_basic(db_connection):
```
sumitmsft, line 1317: Add the following test:
Double cursor.close() desired behavior: Validate no-op. I think cursor.close() shouldn't raises an exception when invoked a second time. I would recommend going with idempotency; add test accordingly.

gargsaumya (dismissed):

sumitmsft (dismissed):

gargsaumya (approved):

sumitmsft (approved):

PR #163, opened 2025-08-05
FEAT: Adding cursor.rownumber - DBAPI 2.0 

jahnvi480:

> [AB#34888](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34888)

### Summary   
This pull request adds a new DB-API extension attribute, `rownumber`, to the `Cursor` class in `mssql_python/cursor.py`, allowing users to track the current 0-based index of the cursor in the result set. It includes full implementation, integration with fetch methods, and comprehensive tests to verify the new functionality and its logging behavior.

### Feature Addition: Cursor Row Number Tracking
* Added the `rownumber` property to the `Cursor` class, which returns the current 0-based index in the result set or `None` if unavailable. This property logs a warning when accessed, as it's a DB-API extension and may not be portable. Supporting methods for managing rownumber state (`_reset_rownumber`, `_increment_rownumber`, `_decrement_rownumber`, `_clear_rownumber`) were also introduced.
* Integrated rownumber management into cursor operations: reset/clear on `execute`, increment on successful fetches in `fetchone`, `fetchmany`, and `fetchall`. 

### Testing Enhancements
* Added new tests in `tests/test_004_cursor.py` to verify basic rownumber functionality, correct progression during fetch operations, warning logging behavior, and correct handling when the cursor is closed.

These changes provide users with a convenient way to track cursor position in result sets and ensure the feature is robustly tested and logged.


mssql_python/cursor.py
```diff
@@ -29,6 +29,7 @@ class Cursor:
         description: Sequence of 7-item sequences describing one result column.
         rowcount: Number of rows produced or affected by the last execute operation.
         arraysize: Number of rows to fetch at a time with fetchmany().
+        rownumber: Current 0-based index of the cursor in the result set.
```
sumitmsft: There are a bunch of missing integration points which must be taken care when `rownumber` is implemented.... I am writing them down one by one below to track them appropriately.. 

1. `__iter__()` and `__next__()` methods
2. `close()` method
3. Error handling in `fetch` methods

mssql_python/cursor.py
```diff
@@ -536,6 +541,70 @@ def _map_data_type(self, sql_type):
...
+            raise InterfaceError("Cannot decrement rownumber: no active result set.")
+        
+    def _clear_rownumber(self):
+        """
+        Clear the rownumber tracking.
+        
+        This should be called when the result set is cleared or when the cursor is reset.
+        """
+        self._rownumber = None
+        self._has_result_set = False
```
sumitmsft, line 641: **Iterator - (`__iter__` and `__next__`)**

If the __iter__ and __next__ methods of the cursor do not increment the rownumber attribute, iterating over the cursor with a for loop will result in an incorrect or static row number value, while fetch methods do keep it updated.

In `mssql-python`, the cursor implements the iterator protocol, but the `__iter__()` and `__next__()` methods don't update `rownumber`. When users iterate over the cursor using:

```python
for row in cursor:
    print(f"Row {cursor.rownumber}: {row}")
```
The rownumber won't be tracked correctly because iteration bypasses the fetch methods that increment the counter.

mssql_python/cursor.py
sumitmsft: **Integrating with `close()` method**

When a cursor is closed, the `rownumber` tracking state should be explicitly cleared. Currently, the `rownumber` property checks if the cursor is closed and returns `None`, but the `close()` method doesn't reset the internal tracking variables.

I don't see any changes to this behavior. 

Also check what will happen in case of concurrent access scenario?

mssql_python/cursor.py
```diff
@@ -772,6 +847,9 @@ def fetchone(self) -> Union[None, Row]:
         if ret == ddbc_sql_const.SQL_NO_DATA.value:
             return None
         
+        # Increment rownumber for successful fetch
+        self._increment_rownumber()
```
sumitmsft: The `_increment_rownumber()` is called after the fetch operation but potentially before complete error checking. If an exception occurs after the increment, the `rownumber` will be incorrect.

Only increment `rownumber` after confirming the fetch was successful and no errors occurred. The same issue exists in `fetchmany()` (line 881) and `fetchall()` (line 902).

mssql_python/cursor.py
```diff
@@ -813,6 +896,11 @@ def fetchall(self) -> List[Row]:
         rows_data = []
         ret = ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
         
+        # Update rownumber for the number of rows fetched
```
sumitmsft: Please check if we are properly handling scenarios where different fetch methods are mixed:
For example:

```python
cursor.execute("SELECT * FROM table") // 10 rows
cursor.fetchone()     # rownumber = 1
cursor.fetchmany(3)   # rownumber = 4 (increments 3 times)
cursor.fetchall()     # What's the final rownumber?
```
Issues:
1. fetchall() after fetchmany() should only count remaining rows, not all rows..... In the above case it should return rownumber 10, not 16. 
2. Need to ensure consistent behavior across all fetch combinations

Write test cases accordingly

mssql_python/cursor.py
```diff
@@ -536,6 +541,88 @@ def _map_data_type(self, sql_type):
...
+
+        # Return None if cursor is closed or no result set is available
+        if self.closed or not self._has_result_set:
+            return None
+        
+        return self._rownumber
+    
+    @property
+    def connection(self):
+        """
```
sumitmsft, line 584: The current implementation sets `_rownumber = 0` for all SELECT statements, including empty result sets. This might be confusing:

```python
cursor.execute("SELECT * FROM empty_table")
cursor.rownumber  # Returns 0, but no rows exist
cursor.fetchone() # Returns None - we never actually fetched a row
cursor.rownumber  # Still 0
```

Just to avoid this confusion, can we set `rownumber` to `None` until the first successful fetch, or document this behavior clearly. The DB-API 2.0 spec doesn't specify behavior for empty result sets as such.

mssql_python/cursor.py
```diff
@@ -74,6 +75,10 @@ def __init__(self, connection) -> None:
         # Is a list instead of a bool coz bools in Python are immutable.
         # Hence, we can't pass around bools by reference & modify them.
         # Therefore, it must be a list with exactly one bool element.
+        
+        # rownumber attribute
```
sumitmsft, line 78: The `_rownumber` and `_has_result_set` attributes are not thread-safe in this implementation. If multiple threads access the same cursor concurrently (say using Thread lib of python), the row number tracking could become corrupted.

**Problem:**
```python
# Thread 1
cursor.fetchone()  # Reads _rownumber, increments

# Thread 2 (concurrent)
cursor.fetchone()  # Reads same _rownumber, increments
```
Add documentation stating cursors are not thread-safe (if that's the design)
Or add thread synchronization (threading.Lock) around rownumber operations - Please check how pyodbc and psycopg have done it??

Documentation needed: Clarify in the rownumber property docstring whether concurrent access is supported.

This could lead us to a bigger problem..

tests/test_004_cursor.py
```diff
@@ -1883,6 +1887,154 @@ def test_execute_chaining_compatibility_examples(cursor, db_connection):
...
+        cursor.execute("CREATE TABLE #test_rownumber (id INT, value VARCHAR(50))")
+        db_connection.commit()
+        
+        # Insert test data
+        for i in range(5):
+            cursor.execute("INSERT INTO #test_rownumber VALUES (?, ?)", i, f"value_{i}")
+        db_connection.commit()
+        
+        # Execute query and check initial rownumber
+        cursor.execute("SELECT * FROM #test_rownumber ORDER BY id")
```
sumitmsft, line 1904: Can you please validate if `executemany()` is propely handled? I do not see any test case for it. Add appropriate `rownumber` handling to `executemany()` for consistency.

sumitmsft (changes requested):
Left a few comments on the PR. Please address them.

mssql_python/cursor.py
jahnvi480: 1. `__iter__()` method should not increment rownumber since it does not fetch data: The `__iter__()` method simply returns `self` to indicate that the cursor is iterable. It doesn't actually fetch any rows or move the cursor position.
`__next__()` method calls `fetchone()` and `fetchone()` already increments the rownumber

2. Added `_clear_rownumber()` for close 

3. Added try and except for all `fetch` methods

jahnvi480: Added `_clear_rownumber()` for close.

mssql_python/cursor.py
jahnvi480: Concurrent Access: The current implementation has several thread safety issues:
 - Multiple threads could access/modify `_rownumber` `_has_result_set` simultaneously
 - Race conditions between checking `closed` state and accessing rownumber
 - Fetch operations could interfere with each other
 
Performance Impact Analysis
 - Adding `threading.RLock` to every rownumber access would:

 - CPU Overhead: Each property access requires lock acquisition/release
 - Memory Overhead: Each cursor gets a lock object
 - Contention: If multiple threads actually used same cursor (anti-pattern)
 - Complexity: Unnecessary for the 99.9% use case (single-threaded cursor usage)
 
How Other Database Drivers Handle Thread Safety:
1. pyodbc
 - No per-cursor locking for rownumber or basic operations
 - Connection-level thread safety warnings only
 - Cursors are not thread-safe by design - users expected to handle concurrency
 - Performance-first approach - minimal overhead
2. psycopg2/psycopg3
 - No locking for cursor.rownumber
 - Thread safety handled at connection level only
 - Cursor operations assumed to be single-threaded per cursor instance
 - Uses atomic operations where needed, not locks

Recommendation:
 - DB-API 2.0 Standard: Cursors are designed for single-thread use
 - Real-World Usage: 99.9% of applications use one cursor per thread
 - Performance Critical: Database drivers prioritize speed over defensive programming
 - Error Prone: Shared cursors between threads is almost always a bug
 - Better Alternative: Users should create separate cursors for each thread

sumitmsft (approved):

PR #166, opened 2025-08-05
FEAT: Support Emojis for data types

gargsaumya:

> [AB#37664](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/37664)

### Summary   
  
This pull request improves Unicode string handling in the SQL type mapping logic and adds comprehensive tests for round-tripping emoji and special characters in the database. The main focus is on ensuring that Unicode strings, including those with emojis and special characters, are correctly mapped and stored in SQL Server.

**Unicode string mapping improvements:**

* Updated the `_map_sql_type` method in `cursor.py` to use the UTF-16 code unit length for Unicode string parameters instead of the Python string length, ensuring correct sizing for both long and short Unicode strings. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R326-R330) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R340-R344)

**Testing enhancements:**

* Added a new test `test_emoji_round_trip` in `test_004_cursor.py` to verify that various emoji, accented, and non-Latin characters can be inserted and retrieved accurately from the database, improving coverage for Unicode edge cases.


tests/test_004_cursor.py
```diff
@@ -1313,6 +1313,44 @@ def test_row_column_mapping(cursor, db_connection):
...
+            cursor.execute("INSERT INTO emoji_test (content) OUTPUT INSERTED.id VALUES (?)", [text])
+            inserted_id = cursor.fetchone()[0]
+            cursor.execute("SELECT content FROM emoji_test WHERE id = ?", [inserted_id])
+            result = cursor.fetchone()
+            assert result is not None, f"No row returned for ID {inserted_id}"
+            assert result[0] == text, f"Mismatch! Sent: {text}, Got: {result[0]}"
+
+        except Exception as e:
+            pytest.fail(f"Error for input {repr(text)}: {e}")
+
```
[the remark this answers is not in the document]
bewithgaurav: just an FYI as a common note - for pytest tables, we should use # prefix since keeps the table as a temporary one in the session, no need to explicitly delete it.
for this one, it can be called #pytest_emoji_test. I'll raise a separate PR later on for refactoring of pytests - will create a function called `create_table_if_not_exists` where you can pass table name only and it'll take care of rest. Link for ADO task - https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/38109 
Approved now, lgtm

bewithgaurav (dismissed):

bewithgaurav (dismissed):

mssql_python/cursor.py
```diff
@@ -323,10 +323,11 @@ def _map_sql_type(self, param, parameters_list, i):
             # TODO: revisit
             if len(param) > 4000:  # Long strings
                 if is_unicode:
+                    utf16_len = len(param.encode("utf-16-le")) // 2
```
sumitmsft, line 335: Can we wrap ` utf16_len = len(param.encode("utf-16-le")) // 2` in a function and then check on the following: Use try-catch for encoding failures (to handle Unicode Encode error)?

Something like:

```
def _calculate_utf16_length(self, param):
   ...
   ...
    try:
        return len(param.encode("utf-16-le")) // 2
    except UnicodeEncodeError as e:                    // or some other exception
        # Log error and fallback to Python length
        logger.warning(f"UTF-16 encoding failed: {e}")
        return len(param)
```

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -54,12 +54,28 @@ using namespace pybind11::literals;
     }
 
     inline std::vector<SQLWCHAR> WStringToSQLWCHAR(const std::wstring& str) {
-        std::vector<SQLWCHAR> result(str.size() + 1, 0);  // +1 for null terminator
-        for (size_t i = 0; i < str.size(); ++i) {
-            result[i] = static_cast<SQLWCHAR>(str[i]);
+    std::vector<SQLWCHAR> result;
```
sumitmsft, line 102: Can we put some comments around Surrogate logic to better understand the code?

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -54,12 +54,28 @@ using namespace pybind11::literals;
...
+
+    for (wchar_t wc : str) {
+        uint32_t codePoint = static_cast<uint32_t>(wc);
+        if (codePoint >= 0xD800 && codePoint <= 0xDFFF) {
```
sumitmsft: Can we make 0xD800, 0xDFFF, 0x10FFFF named constants?

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -54,12 +54,28 @@ using namespace pybind11::literals;
...
+            // Skip invalid lone surrogates (shouldn't occur in well-formed wchar_t strings)
```
sumitmsft: Can you add a test case for "Hello 🚀 World" and "admin🔒user" string and see if there is any corruption happening? Just wanted to check if this goes through and that we do not drop drop surrogates. 

If the surrogates are being dropped, then there could be a chance for SQL injection with these type of queries:

"SELECT * FROM users WHERE id = 1🚀' OR '1'='1"

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -275,14 +275,12 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
 
                 // Reserve space and convert from wstring to SQLWCHAR array
                 sqlwcharBuffer->resize(strParam->size() + 1, 0); // +1 for null terminator
+                std::vector<SQLWCHAR> utf16 = WStringToSQLWCHAR(*strParam);
+                sqlwcharBuffer->assign(utf16.begin(), utf16.end());
```
sumitmsft: WStringToSQLWCHAR() can return vectors of unpredictable size because UTF-16 conversion may expand data significantly and surrogate pairs double the storage requirements for emoji (as emojis needs 2 code units in UTF-16). 

But, `vector::assign()` doesn't validate capacity and will attempt to store all elements regardless of buffer size. 

If utf16.size() > sqlwcharBuffer then we will hit buffer overflow. 

There is no error handling or logging for size mismatches

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1528,7 +1526,10 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
 						if (numCharsInData < dataBuffer.size()) {
                             // SQLGetData will null-terminate the data
 #if defined(__APPLE__) || defined(__linux__)
-                            row.append(SQLWCHARToWString(dataBuffer.data(), SQL_NTS));
+                            auto raw_bytes = reinterpret_cast<const char*>(dataBuffer.data());
+                            py::bytes py_bytes(raw_bytes, dataLen);
```
sumitmsft, line 1719: Currently, `py::bytes py_bytes(raw_bytes, dataLen)` does not validate that dataLen fits within the allocated buffer. If the dataLen is larger than the real buffer, this can leak memory beyond the intended area—exposing potentially sensitive data (like passwords, other users’ data, etc.) to Python. 

Let's check if dataLen is non-negative and does not exceed the actual buffer size before creating the Python bytes object.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -54,12 +54,28 @@ using namespace pybind11::literals;
...
+            continue;
+        } else if (codePoint <= 0xFFFF) {
+            result.push_back(static_cast<SQLWCHAR>(codePoint));
+        } else if (codePoint <= 0x10FFFF) {
+            // Encode as surrogate pair
+            codePoint -= 0x10000;
+            SQLWCHAR highSurrogate = static_cast<SQLWCHAR>((codePoint >> 10) + 0xD800);
```
sumitmsft: The function assumes that each `wchar_t wc` from the input contains a valid Unicode Codepoint, but there's no validation that the wchar_t values are actually valid Unicode.

Since, wchar_t could be different on Win (UTF 16) and Linux (UTF32) it might cause issues. 

wstring may contain invalid or malformed codepoint and may cause issues later on.

mssql_python/pybind/ddbc_bindings.h
gargsaumya: Verified the behavior, it worked as expected. Also added those scenarios to the test suite.

bewithgaurav (dismissed):

mssql_python/cursor.py
```diff
@@ -207,6 +207,14 @@ def _get_numeric_data(self, param):
         numeric_data.val = val
         return numeric_data
 
+    def _calculate_utf16_length(self, param: str) -> int:
+        """Return UTF-16 code unit length of a Python string."""
+        try:
+            return len(param.encode("utf-16-le")) // 2
+        except UnicodeEncodeError as e:
```
sumitmsft: Saw two issues:
1. On UnicodeEncodeError we log the full parameter value with %r. This may leak sensitive user data (PII, secrets) into logs. 
2. Also the fallback returns len(param) which can undercount UTF-16 code units for non-BMP chars, potentially mis-sizing buffers. 

I see this as a risk of data leakage in logs + possible length underestimation for strings containing surrogate pairs.

sumitmsft (approved):

bewithgaurav (approved):

PR #167, opened 2025-08-05
FEAT: Adding cursor.lastrowid attribute

jahnvi480:

> AB#34897

### Summary   
This pull request introduces support for the `lastrowid` property in the `Cursor` class, allowing users to retrieve the ID of the last inserted row after a single `INSERT` operation. The implementation ensures that `lastrowid` is set only for single-row inserts and is reset or set to `None` for other types of operations, in line with DB-API semantics. Comprehensive tests are added to verify correct behavior in various scenarios.

**Core feature addition:**

* Added a read-only `lastrowid` property to the `Cursor` class, which tracks the ID of the last inserted row for single-row `INSERT` operations. The property is set to `None` for other operations or if the database does not support identity columns. 

**Implementation details:**

* In the `execute` method, `lastrowid` is reset at the start of each execution. After a successful single-row `INSERT`, the code queries `@@IDENTITY` to obtain the new row ID. For multi-row inserts or non-`INSERT` statements, `lastrowid` remains `None`. 
* In the `executemany` method, `lastrowid` is explicitly reset to `None`, as its semantics are undefined for batch operations. 

**Testing and validation:**

* Added extensive tests to cover `lastrowid` behavior for single inserts, multiple inserts, batch inserts (`executemany`), non-`INSERT` operations, tables without identity columns, and to verify that `lastrowid` is read-only.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -73,6 +73,25 @@ class Cursor:
         # Is a list instead of a bool coz bools in Python are immutable.
         # Hence, we can't pass around bools by reference & modify them.
         # Therefore, it must be a list with exactly one bool element.
+        self._lastrowid = None  # Track last inserted row ID
+
+    @property
+    def lastrowid(self):
+        """
+        Read-only attribute that provides the rowid of the last modified row.
+        
+        This attribute provides the rowid of the last modified row (most databases 
+        return a rowid only when a single INSERT operation is performed). If the 
+        operation does not set a rowid or if the database does not support rowids, 
+        this attribute should be set to None.
+        
+        The semantics of lastrowid are undefined in case the last executed statement 
+        modified more than one row, e.g. when using INSERT with executemany().
+        
+        Returns:
+            int or None: The last inserted row ID, or None if not applicable
+        """
+        return self._lastrowid
 
     def _is_unicode_string(self, param):
         """
@@ -613,6 +632,42 @@ class Cursor:
         # Initialize description after execution
         self._initialize_description()
 
+        # Reset lastrowid at the start of each execute
+        self._lastrowid = None
+
+        # Check if this was a single INSERT operation that affected exactly one row
+        if (self.rowcount == 1 and 
+            operation.strip().upper().startswith('INSERT') and
+            not any(keyword in operation.upper() for keyword in ['SELECT', '),('])):
+            try:
+                # Use @@IDENTITY which persists across statement boundaries
+                identity_query = "SELECT @@IDENTITY"
+                
+                ret = ddbc_bindings.DDBCSQLExecute(
+                    self.hstmt,
+                    identity_query,
+                    [],
+                    [],
+                    [False],  # Don't prepare this simple query
+                    False,    # Use SQLExecDirectW
+                )
+                
+                # Check if the execution was successful
+                if ret == ddbc_sql_const.SQL_SUCCESS.value or ret == ddbc_sql_const.SQL_SUCCESS_WITH_INFO.value:
+                    # Fetch the result
+                    row_data = []
+                    fetch_ret = ddbc_bindings.DDBCSQLFetchOne(self.hstmt, row_data)
+                    
+                    if (fetch_ret == ddbc_sql_const.SQL_SUCCESS.value and 
+                        row_data and row_data[0] is not None):
+                        self._lastrowid = int(row_data[0])
+                
+            except Exception:
+                # If we can't get the identity, leave lastrowid as None
+                self._lastrowid = None
+        
+        return self
+
     @staticmethod
     def _select_best_sample_value(column):
         """
@@ -682,6 +737,9 @@ class Cursor:
         param_info = ddbc_bindings.ParamInfo
         param_count = len(seq_of_parameters[0])
         parameters_type = []
+
+        # Reset lastrowid - executemany semantics are undefined for lastrowid
+        self._lastrowid = None
 
         for col_index in range(param_count):
             column = [row[col_index] for row in seq_of_parameters]
--- a/tests/test_004_cursor.py
+++ b/tests/test_004_cursor.py
@@ -1313,6 +1313,145 @@ def test_row_column_mapping(cursor, db_connection):
         cursor.execute("DROP TABLE #pytest_row_test")
         db_connection.commit()
 
+def test_lastrowid_single_insert(cursor, db_connection):
+    """Test lastrowid with single INSERT operation"""
+    try:
+        # Create table with identity column
+        cursor.execute("CREATE TABLE #test_lastrowid (id INT IDENTITY(1,1) PRIMARY KEY, name VARCHAR(50))")
+        db_connection.commit()
+        
+        # Test initial state
+        assert cursor.lastrowid is None, "lastrowid should initially be None"
+        
+        # Single INSERT should set lastrowid
+        cursor.execute("INSERT INTO #test_lastrowid (name) VALUES (?)", ["test1"])
+        db_connection.commit()
+        
+        assert cursor.lastrowid is not None, "lastrowid should not be None after INSERT"
+        assert isinstance(cursor.lastrowid, int), "lastrowid should be an integer"
+        assert cursor.lastrowid > 0, "lastrowid should be positive"
+        
+        # Store the first ID
+        first_id = cursor.lastrowid
+        
+        # Another single INSERT should update lastrowid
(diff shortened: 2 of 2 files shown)


mssql_python/cursor.py
```diff
@@ -613,6 +632,42 @@ def execute(
...
+        # Reset lastrowid at the start of each execute
+        self._lastrowid = None
+
+        # Check if this was a single INSERT operation that affected exactly one row
+        if (self.rowcount == 1 and 
+            operation.strip().upper().startswith('INSERT') and
+            not any(keyword in operation.upper() for keyword in ['SELECT', '),('])):
+            try:
+                # Use @@IDENTITY which persists across statement boundaries
+                identity_query = "SELECT @@IDENTITY"
```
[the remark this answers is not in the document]
jahnvi480: Tried this but SCOPE_IDENTITY() only works within the same scope/batch.
That is the reason I used @@IDENTITY, even pyodbc is using IDENTITY and not SCOPE_IDENTITY()

mssql_python/cursor.py
```diff
@@ -613,6 +632,42 @@ def execute(
         # Initialize description after execution
         self._initialize_description()
 
+        # Reset lastrowid at the start of each execute
+        self._lastrowid = None
+
+        # Check if this was a single INSERT operation that affected exactly one row
+        if (self.rowcount == 1 and
```
bewithgaurav: I think we should rethink `lastrowid` implementation. Parsing SQL queries on the client side is pretty fragile—it’s easy to miss edge cases or break with slightly different valid SQL. Just checked, pyodbc doesn’t support lastrowid, probably for these reasons - pls correct me if I'm wrong.

It’s also worth noting that this is partly a limitation of SQL Server and ODBC—they don’t provide a reliable way to get the last inserted ID automatically, which is why we can just let users run SELECT SCOPE_IDENTITY()/@@IDENTITY themselves if they need it.

mssql_python/cursor.py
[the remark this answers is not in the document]
bewithgaurav: Actually its a valid comment, since `@@IDENTITY` doesn't exactly produce expected results sometimes because it goes out of scope, example 
```
-- Suppose TableA and TableB both have identity columns
-- TableA has an INSERT trigger that inserts a row into TableB

INSERT INTO TableA (name) VALUES ('foo');
SELECT @@IDENTITY;      -- Returns TableB's identity (from the trigger), NOT TableA!
SELECT SCOPE_IDENTITY();-- Returns TableA's identity (your insert)
```
I believe pyodbc has not implemented this at all - checked their code, could you double check.

bewithgaurav (commented):
There are a few limitations on ODBC w.r.t implementation and usage of this attribute, added comments in detail. I believe we can discuss this in detail and park it for now.
Additionally, this is an optional DBAPI extension - https://peps.python.org/pep-0249/#lastrowid and is not implemented by pyodbc due to the above mentioned limitations. Lets discuss this in next meeting.

mssql_python/cursor.py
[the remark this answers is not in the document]
sumitmsft: I have the same reasoning for @@IDENTITY vs SCOPE_IDENTITY(). With @@IDENTITY there are chances of getting bugs due to triggers. Let's discuss where you see the challenge @jahnvi480

sumitmsft (changes requested):
Left a comment for @@IDENTITY VS SCOPE_IDENTITY() .. Let's discuss where the challenge is @jahnvi480 `@@IDENTITY` can return unexpected results if there are triggers on the table that perform additional inserts.

mssql_python/cursor.py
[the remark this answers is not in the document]
jahnvi480: @sumitmsft @bewithgaurav Got this from Copilot

Limitations of SCOPE_IDENTITY()
SCOPE_IDENTITY() in SQL Server has several limitations you should be aware of when implementing it in your code:
- Scope Limitation: SCOPE_IDENTITY() only returns identity values from the current scope (batch or procedure). If an identity is generated in a separate scope (like a nested stored procedure call), SCOPE_IDENTITY() won't capture it.
- Connection Boundary: SCOPE_IDENTITY() is session-specific and doesn't work across different connections. This means if your application uses connection pooling and executes an INSERT followed by a SCOPE_IDENTITY() query on different connections, it may not return the expected value.
- Batch Limitation: It only works within the same batch. If you execute multiple statements or batches (separated by GO statements in T-SQL), SCOPE_IDENTITY() won't return values from previous batches.
- Immediate Execution Required: You must retrieve SCOPE_IDENTITY() immediately after the INSERT operation. If other statements that manipulate data are executed between the INSERT and SCOPE_IDENTITY() call, the value might be lost or incorrect.
- No Table Specificity: Unlike IDENT_CURRENT(), SCOPE_IDENTITY() doesn't let you specify which table's identity value you want. It simply returns the last identity value generated in the current scope, regardless of which table it was generated in.
- Asynchronous Operations: In systems using asynchronous operations, SCOPE_IDENTITY() might not behave as expected because the context and timing of execution might differ.
- Concurrency Issues: In high-concurrency environments, if multiple INSERT operations happen simultaneously, there's a risk of getting the wrong identity value if proper transaction isolation isn't maintained.
- Statelessness Challenges: For stateless web applications or microservices, relying on SCOPE_IDENTITY() can be problematic because of the connection context requirements.

mssql_python/cursor.py
```diff
@@ -1024,6 +1043,39 @@ def execute(
             self.description = None
         
         self._reset_inputsizes()  # Reset input sizes after execution
+
+        # Reset lastrowid at the start of each execute
+        self._lastrowid = None
+
+        # Check if this was a single INSERT operation that affected exactly one row
+        if self.rowcount == 1:
```
bewithgaurav: I think no need to implement this filter, can you evaluate rowcount for multiple changes and what's @@identity output for that

mssql_python/cursor.py
```diff
@@ -1024,6 +1043,39 @@ def execute(
...
+                    row_data = []
+                    fetch_ret = ddbc_bindings.DDBCSQLFetchOne(self.hstmt, row_data)
+                    
+                    if (fetch_ret == ddbc_sql_const.SQL_SUCCESS.value and 
+                        row_data and row_data[0] is not None):
+                        self._lastrowid = int(row_data[0])
+                
+            except Exception:
+                # If we can't get the identity, leave lastrowid as None
+                self._lastrowid = None
```
bewithgaurav: need to log the exception error here

bewithgaurav: or better, can remove the  entire try except block, not needed
add a check_error() post execute/fetch calls

jahnvi480: Added log line

tests/test_004_cursor.py
```diff
@@ -11409,6 +11409,133 @@ def test_datetime_string_parameter_binding(cursor, db_connection):
         drop_table_if_exists(cursor, table_name)
         db_connection.commit()
         
+def test_lastrowid_single_insert(cursor, db_connection):
```
sumitmsft, line 11412: Can you check if these tests are added:

- Multi-values insert (should yield None).
- INSERT ... SELECT with multiple rows (should yield None).
- Insert with trigger inserting into another identity table (demonstrates @@IDENTITY vs SCOPE_IDENTITY() difference).
- Failed INSERT (constraint violation) ensuring lastrowid stays None.

bewithgaurav (approved):

mssql_python/cursor.py
```diff
@@ -1024,6 +1043,38 @@ def execute(
             self.description = None
         
         self._reset_inputsizes()  # Reset input sizes after execution
+
+        # Reset lastrowid at the start of each execute
+        self._lastrowid = None
+        
+        try:
+            # Use @@IDENTITY which persists across statement boundaries
+            identity_query = "SELECT @@IDENTITY"
```
sumitmsft, line 1052: Fetching @@IDENTITY before we execute the query is not as per DBAPI standards,
 
- Execute the user’s INSERT statement first (the statement that is supposed to generate the identity).
- Only after that statement has completed successfully (and you know it affected exactly one row) retrieve the identity value that that INSERT produced.
- Then set `cursor.lastrowid` to that value (or leave it None if conditions are not met).

sumitmsft:
@jahnvi480 @bewithgaurav - For `lastrowid` I don't feel confident to ship in GA unless we have a concrete reason to ship it. My reservations are primarily due to non-compliance to DBAPI and the fact that there is a huge possibility of it returning stale or incorrect data.

PR #176, opened 2025-08-13
FEAT: Support for streaming large parameters in execute()

gargsaumya:

> [AB#33395](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/33395)

### Summary   
  
This pull request adds support for streaming large parameters to SQL Server using ODBC's Data At Execution (DAE) mechanism, particularly for long Unicode strings and binary data. The changes update both the Python and C++ layers to correctly identify large parameters, mark them for DAE, and handle the streaming process during execution. Additional refactoring improves parameter type mapping and memory handling for these cases.

**Large parameter streaming (DAE) support:**

* Updated the `_map_sql_type` method in `cursor.py` to return an `is_dae` flag for parameters that require streaming (e.g., long Unicode strings, long binary data), and to calculate the correct size for Unicode strings using UTF-16 encoding. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R227-R231) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R329-R366) [[3]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R376-R383) [[4]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R393-R419)
* Modified `_create_parameter_types_list` in `cursor.py` to set DAE-related fields (`isDAE`, `strLenOrInd`, `dataPtr`) in the parameter info when streaming is needed.

**C++ bindings and execution logic:**

* Extended the `ParamInfo` struct and its Python bindings to include DAE fields (`isDAE`, `strLenOrInd`, `dataPtr`) for use during parameter binding and streaming. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L48-R51) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2509-R2556)
* Added ODBC DAE API function pointers (`SQLParamData`, `SQLPutData`) and integrated their loading and usage into the driver handle setup. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R141-R144) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R785-R787) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L799-R798) [[4]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635R121-R123)
* Refactored parameter binding and execution logic in `ddbc_bindings.cpp` to handle DAE parameters: if a parameter is marked for DAE, the code enters a loop to stream the data in chunks using `SQLParamData` and `SQLPutData`. This is done for large Unicode strings and (potentially) binary data. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R254-R288) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R997-L1007)

**Build and warning improvements:**

* Adjusted MSVC compiler flags to remove the "treat warnings as errors" option, making builds less strict on warnings.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -437,7 +437,13 @@ class Cursor:
             )
 
         # For safety: unknown/unhandled Python types should not silently go to SQL
-        raise TypeError("Unsupported parameter type: The driver cannot safely convert it to a SQL type.")
+        # raise TypeError("Unsupported parameter type: The driver cannot safely convert it to a SQL type.")
+        return (
+            ddbc_sql_const.SQL_VARCHAR.value,
+            ddbc_sql_const.SQL_C_CHAR.value,
+            len(str(param)),
+            0,
+        )
 
     def _initialize_cursor(self) -> None:
         """
@@ -793,6 +799,7 @@ class Cursor:
         except Exception as e:
             log('warning', "Execute failed, resetting cursor: %s", e)
             self._reset_cursor()
+            raise
 
         
         # Capture any diagnostic messages (SQL_SUCCESS_WITH_INFO, etc.)
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -233,10 +233,11 @@ class Cursor:
                 ddbc_sql_const.SQL_C_DEFAULT.value,
                 1,
                 0,
+                False,
             )
 
         if isinstance(param, bool):
-            return ddbc_sql_const.SQL_BIT.value, ddbc_sql_const.SQL_C_BIT.value, 1, 0
+            return ddbc_sql_const.SQL_BIT.value, ddbc_sql_const.SQL_C_BIT.value, 1, 0, False
 
         if isinstance(param, int):
             if 0 <= param <= 255:
@@ -245,6 +246,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_TINYINT.value,
                     3,
                     0,
+                    False,
                 )
             if -32768 <= param <= 32767:
                 return (
@@ -252,6 +254,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_SHORT.value,
                     5,
                     0,
+                    False,
                 )
             if -2147483648 <= param <= 2147483647:
                 return (
@@ -259,12 +262,14 @@ class Cursor:
                     ddbc_sql_const.SQL_C_LONG.value,
                     10,
                     0,
+                    False,
                 )
             return (
                 ddbc_sql_const.SQL_BIGINT.value,
                 ddbc_sql_const.SQL_C_SBIGINT.value,
                 19,
                 0,
+                False,
             )
 
         if isinstance(param, float):
@@ -273,6 +278,7 @@ class Cursor:
                 ddbc_sql_const.SQL_C_DOUBLE.value,
                 15,
                 0,
+                False,
             )
 
         if isinstance(param, decimal.Decimal):
@@ -284,6 +290,7 @@ class Cursor:
                 ddbc_sql_const.SQL_C_NUMERIC.value,
                 parameters_list[i].precision,
                 parameters_list[i].scale,
+                False,
             )
 
         if isinstance(param, str):
@@ -297,6 +304,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_WCHAR.value,
                     len(param),
                     0,
+                    False,
                 )
 
             # Attempt to parse as date, datetime, datetime2, timestamp, smalldatetime or time
@@ -309,6 +317,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_TYPE_DATE.value,
                     10,
                     0,
+                    False,
                 )
             if self._parse_datetime(param):
                 parameters_list[i] = self._parse_datetime(param)
@@ -317,6 +326,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_TYPE_TIMESTAMP.value,
                     26,
                     6,
+                    False,
                 )
             if self._parse_time(param):
                 parameters_list[i] = self._parse_time(param)
@@ -325,6 +335,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_TYPE_TIME.value,
                     8,
                     0,
+                    False,
                 )
 
             # String mapping logic here
@@ -338,12 +349,14 @@ class Cursor:
                         ddbc_sql_const.SQL_C_WCHAR.value,
                         utf16_len,
                         0,
+                        True,
                     )
                 return (
                     ddbc_sql_const.SQL_LONGVARCHAR.value,
                     ddbc_sql_const.SQL_C_CHAR.value,
                     len(param),
                     0,
+                    True,
                 )
             if is_unicode:  # Short Unicode strings
                 utf16_len = len(param.encode("utf-16-le")) // 2
@@ -352,12 +365,14 @@ class Cursor:
                     ddbc_sql_const.SQL_C_WCHAR.value,
                     utf16_len,
                     0,
+                    False,
                 )
             return (
                 ddbc_sql_const.SQL_VARCHAR.value,
                 ddbc_sql_const.SQL_C_CHAR.value,
                 len(param),
                 0,
+                False,
             )
 
         if isinstance(param, bytes):
@@ -367,12 +382,14 @@ class Cursor:
                     ddbc_sql_const.SQL_C_BINARY.value,
                     len(param),
                     0,
+                    True,
                 )
             return (
                 ddbc_sql_const.SQL_BINARY.value,
                 ddbc_sql_const.SQL_C_BINARY.value,
                 len(param),
                 0,
+                False,
             )
 
         if isinstance(param, bytearray):
@@ -382,13 +399,25 @@ class Cursor:
                     ddbc_sql_const.SQL_C_BINARY.value,
                     len(param),
                     0,
+                    True,
                 )
             return (
                 ddbc_sql_const.SQL_BINARY.value,
                 ddbc_sql_const.SQL_C_BINARY.value,
                 len(param),
                 0,
+                False,
             )
+        
+        # if isinstance(param, (bytes, bytearray)):
+        #     is_large = len(param) > 8000
+        #     return (
+        #         ddbc_sql_const.SQL_VARBINARY.value if is_large else ddbc_sql_const.SQL_BINARY.value,
+        #         ddbc_sql_const.SQL_C_BINARY.value,
+        #         len(param),
+        #         0,
+        #         is_large,
+        #     )
 
         if isinstance(param, datetime.datetime):
             return (
@@ -396,6 +425,7 @@ class Cursor:
                 ddbc_sql_const.SQL_C_TYPE_TIMESTAMP.value,
                 26,
                 6,
+                False,
             )
 
         if isinstance(param, datetime.date):
@@ -404,6 +434,7 @@ class Cursor:
                 ddbc_sql_const.SQL_C_TYPE_DATE.value,
                 10,
                 0,
+                False,
             )
 
         if isinstance(param, datetime.time):
@@ -412,6 +443,7 @@ class Cursor:
                 ddbc_sql_const.SQL_C_TYPE_TIME.value,
                 8,
                 0,
+                False,
             )
 
         return (
@@ -419,6 +451,7 @@ class Cursor:
             ddbc_sql_const.SQL_C_CHAR.value,
             len(str(param)),
             0,
+            False,
         )
 
     def _initialize_cursor(self) -> None:
@@ -495,7 +528,7 @@ class Cursor:
             paraminfo.
         """
         paraminfo = param_info()
-        sql_type, c_type, column_size, decimal_digits = self._map_sql_type(
+        sql_type, c_type, column_size, decimal_digits, is_dae = self._map_sql_type(
             parameter, parameters_list, i
         )
         paraminfo.paramCType = c_type
@@ -503,6 +536,12 @@ class Cursor:
         paraminfo.inputOutputType = ddbc_sql_const.SQL_PARAM_INPUT.value
         paraminfo.columnSize = column_size
         paraminfo.decimalDigits = decimal_digits
+        paraminfo.isDAE = is_dae
+
+        if is_dae:
+            paraminfo.strLenOrInd = -1  # Tells ODBC this is streamed data
+            paraminfo.dataPtr = parameter  # Will be converted to py::object* in C++
+
         return paraminfo
 
     def _initialize_description(self):
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -233,7 +233,7 @@ class Cursor:
             return (
                 ddbc_sql_const.SQL_VARCHAR.value, # TODO: Add SQLDescribeParam to get correct type
                 ddbc_sql_const.SQL_C_DEFAULT.value,
-                1,
+                0,
                 0,
                 False,
             )
@@ -513,6 +513,18 @@ class Cursor:
             paraminfo.
         """
         paraminfo = param_info()
+        # Explicit None handling
+        if parameter is None:
+            paraminfo.paramSQLType = ddbc_sql_const.SQL_VARCHAR.value
+            paraminfo.paramCType = ddbc_sql_const.SQL_C_CHAR.value
+            paraminfo.columnSize = 0
+            paraminfo.decimalDigits = 0
+            paraminfo.isDAE = False
+            paraminfo.inputOutputType = ddbc_sql_const.SQL_PARAM_INPUT.value
+            paraminfo.strLenOrInd = ddbc_sql_const.SQL_NULL_DATA.value
+            paraminfo.dataPtr = None
+            return paraminfo
+
         sql_type, c_type, column_size, decimal_digits, is_dae = self._map_sql_type(
             parameter, parameters_list, i
         )
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -18,6 +18,8 @@ from mssql_python import ddbc_bindings
 from mssql_python.exceptions import InterfaceError, NotSupportedError, ProgrammingError
 from .row import Row
 
+# Constants for string handling
+MAX_INLINE_CHAR = 4000  # NVARCHAR/VARCHAR inline limit; this triggers NVARCHAR(MAX)/VARCHAR(MAX) + DAE
 
 class Cursor:
     """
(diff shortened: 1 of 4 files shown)


mssql_python/cursor.py
```diff
@@ -316,37 +326,44 @@ def _map_sql_type(self, param, parameters_list, i):
                     ddbc_sql_const.SQL_C_TYPE_TIME.value,
                     8,
                     0,
+                    False,
                 )
 
             # String mapping logic here
             is_unicode = self._is_unicode_string(param)
             # TODO: revisit
             if len(param) > 4000:  # Long strings
```
sumitmsft: The threshold 4000 is hard‑coded. Please extract a named constant (e.g. MAX_INLINE_CHAR = 4000) with a comment explaining its relationship to NVARCHAR(4000)/VARCHAR(4000) vs switching to (N)[VAR]CHAR(MAX) + Data At Execution. This avoids future drift if the threshold changes.

mssql_python/cursor.py
```diff
@@ -224,10 +224,11 @@ def _map_sql_type(self, param, parameters_list, i):
                 ddbc_sql_const.SQL_C_DEFAULT.value,
                 1,
                 0,
+                False,
```
sumitmsft, line 238: This just caught my eyes. @gargsaumya please validate this and let me know if my understanding is wrong:

In `_map_sql_type()`:
Current None handling returns a VARCHAR with size 1 (and SQL_C_DEFAULT) but does not ensure a true SQL NULL is sent. This risks inserting the literal `None` or an empty value instead of `NULL` semantics (affecting `IS NULL` logic, defaults, constraints). 

Instead what must do: Return a neutral SQL type with column_size = 0 (not 1) and an explicit C type (e.g. SQL_C_CHAR) for clarity. In `_create_parameter_types_list` set `paraminfo.strLenOrInd = SQL_NULL_DATA` (ODBC sentinel) for `None` parameters. 

Finally, add tests to distinguish None vs "" vs "None" across text and numeric columns.

In that case you'll have to make changes to `_create_parameter_types_list` as well:

if parameter is None:
    paraminfo.strLenOrInd = ddbc_sql_const.SQL_NULL_DATA.value  # typically -1

This ensures correct NULL propagation and avoids silent data corruption.

Can you please validate this?? I guess this has to potential to cause data corruption on SQL.

mssql_python/cursor.py
```diff
@@ -316,37 +326,44 @@ def _map_sql_type(self, param, parameters_list, i):
...
                     0,
+                    False,
                 )
 
             # String mapping logic here
             is_unicode = self._is_unicode_string(param)
             # TODO: revisit
             if len(param) > 4000:  # Long strings
                 if is_unicode:
+                    utf16_len = len(param.encode("utf-16-le")) // 2
```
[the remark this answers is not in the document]
sumitmsft: Spoke to @gargsaumya offline, and to both of us this make sense. I drilled a bit deeper and found some concerning picture. 

Unicode length calculation, classification, and streaming prep are intermingled. `_map_sql_type` encodes every large (and even small) string to `UTF‑16 LE` (`param.encode("utf-16-le")`) just to derive `utf16_len`, then later (in C++ streaming) we re-encode the same string again. This has three downsides:

1. Large values are encoded twice (wasted CPU + temporary memory).
2. Unnecessary work for small values: Even non‑streamed values pay the cost of a full encode when a simple len(param) check would suffice to decide threshold.
3. Type classification (which is cheaper op) + expensive preparation (encoding) live together, making code work harder.

Saumya and I have agreed for following to test: 

1. In `_map_sql_type`: Use `len(param)` to decide base sql/c types and whether it should be `is_dae`. At this point we can defer any actual `UTF‑16` encoding (and code unit length calculation) to `_create_parameter_types_list` only for parameters flagged `is_dae` (is_dae is True).

2. Cache the encoded bytes (e.g., `paraminfo.encoded_utf16` and `paraminfo.encoded_len`). C++ then streams that buffer directly (remove second encode).

3. For non‑DAE strings, skip encoding entirely. Let normal binding handle it. 

4. if we really need code unit length to decide between NVARCHAR vs NLONGVARCHAR, compute only once in the DAE path.

mssql_python/cursor.py
```diff
@@ -316,37 +326,44 @@ def _map_sql_type(self, param, parameters_list, i):
...
                 )
             return (
                 ddbc_sql_const.SQL_VARCHAR.value,
                 ddbc_sql_const.SQL_C_CHAR.value,
                 len(param),
                 0,
+                False,
             )
 
         if isinstance(param, bytes):
```
sumitmsft, line 378: You mark large bytes / bytearray (>8000) with is_dae=True, but C++ currently only implements DAE for Unicode strings (raises “DAE only supported for str or bytes” but then lacks bytes handling). Can you please validate it?

mssql_python/cursor.py
```diff
@@ -401,13 +424,15 @@ def _map_sql_type(self, param, parameters_list, i):
                 ddbc_sql_const.SQL_C_TYPE_TIME.value,
                 8,
                 0,
+                False,
             )
 
         return (
```
sumitmsft: Again, I was just going through the code and encountered this code path. I remember I had a discussion with Ram on this issue way back in Feb'25 and we planned to make sure driver doesn’t crash or throw an error when encountering unknown types. Hence, by default unknown type will be sent as VARVCHAR to SQL. 

But, now I feel this was a wrong decision because that means literally any unexpected object can slip through silently to SQL and database stores a meaningless string.  

I feel we should raise `TypeError` to avoid hidden bugs, huge unintended payloads, and leaking internal representations.

mssql_python/cursor.py
```diff
@@ -316,37 +326,44 @@ def _map_sql_type(self, param, parameters_list, i):
                     ddbc_sql_const.SQL_C_TYPE_TIME.value,
                     8,
                     0,
+                    False,
                 )
 
             # String mapping logic here
             is_unicode = self._is_unicode_string(param)
             # TODO: revisit
```
sumitmsft: Can we remove this TODO? I guess it is taken care of already?

mssql_python/cursor.py
```diff
@@ -470,14 +495,20 @@ def _create_parameter_types_list(self, parameter, param_info, parameters_list, i
...
         )
         paraminfo.paramCType = c_type
         paraminfo.paramSQLType = sql_type
         paraminfo.inputOutputType = ddbc_sql_const.SQL_PARAM_INPUT.value
         paraminfo.columnSize = column_size
         paraminfo.decimalDigits = decimal_digits
+        paraminfo.isDAE = is_dae
+
+        if is_dae:
+            paraminfo.strLenOrInd = -1  # Tells ODBC this is streamed data
```
sumitmsft: The DAE setup uses a magic number (-1) for `paraminfo.strLenOrInd` in Python here:
`
//python
paraminfo.strLenOrInd = -1  # Tells ODBC this is streamed data
`
And then the C++ layer overwrites it with SQL_LEN_DATA_AT_EXEC(0), here:
`
//C++
 *strLenOrIndPtr = SQL_LEN_DATA_AT_EXEC(0);
`
This creates a hidden contract and is potentially confusing because:

- In `-1` typically means `SQL_NULL_DATA` (NULL), not “stream later.” Reusing it for DAE is ambiguous and risks collision with real NULL handling.
- Two layers mutate the same meaning (Python assigns a placeholder; C++ silently replaces). Future refactors might remove the overwrite and break streaming or mis-handle NULL.
- ODBC already defines explicit constructs: `SQL_DATA_AT_EXEC` or (preferably) `SQL_LEN_DATA_AT_EXEC(n)`. Use those directly and consistently.

Instead, I propose the following (you may change the approach though):

- Expose and use a named constant (e.g., `ddbc_sql_const.SQL_LEN_DATA_AT_EXEC_0` or a helper) instead of -1.
- Set the final `StrLenOrInd` value in a single place (either Python or C++, not both).
- Reserve `SQL_NULL_DATA` (usually -1) strictly for `NULL` parameters.
- (Optional) If total length is known for DAE (after one encoding pass), use `SQL_LEN_DATA_AT_EXEC(total_len_units)` for clarity.

We can remove the C++ overwrite that forces SQL_LEN_DATA_AT_EXEC(0); instead validate:
`
// C++
assert(strLenOrInd == SQL_LEN_DATA_AT_EXEC(0) || IS_SQL_LEN_DATA_AT_EXEC(strLenOrInd));
`
We can also add a couple of test cases as well:
- A test that a DAE parameter does not get mistaken for NULL.
- A test for a real NULL parameter (SQL_NULL_DATA) to ensure paths diverge.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -244,53 +248,40 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
+                    LOG("Parameter[{}] is marked for DAE streaming", paramIndex);
+                    dataPtr = const_cast<void*>(reinterpret_cast<const void*>(&paramInfos[paramIndex]));
+                    strLenOrIndPtr = AllocateParamBuffer<SQLLEN>(paramBuffers);
+                    *strLenOrIndPtr = SQL_LEN_DATA_AT_EXEC(0);
+                    bufferLength = 0;
+                } else {
+                    // Normal small-string case
+                    std::wstring* strParam =
+                        AllocateParamBuffer<std::wstring>(paramBuffers, param.cast<std::wstring>());
+                    LOG("SQL_C_WCHAR Parameter[{}]: Length={}, Content='{}'",
```
sumitmsft: Seems we are logging raw params here. We shouldn't do this. 

Logging raw (or partially truncated) parameter contents here can leak PII or secrets (passwords, tokens, API keys etc.). Truncation does not guarantee safety; prefixes are often sensitive and logs replicate to multiple systems. 

Let's take safe-by-default approach:

- Remove content logging by default; log only metadata (index, sql type, length, NULL, DAE) or whatever is necessary from metadata perspective can be logged. 
- I would recommend an addition of explicit OPT-IN debug flag/env var (e.g. DDBC_ENABLE_PARAMETER_CONTENT_LOG=1) to allow local diagnostic content logging upon customer's consent only. It is their choice to enable or disable. We will publicly mention it. 
When this flag is enabled:
- Mask\REDACT by default: password|passwd|token|secret|key|auth (case-insensitive).
- For binary data: never print bytes; show length and maybe a hash (SHA256 prefix).
- Emit a one-time WARNING if content logging is enabled so users know the risk.
- We can discuss more on what to be revealed and what to be masked ... Let's discuss in our scrum. 
-
Add tests (or at least manual instructions) ensuring content is absent when flag off and masked when on.

mssql_python/cursor.py
```diff
@@ -356,12 +373,14 @@ def _map_sql_type(self, param, parameters_list, i):
                     ddbc_sql_const.SQL_C_BINARY.value,
                     len(param),
                     0,
+                    True,
```
sumitmsft: @gargsaumya I want you think about the possibility of Denial of Service (DoS) attack. We may have extremely large parameters (huge str or bytes) are accepted without any size validation. 

// Perhaps we can backlog it, but important to consider. 

For large strings we 
(a) keep the original Python object, then 
(b) fully encode it to UTF‑16 LE (duplicating memory), and 
(c) C++ may handle additional buffers when streaming. 

A single massive parameter can therefore transiently occupy 2–3x its size in RAM. An attacker (or accidental misuse) could submit multi‑hundred‑MB/GB values and force OOM or severe GC pressure.

We need to think about the guard rails - possibly check with JDBC and .NET teams how are they tackling this situation? 
Few that comes to my mind:

- Per-parameter cap: e.g. MAX_PARAMETER_CHARS / MAX_PARAMETER_BYTES; raise ValueError early: "Parameter too large (X bytes > limit Y). Enable streaming or increase limit."
- Total parameters cap per execute: sum(sizes) <= MAX_TOTAL_PARAMETER_BYTES.
But both of them are like limiting the things customers can do with SQL Server and we will simply creating a wall in between. But we can certainly park them, if required use them.

- Tests to be added:
Parameter just below limit succeeds.
Parameter above limit raises clear error.
Large streamed parameter does not exceed single-copy memory usage (profiling or mock to assert single encode call).
Log (DEBUG/WARN) only the size, not content, when rejecting.

mssql_python/cursor.py
sumitmsft: The streaming path (aka DAE) fully encodes the entire Python Unicode value into utf16_str before chunking. After that, you loop over utf16_str in fixed-size slices and call SQLPutData with each chunk. This is what I understood. However @gargsaumya will confirm this. 

If this is true then: 

For very large strings this doubles memory (original str + full UTF‑16 bytes) and defeats part of the purpose of streaming (reduced peak memory). It also introduces a large up‑front allocation and copy. This also incurs latency and DoS risk as well. 

Can we get to an implementation where the peak memory ~ original Python str + one chunk buffer (vs entire duplicate encoded buffer)? This way we will be able to keep memory under control?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -995,16 +989,62 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
+                        ThrowStdException("Error encoding string to UTF-16: " + std::string(e.what()));
+                    }
+                    const char* dataPtr = utf16_str.data();
+                    size_t totalBytes = utf16_str.size();
+                    const size_t chunkSize = DAE_CHUNK_SIZE;
+                    for (size_t offset = 0; offset < totalBytes; offset += chunkSize) {
+                        size_t len = std::min(chunkSize, totalBytes - offset);
+                        rc = SQLPutData_ptr(hStmt, (SQLPOINTER)(dataPtr + offset), static_cast<SQLLEN>(len));
+                        if (!SQL_SUCCEEDED(rc)) {
+                            LOG("SQLPutData failed.");
```
sumitmsft: On SQLPutData failure we return immediately without resetting parameter bindings. This can leave the statement handle in a partially executed state. I think we need to do explicit cleanup before returning, we can call:
1. SQLFreeStmt to unbind/reset parameter state
2. SQLCloseCursor if a cursor might have been opened earlier.
3. If possible, log the failure including which chunk offset failed

mssql_python/cursor.py
sumitmsft: I got to know about SQL_SS_LENGTH_UNLIMITED https://learn.microsoft.com/en-us/sql/relational-databases/native-client-odbc-api/sqldescribecol?view=sql-server-ver15 

Theoretically what I understood: 

For large Unicode parameters we set `columnSize = utf16_len` (full length). 

For “MAX” / unbounded cases SQL Server expect `columnSize = 0 (SQL_SS_LENGTH_UNLIMITED)` rather than a huge literal size. Supplying the full length may cause large pre-allocations. 

If length > INLINE_NVARCHAR_LIMIT (current 4000) and we mark is_dae=True, set columnSize to an “unlimited” sentinel (0 / SQL_SS_LENGTH_UNLIMITED) instead of the actual length.

For “normal” small to medium NVARCHAR parameters (say length <= 4000 characters) you supply columnSize = actual character count.

Again, this is what I could sense theoretically, but we need validate it through test cases

mssql_python/pybind/ddbc_bindings.cpp
sumitmsft: On SQLPutData error we log and return rc but skip diagnostics. We can use `SQLGetDiagRec` (loop over records) before cleanup/return to capture SQLSTATE, native error, and message text. Attach that detail to the raised Python exception. This greatly improves root-cause analysis vs a bare return code.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -995,16 +989,62 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
+                // Finding the paramInfo that matches the returned token
+                const ParamInfo* matchedInfo = nullptr;
+                for (auto& info : paramInfos) {
+                    if (reinterpret_cast<SQLPOINTER>(const_cast<ParamInfo*>(&info)) == paramToken) {
+                        matchedInfo = &info;
+                        break;
+                    }
+                }
+                if (!matchedInfo) {
+                    ThrowStdException("Unrecognized paramToken returned by SQLParamData");
```
sumitmsft, line 1169: We can enhance the message with context, wherein we can include expected token (pointer/value you bound), the raw token pointer/value returned, and current parameter index or total parameter count. 

Example: 
LOG_ERROR("Unrecognized paramToken from SQLParamData: got=%p expected=%p paramIndex=%d/%d", token, expectedToken, currentIndex, totalParams). 

This aids diagnosing driver or binding mismatches.

sumitmsft (changes requested):
This is quite an important feature and very complex too. Hence, we need to make sure it's robust, performant and scalable. Have added some comments to be discussed and requesting some changes in the architecture per se.

mssql_python/cursor.py
gargsaumya: This support in C++ will be added in upcoming PR. Marked it false for now.

mssql_python/cursor.py
gargsaumya: Makes sense. Added a work Item for this - TASK 38259

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: We now handle errors from SQLExecute (including DAE streaming failures) entirely on the Python side within the cursor class. We chose Python-side handling because the cursor lifecycle and statement handle management are fully managed in Python; the C++ layer only reports errors via return codes and diagnostic information.

mssql_python/cursor.py
gargsaumya: Sure. Noting that proper None handling isn’t part of this PR, I’ve created a separate Work Item for it here-https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/38293

Apart from updates in _create_parameter_types_list and _map_sql_type, this will require adjustments to the current pytests to handle None, empty strings, and literal "None" values correctly. We’ll tackle this separately in upcoming PR, to keep this PR focused.

sumitmsft (approved):

bewithgaurav (approved):

PR #177, opened 2025-08-13
FEAT: Adding set_attr in connection class

jahnvi480:

> [AB#39058](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39058)

### Summary   
This pull request introduces comprehensive support for setting ODBC connection attributes in the `mssql_python` library, aligning its functionality with pyodbc's `set_attr` API. The changes include new constants for connection attributes, transaction isolation levels, and related options, as well as robust error handling and input validation in both Python and C++ layers. This enables users to configure connection behavior (e.g., autocommit, isolation level, timeouts) in a standardized and secure manner.

### Connection Attribute Support

* Added a wide set of ODBC connection attribute constants, transaction isolation level constants, access mode constants, and related enums to `mssql_python/__init__.py` and `mssql_python/constants.py`, making them available for use in Python code. 
* Implemented the `set_attr` method in the `Connection` Python class, providing pyodbc-compatible functionality for setting connection attributes with detailed input validation and error handling.

### C++ Backend Enhancements

* Exposed `setAttribute` as a public method in the C++ `Connection` class, and added a new `setAttr` method in `ConnectionHandle`, with improved error reporting and range validation for SQLUINTEGER values.
* Registered the new `set_attr` method with the Python bindings, making it accessible from Python code.

### Code Cleanup and Refactoring

* Moved and consolidated connection attribute constants in `ConstantsDDBC` to improve maintainability, and removed legacy/unused constants.

These changes provide a robust interface for configuring ODBC connection attributes, improve compatibility with pyodbc, and enhance error handling for attribute operations.

### DIFF
--- a/mssql_python/__init__.py
+++ b/mssql_python/__init__.py
@@ -5,6 +5,8 @@ This module initializes the mssql_python package.
 """
 import threading
 import locale
+import sys
+import types
 
 # Exceptions
 # https://www.python.org/dev/peps/pep-0249/#exceptions
@@ -31,6 +33,7 @@ class Settings:
         self.lowercase = False
         # Use the pre-determined separator - no locale access here
         self.decimal_separator = _DEFAULT_DECIMAL_SEPARATOR
+        self.native_uuid = False  # Default to False for backwards compatibility
 
 # Global settings instance
 _settings = Settings()
@@ -39,10 +42,7 @@ _settings_lock = threading.Lock()
 def get_settings():
     """Return the global settings object"""
     with _settings_lock:
-        _settings.lowercase = lowercase
         return _settings
-
-lowercase = _settings.lowercase  # Default is False
 
 # Set the initial decimal separator in C++
 from .ddbc_bindings import DDBCSetDecimalSeparator
@@ -188,21 +188,7 @@ def pooling(max_size=100, idle_timeout=600, enabled=True):
     else:
         PoolingManager.enable(max_size, idle_timeout)
 
-import sys
 _original_module_setattr = sys.modules[__name__].__setattr__
-
-def _custom_setattr(name, value):
-    if name == 'lowercase':
-        with _settings_lock:
-            _settings.lowercase = bool(value)
-            # Update the module's lowercase variable
-            _original_module_setattr(name, _settings.lowercase)
-    else:
-        _original_module_setattr(name, value)
-
-# Replace the module's __setattr__ with our custom version
-sys.modules[__name__].__setattr__ = _custom_setattr
-
 
 # Export SQL constants at module level
 SQL_CHAR = ConstantsDDBC.SQL_CHAR.value
@@ -279,3 +265,45 @@ def get_info_constants():
     """
     return {name: member.value for name, member in GetInfoConstants.__members__.items()}
 
+# Create a custom module class that uses properties instead of __setattr__
+class _MSSQLModule(types.ModuleType):
+    @property
+    def native_uuid(self):
+        return _settings.native_uuid
+
+    @native_uuid.setter
+    def native_uuid(self, value):
+        if not isinstance(value, bool):
+            raise ValueError("native_uuid must be a boolean value")
+        with _settings_lock:
+            _settings.native_uuid = value
+
+    @property
+    def lowercase(self):
+        return _settings.lowercase
+
+    @lowercase.setter
+    def lowercase(self, value):
+        if not isinstance(value, bool):
+            raise ValueError("lowercase must be a boolean value")
+        with _settings_lock:
+            _settings.lowercase = value
+
+# Replace the current module with our custom module class
+old_module = sys.modules[__name__]
+new_module = _MSSQLModule(__name__)
+
+# Copy all existing attributes to the new module
+for attr_name in dir(old_module):
+    if attr_name != "__class__":
+        try:
+            setattr(new_module, attr_name, getattr(old_module, attr_name))
+        except AttributeError:
+            pass
+
+# Replace the module in sys.modules
+sys.modules[__name__] = new_module
+
+# Initialize property values
+lowercase = _settings.lowercase
+native_uuid = _settings.native_uuid
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -1034,29 +1034,37 @@ class Cursor:
         # After successful execution, initialize description if there are results
         column_metadata = []
         try:
+            # ODBC specification guarantees that column metadata is available immediately after
+            # a successful SQLExecute/SQLExecDirect for the first result set
             ddbc_bindings.DDBCSQLDescribeCol(self.hstmt, column_metadata)
             self._initialize_description(column_metadata)
         except Exception as e:
             # If describe fails, it's likely there are no results (e.g., for INSERT)
             self.description = None
-        
+
         # Reset rownumber for new result set (only for SELECT statements)
         if self.description:  # If we have column descriptions, it's likely a SELECT
+            # Capture settings snapshot for this result set
+            settings = get_settings()
+            self._settings_snapshot = {
+                'lowercase': settings.lowercase,
+                'native_uuid': settings.native_uuid
+            }
+            # Identify UUID columns based on Python type in description[1]
+            # This relies on _map_data_type correctly mapping SQL_GUID to uuid.UUID
+            self._uuid_indices = []
+            for i, desc in enumerate(self.description):
+                if desc and desc[1] == uuid.UUID:  # Column type code at index 1
+                    self._uuid_indices.append(i)
+                # Verify we have complete description tuples (7 items per PEP-249)
+                elif desc and len(desc) != 7:
+                    log('warning', f"Column description at index {i} has incorrect tuple length: {len(desc)}")
             self.rowcount = -1
             self._reset_rownumber()
         else:
             self.rowcount = ddbc_bindings.DDBCSQLRowCount(self.hstmt)
             self._clear_rownumber()
 
-        # After successful execution, initialize description if there are results
-        column_metadata = []
-        try:
-            ddbc_bindings.DDBCSQLDescribeCol(self.hstmt, column_metadata)
-            self._initialize_description(column_metadata)
-        except Exception as e:
-            # If describe fails, it's likely there are no results (e.g., for INSERT)
-            self.description = None
-        
         self._reset_inputsizes()  # Reset input sizes after execution
         # Return self for method chaining
         return self
@@ -1763,7 +1771,8 @@ class Cursor:
             
             # Create and return a Row object, passing column name map if available
             column_map = getattr(self, '_column_name_map', None)
-            return Row(self, self.description, row_data, column_map)
+            settings_snapshot = getattr(self, '_settings_snapshot', None)
+            return Row(self, self.description, row_data, column_map, settings_snapshot)
         except Exception as e:
             # On error, don't increment rownumber - rethrow the error
             raise e
@@ -1811,7 +1820,8 @@ class Cursor:
             
             # Convert raw data to Row objects
             column_map = getattr(self, '_column_name_map', None)
-            return [Row(self, self.description, row_data, column_map) for row_data in rows_data]
+            settings_snapshot = getattr(self, '_settings_snapshot', None)
+            return [Row(self, self.description, row_data, column_map, settings_snapshot) for row_data in rows_data]
         except Exception as e:
             # On error, don't increment rownumber - rethrow the error
             raise e
@@ -1849,7 +1859,8 @@ class Cursor:
             
             # Convert raw data to Row objects
             column_map = getattr(self, '_column_name_map', None)
-            return [Row(self, self.description, row_data, column_map) for row_data in rows_data]
+            settings_snapshot = getattr(self, '_settings_snapshot', None)
+            return [Row(self, self.description, row_data, column_map, settings_snapshot) for row_data in rows_data]
         except Exception as e:
             # On error, don't increment rownumber - rethrow the error
             raise e
--- a/mssql_python/row.py
+++ b/mssql_python/row.py
@@ -1,19 +1,13 @@
+from mssql_python import get_settings
+from mssql_python.constants import ConstantsDDBC
+import uuid
+
 class Row:
     """
-    A row of data from a cursor fetch operation. Provides both tuple-like indexing
-    and attribute access to column values.
-
-    Column attribute access behavior depends on the global 'lowercase' setting:
-    - When enabled: Case-insensitive attribute access
-    - When disabled (default): Case-sensitive attribute access matching original column names
-
-    Example:
-        row = cursor.fetchone()
-        print(row[0])           # Access by index
-        print(row.column_name)  # Access by column name (case sensitivity varies)
+    A row of data from a cursor fetch operation.
     """
     
-    def __init__(self, cursor, description, values, column_map=None):
+    def __init__(self, cursor, description, values, column_map=None, settings_snapshot=None):
         """
         Initialize a Row object with values and description.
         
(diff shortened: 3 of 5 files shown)


mssql_python/pybind/connection/connection.h
```diff
@@ -42,10 +42,15 @@ class Connection {
     // Allocate a new statement handle on this connection.
     SqlHandlePtr allocStatementHandle();
 
+    // Move setAttribute from private to public
+    SQLRETURN setAttribute(SQLINTEGER attribute, py::object value);
+
+    // Add getter for DBC handle for error reporting
+    SqlHandlePtr getDbcHandle() const { return _dbcHandle; }
```
gargsaumya: Returning SqlHandlePtr by value gives the caller ownership they probably shouldn’t have. I would suggest returning const SqlHandlePtr& instead to keep ownership internal, while still allowing use for error reporting.

mssql_python/pybind/connection/connection.h
```diff
@@ -42,10 +42,15 @@ class Connection {
     // Allocate a new statement handle on this connection.
     SqlHandlePtr allocStatementHandle();
 
+    // Move setAttribute from private to public
```
gargsaumya: This seems to be an irrelevant comment, we can remove this.

mssql_python/pybind/connection/connection.h
```diff
@@ -66,6 +71,7 @@ class ConnectionHandle {
     void setAutocommit(bool enabled);
     bool getAutocommit() const;
     SqlHandlePtr allocStatementHandle();
+    void setAttr(int attribute, py::object value);  // Add this line
```
gargsaumya: same, irrelevant comment

mssql_python/__init__.py
```diff
@@ -52,6 +52,53 @@
...
+SQL_ATTR_LOGIN_TIMEOUT = ConstantsDDBC.SQL_ATTR_LOGIN_TIMEOUT.value
+SQL_ATTR_ODBC_CURSORS = ConstantsDDBC.SQL_ATTR_ODBC_CURSORS.value
+SQL_ATTR_PACKET_SIZE = ConstantsDDBC.SQL_ATTR_PACKET_SIZE.value
+SQL_ATTR_QUIET_MODE = ConstantsDDBC.SQL_ATTR_QUIET_MODE.value
+SQL_ATTR_TXN_ISOLATION = ConstantsDDBC.SQL_ATTR_TXN_ISOLATION.value
+SQL_ATTR_TRACE = ConstantsDDBC.SQL_ATTR_TRACE.value
+SQL_ATTR_TRACEFILE = ConstantsDDBC.SQL_ATTR_TRACEFILE.value
+SQL_ATTR_TRANSLATE_LIB = ConstantsDDBC.SQL_ATTR_TRANSLATE_LIB.value
+SQL_ATTR_TRANSLATE_OPTION = ConstantsDDBC.SQL_ATTR_TRANSLATE_OPTION.value
+SQL_ATTR_CONNECTION_POOLING = ConstantsDDBC.SQL_ATTR_CONNECTION_POOLING.value
```
gargsaumya: **Have a concern here**
Have we verified that all of these constants are actually usable in our context, given that our driver calls the ODBC driver directly and does not go through an ODBC Driver Manager?

Some of these attributes are traditionally handled at the Driver Manager level (e.g. connection pooling, tracing, translation libraries, quiet mode, etc.) and won’t have any effect when set directly on the driver. For example, I know for sure that SQL_ATTR_CONNECTION_POOLING is not supported without the Driver Manager. If we still expose such constants, they may confuse users since they won’t behave as expected.

I think it’s important to either-
(a) trim this list down to attributes we know the SQL Server ODBC driver supports directly, or 
(b) very clearly document which ones are no-ops in our implementation. 

Otherwise, we risk exposing a larger API surface that looks complete but isn’t functionally correct.

tests/test_003_connection.py
```diff
@@ -1328,4 +1328,299 @@ def test_setdecoding_with_unicode_data(db_connection):
...
+        cursor.close()
+
+# ==================== SET_ATTR TEST CASES ====================
+
+def test_set_attr_constants_access():
+    """Test that connection attribute constants are accessible."""
+    
+    
+    # Test that common constants exist and have correct types
+    attr_constants = [
```
gargsaumya: Can we double-check which of these are driver-independent vs. driver-manager–dependent and make sure our tests cover that distinction? That way, we’ll avoid leaking irrelevant constants into the public API and give users a clearer, more reliable experience.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -174,8 +174,18 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
...
+        long long longValue = value.cast<long long>();
+        
+        // Validate range for SQLUINTEGER (0 to 4294967295)
+        if (longValue < 0 || longValue > 4294967295LL) {
+            LOG("Integer value out of SQLUINTEGER range: {}", longValue);
+            return SQL_ERROR;
+        }
+        
+        // Cast to SQLUINTEGER for proper handling
+        SQLUINTEGER uintValue = static_cast<SQLUINTEGER>(longValue);
```
gargsaumya: For integer attributes, we’re currently clamping values to 32-bit (SQLUINTEGER, max 4294967295). On 64-bit builds, ODBC defines these as SQLULEN (64-bit), and pyodbc supports that. Should we use SQLULEN here instead, so we don’t reject valid values on x64?

mssql_python/pybind/connection/connection.cpp
```diff
@@ -174,8 +174,18 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
     SQLINTEGER length = 0;
 
     if (py::isinstance<py::int_>(value)) {
-        int intValue = value.cast<int>();
-        ptr = reinterpret_cast<SQLPOINTER>(static_cast<uintptr_t>(intValue));
+        // Handle large integer values up to SQLUINTEGER range
+        long long longValue = value.cast<long long>();
+        
+        // Validate range for SQLUINTEGER (0 to 4294967295)
+        if (longValue < 0 || longValue > 4294967295LL) {
```
gargsaumya: The 4294967295 literal is hardcoded, it would be better to use either magic numbers or even better + safer to use std::numeric_limits<SQLUINTEGER>::max() to avoid magic numbers.

mssql_python/connection.py
```diff
@@ -520,6 +521,78 @@ def rollback(self) -> None:
...
+        # Validate attribute type and range for SQLUINTEGER compatibility
+        if not isinstance(attribute, int) or attribute < 0:
+            raise ProgrammingError("Connection attribute must be a non-negative integer", f"Invalid attribute: {attribute}")
+
+        # Validate attribute is within SQLUINTEGER range
+        if attribute > 4294967295:  # 2^32 - 1
+            raise ProgrammingError("Connection attribute must be within SQLUINTEGER range (0-4294967295)", f"Attribute out of range: {attribute}")
+
+        # Validate value type - must be integer, bytes, or bytearray
+        if not isinstance(value, (int, bytes, bytearray)):
```
gargsaumya: On the C++ side, we only support int, bytes, and bytearray as attribute values, but not str. That means attributes like SQL_ATTR_CURRENT_CATALOG or SQL_ATTR_TRACEFILE can’t be set, even though pyodbc supports them and the ODBC spec requires string values in these cases.

To be fully pyodbc-compatible, I think we should add handling for py::str in Connection::setAttribute (converting to std::wstring / SQLWCHAR*), and also allow str on the Python side. Otherwise, we’re leaving out a set of valid attributes.

gargsaumya (changes requested):

mssql_python/connection.py
sumitmsft: The set_attr method implements input validation, but please ensure all inputs—especially value when it is bytes or bytearray—are properly sanitized before being passed to the C++ layer.
Please confirm that no user-supplied data can bypass these checks, and that the sanitization logic covers all edge cases.

mssql_python/connection.py
```diff
@@ -520,6 +521,78 @@ def rollback(self) -> None:
...
+            >>> conn.set_attr(SQL_ATTR_TXN_ISOLATION, SQL_TXN_READ_COMMITTED)
+
+        Note:
+            This method is compatible with pyodbc's set_attr functionality.
+            Attribute values must be within valid SQLUINTEGER range (0 to 4294967295).
+        """
+        if self._closed:
+            raise InterfaceError("Cannot set attribute on closed connection", "Connection is closed")
+
+        # Validate attribute type and range for SQLUINTEGER compatibility
```
sumitmsft: While set_attr is not intended to execute SQL directly, it’s important to ensure that no user input (either as an attribute or value) could ever be interpreted as an `SQL command` or inadvertently leak into any SQL execution context.

mssql_python/pybind/connection/connection.cpp
jahnvi480: I have removed the extra casting and just validating that t is not negative the way we have in pyodbc

mssql_python/connection.py
jahnvi480: Added all the checks on Cpp side

mssql_python/connection.py
jahnvi480: Added all the checks on Cpp side

mssql_python/__init__.py
```diff
@@ -52,6 +52,54 @@
...
+SQL_ATTR_ACCESS_MODE = ConstantsDDBC.SQL_ATTR_ACCESS_MODE.value
+SQL_ATTR_AUTOCOMMIT = ConstantsDDBC.SQL_ATTR_AUTOCOMMIT.value
+SQL_ATTR_CONNECTION_TIMEOUT = ConstantsDDBC.SQL_ATTR_CONNECTION_TIMEOUT.value
+SQL_ATTR_CURRENT_CATALOG = ConstantsDDBC.SQL_ATTR_CURRENT_CATALOG.value
+SQL_ATTR_LOGIN_TIMEOUT = ConstantsDDBC.SQL_ATTR_LOGIN_TIMEOUT.value
+SQL_ATTR_ODBC_CURSORS = ConstantsDDBC.SQL_ATTR_ODBC_CURSORS.value
+SQL_ATTR_PACKET_SIZE = ConstantsDDBC.SQL_ATTR_PACKET_SIZE.value
+SQL_ATTR_TXN_ISOLATION = ConstantsDDBC.SQL_ATTR_TXN_ISOLATION.value
+
+# The following attributes are [NO-OP] in this implementation (require Driver Manager):
```
gargsaumya: I'm not sure if we should explicitly mention that these attributes require the Driver Manager.
@sumitmsft what do you suggest here. For context- https://github.com/microsoft/mssql-python/pull/177#discussion_r2287614161

mssql_python/pybind/connection/connection.cpp
```diff
@@ -167,34 +169,291 @@ SqlHandlePtr Connection::allocStatementHandle() {
     return std::make_shared<SqlHandle>(static_cast<SQLSMALLINT>(SQL_HANDLE_STMT), stmt);
 }
 
+// Check if an attribute ID is in a list of sensitive attributes that need special handling
+bool Connection::isSensitiveAttribute(SQLINTEGER attribute) const {
```
gargsaumya: We are not really doing anything here.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -167,34 +169,291 @@ SqlHandlePtr Connection::allocStatementHandle() {
...
+    if (upperValue.find("--") != std::string::npos || 
+        (upperValue.find("/*") != std::string::npos && upperValue.find("*/") != std::string::npos)) {
+        return true;
+    }
+    
+    return false;
+}
+
+// Check if an attribute requires string sanitization
+bool Connection::requiresStringSanitization(SQLINTEGER attribute) const {
```
gargsaumya: Similarly here and at multiple other places. We are not doing anything here.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -167,34 +169,291 @@ SqlHandlePtr Connection::allocStatementHandle() {
...
+        // Add attributes that process binary data and might be sensitive
+        // Example: SQL_COPT_SS_ACCESS_TOKEN
+        1256 // SQL_COPT_SS_ACCESS_TOKEN
+    };
+    
+    return sanitizedAttrs.find(attribute) != sanitizedAttrs.end();
+}
+
+// Check binary data for suspicious patterns
+bool Connection::containsSuspiciousBinaryPatterns(const std::string& data) const {
```
gargsaumya: What do we really plan to achieve here? Let's discuss.

gargsaumya (changes requested):
Let's discuss the current implementation. Want to understand what we plan to do here.
Also, has this been thoroughly tested?

mssql_python/__init__.py
sumitmsft: @gargsaumya @jahnvi480 Is there any requirement of these attributes in current implementation? If yes, what's that? If not, then why are we even adding them? 
If we need them and they require DM then how are we using them? 
Ideally, we shouldn't use any attribute which requires DM

gargsaumya: +1
As I mentioned in my earlier comment, we should trim this list down to only the attributes that the ODBC driver supports directly.

For parity with pyodbc, we can also review the attributes that depend on the Driver Manager and see if we can implement equivalent behavior directly in ddbc_bindings.

sumitmsft: Using Copilot

Below is a categorized rundown of ODBC attributes, highlighting which ones are Driver Manager (DM)–only or DM‑managed, versus those typically implemented (or meaningfully implementable) inside an actual driver. Then I answer whether it makes sense to include (export) DM-only attributes in your code base when you do not ship or depend on an ODBC Driver Manager.

Clear Definitions
Driver Manager (DM)-only (or DM-managed) attribute: An attribute whose semantics are handled entirely (or primarily) by the ODBC Driver Manager layer. A stand‑alone driver (without a DM) cannot give it its intended behavior; at best it can silently accept or reject it.
Driver-level attribute: One that the underlying driver must enforce (transaction isolation, autocommit, current catalog, etc.).
Environment Attributes (Set via SQLSetEnvAttr) – DM Scope These are NOT connection attributes and are managed by the DM:
SQL_ATTR_ODBC_VERSION
SQL_ATTR_CONNECTION_POOLING
SQL_ATTR_CP_MATCH
SQL_ATTR_OUTPUT_NTS 

If you have no DM layer, exposing these makes little sense. They should not appear in a connection.set_attr API.
Connection Attributes Primarily / Historically DM-Managed The following are either: 
(a) Implemented by the DM for portability, or 
(b) Require DM infrastructure (tracing, async event dispatch, pooling), or 
(c) Are largely advisory to the DM rather than to the driver.

DM-centric connection attributes:

SQL_ATTR_TRACE
SQL_ATTR_TRACEFILE
SQL_ATTR_QUIET_MODE
SQL_ATTR_ODBC_CURSORS (DM coordinates cursor library emulation)
SQL_ATTR_ASYNC_DBC_FUNCTIONS_ENABLE
SQL_ATTR_ASYNC_DBC_EVENT
SQL_ATTR_ASYNC_DBC_PCALLBACK
SQL_ATTR_ASYNC_DBC_PCONTEXT
SQL_ATTR_DRIVER_AWARE_POOLING_SUPPORTED (read-only; DM queries capability)
SQL_ATTR_KEEPCONNECTION (used with pooling)
SQL_ATTR_DBC_INFO_TOKEN (internal use with DM-managed pooling / connection resurrection) Potentially DM-intermediated, depending on architecture:
SQL_ATTR_ENLIST_IN_DTC
SQL_ATTR_ENLIST_IN_XA
SQL_ATTR_RESET_CONNECTION (the DM may implement a pooled-reset pattern)
SQL_ATTR_CONNECTION_DEAD (diagnostic/read-only flag; DM may supply logic, driver reports)

Without a DM, most of these either:

Cannot function (async callback/event infra, pooling, tracing), or
Become meaningless (quiet mode, tracing), or
Risk misleading users.
Conclusion: Do not expose these as settable attributes. If you must recognize them, raising a clear NotSupportedError / ProgrammingError is preferable to silently accepting.

@jahnvi480 Let's not expose them at all. Not of much use.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -167,34 +169,291 @@ SqlHandlePtr Connection::allocStatementHandle() {
...
+    // Attribute-specific validation
+    switch (attribute) {
+        // Example validation for connection timeout
+        case SQL_ATTR_CONNECTION_TIMEOUT:
+            // Ensure reasonable timeout values
+            // Allow values from 0 to UINT_MAX (4294967295)
+            return value >= 0 && value <= 4294967295LL; // Use LL suffix for 64-bit literal
+            
+        // Example validation for query timeout
+        case SQL_ATTR_QUERY_TIMEOUT:
```
sumitmsft: The connection-level attribute validation code is currently referencing SQL_ATTR_QUERY_TIMEOUT, which is a statement attribute (set via SQLSetStmtAttr), not a connection attribute. 


- Let's remove SQL_ATTR_QUERY_TIMEOUT (and any other statement attributes) from connection attribute validation 
- Introduce a classification set (e.g. STATEMENT_ATTRS) and in connection.set_attr raise ProgrammingError (or NotSupportedError) with a clear message: "SQL_ATTR_QUERY_TIMEOUT is a statement attribute; set it on a statement handle (not supported yet)." 
- Add a unit test asserting that attempting connection.set_attr(SQL_ATTR_QUERY_TIMEOUT, ...) raises the expected exception. 
- D. Update documentation / docstring to list only the supported connection attributes.

mssql_python/__init__.py
jahnvi480: I have just commented the attributes using DM 
If you want, I'll remove it completely and in our Wiki, we can mention that these attributes are no supported

mssql_python/pybind/connection/connection.cpp
jahnvi480: Will remove it

mssql_python/pybind/connection/connection.cpp
jahnvi480: Will remove it

mssql_python/pybind/connection/connection.cpp
```diff
@@ -167,34 +169,291 @@ SqlHandlePtr Connection::allocStatementHandle() {
...
+            
+        default:
+            // For unknown attributes, just ensure it's not negative
+            // and within reasonable SQLULEN range for safe casting
+            return value >= 0 && value <= 4294967295LL;
+    }
+}
+
+// Check if a string value contains potential SQL injection patterns
+bool Connection::containsSQLInjectionPatterns(const std::string& value) const {
```
sumitmsft: We can remove this check for SQL Injection because SQL will never parse\compile any SQL command as part of Connection attribute.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -167,34 +169,291 @@ SqlHandlePtr Connection::allocStatementHandle() {
...
+    static const std::unordered_set<SQLINTEGER> sanitizedAttrs = {
+        // Add attributes that process string values and might be sensitive
+        // Example: SQL_ATTR_CURRENT_CATALOG
+    };
+    
+    return sanitizedAttrs.find(attribute) != sanitizedAttrs.end();
+}
+
+// Sanitize string values for attributes that need it
+std::string Connection::sanitizeStringValue(const std::string& value) const {
```
sumitmsft: I remember I asked you to sanitize the inputs, but ideally, we should NOT content‑sanitize connection attribute values (no keyword stripping, no regex “SQL injection” filtering etc). Let's only validate:

- The attribute ID is known, supported for a connection.
- The value length is within reasonable, documented bounds (to avoid excessive allocation).
- Integers fall within the permitted set/range for that attribute.
- Binary values meet size constraints

sumitmsft: Let's remove Upper Casing and Regex altogether.

sumitmsft (changes requested):

gargsaumya (approved):

mssql_python/connection.py
```diff
@@ -520,6 +521,63 @@ def rollback(self) -> None:
...
+                           SQL_ATTR_TXN_ISOLATION).
+            value: The value to set for the attribute. Can be an integer, string, 
+                   bytes, or bytearray depending on the attribute type.
+
+        Raises:
+            InterfaceError: If the connection is closed or attribute is invalid.
+            ProgrammingError: If the value type or range is invalid.
+
+        Example:
+            >>> conn.set_attr(SQL_ATTR_AUTOCOMMIT, SQL_AUTOCOMMIT_OFF)
```
sumitmsft: You added a Python method set_attr that lets users change low-level attributes, including SQL_ATTR_AUTOCOMMIT. However, when a caller does: conn.set_attr(SQL_ATTR_AUTOCOMMIT, SQL_AUTOCOMMIT_OFF) the underlying ODBC/driver state changes, but (based on current code) any Python-side attribute or cached state (e.g., conn.autocommit or an internal flag previously used by other methods) is not updated. 

As discussed let's skip this attribute for now.

sumitmsft (dismissed):

PR #178, opened 2025-08-13
FEAT: Adding fetchval() in cursor class

jahnvi480:

> AB#34921

### Summary   
This pull request introduces a new convenience method, `fetchval`, to the `cursor.py` module, designed to simplify retrieval of single values from SQL queries. Comprehensive tests have been added to ensure its correctness, robustness, and compatibility with various SQL scenarios and data types.

Addition of `fetchval` method:

* Added `fetchval` to the `Cursor` class in `mssql_python/cursor.py`, allowing users to easily fetch the first column of the first row from a query result. This is especially useful for aggregate queries or scalar selects and supports method chaining.

Testing and validation of `fetchval`:

* Extensive tests added in `tests/test_004_cursor.py` to validate `fetchval` for basic functionality, multiple data types, NULL values, empty results, multiple columns/rows, aggregate functions, method chaining, closed cursor behavior, rownumber tracking, error scenarios, and common performance patterns.

These changes make it easier and safer for users to retrieve single values from queries, improving developer experience and reliability.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -970,6 +970,46 @@ class Cursor:
 
         return True
 
+    def fetchval(self):
+        """
+        Fetch the first column of the first row if there are results.
+        
+        This is a convenience method for queries that return a single value,
+        such as SELECT COUNT(*) FROM table, SELECT MAX(id) FROM table, etc.
+        
+        Returns:
+            The value of the first column of the first row, or None if no rows
+            are available or the first column value is NULL.
+            
+        Raises:
+            Exception: If the cursor is closed.
+            
+        Example:
+            >>> count = cursor.execute('SELECT COUNT(*) FROM users').fetchval()
+            >>> max_id = cursor.execute('SELECT MAX(id) FROM products').fetchval()
+            >>> name = cursor.execute('SELECT name FROM users WHERE id = ?', user_id).fetchval()
+            
+        Note:
+            This is a convenience extension beyond the DB-API 2.0 specification.
+            After calling fetchval(), the cursor position advances by one row,
+            just like fetchone().
+        """
+        self._check_closed()  # Check if the cursor is closed
+        
+        # Fetch the first row
+        row = self.fetchone()
+        
+        # If no row is available, return None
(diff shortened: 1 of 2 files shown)


mssql_python/cursor.py
```diff
@@ -970,6 +970,46 @@ def nextset(self) -> Union[bool, None]:
...
+        # If no row is available, return None
+        if row is None:
+            return None
+            
+        # If the row has no columns, return None (shouldn't happen in normal cases)
+        if len(row) == 0:
+            return None
+            
+        # Return the first column value (could be None if the column value is NULL)
+        return row[0]
```
gargsaumya: We can skip these multiple lines and the redundant length check (with fetchone()) and keep it simple.

row = self.fetchone()
return None if row is None else row[0].

mssql_python/cursor.py
```diff
@@ -970,6 +970,46 @@ def nextset(self) -> Union[bool, None]:
...
+        
+        # Fetch the first row
+        row = self.fetchone()
+        
+        # If no row is available, return None
+        if row is None:
+            return None
+            
+        # If the row has no columns, return None (shouldn't happen in normal cases)
+        if len(row) == 0:
```
sumitmsft: Agree with Saumya's comment. Another reason is: 

In typical Python DB-API (PEP 249) implementations, when you fetch a row from a cursor (e.g., with fetchone() or fetchall()), each row is always a sequence (like a tuple or list) containing one element for each selected column. If your query is `SELECT col1 FROM table`, then each row will be something like ('value',)—that is, a one-element tuple. If there are no results, fetchone() returns None, not an empty tuple or list.

Therefore if you get a row, it should always have as many columns as your query specifies, and it should never be an empty tuple/list (zero columns). If there are no rows, you get None, not [].

If your code does if len(row) == 0: return None, this is likely unnecessary, because you should never get a row with zero columns under normal DB-API usage.
This check could be removed unless you have a reason (such as a custom or nonstandard driver) where a zero-column row is possible.

sumitmsft (approved):

PR #180, opened 2025-08-14
FEAT: Adding cursor.scoll function

jahnvi480:

> [AB#34892](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34892)

### Summary   
This pull request introduces support for emulated scrolling (relative and absolute) in the `Cursor` class for forward-only result sets, including a new `scroll()` method and comprehensive tests for scroll behavior. It also refines row position tracking, improves error handling, and updates some test assertions for clarity. The most important changes are grouped below.

### Cursor Scrolling Feature

* Added a new `scroll()` method to the `Cursor` class, allowing movement to a specific row position in the result set using relative or absolute modes. This is implemented by consuming rows for forward-only cursors, with robust error handling for unsupported operations and invalid parameters. 
* Implemented helper methods `_scroll_relative`, `_scroll_absolute`, and `_consume_rows_for_scroll` to handle the logic for moving the cursor position and updating internal counters.

### Row Position Tracking Improvements

* Refined row position tracking by introducing `_next_row_index` and updating how `rownumber` is managed after fetch and scroll operations, ensuring consistency with DB-API semantics. 
### Error Handling and Validation

* Improved error handling in cursor methods by providing more descriptive error messages and raising appropriate exceptions for invalid scroll operations or parameters. 
* Updated import statements to include new exception types used for scroll validation.

### Test Coverage for Scrolling

* Added a suite of new tests to verify scroll behavior, including relative and absolute scrolling, backward scroll rejection, empty result set handling, mixed fetch/scroll operations, and edge case validation. 

### Minor Test Assertion Updates

* Updated test assertions and comments in transaction-related tests for clarity and to better reflect the intended behavior.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -81,6 +81,8 @@ class Cursor:
         self._has_result_set = False  # Track if we have an active result set
         self._skip_increment_for_next_fetch = False  # Track if we need to skip incrementing the row index
 
+        self.messages = []  # Store diagnostic messages
+
     def _is_unicode_string(self, param):
         """
         Check if a string contains non-ASCII characters.
@@ -453,6 +455,9 @@ class Cursor:
         if self.closed:
             raise Exception("Cursor is already closed.")
 
+        # Clear messages per DBAPI
+        self.messages = []
+        
         if self.hstmt:
             self.hstmt.free()
             self.hstmt = None
@@ -698,6 +703,9 @@ class Cursor:
         if reset_cursor:
             self._reset_cursor()
 
+        # Clear any previous messages
+        self.messages = []
+
         param_info = ddbc_bindings.ParamInfo
         parameters_type = []
 
@@ -745,7 +753,14 @@ class Cursor:
             self.is_stmt_prepared,
             use_prepare,
         )
+        
+        # Check for errors but don't raise exceptions for info/warning messages
         check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
+        
+        # Capture any diagnostic messages (SQL_SUCCESS_WITH_INFO, etc.)
+        if self.hstmt:
+            self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
+    
         self.last_executed_stmt = operation
 
         # Update rowcount after execution
@@ -827,7 +842,10 @@ class Cursor:
         """
         self._check_closed()
         self._reset_cursor()
-
+        
+        # Clear any previous messages
+        self.messages = []
+        
         if not seq_of_parameters:
             self.rowcount = 0
             return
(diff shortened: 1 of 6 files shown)


mssql_python/cursor.py
```diff
@@ -1073,4 +1081,114 @@ def __del__(self):
...
+        Raises IndexError if end-of-result reached before consuming requested number.
+        """
+        if rows_to_consume <= 0:
+            return
+
+        rows_consumed = 0
+        try:
+            for _ in range(rows_to_consume):
+                row_data = []
+                ret = ddbc_bindings.DDBCSQLFetchOne(self.hstmt, row_data)
```
gargsaumya: The current implementation of scroll() is correct for forward-only cursors, but we have a limitation:

Right now, _consume_rows_for_scroll() repeatedly fetches rows one-by-one until the target position. That works, but this will become inefficient for large offsets (e.g., skipping thousands of rows).
In ODBC there’s a distinction between forward-only cursors (default, what we have now) and scrollable cursors (enabled by setting SQL_ATTR_CURSOR_TYPE and using SQLFetchScroll). With scrollable cursors, the driver itself can jump directly to an absolute or relative row without consuming all intermediate rows.

So this might be okay for now, but we should consider adding proper scrollable cursor support in future, which would make scroll() much more efficient. 
(FYI, pyodbc’s scroll() uses ODBC’s native support for scrollable cursors (SQLFetchScroll) rather than emulating it row-by-row.)

mssql_python/cursor.py
```diff
@@ -1073,4 +1081,114 @@ def __del__(self):
...
+                if ret == ddbc_sql_const.SQL_NO_DATA.value:
```
gargsaumya: So, for a user call like:
cursor.scroll(1_000_000, mode="relative")

the current implementation will fetch one million rows one-by-one from the driver and discard them.
Concern:

1. This is very inefficient for large skips.
2. If rows contain LOBs, we may end up forcing unnecessary allocation/transfer of large objects, even though the data is never used.

We might want to consider leveraging SQLFetchScroll() with scrollable cursors in the future to avoid this overhead. Open for discussion.

mssql_python/cursor.py
sumitmsft: If this is the case, then we must fix it and make it efficient rather than doing row by row operation. @jahnvi480 let's see how we can leverage ODBC's SQLFetchScroll and move away from the row ops.

mssql_python/cursor.py
sumitmsft: Agree! Let's make the change in logic

sumitmsft (changes requested):
I just have the apprehension around scrolling many rows forward may result in high memory usage due to row by row operation. Let's make the correction and do the above mentioned fix.

mssql_python/cursor.py
jahnvi480: I am now using SQLFetchScroll for this implementation. 
Please have a look at it and let me know if there are any changes required

Issue #182, opened 2025-08-15
Make cursor finalizer shutdown cleanup idempotent and lower log level when cursor is already closed


hiroyukinakatsuka:
### Is your feature request related to a problem? Please describe.
At interpreter shutdown, an **ERROR** log is emitted from the cursor finalizer even when the cursor and connection were explicitly closed during normal execution. This is noisy and can be misleading in production diagnostics.

**Observed log (at process exit):**

```
mssql\_python::log\@142: Error during cursor cleanup in **del**: Cursor is already closed.
````

**Environment / Context**
- 
- OS: Windows 10
- Python: 3.12.6
- mssql-python: 0.9.0
- Pooling: disabled
- Logging: `setup_logging("stdout", logging.DEBUG)`

Minimal context (simplified code):

```python
import logging, sys
import mssql_python as DBDRV

DBDRV.setup_logging("stdout", logging.DEBUG)
DBDRV.pooling(enabled=False)

conn = DBDRV.connect("<DSN>")
cur = conn.cursor()
cur.execute("SELECT 1;")
cur.fetchall()

conn.close()
sys.exit(0)  # At shutdown, ERROR log above is printed
````

### Describe the solution you'd like

* Make cursor cleanup **idempotent** in the finalizer path: if the cursor is already closed, treat it as a no-op.
* Downgrade this condition from **ERROR** to **INFO** or **DEBUG**, or suppress it entirely when it occurs during interpreter shutdown.

### Describe alternatives you've considered

* Explicitly closing all cursors/connections, clearing references, forcing `gc.collect()` → functional but still prints the ERROR at shutdown.
* Lowering global log level to suppress the message → also hides genuinely actionable errors.
* Custom `atexit` cleanup → reduces likelihood but does not address the finalizer’s behavior/severity.

### Additional context

This request aims to reduce shutdown-time log noise without changing runtime semantics. It should be backward compatible: double-closing a closed cursor becomes a no-op, and logging severity better reflects the benign nature of the condition.

sumitmsft:
Hi @hiroyukinakatsuka - Thank you for your feedback. Appreciate your time to use this driver and sharing your concern with us. I agree you have a good point and we should be able to take care of this issue. Please allow us sometime before this will be fixed in subsequent releases. 

We are actively listening, so keep sharing your feedback!!

PR #183, opened 2025-08-18
FIX: Closed Attribute in cursor `__del__`

bewithgaurav:

> AB#<WORK_ITEM_ID>

> GitHub Issue: #182 

### Summary   
  
This pull request makes improvements to the cursor resource management and error handling in the `mssql_python/cursor.py` module. The main changes include making the cursor close operation idempotent, refining exception types, and improving logging during cleanup. These updates help ensure more robust and predictable behavior when working with cursors.

Error handling improvements:

* Changed the exception raised when attempting to close an already closed cursor from a generic `Exception` to a more specific `ProgrammingError`, providing clearer feedback to users.
* Updated the import statement to use `ProgrammingError` instead of `InterfaceError`, aligning with the new error handling approach.

Resource cleanup and logging:

* Improved the destructor (`__del__`) to check for the correct `closed` attribute and made sure no exception is raised if the cursor is already closed, enhancing reliability during garbage collection.
* Changed the logging level in the destructor from 'error' to 'debug' for exceptions during cleanup, reducing unnecessary noise in error logs.


mssql_python-0.8.1-cp312-cp312-musllinux_1_2_aarch64.whl
bewithgaurav, line 1: Deleted unwanted whl file which was committed earlier

mssql_python/cursor.py
```diff
@@ -435,13 +435,14 @@ def _reset_cursor(self) -> None:
...
-        Close the cursor now (rather than whenever __del__ is called).
+        Close the cursor, if it is open.
+        Idempotent: subsequent calls have no effect.
 
         Raises:
             Error: If any operation is attempted with the cursor after it is closed.
         """
         if self.closed:
-            raise Exception("Cursor is already closed.")
+            raise ProgrammingError("Cursor is already closed.")
```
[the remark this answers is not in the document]
bewithgaurav: The "idempotency" in this context means that calling `close()` multiple times won't free resources twice or cause corruption - the first call frees resources and subsequent calls raise a predictable exception. This is actually the standard behavior in pyodbc as well. Added a comment here specifying the same.

mssql_python/msvcp140.dll
bewithgaurav, line 1: cleaned this up as well, this is already present inside vcredist

mssql_python/cursor.py
```diff
@@ -794,10 +798,16 @@ def __del__(self):
         Destructor to ensure the cursor is closed when it is no longer needed.
         This is a safety net to ensure resources are cleaned up
         even if close() was not called explicitly.
+        If the cursor is already closed, it will not raise an exception during cleanup.
         """
-        if "_closed" not in self.__dict__ or not self._closed:
+        if "closed" not in self.__dict__ or not self.closed:
```
sumitmsft, line 1189: Can't we define `self.closed` in the constructor, so you can simply check `if not `self.closed:`? 
This may avoid reliance on __dict__, and guards against subtle bugs if the attribute is missing. 

Let me know if I am missing anything wrt to dict if its used anywhere else.

mssql_python/cursor.py
```diff
@@ -457,7 +461,7 @@ def _check_closed(self):
             Error: If the cursor is closed.
         """
         if self.closed:
-            raise Exception("Operation cannot be performed: the cursor is closed.")
+            raise ProgrammingError("Cursor is already closed.", "")
```
sumitmsft: These changes still keeps cursor.close() non‑idempotent while the comments claims it is idempotent. 

close() now calls self._check_closed() at the start.
_check_closed() raises ProgrammingError when self.closed is True.
Therefore a second close() raises instead of being a no‑op. That is NOT idempotent.

First, the issue raised in 183 requires no-op\no error during second close of the cursor. Secondly, DB-API 2.0 convention defines calling .close() multiple times must be safe (no exception). 

We can keep _check_closed() raising ProgrammingError only for operational methods (execute/fetch) after closure - NOT for close() itself.

Fix the tests to make sure double closing of cursor is silent.

sumitmsft (changes requested):
Left a couple of comments.

mssql_python/cursor.py
bewithgaurav: The expected behaviour (also consistent with pyodbc - verified) is raising a ProgrammingError even when cursor.close() called twice, no silent closing. The comments above clarify that idempotency here means that resources are freed only once - subsequent calls won't double-free resources or cause corruption, but will raise a predictable exception instead. The issue currently raised is that it tries to double free memory, which is now fixed.

mssql_python/cursor.py
bewithgaurav: this is already present in line 68, resolving this.

mssql_python/cursor.py
bewithgaurav: Have fixed and cleaned up the flow a bit more, but the logic remains consistent as described above and in comments.

sumitmsft: I’d like to clarify the idempotency point.

Per https://peps.python.org/pep-0249/, close() is explicitly defined as idempotent, which means calling it multiple times should not raise an exception. In other words, once resources are freed on the first call, subsequent calls should be safe no-ops.

The current implementation raises ProgrammingError on a second close(), which makes it non‑idempotent from an API user’s perspective. While I agree that resource cleanup should happen only once (avoiding double free), DB-API requires the public method contract to behave silently beyond the first call.

Regarding pyodbc: if pyodbc raises here, that appears to be its own deviation from DB-API. My argument is that we should follow the DB-API spec first and foremost, so user code that expects standard-compliant behavior works consistently across drivers. And let's document that in our wiki as well about this behavior. 

One thing I agree on is that:
- We can keep _check_closed() raising ProgrammingError for operational methods (execute, fetch, etc.) on closed cursors.
- But update close() to be a safe no-op when self.closed is already True.

Let's talk if this warrant more discussions :)

gargsaumya (approved):

PR #187, opened 2025-08-20
FEAT: Adding lowercase for global variable

jahnvi480:

> [AB#34905](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34905)

### Summary   
This pull request introduces a new global `lowercase` flag to the `mssql_python` package, allowing users to control whether column names in query results are automatically converted to lowercase. The changes include updates to the global settings infrastructure, the `Cursor` and `Row` classes to support case-insensitive access, and comprehensive tests to ensure correct behavior.

**Global lowercase flag and settings:**

* Added a global `lowercase` flag and supporting infrastructure to `mssql_python/__init__.py`, including a `Settings` class and a `get_settings()` accessor. The default value is `False`.
* Exported the `lowercase` flag for external use and updated tests to check its default value. 

**Cursor and Row class enhancements:**

* Modified the `Cursor` class to read the global `lowercase` setting and pass it to `Row` objects. The `_initialize_description` method now lowercases column names in the description if the flag is set.
* Updated the `Row` class to support attribute access by column name, respecting the `lowercase` flag for case-insensitive lookup.
* Adjusted cursor fetch methods to pass the `Cursor` instance to `Row` objects for correct lowercase handling. 

**Testing improvements:**

* Added a comprehensive test for the `lowercase` flag in `tests/test_004_cursor.py`, verifying that column names in the cursor description and row attribute access are correctly lowercased when the flag is enabled.
* Ensured proper cleanup and restoration of global state in the new test.


tests/test_004_cursor.py
```diff
@@ -1313,6 +1314,76 @@ def test_row_column_mapping(cursor, db_connection):
...
+        drop_cursor = db_connection.cursor()
+        
+    finally:
+        # Restore original value
+        mssql_python.lowercase = original_lowercase
+        
+        try:
+            # Use a separate cursor for cleanup
+            if drop_cursor:
+                drop_cursor.execute("DROP TABLE IF EXISTS #pytest_lowercase_test")
```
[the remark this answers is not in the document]
sumitmsft: DROP TABLE IF EXISTS syntax IS supported for temporary tables in SQL Server

mssql_python/cursor.py
```diff
@@ -480,26 +483,32 @@ def _create_parameter_types_list(self, parameter, param_info, parameters_list, i
...
-            )
-            for col in col_metadata
-        ]
+    def _initialize_description(self, column_metadata=None):
+        """Initialize the description attribute from column metadata."""
+        if not column_metadata:
+            self.description = None
+            return
+        import mssql_python
+
```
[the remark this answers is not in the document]
sumitmsft: Agree with this.

mssql_python/cursor.py
```diff
@@ -480,26 +483,32 @@ def _create_parameter_types_list(self, parameter, param_info, parameters_list, i
...
+        description = []
+        for i, col in enumerate(column_metadata):
+            # Get column name - lowercase it if the lowercase flag is set
+            column_name = col["ColumnName"]
+            
+            if mssql_python.lowercase:
```
sumitmsft: There are direct accesses to the global `lowercase` variable in `mssql_python/cursor.py`. For better maintainability and to avoid issues with mutable global state, please ensure all access to this variable goes through the `get_settings()` accessor. This will centralize state management and help prevent subtle bugs or inconsistencies.

sumitmsft (changes requested):
Added a comment.. Please resolve. Rest all looks good

mssql_python/row.py
```diff
@@ -26,20 +29,34 @@ def __init__(self, values, cursor_description):
...
+                col_name = col_desc[0]  # Name is first item in description tuple
+                column_map[col_name] = i
+                
+        self._column_map = column_map
     
     def __getitem__(self, index):
         """Allow accessing by numeric index: row[0]"""
         return self._values[index]
     
     def __getattr__(self, name):
```
sumitmsft, line 49: The current implementation in row.py does indeed have conditional case-insensitive behavior that should be documented for clarity.

The case-insensitive behavior depends on how _column_map is populated, which happens during cursor initialization when the global lowercase flag is enabled. This creates an important behavioral difference that users should be aware of.

Update the docstring to clearly document this behavior:

`    Note: Case sensitivity depends on the global 'lowercase' setting:
    - When lowercase=True: Column names are stored in lowercase, enabling 
      case-insensitive attribute access (e.g., row.NAME, row.name, row.Name all work)
    - When lowercase=False (default): Column names preserve original casing,
      requiring exact case matching for attribute access
`

Additionally, consider updating the class-level docstring to mention this behavior:

`
    A row of data from a cursor fetch operation. Provides both tuple-like indexing
    and attribute access to column values.
    
    Column attribute access behavior depends on the global 'lowercase' setting:
    - When enabled: Case-insensitive attribute access
    - When disabled (default): Case-sensitive attribute access matching original column names
    
    Example:
        row = cursor.fetchone()
        print(row[0])           # Access by index
        print(row.column_name)  # Access by column name (case sensitivity varies)
`

This documentation will help users understand the conditional behavior and avoid confusion when migrating between different lowercase settings.

tests/test_004_cursor.py
```diff
@@ -1313,6 +1314,76 @@ def test_row_column_mapping(cursor, db_connection):
         cursor.execute("DROP TABLE #pytest_row_test")
         db_connection.commit()
 
+def test_lowercase_attribute(cursor, db_connection):
```
sumitmsft, line 1822: I want you to think of following test cases as well... May be for futures:

**Test case for concurrent cursors with different lowercase values**
This would be valuable future work if per-cursor settings are supported. Currently, the implementation uses a global lowercase setting that affects all cursors created after the setting is changed. A test for this scenario could look like:
`
def test_concurrent_cursors_different_lowercase_settings():
    """Test behavior when multiple cursors exist with different lowercase settings"""
    # This test would be relevant if per-cursor lowercase settings are implemented
    # Currently, the global setting affects all new cursors uniformly
    pass  # Future implementation
`

**Test for lowercase setting changes after cursor creation**
The lowercase setting is read at cursor creation time during _initialize_description()
Changing mssql_python.lowercase after a cursor is created but before fetch should not affect that cursor's behavior
The cursor's column name casing is determined when the description is initialized
Here's a suggested test case:

`
def test_lowercase_setting_after_cursor_creation(cursor, db_connection):
    """Test that changing lowercase setting after cursor creation doesn't affect existing cursor"""
    original_lowercase = mssql_python.lowercase
    
    try:
        # Create table and cursor with lowercase=False
        mssql_python.lowercase = False
        cursor.execute("CREATE TABLE #test (UserName VARCHAR(50))")
        cursor.execute("SELECT * FROM #test")
        
        # Change setting after cursor is created but before fetch
        mssql_python.lowercase = True
        
        # The cursor should still use the original casing
        column_names = [desc[0] for desc in cursor.description]
        assert "UserName" in column_names, "Column casing should not change after cursor creation"
        assert "username" not in column_names, "Lowercase should not apply to existing cursor"
        
    finally:
        mssql_python.lowercase = original_lowercase
`

The lowercase setting affects cursors at creation time, not at fetch time. Once a cursor's description is initialized, changing the global setting will not affect that cursor's column name casing.

This behavior should be clearly documented in the docstrings and potentially in user-facing documentation to avoid confusion.

mssql_python/__init__.py
```diff
@@ -6,6 +6,26 @@
...
+    def __init__(self):
+        self.lowercase = False
+
+# Create a global instance
+_settings = Settings()
+
+def get_settings():
+    return _settings
+
+lowercase = _settings.lowercase  # Default is False
```
sumitmsft: The current implementation of the global lowercase setting introduces potential race conditions in multi-threaded environments. Here are the key concerns:

Current Issue
The global lowercase variable is mutable state that can be modified at runtime:

`
This creates a race condition in multi-threaded code
mssql_python.lowercase = True  # Thread A
mssql_python.lowercase = False # Thread B (concurrent modification)
`
Potential Problems
Race conditions: Multiple threads modifying lowercase simultaneously
Inconsistent behavior: Cursors created in different threads may get different column casing unexpectedly
Thread interference: One thread's setting change affects all other threads globally


This I have mentioned in my other comments as well.. But I wanted to enforce this here as well. 

If the current global approach is maintained, the documentation should clearly state:

- Not thread-safe: Concurrent modifications to mssql_python.lowercase can cause undefined behavior

- Recommended usage: Set once at application startup before creating any connections/cursors

- Multi-threading: Use connection-level or cursor-level overrides for thread-safe operation

This is particularly important given that threadsafety = 1 indicates "Threads may share the module, but not connections" in the DB-API specification, yet the global setting affects all connections across all threads.

bewithgaurav (approved):

PR #190, opened 2025-08-21
FEAT: Adding Ouput Converter APIs

jahnvi480:

> [AB#34913](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34913)
> [AB#34914](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34914)
> [AB#34915](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34915)
> [AB#34916](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34916)

### Summary   
This pull request adds a flexible output converter system to the database connection, allowing custom Python functions to be registered for converting SQL data types when fetching results. It introduces new methods for managing these converters and updates the row handling logic to apply them automatically. Comprehensive tests are added to verify correct registration, retrieval, removal, and integration of output converters, including edge cases and chaining behavior.

**Core output converter system:**

* Added methods to `Connection` (`add_output_converter`, `get_output_converter`, `remove_output_converter`, `clear_output_converters`) for registering and managing output converter functions for specific SQL types. These converters are called when values of the registered SQL type are read from the database. (`mssql_python/connection.py`)
* Updated the `Row` class to apply registered output converters automatically to values fetched from the database, with fallback logic for string types and robust error handling. (`mssql_python/row.py`) 

**Testing and validation:**

* Added extensive tests for output converter management, including adding, retrieving, removing, clearing, chaining, temporary replacement, and integration during data fetching. Tests also cover edge cases such as handling `NULL` values and using multiple converters at once. (`tests/test_003_connection.py`)
* Introduced helper converter functions for specific SQL types (e.g., `DATETIMEOFFSET` and custom string handling) to support and validate the new converter system in tests. (`tests/test_003_connection.py`)

**Test and utility enhancements:**

* Imported additional modules (`struct`, `datetime`, `timezone`, and constants) to support new test cases and converter logic. (`tests/test_003_connection.py`)


mssql_python/row.py
```diff
@@ -38,6 +44,55 @@ def __init__(self, cursor, description, values, column_map=None):
...
+                try:
+                    # If value is already a Python type (str, int, etc.), 
+                    # we need to convert it to bytes for our converters
+                    if isinstance(value, str):
+                        # Encode as UTF-16LE for string values (SQL_WVARCHAR format)
+                        value_bytes = value.encode('utf-16-le')
+                        converted_values[i] = converter(value_bytes)
+                    else:
+                        converted_values[i] = converter(value)
+                except Exception as e:
```
sumitmsft: All exceptions are currently suppressed silently. While it's important not to leak sensitive data, it would be beneficial to at least log the exception at a debug or warning level for troubleshooting purposes. This way, issues with converter functions can be diagnosed without exposing sensitive information.

`
except Exception:
    log('debug', 'Exception occurred in output converter', exc_info=True)
`
This will help with debugging, especially if custom converter functions are used, without leaking sensitive data.

mssql_python/connection.py
```diff
@@ -208,6 +208,76 @@ def execute(self, sql, *args):
         cursor = self.cursor()
         cursor.execute(sql, *args)
         return cursor
+    
+    def add_output_converter(self, sqltype, func) -> None:
```
sumitmsft, line 540: The use of `self._output_converters` for storing and managing output converters is clear and straightforward for single-threaded scenarios. However, if the Connection object is accessed or modified from multiple threads, operations like adding, removing, or clearing converters could result in race conditions or inconsistent state.

If multi-threaded access to the Connection object is expected or possible, please consider protecting accesses to `self._output_converters` with a threading lock (e.g., threading.Lock). This will ensure thread safety during concurrent modifications.

tests/test_003_connection.py
```diff
@@ -31,6 +34,26 @@ def drop_table_if_exists(cursor, table_name):
...
+        return None
+        
+    # The format depends on the ODBC driver and how it returns binary data
+    # This matches SQL Server's format for DATETIMEOFFSET
+    tup = struct.unpack("<6hI2h", dto_value)  # e.g., (2017, 3, 16, 10, 35, 18, 500000000, -6, 0)
+    return datetime(
+        tup[0], tup[1], tup[2], tup[3], tup[4], tup[5], tup[6] // 1000,
+        timezone(timedelta(hours=tup[7], minutes=tup[8]))
+    )
+
```
sumitmsft, line 92: As an additional suggestion, please consider adding a test case where a converter function intentionally raises an exception. The test should verify that:

The value for that column falls back to the original (unconverted) value,
No sensitive information is leaked.

This will help ensure the robustness of error handling in the converter logic.

mssql_python/connection.py
```diff
@@ -208,6 +208,76 @@ def execute(self, sql, *args):
...
+        """
+        Register an output converter function that will be called whenever a value
```
sumitmsft, line 542: The output converter registration API allows arbitrary Python functions to be registered and executed on data retrieved from the database. If this API is public or can be accessed by untrusted users, there is a risk of code injection and arbitrary code execution. This risk is mitigated if only trusted application developers register converters. It is important to document this behavior clearly in the API documentation and warn users not to register converters from untrusted sources.

For the safety of all users, please consider updating the documentation (and relevant docstrings) with a clear "WARNING" that registering output converters is equivalent to executing arbitrary Python code. This should include explicit guidance that only trusted code should be registered as converters, and that converter registration should never be exposed to untrusted or external input.

Example warning for documentation:
`
⚠️ Warning: Registering an output converter will cause the supplied Python function to be executed on every matching database value. Do not register converters from untrusted sources, as this can result in arbitrary code execution and security vulnerabilities.
`
Making this risk explicit will help prevent misuse and keep downstream users secure.

sumitmsft (changes requested):
Left a few comments.

mssql_python/connection.py
jahnvi480: Made a note of this, will make sure to add warnings while writing documentation.

tests/test_003_connection.py
jahnvi480: Added a new testcase for this.

mssql_python/connection.py
jahnvi480: Added locking mechanism for this using threading module.

sumitmsft (approved):

sumitmsft (dismissed):

requirements.txt
```diff
@@ -4,3 +4,4 @@ pybind11
 coverage
 unittest-xml-reporting
 setuptools
+psutil
```
bewithgaurav, line 7: @jahnvi480 - is this a needed dependency? not able to see any usage in the code/tests etc.

jahnvi480: We have test which uses psutil

PR #191, opened 2025-08-21
FEAT: Adding timeout attribute

jahnvi480:

> [AB#34910](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34910)

### Summary   
This pull request adds support for per-connection query timeouts to the MSSQL Python driver. Now, you can set a timeout value on a connection, either at creation or later, and all cursors created from that connection will enforce this timeout for query execution. The changes include updates to the connection and cursor classes, integration with the underlying driver, and comprehensive tests for the new functionality.

**Query Timeout Support**

* Added a `timeout` parameter to the `Connection` class and the `connect` function, allowing users to specify a query timeout (in seconds) when establishing a database connection. The timeout can also be set or updated via a property on the `Connection` object. 
* Implemented getter and setter for the `timeout` property in the `Connection` class, including input validation and documentation. Setting the timeout updates all subsequently created cursors.
* Modified the `cursor` method in `Connection` to pass the current timeout value to each new `Cursor` instance.
* Updated the `Cursor` class to accept a timeout parameter and, if set, apply it to each query execution using the underlying driver’s statement attribute API. 
* Exposed the `SQL_ATTR_QUERY_TIMEOUT` constant and the `DDBCSQLSetStmtAttr` function in the C++ driver bindings to support setting the timeout at the driver level.

**Testing and Validation**

* Added comprehensive tests to verify default timeout behavior, setting and getting the timeout property, passing timeout via the constructor, enforcing timeout on long-running queries, and ensuring that updating the connection timeout affects all new cursors.


mssql_python/cursor.py
```diff
@@ -46,14 +46,15 @@ class Cursor:
         setoutputsize(size, column=None) -> None.
     """
 
-    def __init__(self, connection) -> None:
+    def __init__(self, connection, timeout: int = 0) -> None:
```
sumitmsft, line 57: I noticed that if a cursor is created before the timeout is set on the connection, it does not inherit the new timeout value. I suggest making this behavior more explicit in the user-facing documentation as well (e.g., in the README \WIKIS). This will help users avoid confusion and ensure they understand when the timeout value is applied to cursors.

Suggestion:
Please add a note to the relevant user documentation explaining that the timeout setting only affects cursors created after the value is set on the connection. Existing cursors will retain the timeout value that was current at their creation time.

sumitmsft (changes requested):
Left a comment. Rest all looks good

mssql_python/cursor.py
jahnvi480: Will keep a note of it and will update the Wikis accordingly.

sumitmsft (approved):

sumitmsft (dismissed):

sumitmsft (approved):

PR #197, opened 2025-08-22
FEAT: Adding rowID and rowvVer functions

jahnvi480:

> [AB#34929](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34929)
> [AB#34930](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34930)

### Summary   
This pull request adds support for retrieving special columns from a database table using the ODBC `SQLSpecialColumns` function. The main changes include new methods in the Python cursor for accessing row identifier and version columns, as well as the necessary C++ and Python bindings to expose this ODBC functionality. Additionally, relevant constants have been added to support these features.

**New Python cursor methods:**

* Added `rowIdColumns` and `rowVerColumns` methods to the `Cursor` class in `mssql_python/cursor.py`, allowing users to retrieve columns that uniquely identify a row (`SQL_BEST_ROWID`) and columns that are automatically updated when any value in the row is updated (`SQL_ROWVER`). These methods handle ODBC parameter preparation, call the new binding, and return results as row objects.

**ODBC and pybind11 binding support:**

* Defined the `SQLSpecialColumnsFunc` function pointer type and added the corresponding global pointer in `mssql_python/pybind/ddbc_bindings.h` and `mssql_python/pybind/ddbc_bindings.cpp`. 
* Loaded the `SQLSpecialColumnsW` symbol from the ODBC driver in the driver loader, and included it in the function pointer validation check.
* Implemented the `SQLSpecialColumns_wrap` function to call the ODBC API, handling both Unix and Windows string conversions, and exposed it to Python as `DDBCSQLSpecialColumns` in the module definition. 

**Constants:**

* Added new constants for `SQL_SCOPE_CURROW`, `SQL_BEST_ROWID`, `SQL_ROWVER`, `SQL_NO_NULLS`, and `SQL_NULLABLE_UNKNOWN` to the `ConstantsDDBC` enum in `mssql_python/constants.py` to support the new functionality.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -1326,6 +1326,284 @@ class Cursor:
         # Return the cursor itself
         return self
 
+    def statistics(self, table: str, catalog: str = None, schema: str = None, unique: bool = False, quick: bool = True) -> 'Cursor':
+        """
+        Creates a result set of statistics about a single table and the indexes associated 
+        with the table by executing SQLStatistics.
+        
+        Args:
+            table (str): The name of the table.
+            catalog (str, optional): The catalog name. Defaults to None.
+            schema (str, optional): The schema name. Defaults to None.
+            unique (bool, optional): If True, only unique indexes are returned. 
+                                    If False, all indexes are returned. Defaults to False.
+            quick (bool, optional): If True, CARDINALITY and PAGES are returned only 
+                                    if readily available. Defaults to True.
+        
+        Returns:
+            cursor: The cursor itself, containing the result set. Use fetchone(), fetchmany(),
+                   or fetchall() to retrieve the results.
+            
+        Example:
+            # Get statistics for the 'Customers' table
+            stats_cursor = cursor.statistics(table='Customers')
+            
+            # Fetch rows as needed
+            first_stat = stats_cursor.fetchone()
+            next_10_stats = stats_cursor.fetchmany(10)
+            all_remaining = stats_cursor.fetchall()
+        """
+        self._check_closed()
+        
+        # Always reset the cursor first to ensure clean state
+        self._reset_cursor()
+
+        # Table name is required
+        if not table:
+            raise ProgrammingError("Table name is required", "HY000")
+        
+        # Set unique flag (SQL_INDEX_UNIQUE = 0, SQL_INDEX_ALL = 1)
+        unique_option = ddbc_sql_const.SQL_INDEX_UNIQUE.value if unique else ddbc_sql_const.SQL_INDEX_ALL.value
+        
+        # Set quick flag (SQL_QUICK = 0, SQL_ENSURE = 1)
+        reserved_option = ddbc_sql_const.SQL_QUICK.value if quick else ddbc_sql_const.SQL_ENSURE.value
+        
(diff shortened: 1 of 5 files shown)


mssql_python/cursor.py
```diff
@@ -1066,6 +1066,194 @@ def foreignKeys(self, table=None, catalog=None, schema=None, foreignTable=None,
...
+            nullable_flag
+        )
+        check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, retcode)
+        
+        # Initialize description from column metadata
+        column_metadata = []
+        try:
+            ddbc_bindings.DDBCSQLDescribeCol(self.hstmt, column_metadata)
+            self._initialize_description(column_metadata)
+        except Exception:
```
sumitmsft: When ODBC calls (e.g., `DDBCSQLDescribeCol`, `DDBCSQLSpecialColumns`) fail, exceptions are caught but not logged or re-raised with context. 

- Log all exceptions with relevant context using a logging framework (e.g., Python `logging`).
- When raising or reporting exceptions, wrap them in a custom error, or include helpful context.

mssql_python/cursor.py
```diff
@@ -1066,6 +1066,194 @@ def foreignKeys(self, table=None, catalog=None, schema=None, foreignTable=None,
...
```
sumitmsft: When ODBC calls (e.g., DDBCSQLDescribeCol, DDBCSQLSpecialColumns) fail, exceptions are caught but not logged or re-raised with context.

Log all exceptions with relevant context using a logging framework (e.g., Python logging).
When raising or reporting exceptions, wrap them in a custom error, or include helpful context.

mssql_python/cursor.py
```diff
@@ -1066,6 +1066,194 @@ def foreignKeys(self, table=None, catalog=None, schema=None, foreignTable=None,
...
+                - type_name: Type name
+                - column_size: Column size
+                - buffer_length: Buffer length
+                - decimal_digits: Decimal digits
+                - pseudo_column: One of SQL_PC_UNKNOWN, SQL_PC_NOT_PSEUDO, SQL_PC_PSEUDO
+        """
+        self._check_closed()
+        
+        # Always reset the cursor first to ensure clean state
+        self._reset_cursor()
```
sumitmsft, line 1115: Any method that 
- mutates the cursor’s internal state (_reset_cursor, self.description, etc.)
- Calls ODBC functions with shared handles (self.hstmt)

…is a risk for thread-safety when the cursor object is shared across threads. 

Let's add a todo to take care of because when we implement async this will be necessary.

jahnvi480: Added TODO before the init function of cursor under my name.

mssql_python/cursor.py
jahnvi480: Added proper logging

mssql_python/cursor.py
jahnvi480: Added proper logging

PR #198, opened 2025-08-22
FEAT: Adding statistics API in cursor class

jahnvi480:

> AB#34928

### Summary   
This pull request adds support for retrieving table and index statistics via the ODBC `SQLStatistics` API in the `mssql_python` package. The main changes include implementing the `statistics` method on the Python cursor, exposing the required C++ bindings, defining new constants, and introducing comprehensive tests to ensure the new functionality works as expected.

**New statistics feature:**

* Added the `statistics` method to the `Cursor` class in `mssql_python/cursor.py`, allowing users to retrieve index and table statistics for a given table, with options to filter by uniqueness, catalog, schema, and quickness of statistics retrieval.
* Defined new constants (`SQL_INDEX_UNIQUE`, `SQL_INDEX_ALL`, `SQL_QUICK`, `SQL_ENSURE`) in `mssql_python/constants.py` to support the statistics method options.

**C++ binding and integration:**

* Added the `SQLStatisticsFunc` type and related function pointer to `mssql_python/pybind/ddbc_bindings.h` and initialized it in `mssql_python/pybind/ddbc_bindings.cpp`.
* Implemented the `SQLStatistics_wrap` function and exposed it to Python via the `DDBCSQLStatistics` binding in the pybind module. 

**Testing:**

* Added a comprehensive suite of tests in `tests/test_004_cursor.py` for the new `statistics` method, including tests for setup, basic functionality, unique index filtering, empty tables, non-existent tables, result structure, catalog filtering, the `quick` parameter, and cleanup.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -1254,6 +1254,100 @@ class Cursor:
             result_rows.append(row)
         
         return result_rows
+    
+    def statistics(self, table, catalog=None, schema=None, unique=False, quick=True):
+        """
+        Creates a result set of statistics about a single table and the indexes associated 
+        with the table by executing SQLStatistics.
+        
+        Args:
+            table (str): The name of the table.
+            catalog (str, optional): The catalog name. Defaults to None.
+            schema (str, optional): The schema name. Defaults to None.
+            unique (bool, optional): If True, only unique indexes are returned. 
+                                    If False, all indexes are returned. Defaults to False.
+            quick (bool, optional): If True, CARDINALITY and PAGES are returned only 
+                                    if readily available. Defaults to True.
+        
+        Returns:
+            list: A list of Row objects containing statistics about the table and its indexes.
+                  Each row contains: table_cat, table_schem, table_name, non_unique,
+                  index_qualifier, index_name, type, ordinal_position, column_name,
+                  asc_or_desc, cardinality, pages, filter_condition
+        """
+        self._check_closed()
+        
+        # Always reset the cursor first to ensure clean state
+        self._reset_cursor()
+        
+        # Convert None values to empty strings as required by ODBC API
+        catalog_p = "" if catalog is None else catalog
+        schema_p = "" if schema is None else schema
+        table_p = table  # Table name is required
+        
+        # Set unique flag (SQL_INDEX_UNIQUE = 0, SQL_INDEX_ALL = 1)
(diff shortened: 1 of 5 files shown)


mssql_python/cursor.py
```diff
@@ -1254,6 +1254,100 @@ def rowVerColumns(self, table, catalog=None, schema=None, nullable=True):
...
+        Returns:
+            list: A list of Row objects containing statistics about the table and its indexes.
+                  Each row contains: table_cat, table_schem, table_name, non_unique,
+                  index_qualifier, index_name, type, ordinal_position, column_name,
+                  asc_or_desc, cardinality, pages, filter_condition
+        """
+        self._check_closed()
+        
+        # Always reset the cursor first to ensure clean state
+        self._reset_cursor()
```
sumitmsft, line 1298: The method resets the cursor before use, which is good for avoiding stale state. However, if an error occurs during the `ddbc_bindings.DDBCSQLFetchAll` call, there isn't a finally block or context manager to guarantee that resources are properly released. This could lead to resource leaks if exceptions occur, especially in long-running applications.
Can we consider wrapping the fetch logic in a try/finally block, or using a context manager if available, to ensure cursor or statement handles are cleaned up regardless of errors.

mssql_python/cursor.py
```diff
@@ -1254,6 +1254,100 @@ def rowVerColumns(self, table, catalog=None, schema=None, nullable=True):
...
+        # Define column names in ODBC standard order
+        column_names = [
+            "table_cat", "table_schem", "table_name", "non_unique",
+            "index_qualifier", "index_name", "type", "ordinal_position",
+            "column_name", "asc_or_desc", "cardinality", "pages", "filter_condition"
+        ]
+        
+        # Fetch all rows
+        rows_data = []
+        ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
```
sumitmsft: The current implementation fetches all statistics rows for a table at once using `ddbc_bindings.DDBCSQLFetchAll`. While this is efficient for small to moderate tables, it can have significant memory impact for very large tables with many indexes or columns.

mssql_python/cursor.py
jahnvi480: Have removed fetch logic so we don't need to take care of this

bewithgaurav (approved):

PR #199, opened 2025-08-22
FEAT: Adding Columns API in cursor

jahnvi480:

> [AB#34927](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34927)

### Summary   
This pull request adds support for retrieving column metadata from SQL Server using the ODBC `SQLColumns` API, and exposes this functionality to Python through the `columns()` method on the `cursor` class. The implementation includes both backend C++ bindings and a high-level Python interface, as well as careful handling of cursor state and result set mapping.

The most important changes are:

**Python interface:**

* Added a new `columns()` method to the `cursor` class in `mssql_python/cursor.py` to retrieve detailed column metadata for tables using the ODBC `SQLColumns` function. This method includes extensive documentation, result mapping, and warnings about potentially expensive queries. It also wraps and overrides the fetch methods to provide specialized column mappings for the result set.

* Updated the `execute()` method in `cursor.py` to restore the original fetch methods if they were previously overridden, ensuring correct behavior after using `columns()`.

**C++ bindings and backend:**

* Added the `SQLColumnsFunc` typedef, function pointer, and symbol loading for the `SQLColumnsW` ODBC API in `ddbc_bindings.h` and `ddbc_bindings.cpp`. This ensures the backend can call the ODBC function for column metadata.

* Implemented the `SQLColumns_wrap` C++ function in `ddbc_bindings.cpp` to call the ODBC `SQLColumns` API, handling both Windows and Unix platforms, and converting Python objects to wide strings as needed.

* Exposed a new `DDBCSQLColumns` method in the Python extension module, allowing Python code to call the new backend functionality.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -1356,8 +1356,8 @@ class Cursor:
         
         Args:
             table (str, optional): The table name pattern. Default is None (all tables).
-            catalog (str, optional): The catalog name. Default is None.
-            schema (str, optional): The schema name pattern. Default is None.
+            catalog (str, optional): The catalog name. Default is None (current catalog).
+            schema (str, optional): The schema name pattern. Default is None (all schemas).
             column (str, optional): The column name pattern. Default is None (all columns).
         
         Returns:
@@ -1381,6 +1381,13 @@ class Cursor:
                   - ordinal_position: Column sequence number (starting with 1)
                   - is_nullable: "NO" means column does not allow NULL, "YES" means it does
         
+        Warning:
+            Calling this method without any filters (all parameters as None) will enumerate 
+            EVERY column in EVERY table in the database. This can be extremely expensive in 
+            large databases, potentially causing high memory usage, slow execution times, 
+            and in extreme cases, timeout errors. Always use filters (catalog, schema, table, 
+            or column) whenever possible to limit the result set.
+    
         Example:
             # Get all columns in table 'Customers'
             columns = cursor.columns(table='Customers')
@@ -1396,19 +1403,13 @@ class Cursor:
         # Always reset the cursor first to ensure clean state
         self._reset_cursor()
         
-        # Convert None values to empty strings as required by ODBC API
-        catalog_p = "" if catalog is None else catalog
-        schema_p = "" if schema is None else schema
-        table_p = "" if table is None else table
-        column_p = "" if column is None else column
-        
         # Call the SQLColumns function
         retcode = ddbc_bindings.DDBCSQLColumns(
             self.hstmt,
-            catalog_p,
-            schema_p,
-            table_p,
-            column_p
+            catalog,
+            schema,
+            table,
+            column
         )
         check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, retcode)
         
@@ -1417,7 +1418,14 @@ class Cursor:
         try:
             ddbc_bindings.DDBCSQLDescribeCol(self.hstmt, column_metadata)
             self._initialize_description(column_metadata)
-        except Exception:
+        except InterfaceError as e:
+            log('error', f"Driver interface error during metadata retrieval: {e}")
+
+        except Exception as e:
+            # Log the exception with appropriate context
+            log('error', f"Failed to retrieve column metadata: {e}. Using standard ODBC column definitions instead.")
+
+        if not self.description:
             # If describe fails, create a manual description for the standard columns
             column_types = [str, str, str, str, int, str, int, int, int, int, int, str, str, int, int, int, int, str]
             self.description = [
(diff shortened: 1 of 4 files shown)


mssql_python/cursor.py
```diff
@@ -1348,6 +1348,122 @@ def statistics(self, table, catalog=None, schema=None, unique=False, quick=True)
...
+        column_names = [
+            "table_cat", "table_schem", "table_name", "column_name", "data_type",
+            "type_name", "column_size", "buffer_length", "decimal_digits", 
+            "num_prec_radix", "nullable", "remarks", "column_def", "sql_data_type",
+            "sql_datetime_sub", "char_octet_length", "ordinal_position", "is_nullable"
+        ]
+        
+        # Fetch all rows
+        rows_data = []
+        ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
```
sumitmsft: The current implementation of `cursor.columns` always uses `DDBCSQLFetchAll` to load the entire result set of column metadata into memory before returning results. In environments with large catalogs, this can lead to high memory usage and latency.

Please consider refactoring to support streaming or iteration (e.g., via a generator, or chunked fetch) rather than full materialization. This would improve scalability, and allow callers to process results efficiently for large schemas.

If architectural constraints prevent immediate changes, please document this limitation in the docstring, and consider adding a stream flag for future extensibility.

Example enhancements:
Provide an iterator-based approach (reusing fetchone()/fetchmany() patterns)
Add a streaming or chunked mode
Clearly document memory usage risks in the docstring

mssql_python/cursor.py
```diff
@@ -1348,6 +1348,122 @@ def statistics(self, table, catalog=None, schema=None, unique=False, quick=True)
...
+            column_p
+        )
+        check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, retcode)
+        
+        # Initialize description from column metadata
+        column_metadata = []
+        try:
+            ddbc_bindings.DDBCSQLDescribeCol(self.hstmt, column_metadata)
+            self._initialize_description(column_metadata)
+        except Exception:
```
sumitmsft: The try/except block for initialization of column metadata currently catches all exceptions and silently falls back to a hardcoded column description without logging the error or preserving diagnostic information.

This approach can mask problems with the driver, bindings, or environment, making debugging difficult and potentially hiding critical issues.

- Please log the original exception (e.g., using Python's logging module) with relevant context, ideally at WARNING or ERROR level.
- Narrow the exception type if possible (e.g., catch only expected driver exceptions).

mssql_python/cursor.py
```diff
@@ -1348,6 +1348,122 @@ def statistics(self, table, catalog=None, schema=None, unique=False, quick=True)
...
+            
+            # Get column named 'CustomerID' in any table
+            columns = cursor.columns(column='CustomerID')
+        """
+        self._check_closed()
+        
+        # Always reset the cursor first to ensure clean state
+        self._reset_cursor()
+        
+        # Convert None values to empty strings as required by ODBC API
```
sumitmsft: In the `cursor.columns` method, the code currently converts all `None` parameters (for catalog, schema, table, column) to empty strings before passing them to the underlying C++/ODBC binding. 

This approach assumes that the C++ binding (and the ODBC API) will treat empty string as a wildcard ("match all"), which is true for many ODBC drivers, but not universal. Some drivers treat empty string as "match nothing" or as an actual empty string value.

If the binding logic changes in the future, or if a user explicitly wants to search for an empty string (not a wildcard), this semantic coupling could lead to bugs and unexpected results.

I am coming from the POV when we want to migrate to Rust, these things wil start to matter to us. 

If we pass `None` directly from Python and let the C++ binding handle the difference between None (meaning "match all") and "" (meaning "match empty string"), would make more sense.

mssql_python/cursor.py
```diff
@@ -1348,6 +1348,122 @@ def statistics(self, table, catalog=None, schema=None, unique=False, quick=True)
             result_rows.append(row)
         
         return result_rows
+    
+    def columns(self, table=None, catalog=None, schema=None, column=None):
```
sumitmsft, line 1392: The docstring for columns() should include a warning that calling this method without any filters (i.e., with all arguments as None) will enumerate every column in every table, which can be very expensive in large catalogs. This may result in high memory usage and slow execution.

Recommendation: Please add a performance note in the docstring advising users to use filters (catalog, schema, table, column) when possible, and warning about the cost of unfiltered metadata retrieval on large databases.

mssql_python/cursor.py
jahnvi480: This behavior is different from pyodbc, should we make these changes or should we keep the implementation same and just add a docstring?

jahnvi480: Looks like pyodbc return cursor class instead of list of rows, removing the check for memory. 

Made the behavior similar to pyodbc

PR #206, opened 2025-08-28
FEAT: varcharmax streaming support in execute

gargsaumya:

> [AB#33395](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/33395)

### Summary   
  
This pull request focuses on improving type safety and error handling for parameter binding and execution in the MSSQL Python driver, especially for string and binary types. The changes ensure that only supported Python types are bound to SQL parameters and that errors are raised for unsupported types, preventing silent failures. Additionally, the data-at-execution (DAE) streaming path for strings is now more robust and explicit.

**Parameter type safety and error handling:**

* Changed `_map_sql_type` in `cursor.py` to raise a `TypeError` for unsupported parameter types instead of defaulting to `SQL_VARCHAR`, preventing silent type mismatches.
* In `ddbc_bindings.cpp`, added explicit type checks for `SQL_C_CHAR` and `SQL_C_BINARY` bindings, raising an error if the Python object is not a string or bytes-like object.

**DAE (Data At Execution) streaming improvements:**

* Enhanced the DAE streaming logic for `SQL_C_CHAR` in `ddbc_bindings.cpp` to handle string parameters more robustly, including chunked streaming via `SQLPutData`.
* Added support for chunked DAE streaming of Python `str` objects mapped to `SQL_C_CHAR`, ensuring large strings are sent in manageable pieces.

**Platform-specific encoding for wide strings:**

* Ensured that wide string (`SQL_C_WCHAR`) parameters are correctly converted to the platform-specific encoding before streaming.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1174,23 +1194,40 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
-                            return rc;
+                    if (matchedInfo->paramCType == SQL_C_WCHAR) {
+                        std::wstring wstr = pyObj.cast<std::wstring>();
+                        size_t totalChars = wstr.size(); // number of characters
+                        const SQLWCHAR* dataPtr = wstr.c_str();
+                        size_t offset = 0;
+                        size_t chunkChars = DAE_CHUNK_SIZE / sizeof(SQLWCHAR);
+                        while (offset < totalChars) {
+                            size_t len = std::min(chunkChars, totalChars - offset);
+                            rc = SQLPutData_ptr(hStmt, (SQLPOINTER)(dataPtr + offset), static_cast<SQLLEN>(len));
```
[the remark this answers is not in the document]
gargsaumya: For SQL_C_WCHAR, SQLPutData expects the length in characters, not bytes. Multiplying by sizeof(SQLWCHAR) would make the driver interpret too much data, causing right-truncation errors. Our current logic correctly passes the number of characters per chunk.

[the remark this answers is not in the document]
sumitmsft: @gargsaumya SQLPutData (as per https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlputdata-function?view=sql-server-ver15) expects "amount of data". If that is the case then what Copilot suggested seems to be correct because as per ODBC spec SQLPutData expects a byte count for character data (unless using SQL_NTS). Passing just the character count will truncate NVARCHAR(MAX) data (e.g., only half transmitted when sizeof(SQLWCHAR)=2).

What do you think?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1203,23 +1223,40 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
                     continue;
                 }
                 if (py::isinstance<py::str>(pyObj)) {
-                    std::wstring wstr = pyObj.cast<std::wstring>();
-#if defined(__APPLE__) || defined(__linux__)
```
sumitmsft, line 1207: On non-Windows platforms, std::wstring uses UTF-32 encoding while ODBC expects wide-character data (SQLWCHAR) in UTF-16 (2 bytes per character). The platform-specific conversion (`#if defined(__APPLE__) || defined(__linux__)`) ensured correct transformation from UTF-32 to UTF-16 before streaming to SQLPutData.

I see that the above logic\conversion is removed and we are directly using std::wstring::c_str(). Isn't this risks sending incorrectly encoded or misaligned data, which can result in data corruption, truncated/malformed strings, or buffer overruns? 

This is critical to maintain cross-platform correctness, especially for NVARCHAR(MAX) and other Unicode columns containing non-ASCII or multi-byte characters.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1203,23 +1223,40 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
...
+                    } else if (matchedInfo->paramCType == SQL_C_CHAR) {
+                        std::string s = pyObj.cast<std::string>();
+                        size_t totalBytes = s.size();
+                        const char* dataPtr = s.data();
+                        size_t offset = 0;
+                        size_t chunkBytes = DAE_CHUNK_SIZE;
+                        while (offset < totalBytes) {
+                            size_t len = std::min(chunkBytes, totalBytes - offset);
+
+                            rc = SQLPutData_ptr(hStmt, (SQLPOINTER)(dataPtr + offset), static_cast<SQLLEN>(len));
```
sumitmsft, line 1262: When casting a large `size_t` value (such as totalBytes or a chunk size) to `SQLLEN` for the `SQLPutData` length parameter, there is a risk of integer overflow on platforms where `SQLLEN` is 32 bits. If the LOB data is very large (e.g. >2GB), this can result in negative or truncated lengths, causing undefined driver behavior or data corruption. 
Can we add a guard before the cast to check that the value does not exceed and raise a meaningful error if so?

sumitmsft, line 1262: @gargsaumya - Can you please check if we have an explicit handling and a test for zero-length (empty) strings when streaming VARCHAR(MAX) or NVARCHAR(MAX) data? I am unable to see that. 
The backdrop of this check is the same issue we are currently working on. 

Without this, the code may skip sending any indicator for empty inputs, which can lead to the database column being set to NULL. For SQL Server empty string (0 length data) IS NOT EQUAL to NULL.

The streaming logic should ensure that a 0-length check is always performed for empty values, and that this case is covered in the test suite.

sumitmsft (changes requested):

Issue #205, opened 2025-08-27
Error when fetching data with empty values


adolfo-ubs:
When you make a query in Linux (also in Docker containers), if any value from the fetched data contains an empty string value, an error happens:

```
python: /workspace/mssql_python/pybind/ddbc_bindings.cpp:2004: SQLRETURN FetchBatchData(SQLHSTMT, ColumnBuffers&, pybind11::list&, pybind11::list&, SQLUSMALLINT, SQLULEN&): Assertion `dataLen > 0 && "Must be > 0 since SQL_NULL_DATA & SQL_NO_DATA is already handled"' failed.
Abortado
```

If you execute the query "SELECT '' test_colum from mytable" error can be reproduced as it also returns empty string

The error doesn't appear on Windows

### Further technical details
Python version: 3.12
Tried Ubuntu 24.04, SUSE SLES 15 and Docker container. Same error on all of them

jussihi:
From the binding code, it seems clear that the `SQL_NO_DATA` hasn't actually been handled at all. Instead, only `SQL_NULL_DATA` and `SQL_NO_TOTAL` have been handled "properly". (`SQL_NO_TOTAL` is not handled properly at all)

How can this library be at version 0.9 already, if there are still this sort of obvious holes in the whole database interaction logic?

```c++
...
            if (dataLen == SQL_NULL_DATA) {
                row.append(py::none());
                continue;
            }
            // TODO: variable length data needs special handling, this logic wont suffice
            // This value indicates that the driver cannot determine the length of the data
            if (dataLen == SQL_NO_TOTAL) {
                LOG("Cannot determine the length of the data. Returning NULL value instead."
                    "Column ID - {}", col);
                row.append(py::none());
                continue;
            }
            assert(dataLen > 0 && "Must be > 0 since SQL_NULL_DATA & SQL_NO_DATA is already handled");
...
```

bewithgaurav:
@adolfo-ubs @jussihi - Thanks a lot for raising this issue, this has now been fixed in the latest release (0.10.0)
Closing this issue now, please reopen if needed.
Thanks again for your contribution!

PR #212, opened 2025-08-29
FIX: Handle empty data + tests

bewithgaurav:

> AB#<WORK_ITEM_ID>

> GitHub Issue: #205

### Summary   
  
This pull request improves the handling of empty string and binary values in the MSSQL Python bindings, ensuring that empty data is correctly distinguished from NULL values and does not cause assertion failures. It also adds comprehensive tests to verify correct behavior for these edge cases.

**Improvements to empty value handling:**

* Updated `SQLGetData_wrap` in `ddbc_bindings.cpp` to append an empty string (`""`) or empty bytes (`b""`) when the returned data length is zero, instead of causing assertion failures or misinterpreting the value as NULL. This change applies to both string and binary column types. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1703-R1705) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1762-R1764) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1961-R1963)
* Changed assertions in `FetchBatchData` to allow zero-length data, ensuring that empty values are handled gracefully instead of triggering errors.

**Testing improvements:**

* Added new tests in `tests/test_004_cursor.py` to verify correct handling of empty strings and binary data, including distinguishing empty values from NULLs, batch fetching, and various edge cases for empty strings. These tests ensure that the fixes work as intended and prevent regressions.

**Minor code cleanup:**

* Minor formatting and comment improvements in `ddbc_bindings.cpp` for better code clarity.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1778,6 +1778,9 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
                         }
 				    } else if (dataLen == SQL_NULL_DATA) {
 					    row.append(py::none());
+                    } else if (dataLen == 0) {
```
sumitmsft, line 1781: I looked at the code and all the branches where dataLen was handled. What I feel is that dataLen needs explicit exhaustive branching in all fetch paths: SQLGetData_wrap and FetchBatchData. 

Are we handling "Unexpected negative values" which in the current code base are only protected by assert(dataLen == SQL_NO_TOTAL). 
In release builds are we going to assert the unexpected negative? Or will assert turn into a silent “treat as NULL” path? 

I think we are missing on handling "Zero-length values (empty vs NULL)" in FetchBatchData; Here we assert dataLen > 0 which blocking legitimate empty values. 

Can we do something like this:

if (dataLen == SQL_NULL_DATA) → row.append(None)
else if (dataLen == 0) → append empty (str/bytes according to dataType)
else if (dataLen > 0) → normal decode/copy
else if (dataLen == SQL_NO_TOTAL) → (log + None OR raise if unsupported streaming)
else → throw with column + dataType

With this change we can remove assert(dataLen > 0 ...)

Can we add the following tests, if not already done:
- Single-row and batch fetch of empty (not NULL) CHAR, NVARCHAR, VARBINARY.
- Simulated SQL_NO_TOTAL (if possible via mock or forced scenario) → ensures it maps to None consistently.
- Batch fetch returns empty string/bytes, not asserting.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1834,6 +1837,9 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                        // Empty string
```
sumitmsft: // Handle zero-length (non-NULL) data

Empty strings can be misleading.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1778,6 +1778,9 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                        row.append(std::string(""));
                     } else {
                         assert(dataLen == SQL_NO_TOTAL);
```
sumitmsft: Earlier I have suggested to throw an exception rather than raising an assert "assert(dataLen == SQL_NO_TOTAL)" So, when you raise an exception, let make sure that we:
- emit a structured log for unexpected negative indicator values (values other than SQL_NULL_DATA, SQL_NO_TOTAL). 
- Provide limited context (column index, SQL type) when substituting NULL.

This will help us with post‑mortem debugging if a driver or memory issue produces a bad indicator. Make sure we DO NOT log any data bytes/contents (privacy/safety).

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: checked both `SQLGetData_wrap` and `FetchBatchData` - actually, the exhaustive branching is already implemented in both functions:

**SQLGetData_wrap**:
- `dataLen == SQL_NULL_DATA` → `py::none()`  
- `dataLen == 0` → `std::string("")` (explicit empty string)
- `dataLen > 0` → normal decode
- `dataLen == SQL_NO_TOTAL` → log + `py::none()`
- `assert(dataLen == SQL_NO_TOTAL)` will take care of unexpected negative values if not `SQL_NO_TOTAL` - will be changing this to an exception 

**FetchBatchData**:
- `dataLen == SQL_NULL_DATA` → `py::none()`
- `dataLen == SQL_NO_TOTAL` → log + `py::none()`  
- `dataLen == 0` → works implicitly (`std::string(..., 0)` creates empty string)
- `dataLen > 0` → normal decode
- negative values → protected by `assert(dataLen >= 0)` - will be changing to exception

I'll try and refactor the code and make the handling more explicit, also will check adding the additional suggested testcases that can be simulated (most of these cases are already covered)

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: sure, good catch
- will add logs for unexpected dataLen with context

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: refactored it to look more explicit, removed assert and instead added a ThrowStdException for unexpected negative values, added a bunch of new tests to simulate new scenarios

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: added logs with some context

gargsaumya (approved):

PR #215, opened 2025-09-02
FIX: Improvement to parameter type inference and handling

gargsaumya:

> [AB#38383](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38383)

### Summary   
  
This pull request refactors the logic for inferring SQL types from Python parameters in the `mssql_python/cursor.py` module, especially for batch operations using `executemany`. The main improvements are more accurate type detection for integer columns by considering the minimum and maximum values in the data, and a cleaner separation of concerns in the codebase.

**Improvements to type inference and parameter handling:**

* Refactored the type mapping logic in `_map_sql_type` to use `min_val` and `max_val` for integer columns, allowing for more accurate type selection based on the actual range of values in the data.
* Updated `_create_parameter_types_list` to accept and forward `min_val` and `max_val`, supporting the improved type inference for batch operations.
* Replaced the static method `_select_best_sample_value` with a new `_compute_column_type` method, which determines a representative sample value and computes min/max for integer columns, enhancing how types are inferred for each parameter column in `executemany`. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L827-L855) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R848-R873)
* Modified the `executemany` method to use `_compute_column_type` for each parameter column, passing the computed min/max values to `_create_parameter_types_list` for better type assignment.

**Code cleanup:**

* Removed the now-unnecessary `_select_best_sample_value` static method, consolidating logic and reducing code duplication.


mssql_python/cursor.py
```diff
@@ -899,11 +898,13 @@ def executemany(self, operation: str, seq_of_parameters: list) -> None:
 
         for col_index in range(param_count):
             column = [row[col_index] for row in seq_of_parameters]
-            sample_value = self._select_best_sample_value(column)
-            dummy_row = list(seq_of_parameters[0])
-            parameters_type.append(
-                self._create_parameter_types_list(sample_value, param_info, dummy_row, col_index)
+            sample_value, min_val, max_val = self._compute_column_type(column)
```
sumitmsft, line 910: The current `executemany` implementation only applies type conversion (e.g., Decimal to NumericData, date string to date object) to a single sample value per column when building parameter metadata. The actual batch data (`columnwise_params`) still contains the original unconverted values. This differs from the single-row execute path, where conversions are applied to the real parameters list. 

Can you confirm if the native layer expects pre-transformed objects? Because if this correct then it can cause incorrect binding or inconsistent behavior between execute and executemany??

Let's figure out how easy and non performance intrusive - if after determining the column type, apply the same transformation to every value in that column (or normalize all rows first, then infer types) - what implication would it have? 


Test cases: 
Add tests for batch decimals, date/time strings, mixed NULL + Decimal, and parity with repeated single execute calls.

mssql_python/cursor.py
```diff
@@ -899,11 +898,13 @@ def executemany(self, operation: str, seq_of_parameters: list) -> None:
...
+            modified_row = list(seq_of_parameters[0])
+            modified_row[col_index] = sample_value
+            paraminfo = self._create_parameter_types_list(
+                sample_value, param_info, modified_row, col_index, min_val=min_val, max_val=max_val
             )
+            parameters_type.append(paraminfo) 
 
         columnwise_params = self._transpose_rowwise_to_columnwise(seq_of_parameters)
         log('info', "Executing batch query with %d parameter sets:\n%s",
```
sumitmsft: Right now the code logs the actual values of every SQL parameter - both in `execute` (logs full parameter values) and `executemany` (logs entire parameter sets )

Right now, the driver is treating user data as if it were safe diagnostic info, which can violate security and compliance expectations.

If those parameters include things like email addresses, API keys, access tokens, personal names, or any other sensitive data, that information ends up in log files. Logs often go to centralized systems (e.g. Application Insights, ELK, Splunk), where broader teams or automated systems can read them. This is a data leakage risk (PII, secrets). 

We need to figure out what to do: 
- Just change level to debug, OR
- Redact values

May be something else??

sumitmsft (changes requested):
Got this from Copilot:

Missing test coverage for:

Batch decimal precision/scale aggregation
Large string threshold (NVARCHAR vs NVARCHAR(MAX)) including astral (UTF‑16 surrogate pair) edge cases
executemany sampling when a later row crosses the 4000 UTF‑16 unit boundary
Mixed Unicode/non-Unicode large strings (type chosen, DAE flag)
All‑NULL column inference (no crash, consistent default type)
Mixed bool and int values (BIT vs widen to INT)
Mixed numeric & string column fallback (string coercion, no silent truncation)
Large negative integers (e.g. −2^63+1) and 64‑bit boundary extremes (±2^63−1, ±2^31, etc.)
Mixed bool + very large int (widening logic)
Astral-heavy string where len(s) < 4000 but UTF‑16 units > 4000 (correct MAX path)
Large binary (>8000) bytes vs bytearray parity (same SQL type + isDAE)
Binary exactly at 8000 and just over (boundary behavior)
All parameters None in a column (executemany robustness)

Some tests may not be worth, so take your call and select the relevant ones.

mssql_python/cursor.py
```diff
@@ -242,23 +242,27 @@ def _map_sql_type(self, param, parameters_list, i):
...
+            if -32768 <= min_to_check and value_to_check <= 32767:
                 return (
                     ddbc_sql_const.SQL_SMALLINT.value,
                     ddbc_sql_const.SQL_C_SHORT.value,
                     5,
                     0,
                     False,
                 )
-            if -2147483648 <= param <= 2147483647:
+            if -2147483648 <= min_to_check and value_to_check <= 2147483647:
```
bewithgaurav, line 269: this might be unrelated to this PR, but I've seen a lot of constants which need to be added in cursor.py, could you please add a refactoring task to cursor? we need to put these numbers as constants inside constants.py

mssql_python/cursor.py
```diff
@@ -824,35 +826,6 @@ def execute(
...
-        if not non_nulls:
-            return None
-        if all(isinstance(v, int) for v in non_nulls):
-            # Pick the value with the widest range (min/max)
-            return max(non_nulls, key=lambda v: abs(v))
-        if all(isinstance(v, float) for v in non_nulls):
-            return 0.0
-        if all(isinstance(v, decimal.Decimal) for v in non_nulls):
-            return max(non_nulls, key=lambda d: len(d.as_tuple().digits))
-        if all(isinstance(v, str) for v in non_nulls):
```
bewithgaurav, line 857: this was earlier changed to accomodate binary data types, could you please pull in latest main and run tests so we know this removal doesn't cause regression? will re-review post updation

gargsaumya: This function will anyway get removed but I have now pulled in latest main.

mssql_python/cursor.py
gargsaumya: sure, added a work item in our next semester's excel for now.

mssql_python/cursor.py
gargsaumya: Moved parameter logging from info to debug to avoid accidentally exposing sensitive user data in logs for now.

mssql_python/cursor.py
gargsaumya: That's a good point, @sumitmsft . So far, ODBC has accepted the data as we pass it, but it would be better to apply type conversion to all values in executemany to avoid any unexpected behavior. This might have some performance impact, but it will ensure correctness and consistent behavior. 
Also, handling this (for array-binding + streaming) will be another task, so we'll need to decide when to take it up.

sumitmsft (dismissed):

bewithgaurav (approved):

bewithgaurav (dismissed):

sumitmsft (dismissed):

bewithgaurav (approved):

PR #217, opened 2025-09-03
FIX: Executemany String corruption on Unix systems + tests

bewithgaurav:

> [AB#38479](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38479)

## Fix string handling in executemany batch operations on Unix systems

**Issue:** 
`executemany` was corrupting string data on Unix/macOS due to encoding mismatch in batch parameter binding. Strings were either truncated or returned garbage data.

**Root Cause:** 
In `BindParameterArray` function, `std::wstring` (UTF-32 on Unix) was directly copied to `SQLWCHAR` buffers (UTF-16) without proper encoding conversion, while single parameter binding in `execute` correctly used `WStringToSQLWCHAR()`.

**Fix:**
- Added platform-specific string conversion in `BindParameterArray` for `SQL_C_WCHAR` case
- Unix: Use `WStringToSQLWCHAR()` for proper UTF-32 → UTF-16 conversion  
- Windows: Keep existing direct copy (both are UTF-16)

**Impact:**
- All string data (including empty strings) now correctly preserved in executemany
- Eliminates data corruption and truncation on Unix platforms
- Maintains existing Windows behavior
- Consistent string handling between `execute` and `executemany`


tests/test_004_cursor.py
```diff
@@ -759,6 +888,405 @@
...
+            # All empty strings
+            (1, '', ''),
+            # One empty, one not
+            (2, '', 'not empty'),
+            (3, 'not empty', ''),
+            # Special whitespace cases
+            (4, ' ', '  '),  # Single and double space
+            (5, '\t', '\n'),  # Tab and newline
+            # Mixed Unicode and empty
+            # (6, '', '🚀'), #TODO: Uncomment once nvarcharmax, varcharmax and unicode support is implemented for executemany
```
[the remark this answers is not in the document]
bewithgaurav: this is a yet to be implemented usecase, and a known one - hence added a todo here. dismissing this as false positive.

gargsaumya (dismissed):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1356,7 +1356,13 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
                                 std::string offending = WideToUTF8(wstr);
                                 ThrowStdException("Input string exceeds allowed column size at parameter index " + std::to_string(paramIndex));
                             }
+#if defined(__APPLE__) || defined(__linux__)
+                            auto utf16Buf = WStringToSQLWCHAR(wstr);
+                            size_t copySize = std::min(utf16Buf.size(), static_cast<size_t>(info.columnSize + 1));
```
sumitmsft: After converting the string to UTF-16 using WStringToSQLWCHAR(wstr), the code does not check if the number of UTF-16 units fits within the column size (info.columnSize). 

Some Unicode characters (like emoji) use two UTF-16 units, so the converted string can be longer than expected. If we don’t check, the string may get cut off incorrectly or become invalid.

Can you please add a check to make sure the converted UTF-16 length does not exceed the column size before copying.

sumitmsft: What happens when converted UTF-16 string is too long?

`std::min(utf16Buf.size(), info.columnSize + 1)` is used to copy the string, so if the converted UTF-16 string is too long, it just gets cut off. This means some data can be lost without any error. I would propose to raise an exception for oversized input.
We should raise an error if UTF-16 string is too big, instead of silently truncating. This will help prevent data corruption and keep error handling consistent.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1356,7 +1356,13 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
...
+                            std::memcpy(wcharArray + i * (info.columnSize + 1), utf16Buf.data(), copySize * sizeof(SQLWCHAR));
```
sumitmsft: Right now, the code only checks the original string length before conversion, but not after it’s turned into UTF-16 units. If the converted string is longer (because of special characters), it can slip through and overflow the buffer. 
If copySize < info.columnSize + 1 due to truncation, the copied buffer may not end with a null terminator (SQLWCHAR(0)). This risks reading past the buffer when downstream code expects a null-terminated string.

Please add a check on the final UTF-16 length after conversion to catch oversize strings correctly.

tests/test_004_cursor.py
```diff
@@ -887,6 +887,384 @@ def test_execute_many(cursor, db_connection):
...
+        
+        cursor.execute("SELECT COUNT(*) FROM #pytest_null_vs_empty WHERE data = ''")
+        empty_count = cursor.fetchone()[0]
+        assert empty_count == 2, f"Expected 2 empty strings, got {empty_count}"
+    except Exception as e:
+        pytest.fail(f"Executemany NULL vs empty string test failed: {e}") 
+    finally:
+        cursor.execute("DROP TABLE IF EXISTS #pytest_null_vs_empty")
+        db_connection.commit()
+
```
sumitmsft, line 1267: Can you see if we can add a case where the string’s UTF-16 length is different from its code point length, especially near the column size limit? 
Try to insert an emoji (which is 2 UTF-16 units) into a column that only accepts 1 unit.


Also, there’s no test that checks if trying to insert an oversized multi-unit character (like an emoji) raises an error.

Please add tests for these scenarios to make sure future changes don’t break this behavior.

sumitmsft (changes requested):
LEft a few comments.

tests/test_004_cursor.py
bewithgaurav: added tests for this

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: resolved and added tests for this

sumitmsft (approved):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1352,11 +1352,25 @@
...
+                            // Convert to UTF-16 first, then check the actual UTF-16 length
+                            auto utf16Buf = WStringToSQLWCHAR(wstr);
+                            // Check UTF-16 length (excluding null terminator) against column size
+                            if (utf16Buf.size() > 0 && (utf16Buf.size() - 1) > info.columnSize) {
+                                std::string offending = WideToUTF8(wstr);
+                                ThrowStdException("Input string UTF-16 length exceeds allowed column size at parameter index " + std::to_string(paramIndex) + 
+                                    ". UTF-16 length: " + std::to_string(utf16Buf.size() - 1) + ", Column size: " + std::to_string(info.columnSize));
+                            }
+                            // If we reach here, the UTF-16 string fits - copy it completely
+                            std::memcpy(wcharArray + i * (info.columnSize + 1), utf16Buf.data(), utf16Buf.size() * sizeof(SQLWCHAR));
```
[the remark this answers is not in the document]
bewithgaurav: this is to be taken up separately, ADO task:
- https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/38483

sumitmsft (approved):

PR #218, opened 2025-09-03
FIX: Binary data padding + tests

bewithgaurav:

> [AB#38480](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38480)

### Summary   
This pull request improves how binary data (Python `bytes` and `bytearray`) is handled in the MSSQL Python driver, especially for edge cases like empty values, mixed types, and large binaries. It also significantly expands the test suite to cover these scenarios and documents current driver limitations regarding parameter and fetch buffer sizes.

**Binary Data Handling Improvements**

* Updated `_map_sql_type` in `cursor.py` to always use `VARBINARY` for Python `bytes`/`bytearray`, avoiding storage waste and ensuring correct handling of variable-length data. This removes previous logic that sometimes used fixed-length `BINARY` and simplifies type mapping.
* Improved sample value selection in `_select_best_sample_value` to correctly handle columns that contain only binary types (`bytes` or `bytearray`).

**Test Suite Enhancements for Binary Data**

* Fixed and clarified the `test_longvarbinary` test to expect the correct number of rows and removed assumptions about zero-padding in returned binary data.
* Added new tests covering edge cases:
  - Executemany with binary data and empty byte arrays, ensuring correct insertion and retrieval of empty, regular, and null binary values.
  - Binary data over 8000 bytes, documenting driver limitations (8192 bytes for parameters, 4096 bytes for fetch buffer) and verifying correct error handling and data retrieval for supported sizes.
  - All empty binaries: verifies that multiple empty binary rows are handled and returned as zero-length `bytes`.
  - Mixing `bytes` and `bytearray` types in the same column, confirming consistent storage and retrieval as `bytes`.
  - Binary columns with mostly small/empty values and one large value, ensuring the large value is handled within driver limits.
  - Table with only NULL and empty binary values, verifying clear distinction between NULL and empty values and correct query results.


mssql_python/cursor.py
```diff
@@ -380,33 +380,21 @@ def _map_sql_type(self, param, parameters_list, i):
...
-                    0,
-                    True,
-                )
+            # Use VARBINARY for Python bytes/bytearray since they are variable-length by nature.
+            # This avoids storage waste from BINARY's zero-padding and matches Python's semantics.
             return (
-                ddbc_sql_const.SQL_BINARY.value,
+                ddbc_sql_const.SQL_VARBINARY.value,
                 ddbc_sql_const.SQL_C_BINARY.value,
                 len(param),
```
sumitmsft, line 399: Same as above: 
What happens when param is say, b'' -->empty bytes/bytearray, then len(param) is 0.
This means the returned column size for an empty binary is still 0, which may cause "Invalid precision" errors. 

We can also add a regression test with all batch values as b''/None to ensure no precision errors occur.

mssql_python/cursor.py
```diff
@@ -380,33 +380,21 @@ def _map_sql_type(self, param, parameters_list, i):
...
-                    0,
-                    False,
-                )
+            # Use VARBINARY for Python bytes/bytearray since they are variable-length by nature.
+            # This avoids storage waste from BINARY's zero-padding and matches Python's semantics.
             return (
-                ddbc_sql_const.SQL_BINARY.value,
+                ddbc_sql_const.SQL_VARBINARY.value,
                 ddbc_sql_const.SQL_C_BINARY.value,
                 len(param),
```
sumitmsft, line 388: What happens when param is say, b'' -->empty bytes/bytearray, then len(param) is 0.
This means the returned column size for an empty binary is still 0, which may cause "Invalid precision" errors.

We can also add a regression test with all batch values as b''/None to ensure no precision errors occur.

mssql_python/cursor.py
```diff
@@ -380,33 +380,21 @@ def _map_sql_type(self, param, parameters_list, i):
...
                 0,
                 False,
             )
 
         if isinstance(param, bytearray):
-            if len(param) > 8000:  # Assuming VARBINARY(MAX) for long byte arrays
```
sumitmsft, line 400: The old code did not truly support VARBINARY(MAX)—it just used SQL_VARBINARY with a large length, which only works if the DB column is declared as VARBINARY(MAX).

The new code also always uses SQL_VARBINARY and sets length to len(param), so it behaves the same for large binaries.

If the DB column is not VARBINARY(MAX), inserting values >8000 bytes may fail or truncate. --- Correct? 

This code still doesn't fully handle large binary values unless the DB schema is correct. Is it on purpose because Saumya has already handled VARBINARY(MAX) in her code?

tests/test_004_cursor.py
```diff
@@ -1265,6 +1265,56 @@ def test_executemany_null_vs_empty_string(cursor, db_connection):
         cursor.execute("DROP TABLE IF EXISTS #pytest_null_vs_empty")
         db_connection.commit()
 
+def test_executemany_binary_data_edge_cases(cursor, db_connection):
```
sumitmsft, line 1268: Can you check if we have following tests:

- Test for value > 8000 bytes.
- Test for all-empty binaries
- Insert data into a binary column where some rows use Python’s bytes type and others use bytearray for the same column.
- Insert several rows into a binary column, most rows have small or empty binary values (b''), but one row contains a very large binary value (e.g., 10,000 bytes).
- Test for only NULL + empty (ensures fallback doesn’t produce size=0).

mssql_python/cursor.py
bewithgaurav: blank/none binary testcase is already present, resolving this

tests/test_004_cursor.py
bewithgaurav: Added tests for cases which are not dependent on LOB streaming support, some are already present but added a few more

mssql_python/cursor.py
bewithgaurav: Yes, correct - more LOB cases will be upcoming in new PRs

mssql_python/cursor.py
bewithgaurav: regarding invalid precision on blank values, there will be no errors in that case - what we send is what we get, even if it is blank
added testcase `test_all_empty_binaries` to ensure regression failures

sumitmsft (approved):

PR #219, opened 2025-09-03
FEAT: streaming support in fetchone for varcharmax data type

gargsaumya:

> [AB#38110](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38110)
[AB#34162](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34162)

### Summary   
  
This pull request significantly improves the handling of large object (LOB) data types (such as large strings and binary data) in the MSSQL Python driver, especially for fetching and streaming variable-length data. The changes introduce robust streaming logic for LOB columns, prevent data truncation, and ensure correct type handling for both single-row and batch fetches. Additionally, the code now detects LOB columns and automatically switches to per-row streaming when necessary, improving reliability and correctness for large datasets.

**LOB Streaming and Fetching Improvements:**

* Introduced the `FetchLobColumnData` function in `ddbc_bindings.cpp` to stream LOB data (CHAR, WCHAR, and BINARY types) in chunks, correctly handling nulls, null-terminators, and platform-specific encoding. This prevents truncation and errors when fetching large columns.
* Updated `SQLGetData_wrap` to use streaming for LOB columns or when data length is unknown/too large, for both narrow and wide character types, as well as binary data. This ensures correct retrieval of all data regardless of size. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1749-L1771) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1782-L1800) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1916-R1967)

**Batch Fetch Logic Enhancements:**

* Modified `FetchBatchData` to detect LOB columns and use streaming fetch for those columns, avoiding exceptions and ensuring all data is retrieved for large columns in batch operations. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2319-R2415) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2379-R2487) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R2499-R2501) [[4]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2424-R2515) [[5]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2516-R2609)
* Updated `FetchMany_wrap` to pre-scan columns for LOB types and, if any are found, fall back to row-by-row streaming fetch for those rows; otherwise, it proceeds with standard batch fetching.

**Type Mapping and Constants:**

* Adjusted `_map_sql_type` in `cursor.py` to map long string types to `SQL_WVARCHAR`/`SQL_VARCHAR` with length 0 for streaming, aligning with the new LOB streaming logic.
* Defined `SQL_MAX_LOB_SIZE` (8000) as the threshold for LOB streaming, centralizing the logic for when to treat columns as LOBs.

These changes collectively make LOB handling more robust, reduce the risk of data truncation, and improve compatibility across platforms.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -1860,8 +1860,6 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
             row.append(py::none());
             continue;
         }
-        LOG("Fetching column {}: size={}, type={}", i, columnSize, dataType);
-
 
         switch (dataType) {
             case SQL_CHAR:
@@ -1922,7 +1920,6 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
             case SQL_WVARCHAR:
             case SQL_WLONGVARCHAR: {
                 if (columnSize == SQL_NO_TOTAL || columnSize > 4000) {
-                    std::cout << "Column size: " << columnSize << std::endl;
                     LOG("Streaming LOB for column {} (NVARCHAR)", i);
                     row.append(FetchLobColumnData(hStmt, i, SQL_C_WCHAR, true, false));
                 } else {
--- a/eng/pipelines/pr-validation-pipeline.yml
+++ b/eng/pipelines/pr-validation-pipeline.yml
@@ -7,76 +7,76 @@ trigger:
       - main
 
 jobs:
-# - job: PytestOnWindows
-#   displayName: 'Windows x64'
-#   pool:
-#     vmImage: 'windows-latest'
+- job: PytestOnWindows
+  displayName: 'Windows x64'
+  pool:
+    vmImage: 'windows-latest'
   
-#   steps:
-#   - task: UsePythonVersion@0
-#     inputs:
-#       versionSpec: '3.13'
-#       addToPath: true
-#       githubToken: $(GITHUB_TOKEN)
-#     displayName: 'Use Python 3.13'
+  steps:
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '3.13'
+      addToPath: true
+      githubToken: $(GITHUB_TOKEN)
+    displayName: 'Use Python 3.13'
 
-#   - script: |
-#       python -m pip install --upgrade pip
-#       pip install -r requirements.txt
-#     displayName: 'Install dependencies'
+  - script: |
+      python -m pip install --upgrade pip
+      pip install -r requirements.txt
+    displayName: 'Install dependencies'
 
-#   # Start LocalDB instance
-#   - powershell: |
-#       sqllocaldb create MSSQLLocalDB
-#       sqllocaldb start MSSQLLocalDB
-#     displayName: 'Start LocalDB instance'
+  # Start LocalDB instance
+  - powershell: |
+      sqllocaldb create MSSQLLocalDB
+      sqllocaldb start MSSQLLocalDB
+    displayName: 'Start LocalDB instance'
 
-#   # Create database and user
-#   - powershell: |
-#       sqlcmd -S "(localdb)\MSSQLLocalDB" -Q "CREATE DATABASE TestDB"
-#       sqlcmd -S "(localdb)\MSSQLLocalDB" -Q "CREATE LOGIN testuser WITH PASSWORD = '$(DB_PASSWORD)'"
-#       sqlcmd -S "(localdb)\MSSQLLocalDB" -d TestDB -Q "CREATE USER testuser FOR LOGIN testuser"
-#       sqlcmd -S "(localdb)\MSSQLLocalDB" -d TestDB -Q "ALTER ROLE db_owner ADD MEMBER testuser"
-#     displayName: 'Setup database and user'
-#     env:
-#       DB_PASSWORD: $(DB_PASSWORD)
+  # Create database and user
+  - powershell: |
+      sqlcmd -S "(localdb)\MSSQLLocalDB" -Q "CREATE DATABASE TestDB"
+      sqlcmd -S "(localdb)\MSSQLLocalDB" -Q "CREATE LOGIN testuser WITH PASSWORD = '$(DB_PASSWORD)'"
+      sqlcmd -S "(localdb)\MSSQLLocalDB" -d TestDB -Q "CREATE USER testuser FOR LOGIN testuser"
+      sqlcmd -S "(localdb)\MSSQLLocalDB" -d TestDB -Q "ALTER ROLE db_owner ADD MEMBER testuser"
+    displayName: 'Setup database and user'
+    env:
+      DB_PASSWORD: $(DB_PASSWORD)
 
-#   - script: |
-#       cd mssql_python\pybind
-#       build.bat x64
-#     displayName: 'Build .pyd file'
+  - script: |
+      cd mssql_python\pybind
+      build.bat x64
+    displayName: 'Build .pyd file'
 
-#   - script: |
-#       python -m pytest -v --junitxml=test-results.xml --cov=. --cov-report=xml --capture=tee-sys --cache-clear
-#     displayName: 'Run tests with coverage'
-#     env:
-#       DB_CONNECTION_STRING: 'Server=(localdb)\MSSQLLocalDB;Database=TestDB;Uid=testuser;Pwd=$(DB_PASSWORD);TrustServerCertificate=yes'
+  - script: |
+      python -m pytest -v --junitxml=test-results.xml --cov=. --cov-report=xml --capture=tee-sys --cache-clear
+    displayName: 'Run tests with coverage'
+    env:
+      DB_CONNECTION_STRING: 'Server=(localdb)\MSSQLLocalDB;Database=TestDB;Uid=testuser;Pwd=$(DB_PASSWORD);TrustServerCertificate=yes'
 
-#   - task: PublishBuildArtifacts@1
-#     inputs:
-#       PathtoPublish: 'mssql_python/ddbc_bindings.cp313-amd64.pyd'
-#       ArtifactName: 'ddbc_bindings'
-#       publishLocation: 'Container'
-#     displayName: 'Publish pyd file as artifact'
+  - task: PublishBuildArtifacts@1
+    inputs:
+      PathtoPublish: 'mssql_python/ddbc_bindings.cp313-amd64.pyd'
+      ArtifactName: 'ddbc_bindings'
+      publishLocation: 'Container'
+    displayName: 'Publish pyd file as artifact'
 
-#   - task: PublishBuildArtifacts@1
-#     inputs:
-#       PathtoPublish: 'mssql_python/ddbc_bindings.cp313-amd64.pdb'
-#       ArtifactName: 'ddbc_bindings'
-#       publishLocation: 'Container'
-#     displayName: 'Publish pdb file as artifact'
+  - task: PublishBuildArtifacts@1
+    inputs:
+      PathtoPublish: 'mssql_python/ddbc_bindings.cp313-amd64.pdb'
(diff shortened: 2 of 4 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1725,8 +1725,88 @@ SQLRETURN SQLFetch_wrap(SqlHandlePtr StatementHandle) {
...
+            LOG("Loop {}: Column {} is NULL", loopCount, colIndex);
+            return py::none();
+        }
+        if (!SQL_SUCCEEDED(ret) && ret != SQL_SUCCESS_WITH_INFO) {
+            LOG("Loop {}: Error fetching col={} with cType={} ret={}", loopCount, colIndex, cType, ret);
+            return py::none();
+        }
+        SQLLEN copyCount = 0;
+        if (indicator > 0 && indicator != SQL_NO_TOTAL) {
+            copyCount = std::min<SQLLEN>(indicator, DAE_CHUNK_SIZE);
```
sumitmsft: I noticed that the calculation of `copyCount` relies on the indicator value:

```
if (indicator > 0 && indicator != SQL_NO_TOTAL) {
    copyCount = std::min<SQLLEN>(indicator, DAE_CHUNK_SIZE);
} else {
    copyCount = DAE_CHUNK_SIZE;
}
```

[Indicator: 
](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/using-length-and-indicator-values?view=sql-server-ver16)

So, the indicator reflects the "total column length" (or SQL_NO_TOTAL), not the number of bytes returned for this SQLGetData call. 

This means that on the "final chunk", when less than DAE_CHUNK_SIZE bytes remain, we may still copy the full buffer which means there is a chance to pulling garbage data as well.

To avoid this, we should determine copyCount based on the actual bytes returned in this call rather than indicator.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1725,8 +1725,88 @@ SQLRETURN SQLFetch_wrap(SqlHandlePtr StatementHandle) {
...
+    }
+    LOG("FetchLobColumnData: Total bytes collected = {}", buffer.size());
+
+    if (indicator == 0 || buffer.empty()) {
+        LOG("FetchLobColumnData: Returning empty string for col {}", colIndex);
+        return py::str("");
+    }
+
+    if (isWideChar) {
+        std::wstring wstr(reinterpret_cast<const wchar_t*>(buffer.data()),
```
sumitmsft: @gargsaumya Can you please validate if casting SQLWCHAR* to wchar_t* is safe on Linux and Mac? I guess SQLWCHAR and wchar_t doesn't match in the sizes and encodings. 

// From Web:
`To convert SQLWCHAR* to wchar_t* on Linux and Mac, you need to understand that SQLWCHAR is typically defined as unsigned short in ODBC implementations, while wchar_t is a platform-dependent type. On Linux and Mac, wchar_t is usually a 4-byte type (UTF-32), whereas SQLWCHAR is a 2-byte type (UTF-16). `

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1725,8 +1725,88 @@ SQLRETURN SQLFetch_wrap(SqlHandlePtr StatementHandle) {
...
+            copyCount = DAE_CHUNK_SIZE;
+        }
+
+        // Check if last byte(s) is a null terminator
+        if (copyCount > 0) {
+            if (!isWideChar && chunk[copyCount - 1] == '\0') {
+                --copyCount;
+                LOG("Loop {}: Trimmed null terminator (narrow)", loopCount);
+            } else if (copyCount >= sizeof(wchar_t)) {
+                auto wcharBuf = reinterpret_cast<const wchar_t*>(chunk.data());
```
sumitmsft: //Copy paste of my comment down below. 

@gargsaumya Can you please validate if casting SQLWCHAR* to wchar_t* is safe on Linux and Mac? I guess SQLWCHAR and wchar_t doesn't match in the sizes and encodings.

// From Web:
To convert SQLWCHAR* to wchar_t* on Linux and Mac, you need to understand that SQLWCHAR is typically defined as unsigned short in ODBC implementations, while wchar_t is a platform-dependent type. On Linux and Mac, wchar_t is usually a 4-byte type (UTF-32), whereas SQLWCHAR is a 2-byte type (UTF-16).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1725,8 +1725,88 @@ SQLRETURN SQLFetch_wrap(SqlHandlePtr StatementHandle) {
...
+                if (wcharBuf[(copyCount / sizeof(wchar_t)) - 1] == L'\0') {
```
sumitmsft: What if `copyCount` is not an exact multiple of sizeof(wchar_t)? For example, if copyCount is 7 bytes, sizeof(wchar_t) is 2), dividing will give a non-integer result (gets truncated to 3 units, which is actually only 6 bytes: the last byte is left out).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1725,8 +1725,88 @@ SQLRETURN SQLFetch_wrap(SqlHandlePtr StatementHandle) {
...
+    LOG("FetchLobColumnData: Total bytes collected = {}", buffer.size());
+
+    if (indicator == 0 || buffer.empty()) {
+        LOG("FetchLobColumnData: Returning empty string for col {}", colIndex);
+        return py::str("");
+    }
+
+    if (isWideChar) {
+        std::wstring wstr(reinterpret_cast<const wchar_t*>(buffer.data()),
+                          buffer.size() / sizeof(wchar_t));
```
sumitmsft: What if bufferis not an exact multiple of sizeof(wchar_t)?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1725,8 +1725,88 @@ SQLRETURN SQLFetch_wrap(SqlHandlePtr StatementHandle) {
...
+        }
+    }
+    LOG("FetchLobColumnData: Total bytes collected = {}", buffer.size());
+
+    if (indicator == 0 || buffer.empty()) {
+        LOG("FetchLobColumnData: Returning empty string for col {}", colIndex);
+        return py::str("");
+    }
+
+    if (isWideChar) {
```
sumitmsft, line 1815: Is there any possibility of trailing null terminator here?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1725,8 +1725,88 @@ SQLRETURN SQLFetch_wrap(SqlHandlePtr StatementHandle) {
...
+            DAE_CHUNK_SIZE,
+            &indicator
+        );
+        if (indicator == SQL_NULL_DATA) {
+            LOG("Loop {}: Column {} is NULL", loopCount, colIndex);
+            return py::none();
+        }
+        if (!SQL_SUCCEEDED(ret) && ret != SQL_SUCCESS_WITH_INFO) {
+            LOG("Loop {}: Error fetching col={} with cType={} ret={}", loopCount, colIndex, cType, ret);
+            return py::none();
```
sumitmsft, line 1758: Can we raise an exception here, instead of `py::none()`?
This is ambiguous with the one you're returning above when `indicator == SQL_NULL_DATA`

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Yes, the earlier issue has been addressed in the updated implementation. In the new implementation, we’ve replaced this with the use of actualRead, which reflects the exact number of bytes returned by SQLGetData for each call. We determine bytesRead from actualRead, and only that many bytes are appended to the buffer. Also, we trim trailing null terminators for character data before appending, ensuring that only valid data is processed.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Addressed.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Addressed.

tests/test_004_cursor.py
```diff
@@ -5305,6 +5305,119 @@ def test_empty_string_chunk(cursor, db_connection):
...
+        cursor.execute("SELECT col FROM #pytest_varcharmax WHERE col = ?", [short_str])
+        assert cursor.fetchone()[0] == short_str
+    finally:
+        cursor.execute("DROP TABLE IF EXISTS #pytest_varcharmax")
+        db_connection.commit()
+
+
+def test_varcharmax_boundary(cursor, db_connection):
+    try:
+        cursor.execute("DROP TABLE IF EXISTS #pytest_varcharmax")
```
bewithgaurav: could you add some comments on the tests? (similar to others) this will simplify whats being tested

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1735,60 +1840,53 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                                LOG("macOS/Linux: Appended CHAR string of length {} to result row", fullStr.length());
+    #else
+                                row.append(std::string(reinterpret_cast<char*>(dataBuffer.data())));
+    #endif
+                            }
+                        } else if (dataLen == SQL_NULL_DATA) {
+                            LOG("Column {} is NULL (CHAR)", i);
+                            row.append(py::none());
+                        } else if (dataLen == 0) {
+                            row.append(py::str(""));
```
bewithgaurav, line 1899: I believe this logic has been a bit more simplified in earlier PRs, please pull in main, will need review once post resolving conflicts in this one

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: The concern about copyCount not being an exact multiple of sizeof(wchar_t) is not applicable now. We now use actualRead from SQLGetData to calculate bytesRead, ensuring we only process the exact number of bytes returned. For wide characters, we explicitly check that bytesRead is at least sizeof(SQLWCHAR) before trimming trailing null terminators and only operate on whole character units. Any incomplete characters are safely ignored, and this approach aligns with how ODBC delivers data in complete chunk.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: This concern is addressed in the current implementation. We calculate the number of characters (wcharCount) by dividing the total bytes in the buffer by sizeof(SQLWCHAR). This truncates any leftover bytes that don’t form a full character, ensuring that only complete characters are processed. Moreover, according to ODBC’s specification, data is delivered in valid chunks, so incomplete characters should not occur.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: The current implementation explicitly trims trailing null terminators from both narrow and wide-character data chunks, ensuring that such padding does not corrupt the final assembled string. (this approach is aligned with ODBC’s specifications.)

bewithgaurav (approved):

sumitmsft (approved):

jahnvi480 (approved):

PR #220, opened 2025-09-03
FEAT: streaming support in fetchone for nvarcharmax data type

gargsaumya:

> [AB#38110](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38110)
[AB#34162](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34162)

### Summary   
  
This pull request improves NVARCHAR data handling in the SQL Server Python bindings and adds comprehensive tests for NVARCHAR(MAX) scenarios. The main changes include switching to streaming for large NVARCHAR values, optimizing direct fetch for smaller values, and adding tests for edge cases and boundaries to ensure correctness.

**NVARCHAR data handling improvements:**

* Updated the logic in `ddbc_bindings.cpp` to use streaming for large NVARCHAR/NCHAR columns (over 4000 characters or unknown size) and direct fetch for smaller values, optimizing performance and reliability.
* Refactored data conversion for NVARCHAR fetches, using `std::wstring` for conversion and simplifying platform-specific handling for both macOS/Linux and Windows.
* Improved handling of empty strings and NULLs for NVARCHAR columns, ensuring correct Python types are returned and logging is more descriptive.

**Testing enhancements:**

* Added new tests in `test_004_cursor.py` for NVARCHAR(MAX) covering short strings, boundary conditions (4000 chars), streaming (4100+ chars), large values (100,000 chars), empty strings, NULLs, and transaction rollback scenarios to verify correct behavior across all edge cases.

**VARCHAR/CHAR fetch improvements:**

* Improved direct fetch logic for small VARCHAR/CHAR columns and fixed string conversion to use the actual data length, preventing potential issues with null-termination and buffer size. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1825-R1830) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1841-L1850)

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -1884,6 +1884,10 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
     #else
                                 row.append(std::string(reinterpret_cast<char*>(dataBuffer.data())));
     #endif
+                            } else {
+                                // Buffer too small, fallback to streaming
+                                LOG("CHAR column {} data truncated, using streaming LOB", i);
+                                row.append(FetchLobColumnData(hStmt, i, SQL_C_CHAR, false, false));
                             }
                         } else if (dataLen == SQL_NULL_DATA) {
                             LOG("Column {} is NULL (CHAR)", i);
@@ -1924,14 +1928,20 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
                         if (dataLen > 0) {
                             uint64_t numCharsInData = dataLen / sizeof(SQLWCHAR);
                             if (numCharsInData < dataBuffer.size()) {
-            #if defined(__APPLE__) || defined(__linux__)
-                                std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()), numCharsInData);
-                                row.append(py::cast(wstr));
-            #else
+#if defined(__APPLE__) || defined(__linux__)
+                                const SQLWCHAR* sqlwBuf = reinterpret_cast<const SQLWCHAR*>(dataBuffer.data());
+                                std::wstring wstr = SQLWCHARToWString(sqlwBuf, numCharsInData);
+                                std::string utf8str = WideToUTF8(wstr);
+                                row.append(py::str(utf8str));
+#else
                                 std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()));
                                 row.append(py::cast(wstr));
-            #endif
+#endif
                                 LOG("Appended NVARCHAR string of length {} to result row", numCharsInData);
+                            }  else {
+                                // Buffer too small, fallback to streaming
+                                LOG("NVARCHAR column {} data truncated, using streaming LOB", i);
+                                row.append(FetchLobColumnData(hStmt, i, SQL_C_WCHAR, true, false));
                             }
                         } else if (dataLen == SQL_NULL_DATA) {
                             LOG("Column {} is NULL (CHAR)", i);
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -1862,12 +1862,10 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
             case SQL_CHAR:
             case SQL_VARCHAR:
             case SQL_LONGVARCHAR: {
-                // Use streaming for large VARCHAR / CHAR
                 if (columnSize == SQL_NO_TOTAL || columnSize == 0 || columnSize > 8000) {
                     LOG("Streaming LOB for column {}", i);
                     row.append(FetchLobColumnData(hStmt, i, SQL_C_CHAR, false, false));
                 } else {
-                    // Small VARCHAR, fetch directly
                     uint64_t fetchBufferSize = columnSize + 1 /* null-termination */;
                     std::vector<SQLCHAR> dataBuffer(fetchBufferSize);
                     SQLLEN dataLen;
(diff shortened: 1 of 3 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1881,62 +1880,42 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                    uint64_t fetchBufferSize = (columnSize + 1) * sizeof(SQLWCHAR);  // +1 for null terminator
+                    std::vector<SQLWCHAR> dataBuffer(columnSize + 1);
+                    SQLLEN dataLen;
+                    ret = SQLGetData_ptr(hStmt, i, SQL_C_WCHAR, dataBuffer.data(), fetchBufferSize, &dataLen);
+                    if (SQL_SUCCEEDED(ret)) {
+                        if (dataLen > 0) {
+                            uint64_t numCharsInData = dataLen / sizeof(SQLWCHAR);
+                            if (numCharsInData < dataBuffer.size()) {
+            #if defined(__APPLE__) || defined(__linux__)
+                                std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()), numCharsInData);
```
sumitmsft: Can you check on this casting? 
On macOS/Linux, wchar_t is 4 bytes, whereas SQLWCHAR is typically 2 bytes. This mismatch will make std::wstring read beyond the allocated 2-byte buffer (buffer over-read and potential crash), and garble characters. 

```std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()), numCharsInData);```

and

```std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()));```

File: mssql_python/pybind/ddbc_bindings.cpp [hunk 2]
```diff
@@ -1852,16 +1854,13 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                                std::string fullStr(reinterpret_cast<char*>(dataBuffer.data()), dataLen);
                                 row.append(fullStr);
                                 LOG("macOS/Linux: Appended CHAR string of length {} to result row", fullStr.length());
     #else
                                 row.append(std::string(reinterpret_cast<char*>(dataBuffer.data())));
     #endif
                             }
-                        } else if (dataLen == SQL_NULL_DATA) {
-                            LOG("Column {} is NULL (CHAR)", i);
-                            row.append(py::none());
```
sumitmsft: This code checked for `dataLen == SQL_NULL_DATA` and appended `py::none()`.
This explicit check was removed in the PR, so when the data is NULL, nothing is appended for that column? 

Now, if `dataLen == SQL_NULL_DATA`, the code does not append anything for that column. This breaks the invariant that every column should always have a value (even if it’s None for NULL).

I think this was the correct logic.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1881,62 +1880,42 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
-                            // TODO: Revisit
-                            std::ostringstream oss;
-                            oss << "Buffer length for fetch (" << dataBuffer.size()-1 << ") is smaller, & data "
-                                << "to be retrieved is longer (" << numCharsInData << "). ColumnID - "
-                                << i << ", datatype - " << dataType;
-                            ThrowStdException(oss.str());
+                        } else if (dataLen == 0) {
+                            row.append(py::str(""));
+                        } else {
+                            assert(dataLen == SQL_NO_TOTAL);
```
sumitmsft: So, this `else` is satisfying `dataLen < 0` The code treats all negative values as `SQL_NO_TOTAL` (unknown length), asserts, and appends python `None`. 

This condition doesn't distinguish between `SQL_NULL_DATA` (actually NULL) and `SQL_NO_TOTAL` (just unknown length).

This means if the value is actually `NULL (SQL_NULL_DATA)`, it may not append anything, or may hit an assertion failure. Ideally we must check for `dataLen == SQL_NULL_DATA` and append `py::none()`.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1881,62 +1880,42 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                } else {
+                    // Small NVARCHAR, fetch directly
+                    uint64_t fetchBufferSize = (columnSize + 1) * sizeof(SQLWCHAR);  // +1 for null terminator
+                    std::vector<SQLWCHAR> dataBuffer(columnSize + 1);
+                    SQLLEN dataLen;
+                    ret = SQLGetData_ptr(hStmt, i, SQL_C_WCHAR, dataBuffer.data(), fetchBufferSize, &dataLen);
+                    if (SQL_SUCCEEDED(ret)) {
+                        if (dataLen > 0) {
+                            uint64_t numCharsInData = dataLen / sizeof(SQLWCHAR);
+                            if (numCharsInData < dataBuffer.size()) {
```
sumitmsft, line 1936: I think we are not covering `if (numCharsInData >= dataBuffer.size()), which means truncation likely occurred because the data is more than buffer size. We are not appending value for this column.
The code just falls out of the if block and does nothing for this column.
The row may end up with missing values for columns where truncation happened.

sumitmsft (changes requested):
Left a few comments

.coveragerc
```diff
@@ -1,6 +1,8 @@
 [run]
 omit =
-    mssql_python/testing_ddbc_bindings.py
+    main.py
```
bewithgaurav: I believe this is also in main branch now, will wait for merging the latest change

.github/workflows/pr-format-check.yml
```diff
@@ -57,9 +57,9 @@ jobs:
               // Extract the summary content
               const summaryContent = summaryMatch[1];
 
-              // Remove all HTML comments including the template placeholder
+              // Remove all HTML comments including unclosed ones (template placeholders)
               const contentWithoutComments = 
-                summaryContent.replace(/<!--[\s\S]*?-->/g, '');
+                summaryContent.replace(/<!--[\s\S]*?(?:-->|$)/g, '');
```
bewithgaurav: same

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: +1, this is mistakenly present here since this is simplified in previous PRs, will be gone post pulling latest changes

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Yes, there were merge conflicts that led to this. This is now resolved.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Thanks @sumitmsft, makes sense. Also added support for this edge case for both CHAR and NVARCHAR columns.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Addressed.

PR #230, opened 2025-09-09
FEAT: Complex Data Type Support- money & smallmoney

gargsaumya:

> [AB#34936](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34936)
[AB#34937](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34937)

### Summary   
  
This pull request enhances support for SQL Server MONEY and SMALLMONEY types in the Python driver, ensuring correct handling of boundary values, NULLs, and round-trip conversions. It also improves the encoding and decoding of decimal values and adds comprehensive tests to validate the new behavior.

**MONEY and SMALLMONEY Type Handling:**

* Updated `_map_sql_type` in `cursor.py` to detect decimal values within MONEY and SMALLMONEY ranges and bind them as strings, falling back to generic numeric binding otherwise.

**Decimal Value Encoding/Decoding:**

* Improved encoding of decimal values in `ddbc_bindings.cpp` by writing them as little-endian byte arrays, ensuring compatibility with SQL_NUMERIC_STRUCT.
* Enhanced decoding of SQL DECIMAL/NUMERIC types to Python `decimal.Decimal`, with robust NULL and error handling during result fetching.

**Testing Enhancements:**

* Added comprehensive tests in `test_004_cursor.py` to verify MONEY and SMALLMONEY handling, including:
  - Insert/fetch of typical, boundary, and NULL values
  - Round-trip conversion checks
  - Boundary value assertions
  - Validation that out-of-range and invalid values raise errors

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -293,6 +293,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_CHAR.value,
                     len(parameters_list[i]),
                     0,
+                    False,
                 )
             elif -922337203685477.5808 <= param <= 922337203685477.5807:
                 # money
@@ -302,6 +303,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_CHAR.value,
                     len(parameters_list[i]),
                     0,
+                    False,
                 )
             else:
                 # fallback to generic numeric binding
@@ -311,6 +313,7 @@ class Cursor:
                     ddbc_sql_const.SQL_C_NUMERIC.value,
                     parameters_list[i].precision,
                     parameters_list[i].scale,
+                    False,
                 )
 
         
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -20,6 +20,10 @@ from .row import Row
 
 # Constants for string handling
 MAX_INLINE_CHAR = 4000  # NVARCHAR/VARCHAR inline limit; this triggers NVARCHAR(MAX)/VARCHAR(MAX) + DAE
+SMALLMONEY_MIN = decimal.Decimal('-214748.3648')
+SMALLMONEY_MAX = decimal.Decimal('214748.3647')
+MONEY_MIN = decimal.Decimal('-922337203685477.5808')
+MONEY_MAX = decimal.Decimal('922337203685477.5807')
 
 class Cursor:
     """
@@ -285,7 +289,7 @@ class Cursor:
         
         if isinstance(param, decimal.Decimal):
         # Detect MONEY / SMALLMONEY range
-            if -214748.3648 <= param <= 214748.3647:
+            if SMALLMONEY_MIN  <= param <= SMALLMONEY_MAX:
                 # smallmoney
                 parameters_list[i] = str(param)
                 return (
@@ -295,7 +299,7 @@ class Cursor:
                     0,
                     False,
                 )
-            elif -922337203685477.5808 <= param <= 922337203685477.5807:
+            elif MONEY_MIN <= param <= MONEY_MAX:
                 # money
                 parameters_list[i] = str(param)
                 return (
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -282,18 +282,49 @@ class Cursor:
                 0,
                 False,
             )
-
+        
         if isinstance(param, decimal.Decimal):
-            parameters_list[i] = self._get_numeric_data(
-                param
-            )  # Replace the parameter with the dictionary
-            return (
-                ddbc_sql_const.SQL_NUMERIC.value,
-                ddbc_sql_const.SQL_C_NUMERIC.value,
-                parameters_list[i].precision,
-                parameters_list[i].scale,
-                False,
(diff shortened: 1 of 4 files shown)


mssql_python/cursor.py
```diff
@@ -282,18 +282,39 @@ def _map_sql_type(self, param, parameters_list, i):
...
-            )  # Replace the parameter with the dictionary
-            return (
-                ddbc_sql_const.SQL_NUMERIC.value,
-                ddbc_sql_const.SQL_C_NUMERIC.value,
-                parameters_list[i].precision,
-                parameters_list[i].scale,
-                False,
-            )
+        # Detect MONEY / SMALLMONEY range
+            if -214748.3648 <= param <= 214748.3647:
```
sumitmsft: The binding for MONEY/SMALLMONEY returns scale=0. SQL Server defines both with a fixed scale of 4 (SMALLMONEY precision 10, scale 4; MONEY precision 19, scale 4). Please correct precision/scale (and preferably bind as numeric rather than VARCHAR) to avoid implicit conversions and possible plan inefficiencies.

Just to give some context:
- the MONEY and SMALLMONEY paths return tuples with scale = 0 and a precision implicitly derived from len(str(param)), but SQL Server defines:
SMALLMONEY: precision 10, scale 4
MONEY: precision 19, scale 4

- Incorrect metadata can force SQL Server to treat the parameter as a string → implicit VARCHAR → MONEY conversion per execution. This may cause SQL query plan inefficiencies

mssql_python/cursor.py
```diff
@@ -282,18 +282,39 @@ def _map_sql_type(self, param, parameters_list, i):
...
+                # smallmoney
+                parameters_list[i] = str(param)
+                return (
+                    ddbc_sql_const.SQL_VARCHAR.value,
+                    ddbc_sql_const.SQL_C_CHAR.value,
+                    len(parameters_list[i]),
```
sumitmsft, line 298: Precision is currently derived via len(str(param)), which is incorrect. Precision = total number of digits (excluding sign and decimal point). Please use fixed MONEY/SMALLMONEY precisions (19 / 10) or compute digit count from the Decimal tuple instead of string length

mssql_python/cursor.py
sumitmsft: Boundary checks for MONEY/SMALLMONEY currently use float literals (e.g. -214748.3648) against Decimal parameters. This introduces an unnecessary float→Decimal conversion with possible rounding artifacts at exact financial limits (especially the large MONEY bounds).

Can we pre-declare boundary constants using precise Decimal string constructors:

MONEY_MIN = Decimal('-922337203685477.5808')
MONEY_MAX = Decimal('922337203685477.5807')
SMALLMONEY_MIN = Decimal('-214748.3648')
SMALLMONEY_MAX = Decimal('214748.3647')

This way we compare only Decimal to Decimal.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -474,15 +474,16 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
                 decimalPtr->scale = decimalParam.scale;
                 decimalPtr->sign = decimalParam.sign;
-                // Convert the integer decimalParam.val to char array
-                std::memset(static_cast<void*>(decimalPtr->val), 0, sizeof(decimalPtr->val));
-                std::memcpy(static_cast<void*>(decimalPtr->val),
-			    reinterpret_cast<char*>(&decimalParam.val),
-                            sizeof(decimalParam.val));
+                // Zero out the byte array
+                std::memset(decimalPtr->val, 0, sizeof(decimalPtr->val));
+
```
sumitmsft: This code copies the unscaled integer `decimalParam.val` into the SQL_NUMERIC_STRUCT buffer without first confirming that:

- The digit count of the unscaled value fits the declared precision. `digits <= precision`
- `scale <= precision`
- There’s no early exit with a clear error if the value exceeds precision leading to potential silent mismatch between declared metadata and actual value.

While current uint64_t values won’t overflow the 16‑byte array, a value with more digits than precision will still be sent and this could be a data integrity risk for numeric correctness.

For this we can some tests as well:

- A test that constructs a NumericData whose val has more digits than precision and assert it raises / returns error.

sumitmsft (changes requested):
Left a few comments

sumitmsft (changes requested):
Left a few comments

mssql_python/cursor.py
bewithgaurav: just correcting the context here - the len(parameters_list[i]) is column size, not precision 
`map_sql_type` returns the below
- `A tuple containing the SQL type, C type, column size, and decimal digits.`
@gargsaumya the column size will be length only right, please correct if I'm wrong
the tests look correct and they're passing for precision cases

bewithgaurav (approved):

Issue #234, opened 2025-09-11
Datetime conversion error


arvis108:
### Describe the bug
mssql_python throws a datetime conversion error when performing string-to-string comparisons in SQL WHERE clauses, even though no datetime conversion should be involved. The same query works fine with pymssql and pyodbc.

The error occurs when passing a string parameter to compare against a string result from SQL functions like RIGHT(). The library incorrectly attempts datetime conversion instead of treating it as a simple string comparison.

### To reproduce

**test_mssql-python.py (fails):**
```python
import mssql_python

# Establish a connection
connection_string = ""
connection = mssql_python.connect(connection_string)
cursor = connection.cursor()
date_str = '2025-08-12'
# This fails with datetime conversion error
cursor.execute("SELECT a_column FROM a_table WHERE RIGHT(a_column,10) = ?", (str(date_str)))
rows = cursor.fetchall()

for row in rows:
    print(row)

connection.close()
```

### Expected behavior
The query should execute successfully since it's a simple string-to-string comparison. `RIGHT(a_column,10)` returns a string, and the parameter is a string, so no datetime conversion should occur.

**Workaround:** The query works when explicitly casting the parameter: 
```python
cursor.execute("SELECT a_column FROM a_table WHERE RIGHT(a_column,10) = cast(? as nvarchar(10))", (str(date_str)))
```

### Further technical details
Python version: 3.13.7
Operating system: Windows 10

bewithgaurav:
FYI, tried running this and works well
```python
date_str = '2025-08-12'

cursor.execute("DROP TABLE IF EXISTS a_table;")
cursor.execute("CREATE TABLE a_table (a_column VARCHAR(20));") #Also works fine with a DATE column
cursor.execute("DELETE FROM a_table;")
cursor.execute("INSERT INTO a_table (a_column) VALUES ('2025-08-12'), ('2025-08-13'), ('2025-08-14');")
conn.commit()

cursor.execute("SELECT a_column FROM a_table WHERE RIGHT(a_column,10) = ?", (str(date_str),))
rows = cursor.fetchall()
print(rows)

# Output - [('gaurav2025-08-12',)] - with VARCHAR
# Output - [(datetime.date(2025, 8, 12),)] - with VARCHAR
```

bewithgaurav:
@arvis108 – We’ve included this fix in our 0.12.0 release. Thank you for your proactive support! 
[...]

PR #235, opened 2025-09-12
FIX: Removed aggressive datetime parsing

arvis108:

> GitHub Issue: https://github.com/microsoft/mssql-python/issues/234

### Summary

This PR fixes a critical issue where string parameters were being automatically converted to datetime objects, causing failures in SQL queries that require string-to-string comparisons. The fix removes aggressive content-based datetime parsing from the `_map_sql_type()` method while preserving all existing functionality for actual datetime objects.

### Problem

The mssql-python driver was failing on queries like:
```python
cursor.execute("SELECT * FROM table WHERE RIGHT(column, 10) = ?", ('2025-08-12',))
```

This would throw datetime conversion errors because the driver was automatically trying to parse the string `'2025-08-12'` as a date object, even when it should remain a string for the RIGHT() function comparison.

### Problematic Code (Before Fix):
```python
if isinstance(param, str):
    # Attempt to parse as date, datetime, datetime2, timestamp, smalldatetime or time
    if self._parse_date(param):
        parameters_list[i] = self._parse_date(param)  # MUTATION!
        return (SQL_DATE, SQL_C_TYPE_DATE, 10, 0, False)
    if self._parse_datetime(param):
        parameters_list[i] = self._parse_datetime(param)  # MUTATION!
        return (SQL_TIMESTAMP, SQL_C_TYPE_TIMESTAMP, 26, 6, False)
    # ... more parsing attempts
```

### Why This Was Wrong:
1. **Violated separation of concerns**: Type mapping shouldn't do content parsing
2. **Unpredictable behavior**: Same string could be treated differently in different contexts
3. **Data mutation**: Modified user's parameter data without permission
4. **Against DB-API principles**: Parameter binding should be based on Python type, not content

### Solution

**Approach**: Follow PyMSSQL's proven type-based parameter mapping approach.

**Key Changes**:
1. **Removed automatic datetime parsing** from `_map_sql_type()` for string parameters
2. **Strings now always map to VARCHAR/NVARCHAR** based purely on Python type

**Files Modified**:
- `mssql_python/cursor.py`: Simplified string parameter handling in `_map_sql_type()`
- `tests/test_004_cursor.py`: Added test for string parameter behavior

### Before vs After

**Before** (Failed):
```python
date_str = '2025-08-12'
cursor.execute("SELECT * WHERE RIGHT(column, 10) = ?", (date_str,))  # DateTime conversion error
```

**After** (Works):
```python  
date_str = '2025-08-12'
cursor.execute("SELECT * WHERE RIGHT(column, 10) = ?", (date_str,))  # Works correctly
```

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -362,7 +362,7 @@ class Cursor:
                     0,
                     False,
                 )
-            
+                
             try:
                 val = uuid.UUID(param)
                 parameters_list[i] = val.bytes_le
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -376,38 +376,6 @@ class Cursor:
             except ValueError:
                 pass
 
-
-            # Attempt to parse as date, datetime, datetime2, timestamp, smalldatetime or time
-            if self._parse_date(param):
-                parameters_list[i] = self._parse_date(
-                    param
-                )  # Replace the parameter with the date object
-                return (
-                    ddbc_sql_const.SQL_DATE.value,
-                    ddbc_sql_const.SQL_C_TYPE_DATE.value,
-                    10,
-                    0,
-                    False,
-                )
-            if self._parse_datetime(param):
-                parameters_list[i] = self._parse_datetime(param)
-                return (
-                    ddbc_sql_const.SQL_TIMESTAMP.value,
-                    ddbc_sql_const.SQL_C_TYPE_TIMESTAMP.value,
-                    26,
-                    6,
-                    False,
-                )
-            if self._parse_time(param):
-                parameters_list[i] = self._parse_time(param)
-                return (
-                    ddbc_sql_const.SQL_TIME.value,
-                    ddbc_sql_const.SQL_C_TYPE_TIME.value,
-                    8,
-                    0,
-                    False,
-                )
-
             # String mapping logic here
             is_unicode = self._is_unicode_string(param)
 
--- a/tests/test_004_cursor.py
+++ b/tests/test_004_cursor.py
@@ -10661,6 +10661,89 @@ def test_decimal_separator_calculations(cursor, db_connection):
         # Cleanup
         cursor.execute("DROP TABLE IF EXISTS #pytest_decimal_calc_test")
 
+def test_date_string_parameter_binding(cursor, db_connection):
+    """Verify that date-like strings are treated as strings in parameter binding"""
+    table_name = "#pytest_date_string"
+    try:
+        drop_table_if_exists(cursor, table_name)
+        cursor.execute(f"""
+            CREATE TABLE {table_name} (
+                a_column VARCHAR(20)
+            )
+        """)
+        cursor.execute(f"INSERT INTO {table_name} (a_column) VALUES ('string1'), ('string2')")
+        db_connection.commit()
+
+        date_str = "2025-08-12"
+
+        # Should fail to match anything, since binding may treat it as DATE not VARCHAR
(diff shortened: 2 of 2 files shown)


mssql_python/cursor.py
```diff
@@ -334,38 +334,6 @@ def _map_sql_type(self, param, parameters_list, i):
...
-                parameters_list[i] = self._parse_time(param)
-                return (
-                    ddbc_sql_const.SQL_TIME.value,
-                    ddbc_sql_const.SQL_C_TYPE_TIME.value,
-                    8,
-                    0,
-                    False,
-                )
-
-            # String mapping logic here
```
bewithgaurav, line 411: can we let this comment stay as is? since this is unrelated to datetime

tests/test_004_cursor.py
```diff
@@ -6909,6 +6909,147 @@ def test_money_smallmoney_invalid_values(cursor, db_connection):
         drop_table_if_exists(cursor, "dbo.money_test")
         db_connection.commit()
 
+def test_string_vs_datetime_parameter_behavior(cursor, db_connection):
```
bewithgaurav: I believe all these test scenarios are already covered in tests for datetime/date/time.. could you please remove these tests and add below tests for the below scenarios as we discussed in #234 

Test Cases we can add:

1. **`test_date_string_parameter_binding`** - Verifies date-like strings work as string parameters (as posted above)
```python
date_str = '2025-08-12'
cursor.execute("DROP TABLE IF EXISTS a_table;")
cursor.execute("CREATE TABLE a_table (a_column VARCHAR(20));")
cursor.execute("DELETE FROM a_table;")
cursor.execute("INSERT INTO a_table (a_column) VALUES ('string1'), ('string2');")
conn.commit()
cursor.execute("SELECT a_column FROM a_table WHERE RIGHT(a_column,10) = ?", (str(date_str),))
```

2. **`test_time_string_parameter_binding`** - Verifies time-like strings work as string parameters  (this issue will be exactly same for time)
```python
time_str = '14:30:45'
cursor.execute("DROP TABLE IF EXISTS time_table;")
cursor.execute("CREATE TABLE time_table (time_col VARCHAR(20));")
cursor.execute("INSERT INTO time_table (time_col) VALUES ('prefix_14:30:45_suffix');")
conn.commit()

# This should fail because '14:30:45' gets converted to TIME type
# and SQL Server can't compare TIME against VARCHAR with prefix/suffix
cursor.execute("SELECT time_col FROM time_table WHERE time_col = ?", (time_str,))
rows = cursor.fetchall()
```

3. **`test_datetime_string_parameter_binding`** - Verifies datetime-like strings work as string parameters (same for datetime
```python
datetime_str = '2025-08-12T14:30:45'
cursor.execute("DROP TABLE IF EXISTS datetime_table;")
cursor.execute("CREATE TABLE datetime_table (datetime_col VARCHAR(30));")
cursor.execute("INSERT INTO datetime_table (datetime_col) VALUES ('prefix_2025-08-12T14:30:45_suffix');")
conn.commit()

# This should fail because '2025-08-12T14:30:45' gets converted to TIMESTAMP type
# and SQL Server can't compare TIMESTAMP against VARCHAR with prefix/suffix
cursor.execute("SELECT datetime_col FROM datetime_table WHERE datetime_col = ?", (datetime_str,))
rows = cursor.fetchall()
```

arvis108: Removed original test cases and added the ones you mentioned

bewithgaurav (approved):
This is great! lets fix the conflicts and we should be good to go :)

gargsaumya (dismissed):

gargsaumya (approved):

PR #236, opened 2025-09-15
FEAT: Complex data type support- uniqueidentifier

gargsaumya:

> [AB#34945](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34945)

> GitHub Issue: #241 

### Summary   
  

This pull request adds comprehensive support for Python's `uuid.UUID` type when interacting with SQL Server `UNIQUEIDENTIFIER` columns. It updates both the parameter binding and result retrieval logic to handle UUIDs as native Python objects, ensuring correct type mapping, byte order, and conversion in both directions. Additionally, new tests are introduced to verify correct insertion and retrieval of UUID values, including edge cases with `None`.

**UUID/UNIQUEIDENTIFIER Support Improvements:**

* Added logic in `mssql_python/cursor.py` to map Python `uuid.UUID` objects to the correct SQL types and convert them to raw bytes before sending to SQL Server. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R429-R437) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R777-R778)
* Updated the C++ binding layer (`mssql_python/pybind/ddbc_bindings.cpp`) to correctly marshal UUIDs between Python and SQL Server, including proper byte order conversion and error handling for incorrect UUID sizes.
* Enhanced result fetching in the C++ layer to convert SQL Server `UNIQUEIDENTIFIER` columns back into Python `uuid.UUID` objects, both for single-row and batch fetches, and to handle `NULL` values gracefully. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2202-R2249) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2593-R2639)

**Testing Enhancements:**

* Added new tests in `tests/test_004_cursor.py` to verify correct insertion, retrieval, and handling of `None` values in UUID columns, as well as round-trip integrity for multiple UUIDs. [[1]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69R6827-R6923) [[2]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69R17)

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -124,6 +124,14 @@ SQLBindColFunc SQLBindCol_ptr = nullptr;
 SQLDescribeColFunc SQLDescribeCol_ptr = nullptr;
 SQLMoreResultsFunc SQLMoreResults_ptr = nullptr;
 SQLColAttributeFunc SQLColAttribute_ptr = nullptr;
+SQLGetTypeInfoFunc SQLGetTypeInfo_ptr = nullptr;
+SQLProceduresFunc SQLProcedures_ptr = nullptr;
+SQLForeignKeysFunc SQLForeignKeys_ptr = nullptr;
+SQLPrimaryKeysFunc SQLPrimaryKeys_ptr = nullptr;
+SQLSpecialColumnsFunc SQLSpecialColumns_ptr = nullptr;
+SQLStatisticsFunc SQLStatistics_ptr = nullptr;
+SQLColumnsFunc SQLColumns_ptr = nullptr;
+SQLGetInfoFunc SQLGetInfo_ptr = nullptr;
 
 // Transaction APIs
 SQLEndTranFunc SQLEndTran_ptr = nullptr;
@@ -254,17 +262,29 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                     !py::isinstance<py::bytes>(param)) {
                     ThrowStdException(MakeParamMismatchErrorStr(paramInfo.paramCType, paramIndex));
                 }
-                std::string* strParam =
-                    AllocateParamBuffer<std::string>(paramBuffers, param.cast<std::string>());
-                if (strParam->size() > 8192 /* TODO: Fix max length */) {
-                    ThrowStdException(
-                        "Streaming parameters is not yet supported. Parameter size"
-                        " must be less than 8192 bytes");
+                if (paramInfo.isDAE) {
+                    // Deferred execution for VARBINARY(MAX)
+                    LOG("Parameter[{}] is marked for DAE streaming (VARBINARY(MAX))", paramIndex);
+                    dataPtr = const_cast<void*>(reinterpret_cast<const void*>(&paramInfos[paramIndex]));
+                    strLenOrIndPtr = AllocateParamBuffer<SQLLEN>(paramBuffers);
+                    *strLenOrIndPtr = SQL_LEN_DATA_AT_EXEC(0);
+                    bufferLength = 0;
+                } else {
+                    // small binary
+                    std::string binData;
+                    if (py::isinstance<py::bytes>(param)) {
+                        binData = param.cast<std::string>();
+                    } else {
+                        // bytearray
+                        binData = std::string(reinterpret_cast<const char*>(PyByteArray_AsString(param.ptr())),
+                                            PyByteArray_Size(param.ptr()));
+                    }
+                    std::string* binBuffer = AllocateParamBuffer<std::string>(paramBuffers, binData);
+                    dataPtr = const_cast<void*>(static_cast<const void*>(binBuffer->data()));
+                    bufferLength = static_cast<SQLLEN>(binBuffer->size());
+                    strLenOrIndPtr = AllocateParamBuffer<SQLLEN>(paramBuffers);
+                    *strLenOrIndPtr = bufferLength;
                 }
-                dataPtr = const_cast<void*>(static_cast<const void*>(strParam->c_str()));
-                bufferLength = strParam->size() + 1 /* null terminator */;
-                strLenOrIndPtr = AllocateParamBuffer<SQLLEN>(paramBuffers);
-                *strLenOrIndPtr = SQL_NTS;
                 break;
             }
             case SQL_C_WCHAR: {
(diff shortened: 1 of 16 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -504,7 +504,32 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                 break;
             }
             case SQL_C_GUID: {
-                // TODO
+                py::object uuid_obj = param;
+                py::object uuid_cls = py::module_::import("uuid").attr("UUID");
+                if (!py::isinstance(uuid_obj, uuid_cls)) {
+                    ThrowStdException(MakeParamMismatchErrorStr(paramInfo.paramCType, paramIndex));
+                }
+                py::bytes uuid_bytes = uuid_obj.attr("bytes").cast<py::bytes>();
```
sumitmsft: UUIDs are manually packed and unpacked between Python and ODBC’s SQLGUID structure using bitwise operations and direct memory copying. This approach is very sensitive to byte order (endianness), and if handled incorrectly, can lead to corrupted UUIDs and mismatches between what’s written and read from the database (because you need to match the exact byte order expected by both Python and ODBC—otherwise, your UUIDs will be incorrect.)

What I want you to comment and validate - that packing from Python’s `uuid.UUID` to ODBC’s `SQLGUID` always uses the correct byte order as required by the ODBC specification (little-endian for Data1, Data2, Data3; direct copy for Data4).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2957,9 +2988,15 @@ SQLRETURN FetchBatchData(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& colum
                     break;
                 }
                 case SQL_GUID: {
-                    row.append(
-                        py::bytes(reinterpret_cast<const char*>(&buffers.guidBuffers[col - 1][i]),
-                                  sizeof(SQLGUID)));
+                    SQLGUID* guidValue = &buffers.guidBuffers[col - 1][i];
+                    std::vector<char> guid_bytes(16);
+                    std::memcpy(guid_bytes.data(), guidValue, sizeof(SQLGUID));
```
sumitmsft: OK, so based on my understanding, we are copying the raw memory of the SQLGUID struct directly into a bytes array. 
Python’s uuid.UUID(bytes=...) expects the 16 bytes in big-endian order.
But, ODBC’s SQLGUID stores some fields (Data1, Data2, Data3) in little-endian order.

I guess the direct memcpy will only work if the memory layout matches Python’s expectation, which is NOT safe or portable. This means that one some platforms, the resulting UUID in Python could have the wrong byte order and not match the value in the database. Do you agree? 

If we reconstruct the UUID bytes explicitly, converting the little-endian fields to big-endian order before passing to Python would make sure we always have a right data with us. 

I guess you've done similar stuff in SQLGetData_wrap()

Please correct me if my understanding is different from your logic.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -504,7 +504,32 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
             case SQL_C_GUID: {
-                // TODO
+                py::object uuid_obj = param;
+                py::object uuid_cls = py::module_::import("uuid").attr("UUID");
+                if (!py::isinstance(uuid_obj, uuid_cls)) {
+                    ThrowStdException(MakeParamMismatchErrorStr(paramInfo.paramCType, paramIndex));
+                }
+                py::bytes uuid_bytes = uuid_obj.attr("bytes").cast<py::bytes>();
+                if (PyBytes_GET_SIZE(uuid_bytes.ptr()) != 16) {
+                    ThrowStdException("UUID binary data must be exactly 16 bytes long.");
```
sumitmsft, line 514: If the UUID is malformed (not 16 bytes), an exception is thrown but not logged, and details are not included in any log output. This may hinder debugging.
Can we Log the actual parameter and index before throwing, and ensure error messages do not leak sensitive content.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -504,7 +504,32 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
+                }
+                const unsigned char* b = reinterpret_cast<const unsigned char*>(PyBytes_AS_STRING(uuid_bytes.ptr()));
+                SQLGUID* guid_data_ptr = AllocateParamBuffer<SQLGUID>(paramBuffers);
```
sumitmsft, line 516: `AllocateParamBuffer<SQLGUID>` and `AllocateParamBuffer<SQLLEN>` (**below**) allocate memory for each UUID. Ensure these are released appropriately after use, especially in error paths. If not, this can lead to leaks.

tests/test_004_cursor.py
```diff
@@ -6942,6 +6943,103 @@ def test_money_smallmoney_invalid_values(cursor, db_connection):
         drop_table_if_exists(cursor, "dbo.money_test")
         db_connection.commit()
 
+def test_uuid_insert_and_select_none(cursor, db_connection):
```
sumitmsft, line 6947: Check for the following tests:
invalid UUID values (bad length, not a UUID instance, wrong type).
duplicate UUIDs (primary key violation).
extreme valid UUIDs (e.g., all zeros, all ones).

sumitmsft (changes requested):
Left a few comments

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: AllocateParamBuffer uses scope-managed buffers; all allocated memory is cleaned up appropriately when the parameter buffers go out of scope, so no leaks are expected.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Right.
Our current solution explicitly reconstructs the UUID bytes from SQLGUID, converting Data1, Data2, Data3 to big-endian order before passing to Python. This guarantees that the Python uuid.UUID object always matches the value stored in the database, regardless of platform or compiler endianness.

tests/test_004_cursor.py
gargsaumya: Added tests for:

1. Invalid UUID values (wrong type, bad length, non-UUID objects)
2. Duplicate UUIDs (primary key violation)
3. Extreme valid UUIDs (all zeros, all ones)

All tests pass, and the driver behaves correctly in each case.

bewithgaurav (approved):

PR #243, opened 2025-09-18
FEAT: Complex Data Type Support - DATETIMEOFFSET

gargsaumya:

> [AB#34944](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34944)

### Summary   
  
This pull request adds comprehensive support for the SQL Server `DATETIMEOFFSET` type in the driver, including full round-trip handling of timezone-aware Python `datetime` objects. The changes span the C++ bindings, Python type mapping, and test coverage to ensure correct reading and writing of `DATETIMEOFFSET` values, as well as validation for timezone information.

**DATETIMEOFFSET support in C++ bindings:**

* Defined new constants and a `DateTimeOffset` struct in `ddbc_bindings.cpp` to represent and handle the SQL Server `DATETIMEOFFSET` type and its C type mapping (`SQL_SS_TIMESTAMPOFFSET`, `SQL_C_SS_TIMESTAMPOFFSET`). [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L15-R22) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R97-R110)
* Implemented logic in parameter binding to convert Python timezone-aware `datetime` objects into the `DateTimeOffset` struct, including extracting and validating timezone offsets. Naive datetimes are rejected for this type.
* Added logic to fetch `DATETIMEOFFSET` values from the database and convert them back into Python `datetime` objects with the correct timezone information.

**Python type mapping and constants:**

* Added `SQL_DATETIMEOFFSET` and `SQL_C_SS_TIMESTAMPOFFSET` to the driver's constants in `constants.py`.
* Updated the cursor's type mapping logic to use `DATETIMEOFFSET` for timezone-aware `datetime` objects, and `TIMESTAMP` for naive ones.

**Testing improvements:**

* Introduced a new test, `test_datetimeoffset_read_write`, which verifies correct round-trip handling of `DATETIMEOFFSET` values, including various timezone offsets and microsecond precision, and ensures that naive datetimes are rejected.
* Added necessary imports for timezone handling in the test module.

These changes ensure that the driver now robustly supports SQL Server's `DATETIMEOFFSET` type, providing accurate and safe handling of timezone-aware datetimes between Python and the database.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -561,7 +561,33 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                 break;
             }
             case SQL_C_GUID: {
-                // TODO
+                if (!py::isinstance<py::bytes>(param)) {
+                    ThrowStdException(MakeParamMismatchErrorStr(paramInfo.paramCType, paramIndex));
+                }
+                py::bytes uuid_bytes = param.cast<py::bytes>();
+                const unsigned char* uuid_data = reinterpret_cast<const unsigned char*>(PyBytes_AS_STRING(uuid_bytes.ptr()));
+                if (PyBytes_GET_SIZE(uuid_bytes.ptr()) != 16) {
+                    LOG("Invalid UUID parameter at index {}: expected 16 bytes, got {} bytes, type {}", paramIndex, PyBytes_GET_SIZE(uuid_bytes.ptr()), paramInfo.paramCType);
+                    ThrowStdException("UUID binary data must be exactly 16 bytes long.");
+                }
+                SQLGUID* guid_data_ptr = AllocateParamBuffer<SQLGUID>(paramBuffers);
+                guid_data_ptr->Data1 =
+                    (static_cast<uint32_t>(uuid_data[3]) << 24) |
+                    (static_cast<uint32_t>(uuid_data[2]) << 16) |
+                    (static_cast<uint32_t>(uuid_data[1]) << 8)  |
+                    (static_cast<uint32_t>(uuid_data[0]));
+                guid_data_ptr->Data2 =
+                    (static_cast<uint16_t>(uuid_data[5]) << 8) |
+                    (static_cast<uint16_t>(uuid_data[4]));
+                guid_data_ptr->Data3 =
+                    (static_cast<uint16_t>(uuid_data[7]) << 8) |
+                    (static_cast<uint16_t>(uuid_data[6]));
+                std::memcpy(guid_data_ptr->Data4, &uuid_data[8], 8);
+                dataPtr = static_cast<void*>(guid_data_ptr);
+                bufferLength = sizeof(SQLGUID);
+                strLenOrIndPtr = AllocateParamBuffer<SQLLEN>(paramBuffers);
+                *strLenOrIndPtr = sizeof(SQLGUID);
+                break;
             }
             default: {
                 std::ostringstream errorString;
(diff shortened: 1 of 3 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -463,6 +477,39 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                 dataPtr = static_cast<void*>(sqlTimePtr);
                 break;
             }
+            case SQL_C_SS_TIMESTAMPOFFSET: {
```
sumitmsft, line 480: Can you confirm that tzinfo is not only present, but also that utcoffset() returns a valid value?

Also, pls check that the timezone offset falls within SQL Server’s allowed range (±14 hours). https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql?view=sql-server-ver17

What is there is a large/small offsets? Let's handle them accordingly.

Check for any overflow or underflow issues when converting offsets.

tests/test_004_cursor.py
```diff
@@ -7356,6 +7356,65 @@ def test_decimal_separator_calculations(cursor, db_connection):
         cursor.execute("DROP TABLE IF EXISTS #pytest_decimal_calc_test")
         db_connection.commit()
 
+def test_datetimeoffset_read_write(cursor, db_connection):
```
sumitmsft, line 7563: Pls check for:

- maximum and minimum allowed offset values.
- invalid offsets (e.g., offset > ±14 hours).
- leap seconds and DST transitions.
  - We have to make sure the driver does not crash or behave unexpectedly if given a leap second value.
  - Does not silently produce incorrect results or throw unexpected errors on DST
- string conversion edge cases (e.g., malformed input).

sumitmsft (changes requested):
Added a couple of comments

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: yes, we validate that tzinfo is present and utcoffset() returns a valid timedelta. Offsets are checked to fall within SQL Server’s ±14-hour range and anything outside this range raises an exception. 
Also, both positive and negative offsets are handled correctly. This is covered by test_datetimeoffset_invalid_offsets.

tests/test_004_cursor.py
gargsaumya: Added tests for maximum/minimum offsets, invalid offsets (>±14 hours), leap-second, DST transitions, and malformed strings. Verified that the driver handles all cases correctly.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2180,6 +2226,55 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                    LOG("[Fetch] Retrieved DTO: {}-{}-{} {}:{}:{}, fraction(ns)={}, tz_hour={}, tz_minute={}",
+                        dtoValue.year, dtoValue.month, dtoValue.day,
+                        dtoValue.hour, dtoValue.minute, dtoValue.second,
+                        dtoValue.fraction,
+                        dtoValue.timezone_hour, dtoValue.timezone_minute
+                    );
+
+                    int totalMinutes = dtoValue.timezone_hour * 60 + dtoValue.timezone_minute;
+                    // Validating offset
+                    if (totalMinutes < -24 * 60 || totalMinutes > 24 * 60) {
```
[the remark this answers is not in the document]
gargsaumya: SQL Server allows offsets from −14:00 to +14:00 (not −12:00)

tests/test_004_cursor.py
```diff
@@ -7355,6 +7356,66 @@ def test_decimal_separator_calculations(cursor, db_connection):
...
+                        tzinfo=timezone(timedelta(hours=-8)))
+            ),
+            (
+                "2023-10-28 20:00:05.9876543 +00:00",
+                datetime(2023, 10, 28, 20, 0, 5, 987654,
+                        tzinfo=timezone(timedelta(hours=0)))
+            ),
+            (
+                "invalid", # Placeholder for the SQL string
+                datetime(2023, 10, 29, 10, 0)
```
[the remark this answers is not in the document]
gargsaumya: Outdated file. This is already being tested in test_datetimeoffset_invalid_offsets.

sumitmsft (dismissed):

tests/test_004_cursor.py
```diff
@@ -7356,6 +7356,172 @@ def test_decimal_separator_calculations(cursor, db_connection):
...
+        cursor.execute("DROP TABLE IF EXISTS #dto_invalid;")
+        db_connection.commit()
+
+def test_datetimeoffset_dst_transitions(cursor, db_connection):
+    """
+    Test inserting and retrieving DATETIMEOFFSET values around DST transitions.
+    Ensures that driver handles DST correctly and does not crash.
+    """
+    try:
+        cursor.execute("IF OBJECT_ID('tempdb..#dto_dst', 'U') IS NOT NULL DROP TABLE #dto_dst;")
```
bewithgaurav: these are temp tables, so no need to drop them before starting the tests, can remove this line from all places

tests/test_004_cursor.py
```diff
@@ -7356,6 +7356,172 @@ def test_decimal_separator_calculations(cursor, db_connection):
...
+        cursor.execute("CREATE TABLE #dto_dst (id INT PRIMARY KEY, dto_column DATETIMEOFFSET);")
```
bewithgaurav: better to use formatting of test table names same as other ones, starting from "pytest_"

bewithgaurav (changes requested):
minor formatting changes under tests

PR #249, opened 2025-09-21
FIX: Correcting getinfo API

jahnvi480:

> [AB#34911](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34911)

### Summary   
This pull request refactors and improves the handling of ODBC `SQLGetInfo` metadata retrieval in the `mssql_python` driver. The main changes include a more robust and consistent method for returning info values from the C++ layer, enhanced decoding and type handling in Python, and better exposure of constants for users. These updates should make metadata access more reliable and easier to use, especially across different drivers and platforms.

**ODBC Info Retrieval Refactor**

* The C++ `getInfo` method in `connection.cpp` now always returns a dictionary containing the raw bytes, length, and info type, instead of attempting to interpret the result as a string or integer. This enables more consistent handling in Python and avoids issues with ambiguous types.
* The Python `getinfo` method in `connection.py` has been rewritten to robustly interpret the raw result dictionary, properly decode string and Y/N types, and handle various numeric formats. It includes improved error handling and logging for easier debugging.

**Constants and Metadata Exposure**

* `GetInfoConstants` is now imported and its members (such as `SQL_DRIVER_NAME`, `SQL_SERVER_NAME`, etc.) are exported at the module level in `__init__.py`. This makes it easier for users to reference standard info types. 
* A new `get_info_constants()` function is provided to return all available `GetInfoConstants` as a dictionary, simplifying programmatic access.
* Added missing constant `SQL_PROCEDURE_TERM` to the `GetInfoConstants` enum.

**Testing Improvements**

* Added debug print statements in `test_getinfo_standard_types` to help diagnose info type values during test runs.

These changes collectively improve the reliability, usability, and maintainability of metadata access in the driver.

### DIFF
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -874,8 +874,22 @@ class Connection:
                 ddbc_error="Cannot get info on closed connection",
             )
         
+        # Check that info_type is an integer
+        if not isinstance(info_type, int):
+            raise ValueError(f"info_type must be an integer, got {type(info_type).__name__}")
+        
+        # Check for invalid info_type values
+        if info_type < 0:
+            log('warning', f"Invalid info_type: {info_type}. Must be a positive integer.")
+            return None
+        
         # Get the raw result from the C++ layer
-        raw_result = self._conn.get_info(info_type)
+        try:
+            raw_result = self._conn.get_info(info_type)
+        except Exception as e:
+            # Log the error and return None for invalid info types
+            log('warning', f"getinfo({info_type}) failed: {e}")
+            return None
         
         if raw_result is None:
             return None
@@ -924,10 +938,28 @@ class Connection:
                 GetInfoConstants.SQL_NEED_LONG_DATA_LEN.value,
                 GetInfoConstants.SQL_PROCEDURES.value
             }
-
+    
+            # Numeric type constants that return integers
+            numeric_type_constants = {
+                GetInfoConstants.SQL_MAX_COLUMN_NAME_LEN.value,
+                GetInfoConstants.SQL_MAX_TABLE_NAME_LEN.value,
+                GetInfoConstants.SQL_MAX_SCHEMA_NAME_LEN.value,
+                GetInfoConstants.SQL_MAX_CATALOG_NAME_LEN.value,
+                GetInfoConstants.SQL_MAX_IDENTIFIER_LEN.value,
+                GetInfoConstants.SQL_MAX_STATEMENT_LEN.value,
+                GetInfoConstants.SQL_MAX_DRIVER_CONNECTIONS.value,
+                GetInfoConstants.SQL_NUMERIC_FUNCTIONS.value,
+                GetInfoConstants.SQL_STRING_FUNCTIONS.value,
+                GetInfoConstants.SQL_DATETIME_FUNCTIONS.value,
+                GetInfoConstants.SQL_TXN_CAPABLE.value,
+                GetInfoConstants.SQL_DEFAULT_TXN_ISOLATION.value,
+                GetInfoConstants.SQL_CURSOR_COMMIT_BEHAVIOR.value
+            }
+    
             # Determine the type of information we're dealing with
             is_string_type = info_type > INFO_TYPE_STRING_THRESHOLD or info_type in string_type_constants
             is_yn_type = info_type in yn_type_constants
+            is_numeric_type = info_type in numeric_type_constants
             
             # Process the data based on type
             if is_string_type:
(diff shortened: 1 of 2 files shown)


mssql_python/pybind/connection/connection.cpp
```diff
@@ -322,183 +322,51 @@ py::object Connection::getInfo(SQLUSMALLINT infoType) const {
...
+    }
     
-    // First try to get the info - handle SQLSMALLINT size limit
-    SQLSMALLINT bufferSize = (buffer.size() <= SQL_MAX_SMALL_INT) 
-        ? static_cast<SQLSMALLINT>(buffer.size()) 
-        : SQL_MAX_SMALL_INT;
-        
-    ret = SQLGetInfo_ptr(_dbcHandle->get(), infoType, buffer.data(), bufferSize, &actualLength);
+    // Allocate buffer with extra space
+    std::vector<char> buffer(requiredLen + 10, 0);  // Extra padding for safety
```
sumitmsft: Currently, the buffer for `SQLGetInfo` is allocated based directly on the value reported by ODBC (`requiredLen + 10`). 
If ODBC returns an unexpectedly large number (due to a bug or malicious driver), this can cause excessive memory usage and even out-of-memory crashes, even though the size passed to `SQLGetInfo` is capped. 

Please cap the buffer allocation itself to a reasonable maximum (such as SQL_MAX_SMALL_INT) to prevent this risk.

mssql_python/connection.py
```diff
@@ -872,7 +872,137 @@ def getinfo(self, info_type):
...
+                            return data[:length]
+                        except Exception:
+                            # Default fallback
+                            return data[:length]
+                else:
+                    # Return the data as-is if not bytes
+                    return data
+    
+        # If we get here, the result is in an unexpected format
+        log('warning', f"Unexpected result format from getInfo: {type(raw_result)}")
```
sumitmsft: Logging statements may log raw data from the server. If any server metadata is sensitive, this could leak information. Can you please redact the sensitive data before logging?

mssql_python/connection.py
```diff
@@ -872,7 +872,137 @@ def getinfo(self, info_type):
...
+                        return actual_data.decode('utf-8').rstrip('\0')
+                    except UnicodeDecodeError:
+                        # Fall back to a different encoding if UTF-8 fails
+                        try:
+                            return actual_data.decode('latin1').rstrip('\0')
+                        except Exception as e:
+                            log('warning', f"Failed to decode string in getinfo: {e}")
+                            # As last resort, strip nulls and convert remaining bytes
+                            cleaned_data = actual_data.replace(b'\x00', b'')
+                            return cleaned_data.decode('latin1', errors='replace')
```
sumitmsft: This code attempts to decode binary data using UTF-8, then Latin-1, and finally Latin-1 with `errors='replace'`. If all decoding attempts fail, the result may be a string with corrupted or misleading content (replacement characters). This can silently hide problems if the data is not actually valid text.

To avoid silent data corruption and make errors more visible, I suggest logging a clear error when decoding fails and either raising an exception or returning a more explicit error value (such as `None` or a custom error object), rather than a potentially invalid string. This will help downstream code handle such cases correctly and make debugging easier.

tests/test_003_connection.py
```diff
@@ -4744,6 +4744,7 @@ def test_timeout_affects_all_cursors(db_connection):
     finally:
         # Reset timeout
         db_connection.timeout = original_timeout
+
```
sumitmsft, line 4747: Can you add a few test cases:

- Pass binary data to getinfo that is not valid UTF-8 or Latin-1 and verify correct error handling/logging, not silent corruption.
- Can we simulate ODBC returning requiredLen = 0 and check that the returned value is an empty string, None, or handled as expected and no crash occurs?
- Pass data to getinfo that is neither bytes, string, int, nor bool (e.g., a list, a custom object). Ensure the function logs a warning and returns the value in a safe way.
- Pass bytes representing both 'Y', 'N', and other values, and ensure only 'Y'/'y'/1 is treated as 'Y'. Everything else is 'N'.
- Pass numeric bytes of different lengths (1, 2, 4, 8) and verify correct conversion to int. check robust fallback for non-numeric bytes.

mssql_python/connection.py
```diff
@@ -872,7 +872,137 @@ def getinfo(self, info_type):
...
+                    if byte_val in (b'Y'[0], b'y'[0], 1):
+                        return 'Y'
+                    else:
+                        return 'N'
+                else:
+                    # If it's not a byte or we can't determine, default to 'N'
+                    return 'Y' if data else 'N'
+            else:
+                # Handle numeric types based on length
+                if isinstance(data, bytes):
```
bewithgaurav: can we refactor things in this else block? we should use a bit better variable names and some methods..  something suggested by copilot
```
# Map byte length → signed int size
    int_sizes = {
        1: lambda d: d[0],
        2: lambda d: int.from_bytes(d[:2], "little", signed=True),
        4: lambda d: int.from_bytes(d[:4], "little", signed=True),
        8: lambda d: int.from_bytes(d[:8], "little", signed=True),
    }

    # Direct numeric conversion if supported length
    if length in int_sizes:
        return int_sizes[length](data)

    # Helper: check if all chars are digits
    def is_digit_bytes(b: bytes) -> bool:
        return all(c in b"0123456789" for c in b)

    # Helper: check if bytes are ASCII-printable or NUL padded
    def is_printable_bytes(b: bytes) -> bool:
        return all(32 <= c <= 126 or c == 0 for c in b)

    chunk = data[:length]

    # Try interpret as integer string
    if is_digit_bytes(chunk):
        return int(chunk)

    # Try decode as ASCII/UTF-8 string
    if is_printable_bytes(chunk):
        str_val = chunk.decode("utf-8", errors="replace").rstrip("\0")
        return int(str_val) if str_val.isdigit() else str_val

    # Fallback: raw bytes
    return chunk
```

mssql_python/connection.py
```diff
@@ -872,7 +872,137 @@ def getinfo(self, info_type):
...
+                GetInfoConstants.SQL_DATA_SOURCE_READ_ONLY.value,
+                GetInfoConstants.SQL_EXPRESSIONS_IN_ORDERBY.value,
+                GetInfoConstants.SQL_LIKE_ESCAPE_CLAUSE.value,
+                GetInfoConstants.SQL_MULTIPLE_ACTIVE_TXN.value,
+                GetInfoConstants.SQL_NEED_LONG_DATA_LEN.value,
+                GetInfoConstants.SQL_PROCEDURES.value
+            }
+            
+            # Determine the type of information we're dealing with
+            is_string_type = info_type > 10000 or info_type in string_type_constants
```
bewithgaurav: can we add 10000 as a constant? looks like an imp threshold

bewithgaurav (changes requested):
some refactoring needed

sumitmsft (approved):

gargsaumya (approved):

PR #252, opened 2025-09-24
FIX: Fixed and Added tests for Type Inference Bug

gargsaumya:

> AB#<WORK_ITEM_ID>

> GitHub Issue: #253

### Summary   
  
This pull request removes the `_select_best_sample_value` static method and updates how sample values are selected for type inference in the `executemany` method. Instead of using the old method, the code now relies on `_compute_column_type` to determine sample values and type information.

**Refactoring and Type Inference Updates:**

* Removed the `_select_best_sample_value` static method from `cursor.py`, consolidating the logic for type inference.
* Updated the `executemany` method to use `_compute_column_type` for determining sample values, minimum, and maximum values for each column, improving clarity and maintainability.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -2013,6 +2013,49 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
                     bufferLength = sizeof(SQL_NUMERIC_STRUCT);
                     break;
                 }
+                case SQL_C_GUID: {
+                    SQLGUID* guidArray = AllocateParamBufferArray<SQLGUID>(tempBuffers, paramSetSize);
+                    strLenOrIndArray = AllocateParamBufferArray<SQLLEN>(tempBuffers, paramSetSize);
+
+                    static py::module_ uuid_mod = py::module_::import("uuid");
+                    static py::object uuid_class = uuid_mod.attr("UUID");
+                    for (size_t i = 0; i < paramSetSize; ++i) {
+                        const py::handle& element = columnValues[i];
+                        std::array<unsigned char, 16> uuid_bytes;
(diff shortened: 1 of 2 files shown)


bewithgaurav:
merging main here to test and check code coverage

mssql_python/cursor.py
```diff
@@ -1643,11 +1614,7 @@ def executemany(self, operation: str, seq_of_parameters: list) -> None:
             else:
                 # Use auto-detection for columns without explicit types
                 column = [row[col_index] for row in seq_of_parameters] if hasattr(seq_of_parameters, '__getitem__') else []
-                if not column:
-                    # For generators, use the sample row for inference
-                    sample_value = sample_row[col_index]
-                else:
-                    sample_value = self._select_best_sample_value(column)
+                sample_value, min_val, max_val = self._compute_column_type(column)
```
sumitmsft, line 1615: Can you add\validate the test cases for type inference and batch execution:

- Mixed positive/negative ints (e.g., [1, -5, 3])
- Very large/small ints (test 32/64-bit edges)
- Mix of bools and ints (e.g., [True, False, 2])
- Ints with None values
- Strings of different lengths (e.g., ["a", "abcd", "abc"])
- Bytes values (e.g., [b"a", b"abcdef"])
- All None values (e.g., [None, None])
- Empty parameter list to executemany

sumitmsft: Also add: In the list the first non-empty value which doesn't have a length (like a decimal value), but a later value does like a string.

sumitmsft (changes requested):
Left a few test cases to add

mssql_python/cursor.py
gargsaumya: Done

bewithgaurav (approved):

Issue #254, opened 2025-09-24
Python segfault during shutdown after exceptions


bewithgaurav:
### Describe the bug
Python segmentation fault occurs during interpreter shutdown when failed cursor operations are encountered. The crash happens when C++ destructors in the native extension try to call ODBC driver functions and Python logging APIs after Python has begun shutting down, causing invalid memory access and segfaults.

**Exception details:**
```
EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: KERN_INVALID_ADDRESS at 0x0000000000000010
Termination Reason: Namespace SIGNAL, Code 11 Segmentation fault: 11

Key stack trace elements:
...
2   ddbc_bindings.cp313-universal2.so    SqlHandle::free() + 232
...
13  Python                               gc_collect_main + 2272
14  Python                               finalize_modules + 3048
15  Python                               _Py_Finalize + 936
```

### To reproduce
Include a complete code listing (or project/solution) that we can run to reproduce the issue.

```python

cursor = conn.cursor()
cursor.execute("syntax error") #Segfault happens here itself

row = cursor.fetchall()
print(row)
```

### Expected behavior
The script should execute, catch the failed cursor operation, clean up resources properly, and exit without any segmentation fault. Python should shut down cleanly even after cursor operation failures occur.

### Further technical details
Python version: 3.13  
SQL Server version: SQL Server (any version with ODBC driver 18)  
Operating system: macOS 14.7.1 (also reproducible on Linux)

**Additional context**

**Root Cause Analysis:**
The segmentation fault occurs during Python interpreter shutdown due to improper cleanup ordering:

1. **Python Shutdown Ordering**: During `Py_Finalize()`, Python's garbage collector runs and finalizes C++ objects
2. **Invalid Driver State**: The ODBC driver library may be unloaded or become invalid during shutdown
3. **Python API Corruption**: `sys.meta_path` becomes `None` during finalization, breaking import mechanisms
4. **Destructor Calls**: C++ destructors (`SqlHandle::free()` and `LOG()` function) try to call Python APIs and ODBC functions after Python has begun shutting down

**Technical Details:**
- Crash occurs in `SqlHandle::free()` when calling `SQLFreeHandle_ptr()`
- Also affects `LOG()` function which tries to import Python modules during shutdown
- The issue manifests as `EXC_BAD_ACCESS` at memory address `0x0000000000000010`
[...]

PR #255, opened 2025-09-24
FIX: Resource cleanup on Python shutdown to avoid segfaults

bewithgaurav:

> [AB#38824](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38824)

> GitHub Issue: #254 

### Summary   
  
This pull request addresses a critical segfault issue that occurred during Python interpreter shutdown when using the `mssql_python` package, especially in scenarios with unclosed cursors, syntax errors, or concurrent operations. The changes introduce robust detection of Python's shutdown state to prevent unsafe resource cleanup and improve test coverage for these edge cases.

Key changes include:

**Python Shutdown Safety & Resource Cleanup**
* Added a centralized helper function `is_python_finalizing()` in `ddbc_bindings.cpp` to reliably detect when Python is shutting down or finalizing, using both `Py_IsInitialized()` and the `sys._is_finalizing` attribute for compatibility across Python versions.
* Updated the `LOG` function to skip logging if Python is shutting down, and to handle exceptions gracefully to avoid crashes during interpreter cleanup. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R665-R700) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R717-R725)
* Modified the `SqlHandle::free()` method to prevent freeing `SQL_HANDLE_STMT` handles during shutdown, as their parent DBC handles may already be freed, avoiding double-free and segfaults. Handles are marked as freed instead.

**Test Coverage for Shutdown and Concurrency Edge Cases**
* Added multiple new tests in `test_005_connection_cursor_lifecycle.py` to ensure no segfaults occur during Python shutdown, including:
  - Tests for syntax errors without explicit resource cleanup.
  - Tests for multiple concurrent cursor operations and abrupt interpreter exits.
  - Tests for closing connections with pending results and aggressive threading scenarios.

These changes significantly improve the stability of the `mssql_python` package in complex and error-prone scenarios, especially during interpreter shutdown and multithreaded usage.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -665,21 +665,43 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
...
+        }
+        
+        py::gil_scoped_acquire gil;  // <---- this ensures safe Python API usage
+
+        // Check if sys module is available and not finalizing
+        py::object sys_module = py::module_::import("sys");
+        if (!sys_module.is_none()) {
+            py::object finalizing_func = sys_module.attr("_is_finalizing");
+            if (!finalizing_func.is_none() && finalizing_func().cast<bool>()) {
+                return; // Python is finalizing, don't log
```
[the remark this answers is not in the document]
bewithgaurav: we support python 10+, not a valid comment

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -662,24 +662,66 @@ void HandleZeroColumnSizeAtFetch(SQLULEN& columnSize) {
...
+            // Check if the attribute exists before accessing it (for Python version compatibility)
+            if (py::hasattr(sys_module, "_is_finalizing")) {
+                py::object finalizing_func = sys_module.attr("_is_finalizing");
+                if (!finalizing_func.is_none() && finalizing_func().cast<bool>()) {
+                    return true; // Python is finalizing
+                }
+            }
+        }
+        return false;
+    } catch (...) {
```
sumitmsft, line 688: Can we consider logging or at least debugging such exceptions if they become frequent (possibly with a debug flag)?

bewithgaurav: Actually, its not recommended to put any loggers on python shutdown, so we can discard putting our standard loggers here.
This catch is in case python doesn't get what is `sys` module etc., so this is absolutely unexpected but viable during shutdown scenarios. This catch further prevents a core dump. 
But this raises a good point, we or the user does need to know if we're getting unexpected exceptions during shutdown.
I guess we can put an std::cerr here which will do an stdout instead of logging, will make a small change here.

gargsaumya: Yes, that makes sense. Using std::cerr here is a good compromise, it avoids unsafe logging during Python shutdown while still surfacing unexpected exceptions for debugging.

sumitmsft (approved):
LGTM. Approved

PR #260, opened 2025-09-25
FEAT: datetimeoffset support in executemany()

gargsaumya:

> [AB#34944](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/34944)

### Summary   
  

This pull request adds comprehensive support for the SQL Server `DATETIMEOFFSET` type in both the Python and C++ layers of the codebase. It introduces new constants, enhances the datetime parsing logic to handle timezone-aware datetimes, and updates the parameter binding and data fetching logic to correctly handle `DATETIMEOFFSET` values. The changes ensure accurate round-trip of timezone-aware datetime values between Python and SQL Server.

**Support for DATETIMEOFFSET type:**

* Added `SQL_DATETIMEOFFSET` and `SQL_C_SS_TIMESTAMPOFFSET` constants to both `mssql_python/constants.py` and the C++ bindings (`ddbc_bindings.cpp`). [[1]](diffhunk://#diff-e6d80f1000af6fd5afca05f435b11fd82df7f5c3e75ecf5763f85d3aacdbe758R127-R128) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L15-R22)
* Introduced a new `DateTimeOffset` C++ struct and integrated it into buffer management for column data. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R67-R80) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R95) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R109)
* Updated the C++ parameter binding logic to correctly marshal Python timezone-aware `datetime` objects as `DATETIMEOFFSET`, including array binding for `executemany`. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R482-R514) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1881) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1904-R1964)
* Implemented fetching and conversion logic for `DATETIMEOFFSET` columns, ensuring Python receives timezone-aware `datetime` objects in UTC. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R2598-R2647) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R2904-R2910) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3126-R3162)

**Datetime parsing and mapping improvements:**

* Enhanced `_parse_datetime` in `mssql_python/cursor.py` to support parsing timezone-aware datetime strings, improving compatibility with `DATETIMEOFFSET`.
* Updated SQL type mapping logic to distinguish between naive and timezone-aware `datetime` objects, mapping the latter to `DATETIMEOFFSET`.

**Refactoring and cleanup:**

* Removed the now-unused `_select_best_sample_value` static method from `mssql_python/cursor.py` as type inference is now handled differently.
* Adjusted type inference during `executemany` to use the new `_compute_column_type` method, aligning with the improved datetime handling.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1830,6 +1878,7 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
                     break;
                 }
                 case SQL_C_TYPE_TIMESTAMP: {
+                    std::cout<<"Binding Timestamp param at index "<<paramIndex<<std::endl;
```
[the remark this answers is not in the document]
bewithgaurav: @gargsaumya - there are debug statements left in the code, could you please remove them?

mssql_python/cursor.py
```diff
@@ -140,27 +140,26 @@ def _parse_date(self, param):
...
-            A datetime.datetime object if parsing is successful, else None.
+        Attempt to parse a string as a datetime, datetime2, timestamp, or smalldatetime.
+        Handles both naive and timezone-aware datetime strings.
         """
         formats = [
-            "%Y-%m-%dT%H:%M:%S.%f",  # ISO 8601 datetime with fractional seconds
-            "%Y-%m-%dT%H:%M:%S",  # ISO 8601 datetime
-            "%Y-%m-%d %H:%M:%S.%f",  # Datetime with fractional seconds
-            "%Y-%m-%d %H:%M:%S",  # Datetime without fractional seconds
+            "%Y-%m-%dT%H:%M:%S.%f%z",  # ISO 8601 with fractional seconds + timezone
```
bewithgaurav: we'll probably be removing the `_parse_date/time/datetime` methods FYI since they're not needed and cause some issues - please check #235 for more context, the changes here can be undone in that case

gargsaumya: yeah no issues, they are not needed even here. Need to do a cleanup of this PR.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1934,6 +1901,67 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
...
+                                ThrowStdException(MakeParamMismatchErrorStr(info.paramCType, paramIndex));
+                            }
+
+                            py::object tzinfo = param.attr("tzinfo");
+                            if (tzinfo.is_none()) {
+                                ThrowStdException("Datetime object must have tzinfo for SQL_C_SS_TIMESTAMPOFFSET at paramIndex " +
+                                    std::to_string(paramIndex));
+                            }
+
+                            // Convert the Python datetime object to UTC before binding.
```
sumitmsft: In `case SQL_C_SS_TIMESTAMPOFFSET` block, the single-row path in `execute()` preserves the original datetime’s offset, while `executemany()` path converts the datetime to UTC and forces the offset fields to 0. Any specific reason? 

Can we choose this uniformly and preserve original offsets for both paths?

sumitmsft: Add the tests accordingly

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -499,17 +501,7 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                 dtoPtr->fraction = static_cast<SQLUINTEGER>(param.attr("microsecond").cast<int>() * 1000);
```
sumitmsft, line 502: Here we are setting `fraction = microsecond * 1000` wherein we are converting Python microseconds to nanoseconds. That preserves only microsecond precision (last three digits of the nanosecond field will always be 000).

However, our SQL type is with scale=7 in `_map_sql_type` for timezone-aware datetimes, which implies 100‑nanosecond resolution because SQL Server DATETIMEOFFSET supports up to 7 fractional digits = 100ns ticks. 

Isn't this mean we are signaling max precision while only delivering microsecond precision? I guess this will cause data truncation in some cases.

tests/test_004_cursor.py
```diff
@@ -7661,166 +7357,119 @@ def test_decimal_separator_calculations(cursor, db_connection):
         db_connection.commit()
 
 def test_datetimeoffset_read_write(cursor, db_connection):
```
sumitmsft, line 7663: Can use following test cases - (got from gpt):

Mixed non-NULL and None in executemany to ensure NULL indicators handled; non-null values round-trip correctly.
Unusual quarter/half-hour offsets (+05:45, +12:45, -03:30, +08:45) to preserve minute portion, no truncation.
7-digit fractional literal (...1234567+00:00) to lock current truncation/rounding policy (microsecond granularity).
NULL as last row of a fetch batch to ensure indicator handling doesn’t reuse previous struct data.
Same UTC instant, different offsets (10:00+02:00 vs 08:00+00:00) to detect policy drift (preserve vs normalize).

tests/test_003_connection.py
```diff
@@ -3099,7 +3099,6 @@ def test_execute_with_large_parameters(db_connection):
     - Working with parameters near but under the size limit
     - Processing large result sets
     """
-    import time
```
sumitmsft, line 3102: Can you add the following tests - got from gpt:

execute() vs executemany consistency (same aware datetime) to check stored textual value must match once binding semantics unified.
Extreme legal offsets (+14:00, -12:00) to check boundary offset handling & round-trip consistency.

tests/test_004_cursor.py
sumitmsft: Max fractional boundary (23:59:59.9999999+00:00) to check no overflow; correct truncation.

sumitmsft (changes requested):
Left a few comments

bewithgaurav (approved):

sumitmsft (approved):

PR #264, opened 2025-09-29
FIX: Memory Leak and Security Risk with Static Token Buffer

gargsaumya:

> [AB#37606](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/37606)

### Summary   
  

This pull request improves the handling of sensitive data when setting SQL connection attributes in `connection.cpp`. The main change is replacing a static buffer with a stack-allocated buffer to better control memory and securely erase sensitive data after use.

Sensitive data handling:

* Replaced the static vector of buffers with a stack-allocated `std::string` (`buffer`) to temporarily hold sensitive data when setting SQL attributes. This avoids unnecessary retention of sensitive data in memory.
* Added logic to zero out the contents of `buffer` after it is used, ensuring sensitive data does not remain in memory.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -1613,6 +1613,7 @@ class Cursor:
                 # Use auto-detection for columns without explicit types
                 column = [row[col_index] for row in seq_of_parameters] if hasattr(seq_of_parameters, '__getitem__') else []
                 sample_value, min_val, max_val = self._compute_column_type(column)
+
                 dummy_row = list(sample_row)
                 paraminfo = self._create_parameter_types_list(
                     sample_value, param_info, dummy_row, col_index, min_val=min_val, max_val=max_val
--- a/tests/test_004_cursor.py
+++ b/tests/test_004_cursor.py
@@ -1262,6 +1262,118 @@ def test_executemany_binary_data_edge_cases(cursor, db_connection):
         cursor.execute("DROP TABLE IF EXISTS #pytest_binary_test")
         db_connection.commit()
 
+def test_executemany_mixed_ints(cursor, db_connection):
+    """Test executemany with mixed positive and negative integers."""
+    try:
+        cursor.execute("CREATE TABLE #pytest_mixed_ints (val INT)")
+        data = [(1,), (-5,), (3,)]
+        cursor.executemany("INSERT INTO #pytest_mixed_ints VALUES (?)", data)
+        db_connection.commit()
+
+        cursor.execute("SELECT val FROM #pytest_mixed_ints ORDER BY val")
+        results = [row[0] for row in cursor.fetchall()]
+        assert sorted(results) == [-5, 1, 3]
+    finally:
+        cursor.execute("DROP TABLE IF EXISTS #pytest_mixed_ints")
+        db_connection.commit()
+
+
+def test_executemany_int_edge_cases(cursor, db_connection):
+    """Test executemany with very large and very small integers."""
+    try:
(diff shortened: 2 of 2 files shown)


bewithgaurav (changes requested):
@gargsaumya - as mentioned in the code coverage comment and the full report as well, I believe we haven't covered this whole function inside connection.cpp - `Connection::setAttribute(SQLINTEGER attribute, py::object value)`
It would be a bit more helpful for the coverage if we can include some tests inside test_connection.py. could you please include a couple of tests with this PR?
If its already covered, and code coverage is not detecting please let me know.

gargsaumya:
> @gargsaumya - as mentioned in the code coverage comment and the full report as well, I believe we haven't covered this whole function inside connection.cpp - `Connection::setAttribute(SQLINTEGER attribute, py::object value)` It would be a bit more helpful for the coverage if we can include some tests inside test_connection.py. could you please include a couple of tests with this PR? If its already covered, and code coverage is not detecting please let me know.

As we discussed offline, since this is an internal function, we don’t add unit tests for it. The coverage gap related to this function can be taken up in a separate PR to keep the scope of this one focused.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -173,16 +173,16 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
...
     SQLINTEGER length = 0;
+    std::string buffer; // to hold sensitive data temporarily
 
     if (py::isinstance<py::int_>(value)) {
         int intValue = value.cast<int>();
         ptr = reinterpret_cast<SQLPOINTER>(static_cast<uintptr_t>(intValue));
         length = SQL_IS_INTEGER;
     } else if (py::isinstance<py::bytes>(value) || py::isinstance<py::bytearray>(value)) {
-        static std::vector<std::string> buffers;
-        buffers.emplace_back(value.cast<std::string>());
```
subrata-ms, line 183: This line supports multiple string inside a vector. But new change is only for one string. Did we validate the impact for all scenario?

mssql_python/pybind/connection/connection.cpp
```diff
@@ -173,16 +173,16 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
...
     SQLPOINTER ptr = nullptr;
     SQLINTEGER length = 0;
+    std::string buffer; // to hold sensitive data temporarily
 
     if (py::isinstance<py::int_>(value)) {
         int intValue = value.cast<int>();
         ptr = reinterpret_cast<SQLPOINTER>(static_cast<uintptr_t>(intValue));
         length = SQL_IS_INTEGER;
     } else if (py::isinstance<py::bytes>(value) || py::isinstance<py::bytearray>(value)) {
-        static std::vector<std::string> buffers;
```
subrata-ms, line 182: removing Static keyword and defining the vector outside the loop may solve the issue??

mssql_python/pybind/connection/connection.cpp
```diff
@@ -195,6 +195,11 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
     else {
         LOG("Set attribute successfully");
     }
+    
+    // Zero out sensitive data if used
+    if (!buffer.empty()) {
+        std::fill(buffer.begin(), buffer.end(), static_cast<char>(0));
```
subrata-ms, line 201: why fill?   buffer is not allocated via "new". buffer is also a local variable.

Issue #250, opened 2025-09-23
cursor.fetchall() cannot parse not utf-8 character


KleinPan:
e = UnicodeDecodeError('utf-8', b'\xd0\xb4\xba\xc5PCBA-SN', 2, 3, 'invalid start byte')

conn.setdecoding(mssql_python.SQL_CHAR, encoding='gbk')
conn.setdecoding(mssql_python.SQL_WCHAR, encoding='gbk')  not work!

jahnvi480:
@KleinPan This issue has been fixed and will be part of next release

PR #265, opened 2025-09-30
FIX: Encoding Decoding

jahnvi480:

> AB#<WORK_ITEM_ID>

> GitHub Issue: #250 

### Summary   
This pull request improves the handling of character encoding and decoding in the `mssql_python/cursor.py` module. The main changes ensure that encoding and decoding settings are dynamically retrieved from the connection, allowing for more robust and flexible support for different character sets when executing queries and fetching results.

**Encoding and decoding settings improvements:**

* Added `_get_encoding_settings` and `_get_decoding_settings` helper methods to retrieve encoding and decoding configurations from the connection, with sensible fallbacks if unavailable.

**Query execution enhancements:**

* Updated the `execute` and `executemany` methods to use dynamic encoding and character type settings when calling the underlying DDBC bindings, ensuring queries are sent with the correct encoding.
**Result fetching improvements:**

* Modified `fetchone`, `fetchmany`, and `fetchall` methods to use dynamic decoding settings for character and wide character data, improving reliability and compatibility when reading results from the database.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -104,6 +104,49 @@ class Cursor:
 
         self.messages = []  # Store diagnostic messages
 
+    def _get_encoding_settings(self):
+        """
+        Get the encoding settings from the connection.
+        
+        Returns:
+            dict: A dictionary with 'encoding' and 'ctype' keys, or default settings if not available
+        """
+        if hasattr(self._connection, 'getencoding'):
+            try:
+                return self._connection.getencoding()
+            except:
+                # Return default encoding settings if there's an error
+                return {
+                    'encoding': 'utf-16le', 
+                    'ctype': ddbc_sql_const.SQL_WCHAR.value
+                }
+        # Return default encoding settings if getencoding is not available
+        return {
+            'encoding': 'utf-16le',
+            'ctype': ddbc_sql_const.SQL_WCHAR.value
+        }
+
+    def _get_decoding_settings(self, sql_type):
+        """
+        Get decoding settings for a specific SQL type.
+        
+        Args:
+            sql_type: SQL type constant (SQL_CHAR, SQL_WCHAR, etc.)
+            
+        Returns:
+            Dictionary containing the decoding settings.
+        """
+        try:
+            # Get decoding settings from connection for this SQL type
+            return self._connection.getdecoding(sql_type)
+        except Exception as e:
+            # If anything goes wrong, return default settings
+            log('warning', f"Failed to get decoding settings for SQL type {sql_type}: {e}")
+            if sql_type == SQL_WCHAR:
+                return {'encoding': 'utf-16le', 'ctype': SQL_WCHAR}
+            else:
+                return {'encoding': 'utf-8', 'ctype': SQL_CHAR}
+
     def _is_unicode_string(self, param):
         """
         Check if a string contains non-ASCII characters.
@@ -979,6 +1022,8 @@ class Cursor:
                     parameters_type[i].decimalDigits,
                     parameters_type[i].inputOutputType,
                 )
+        
+        encoding_settings = self._get_encoding_settings()
 
         ret = ddbc_bindings.DDBCSQLExecute(
             self.hstmt,
@@ -987,6 +1032,8 @@ class Cursor:
             parameters_type,
             self.is_stmt_prepared,
             use_prepare,
+            encoding_settings.get('encoding'),
+            encoding_settings.get('ctype')
         )
         # Check return code
         try:
@@ -1678,12 +1725,16 @@ class Cursor:
             len(seq_of_parameters), "\n".join(f"  {i+1}: {tuple(p) if isinstance(p, (list, tuple)) else p}" for i, p in enumerate(seq_of_parameters[:5]))  # Limit to first 5 rows for large batches
         )
 
+        encoding_settings = self._get_encoding_settings()
+
         ret = ddbc_bindings.SQLExecuteMany(
             self.hstmt,
             operation,
             columnwise_params,
             parameters_type,
-            row_count
+            row_count,
+            encoding_settings.get('encoding'),
+            encoding_settings.get('ctype')
         )
         
         # Capture any diagnostic messages after execution
@@ -1715,10 +1766,14 @@ class Cursor:
         """
         self._check_closed()  # Check if the cursor is closed
 
+        # Get decoding settings for character data
+        char_decoding = self._get_decoding_settings(ddbc_sql_const.SQL_CHAR.value)
+        wchar_decoding = self._get_decoding_settings(ddbc_sql_const.SQL_WCHAR.value)
+
         # Fetch raw data
         row_data = []
         try:
-            ret = ddbc_bindings.DDBCSQLFetchOne(self.hstmt, row_data)
+            ret = ddbc_bindings.DDBCSQLFetchOne(self.hstmt, row_data, char_decoding.get('encoding', 'utf-8'), wchar_decoding.get('encoding', 'utf-16le'))
             
             if self.hstmt:
                 self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
@@ -1759,11 +1814,16 @@ class Cursor:
 
         if size <= 0:
             return []
+
+        # Get decoding settings for character data
+        char_decoding = self._get_decoding_settings(ddbc_sql_const.SQL_CHAR.value)
+        wchar_decoding = self._get_decoding_settings(ddbc_sql_const.SQL_WCHAR.value)
         
         # Fetch raw data
         rows_data = []
         try:
-            ret = ddbc_bindings.DDBCSQLFetchMany(self.hstmt, rows_data, size)
+            ret = ddbc_bindings.DDBCSQLFetchMany(self.hstmt, rows_data, size, char_decoding.get('encoding', 'utf-8'), wchar_decoding.get('encoding', 'utf-16le'))
+
 
             if self.hstmt:
                 self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
@@ -1793,10 +1853,14 @@ class Cursor:
         if not self._has_result_set and self.description:
             self._reset_rownumber()
 
+        # Get decoding settings for character data
+        char_decoding = self._get_decoding_settings(ddbc_sql_const.SQL_CHAR.value)
+        wchar_decoding = self._get_decoding_settings(ddbc_sql_const.SQL_WCHAR.value)
+
         # Fetch raw data
         rows_data = []
         try:
-            ret = ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
+            ret = ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data, char_decoding.get('encoding', 'utf-8'), wchar_decoding.get('encoding', 'utf-16le'))
 
             if self.hstmt:
                 self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -167,6 +167,151 @@ SQLTablesFunc SQLTables_ptr = nullptr;
 
 SQLDescribeParamFunc SQLDescribeParam_ptr = nullptr;
 
+py::object DecodeString(const void* data, SQLLEN dataLen, const std::string& encoding, bool isWideChar) {
+    if (data == nullptr || dataLen <= 0) {
+        return py::none();
+    }
+    
+    // Create a bytes object with the raw binary data
+    py::bytes bytes_obj(static_cast<const char*>(data), dataLen);
+    
+    try {
+        // Import the codecs module
+        py::module_ codecs = py::module_::import("codecs");
+        
+        // For wide character data from SQL Server (always UTF-16LE)
+        if (isWideChar) {
+            return codecs.attr("decode")(bytes_obj, py::str("utf-16le"), py::str("strict"));
+        }
+        // For regular character data, use the specified encoding
+        else {
+            return codecs.attr("decode")(bytes_obj, py::str(encoding), py::str("strict"));
+        }
+    }
+    catch (const std::exception& e) {
+        // Log the error
+        LOG("DecodeString error: {}", e.what());
+        
+        // Try with replace error handler
+        try {
+            py::module_ codecs = py::module_::import("codecs");
+            if (isWideChar) {
+                return codecs.attr("decode")(bytes_obj, py::str("utf-16le"), py::str("replace"));
+            } else {
+                return codecs.attr("decode")(bytes_obj, py::str(encoding), py::str("replace"));
+            }
+        }
+        catch (const std::exception&) {
+            // Last resort: return error message
+            return py::str("[Decoding Error]");
+        }
+    }
+}
+
+py::bytes EncodeString(const std::string& text, const std::string& encoding, bool toWideChar) {
+    // Import Python's codecs module
+    py::module_ codecs = py::module_::import("codecs");
+    
+    // Detailed logging for debugging
+    std::cout << "========== EncodeString DEBUG ==========" << std::endl;
+    std::cout << "Input text: '" << text << "'" << std::endl;
+    std::cout << "Requested encoding: " << encoding << std::endl;
+    std::cout << "toWideChar flag: " << (toWideChar ? "true" : "false") << std::endl;
+    
+    try {
+        py::bytes result;
+        
+        if (toWideChar) {
+            std::cout << "Processing for SQL_C_WCHAR (wide character)" << std::endl;
+            
+            // For East Asian encodings that need special handling
+            if (encoding == "gbk" || encoding == "gb2312" || encoding == "gb18030" || 
+                encoding == "cp936" || encoding == "big5" || encoding == "cp950" || 
+                encoding == "shift_jis" || encoding == "cp932" || encoding == "euc_kr" ||
+                encoding == "cp949" || encoding == "euc_jp") {
+                
+                std::cout << "Using East Asian encoding: " << encoding << std::endl;
+                
+                // First decode the string using the specified encoding to get Unicode
+                py::object unicode_str = codecs.attr("decode")(
+                    py::bytes(text.data(), text.size()), 
+                    py::str(encoding),
+                    py::str("strict")
+                );
+                
+                std::cout << "Successfully decoded with " << encoding << std::endl;
+                
+                // Now encode as UTF-16LE for SQL Server
+                result = codecs.attr("encode")(unicode_str, py::str("utf-16le"), py::str("strict"));
+                std::cout << "Re-encoded to UTF-16LE for SQL Server" << std::endl;
+            } 
+            else {
+                // For all other encodings with wide chars, use UTF-16LE
+                std::cout << "Using UTF-16LE for wide character data" << std::endl;
+                result = codecs.attr("encode")(py::str(text), py::str("utf-16le"), py::str("strict"));
+            }
+        }
+        else {
+            // For SQL_C_CHAR, use the specified encoding directly
+            std::cout << "Processing for SQL_C_CHAR (narrow character)" << std::endl;
+            std::cout << "Using specified encoding: " << encoding << std::endl;
+            result = codecs.attr("encode")(py::str(text), py::str(encoding), py::str("strict"));
+        }
+        
+        // Log the result size
+        size_t result_size = PyBytes_Size(result.ptr());
+        std::cout << "Encoded result size: " << result_size << " bytes" << std::endl;
+        
+        // Debug first few bytes of the result
+        const char* data = PyBytes_AsString(result.ptr());
+        std::cout << "First bytes (hex): ";
+        for (size_t i = 0; i < std::min(result_size, size_t(16)); ++i) {
+            std::cout << std::hex << std::setw(2) << std::setfill('0') 
+                      << (static_cast<int>(data[i]) & 0xFF) << " ";
+        }
+        std::cout << std::dec << std::endl;
+        
+        std::cout << "EncodeString completed successfully" << std::endl;
+        std::cout << "=======================================" << std::endl;
+        return result;
+    } 
+    catch (const std::exception& e) {
+        // Log the error
+        std::cout << "ERROR in EncodeString: " << e.what() << std::endl;
+        LOG("EncodeString error: {}", e.what());
+        
+        try {
+            // Fallback with replace error handler
+            std::cout << "Attempting fallback encoding..." << std::endl;
+            py::bytes result;
+            
+            if (toWideChar) {
+                result = codecs.attr("encode")(py::str(text), py::str("utf-16le"), py::str("replace"));
+                std::cout << "Fallback: Encoded with utf-16le and replace error handler" << std::endl;
+            } 
+            else {
+                result = codecs.attr("encode")(py::str(text), py::str(encoding), py::str("replace"));
+                std::cout << "Fallback: Encoded with " << encoding << " and replace error handler" << std::endl;
+            }
+            
+            std::cout << "Fallback encoding successful" << std::endl;
+            std::cout << "=======================================" << std::endl;
+            return result;
+        } 
+        catch (const std::exception& e2) {
+            // Ultimate fallback
+            std::cout << "ERROR in fallback encoding: " << e2.what() << std::endl;
+            std::cout << "Using ultimate fallback to UTF-8" << std::endl;
+            LOG("Fallback encoding error: {}", e2.what());
+            
+            py::bytes result = codecs.attr("encode")(py::str(text), py::str("utf-8"), py::str("replace"));
+            std::cout << "Ultimate fallback completed" << std::endl;
+            std::cout << "=======================================" << std::endl;
+            return result;
+        }
+    }
+}
+
 namespace {
 
 const char* GetSqlCTypeAsString(const SQLSMALLINT cType) {
@@ -240,7 +385,9 @@ std::string DescribeChar(unsigned char ch) {
 // appropriate arguments
 SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                          std::vector<ParamInfo>& paramInfos,
-                         std::vector<std::shared_ptr<void>>& paramBuffers) {
+                         std::vector<std::shared_ptr<void>>& paramBuffers,
+                         const std::string& encoding = "utf-16le",
+                         int /* ctype */ = SQL_WCHAR) {
     LOG("Starting parameter binding. Number of parameters: {}", params.size());
     for (int paramIndex = 0; paramIndex < params.size(); paramIndex++) {
         const auto& param = params[paramIndex];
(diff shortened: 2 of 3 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,151 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+        }
+    }
+}
+
+py::bytes EncodeString(const std::string& text, const std::string& encoding, bool toWideChar) {
+    // Import Python's codecs module
+    py::module_ codecs = py::module_::import("codecs");
+    
+    // Detailed logging for debugging
+    std::cout << "========== EncodeString DEBUG ==========" << std::endl;
```
sumitmsft: The C++ binding layer introduces extensive `std::cout` debug logging that may be printing raw parameter data, encoded byte dumps, interim transformations, and decoding results. This is a critical security & performance concern:

- Potential leakage of PII / secrets / proprietary SQL text into logs.
- Large performance overhead in tight loops (parameter binding, fetch paths, batch operations).
- Noise obscures real diagnostic messages.
- Inconsistent with existing logging abstraction (LOG macro + sanitization).

Required Remediation:
- Remove or wrap all `std::cout` debug statements with `#ifdef DDBC_DEBUG` (compile‑time guard) or a runtime verbosity flag.
- Replace remaining necessary diagnostics with `LOG("debug", ...)` (or equivalent) using sanitized & truncated payloads (e.g., show length, first N (<=32) bytes/characters, redact binary).
- Avoid logging full parameter values or entire hex sequences—log metadata only (length, type, ctype, encoding).
- Ensure any user input passed to logs is sanitized (reuse existing sanitize_user_input helper if available).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,151 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+    std::cout << "Input text: '" << text << "'" << std::endl;
```
sumitmsft: What is input text here?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -257,15 +404,62 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
+                                      << (static_cast<int>(text_to_encode[i]) & 0xFF) << " ";
+                        }
+                        std::cout << std::dec << std::endl;
+                        
+                        py::bytes encoded = EncodeString(text_to_encode, encoding, false);
+                        std::string encoded_str = encoded.cast<std::string>();
+                        strParam = AllocateParamBuffer<std::string>(paramBuffers, encoded_str);
+                        
+                        std::cout << "  Encoded length: " << encoded_str.size() << " bytes" << std::endl;
+                        std::cout << "  Encoded bytes: ";
```
sumitmsft: Is this required to print?

mssql_python/connection.py
```diff
@@ -342,7 +340,7 @@ def setencoding(self, encoding=None, ctype=None):
         
         # Set default encoding if not provided
         if encoding is None:
-            encoding = 'utf-16le'
+            encoding = 'utf-8'
```
sumitmsft, line 356: `connection.setencoding()` now defaults to UTF-8 (`SQL_CHAR`), but the cursor fallback in `_get_encoding_settings()` (cursor.py) still hardcodes `'utf-16le'` + `SQL_WCHAR` if `getencoding()` is missing or raises. This creates two conflicting "defaults."

This is where the default was changed. Any fallback elsewhere must match this (or you revert this change) to avoid divergence.

Why this matters: If `cursor._get_encoding_settings()` hits its exception path (e.g., transient attribute error, race during initialization, or a future refactor), the cursor will encode parameters and decode results as UTF‑16LE while the connection layer expects UTF‑8.

This mismatch is silent: ASCII tests will pass, but multi‑byte characters (e.g., East Asian, emoji) can become corrupted or misinterpreted. Debugging will be difficult because corruption only appears when the fallback path triggers.

mssql_python/cursor.py
```diff
@@ -104,6 +104,49 @@ def __init__(self, connection, timeout: int = 0) -> None:
...
+                return self._connection.getencoding()
+            except:
+                # Return default encoding settings if there's an error
+                return {
+                    'encoding': 'utf-16le', 
+                    'ctype': ddbc_sql_const.SQL_WCHAR.value
+                }
+        # Return default encoding settings if getencoding is not available
+        return {
+            'encoding': 'utf-16le',
```
sumitmsft: This is the conflicting fallback that should use the same default as `connection.setencoding` (currently UTF‑8/SQL_CHAR) or be refactored to pull from a single shared constant.

mssql_python/connection.py
jahnvi480: Making default as utf-8 and SQL_CHAR since we are not supporting encoding and decoding for SQL_WCHAR

mssql_python/connection.py
```diff
@@ -373,6 +371,13 @@ def setencoding(self, encoding=None, ctype=None):
                 ddbc_error=f"ctype must be SQL_CHAR ({ConstantsDDBC.SQL_CHAR.value}) or SQL_WCHAR ({ConstantsDDBC.SQL_WCHAR.value})",
             )
         
+        # Enforce UTF-16LE for SQL_WCHAR
+        if ctype == ConstantsDDBC.SQL_WCHAR.value and encoding not in UTF16_ENCODINGS:
+            raise ValueError(
```
sumitmsft: Replace ValueError with ProgrammingError (or introduce a specific EncodingError subclass that inherits from ProgrammingError to preserve compatibility) to be DBAPI compliant.
Ensure the error message is split into driver_error and ddbc_error (matching existing pattern) for consistency.

mssql_python/connection.py
```diff
@@ -458,39 +459,49 @@ def setdecoding(self, sqltype, encoding=None, ctype=None):
                 ddbc_error=f"sqltype must be SQL_CHAR ({ConstantsDDBC.SQL_CHAR.value}), SQL_WCHAR ({ConstantsDDBC.SQL_WCHAR.value}), or SQL_WMETADATA ({SQL_WMETADATA})",
             )
         
-        # Set default encoding based on sqltype if not provided
-        if encoding is None:
-            if sqltype == ConstantsDDBC.SQL_CHAR.value:
+        # For SQL_WCHAR and SQL_WMETADATA, enforce UTF-16LE encoding and SQL_WCHAR ctype
+        if sqltype in (ConstantsDDBC.SQL_WCHAR.value, SQL_WMETADATA):
+            if encoding is not None and encoding.lower() not in UTF16_ENCODINGS:
+                raise ValueError(
```
sumitmsft: Replace ValueError with ProgrammingError (or introduce a specific EncodingError subclass that inherits from ProgrammingError to preserve compatibility) to be DBAPI compliant.
Ensure the error message is split into driver_error and ddbc_error (matching existing pattern) for consistency.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1764,37 +2075,51 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
...
+                            std::string bytesStr = value.cast<std::string>();
+                            // Use Python's str() to get a string representation
+                            py::object pyStr = py::str(bytesStr);
+                            // Use EncodeString to properly handle the encoding to UTF-16LE
+                            py::bytes encoded = EncodeString(pyStr.cast<std::string>(), encoding, true);
+                            // Convert to wstring
+                            wstr = py::str(encoded).cast<std::wstring>();
                         }
+                        
+                        size_t copySize = std::min(wstr.size(), info.columnSize);
```
sumitmsft: for wide (WCHAR) and binary/char parameters. This silently chops (truncates) any value that’s too long for the column without warning or error. Silent truncation can corrupt business data and is harder to debug later.

The old behavior raised an error if data was too long. Now we could lose data and not notice.

I think we should restore a length check: if the string/bytes length is greater than the column size, raise an error (like before).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1844,17 +2169,29 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
...
+                        if (py::isinstance<py::str>(value)) {
+                            // Use EncodeString function with false for toWideChar (not wide char)
+                            py::bytes encoded = EncodeString(value.cast<std::string>(), encoding, false);
+                            str = encoded.cast<std::string>();
+                        } else if (py::isinstance<py::bytes>(value) || py::isinstance<py::bytearray>(value)) {
+                            // For bytes/bytearray, use as-is
+                            str = value.cast<std::string>();
                         }
+                        
+                        size_t copySize = std::min(str.size(), info.columnSize);
```
sumitmsft: Here also we silently chops (truncates) any value that’s too long for the column without warning or error. Silent truncation can corrupt business data and is harder to debug later.

The old behavior raised an error if data was too long. Now we could lose data and not notice.

I think we should restore a length check: if the string/bytes length is greater than the column size, raise an error (like before).

mssql_python/cursor.py
```diff
@@ -104,6 +104,49 @@ def __init__(self, connection, timeout: int = 0) -> None:
...
+        """
+        Get the encoding settings from the connection.
+        
+        Returns:
+            dict: A dictionary with 'encoding' and 'ctype' keys, or default settings if not available
+        """
+        if hasattr(self._connection, 'getencoding'):
+            try:
+                return self._connection.getencoding()
+            except:
```
sumitmsft: This except is too broad. If getencoding() or getdecoding() fails due to a programming bug (AttributeError, logic error, etc.), we silently force a fallback encoding (utf-16le), which can mask real defects.

mssql_python/cursor.py
```diff
@@ -104,6 +104,49 @@ def __init__(self, connection, timeout: int = 0) -> None:
...
+        Args:
+            sql_type: SQL type constant (SQL_CHAR, SQL_WCHAR, etc.)
+            
+        Returns:
+            Dictionary containing the decoding settings.
+        """
+        try:
+            # Get decoding settings from connection for this SQL type
+            return self._connection.getdecoding(sql_type)
+        except Exception as e:
```
sumitmsft: This too... This except is too broad. If getencoding() or getdecoding() fails due to a programming bug (AttributeError, logic error, etc.), we silently force a fallback encoding (utf-16le), which can mask real defects.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,151 @@ SQLTablesFunc SQLTables_ptr = nullptr;
 
 SQLDescribeParamFunc SQLDescribeParam_ptr = nullptr;
 
+py::object DecodeString(const void* data, SQLLEN dataLen, const std::string& encoding, bool isWideChar) {
+    if (data == nullptr || dataLen <= 0) {
+        return py::none();
+    }
+    
+    // Create a bytes object with the raw binary data
+    py::bytes bytes_obj(static_cast<const char*>(data), dataLen);
```
sumitmsft: In `EncodeString` and `DecodeString` we call `py::module_::import("codecs")` every time, and we often encode then decode then re‑encode the same string. These functions are used for each parameter bind & each fetched column, so the overhead scales with row/parameter count.

This can cause significant slowdown for large executemany batches or wide result sets.

I would like you to explore:
- Cache the codecs module (and common encoder/decoder callables) once instead of importing on every call.
- Use direct CPython APIs (PyUnicode_FromStringAndSize, PyUnicode_AsWideCharString,
- PyUnicode_DecodeUTF16) to skip extra round trips through codecs.
- Remove redundant encode>> decode >> re‑encode operations for wide char paths, encode directly to UTF‑16LE once.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2540,37 +2867,29 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
-                                std::string utf8str = WideToUTF8(wstr);
-                                row.append(py::str(utf8str));
-#else
-                                std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()));
-                                row.append(py::cast(wstr));
-#endif
-                                LOG("Appended NVARCHAR string of length {} to result row", numCharsInData);
-                            }  else {
+                                // Use the common decoding function
+                                row.append(DecodeString(dataBuffer.data(), dataLen, wcharEncoding, true));
```
sumitmsft: In the wide (WCHAR/NVARCHAR) fetch paths we compute numCharsInData = dataLen / sizeof(SQLWCHAR) and then sometimes pass either dataLen or a recomputed numCharsInData * sizeof(SQLWCHAR) into DecodeString. If the buffer includes a null terminator or unused capacity, or if dataLen is odd or cuts a surrogate pair in half, we can end up mis-decoding or silently corrupting data (especially for emoji / supplementary plane characters).

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2522,15 +2849,15 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                }
                 break;
             }
             case SQL_WCHAR:
             case SQL_WVARCHAR:
             case SQL_WLONGVARCHAR: {
                 if (columnSize == SQL_NO_TOTAL || columnSize > 4000) {
                     LOG("Streaming LOB for column {} (NVARCHAR)", i);
-                    row.append(FetchLobColumnData(hStmt, i, SQL_C_WCHAR, true, false));
+                    row.append(FetchLobColumnData(hStmt, i, SQL_C_WCHAR, true, false, charEncoding, wcharEncoding));
```
sumitmsft, line 3086: Confirm FetchLobColumnData passes the actual ODBC-reported byte length (not buffer capacity) into DecodeString; wide length must remain in bytes.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3194,20 +3515,21 @@ SQLRETURN FetchBatchData(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& colum
...
-                        // On Windows, wchar_t and SQLWCHAR are both 2 bytes, so direct cast works
-                        row.append(std::wstring(
-                            reinterpret_cast<wchar_t*>(&buffers.wcharBuffers[col - 1][i * fetchBufferSize]),
-                            numCharsInData));
-#endif
+                        #if defined(__APPLE__) || defined(__linux__)
+                            // Use unix-specific conversion to handle the wchar_t/SQLWCHAR size difference
+                            SQLWCHAR* wcharData = &buffers.wcharBuffers[col - 1][i * fetchBufferSize];
+                            // Use DecodeString directly with the raw data
+                            py::object decodedStr = DecodeString(wcharData, numCharsInData * sizeof(SQLWCHAR), wcharEncoding, true);
```
sumitmsft: Here we reconstruct byte length as numCharsInData * sizeof(SQLWCHAR). If numCharsInData was derived from a buffer that includes a null terminator or unused capacity, or if a surrogate pair boundary was truncated, we could mis-decode. Prefer passing the original ODBC dataLen (in bytes) captured per row/column rather than recomputing. Also add a defensive check: if (bytesLen % 2 != 0) log & fail for wide data.

Please talk to @gargsaumya for these implementations

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,151 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+            }
+        }
+        catch (const std::exception&) {
+            // Last resort: return error message
+            return py::str("[Decoding Error]");
+        }
+    }
+}
+
+py::bytes EncodeString(const std::string& text, const std::string& encoding, bool toWideChar) {
```
sumitmsft: For Python str parameters we already have Unicode; avoid legacy round‑trip (decode >> encode >> decode>> encode). Directly encode once (UTF‑16LE for wide, requested encoding for narrow). This will remove multiple allocations and potential lossy conversions.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,151 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+    // Import Python's codecs module
+    py::module_ codecs = py::module_::import("codecs");
```
sumitmsft: cache codecs module (static) or use CPython APIs (PyUnicode_AsEncodedString, PyUnicode_DecodeUTF16) to avoid re-import cost each parameter

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,134 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+        py::object decode_func;
+        py::object encode_func;
+        CodecCache() {
+            codecs_module = py::module_::import("codecs");
+            decode_func = codecs_module.attr("decode");
+            encode_func = codecs_module.attr("encode");
+        }
+    };
+    CodecCache& get_codec_cache() {
+        static CodecCache cache;
```
Ramudaykumar: get_codec_cache() uses a static local variable. This is thread-safe in C++11 and later, but if multiple interpreters or sub-interpreters are used, this could cause issues.
Document that this is safe only for single-interpreter scenarios or consider per-interpreter caching.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,134 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+py::bytes EncodeString(const std::string& text, const std::string& encoding, bool toWideChar) {
+    try {
+        if (toWideChar) {
+            // Encode directly to UTF-16LE using CPython API
+            // First, create a Unicode object from the input string
+            PyObject* unicode = PyUnicode_FromStringAndSize(text.data(), static_cast<Py_ssize_t>(text.size()));
+            if (!unicode) throw py::error_already_set();
+            // Encode to UTF-16LE (no BOM)
+            PyObject* encoded = PyUnicode_AsEncodedString(unicode, "utf-16le", "strict");
+            Py_DECREF(unicode);
```
Ramudaykumar: if an exception occurs before Py_DECREF, it could leak. Use py::handle or RAII wrappers consistently to avoid manual decref. you may use Try-Catch as well.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,134 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+            if (!encoded) throw py::error_already_set();
+            py::bytes result = py::reinterpret_steal<py::bytes>(encoded);
+            return result;
+        } else {
+            // For SQL_C_CHAR, use CPython API for utf-8/ascii, else fallback to codecs.encode
+            if (encoding == "utf-8" || encoding == "ascii") {
+                PyObject* unicode = PyUnicode_FromStringAndSize(text.data(), static_cast<Py_ssize_t>(text.size()));
+                if (!unicode) throw py::error_already_set();
+                PyObject* encoded = PyUnicode_AsEncodedString(unicode, encoding.c_str(), "strict");
+                Py_DECREF(unicode);
```
Ramudaykumar: Same as above

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,134 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+    }
+    catch (const std::exception& e) {
+        LOG("EncodeString error: {}", e.what());
+        // Fallback with "replace" error handler
+        try {
+            if (toWideChar) {
+                PyObject* unicode = PyUnicode_FromStringAndSize(text.data(), static_cast<Py_ssize_t>(text.size()));
+                if (!unicode) throw py::error_already_set();
+                PyObject* encoded = PyUnicode_AsEncodedString(unicode, "utf-16le", "replace");
+                Py_DECREF(unicode);
```
Ramudaykumar: Same here.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -317,16 +491,75 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
+                    std::wstring* strParam = nullptr;
+                    
+                    if (py::isinstance<py::str>(param)) {
+                        // For Python strings, convert to wstring using EncodeString
+                        std::string text_to_encode = param.cast<std::string>();
+                        py::bytes encoded = EncodeString(text_to_encode, encoding, true); // true for wide character
+                        py::object decoded = py::module_::import("codecs").attr("decode")(encoded, py::str("utf-16le"), py::str("strict"));
+                        std::wstring wstr = decoded.cast<std::wstring>();
+                        
+                        // Check if data would be truncated and raise error
```
Ramudaykumar, line 631: Double Conversion Overhead

Current flow for Python str:
py::str → std::string → EncodeString (UTF-16LE) → py::bytes → codecs.decode → py::str → std::wstring

This is inefficient because you’re decoding twice (once in Python, once in C++).
Better Approach: Directly cast param to std::wstring using param.cast<std::wstring>() (Pybind11 supports this).
Then validate and allocate without round-tripping through UTF-16LE bytes.

Ramudaykumar:
This comment is not just this PR specific, rather binding.cpp file specific. I would suggest to create a task to track these.

1. String/Wide String Conversion Overhead

For wide-char parameters, the code sometimes round-trips through Python encode/decode, which is inefficient.
For py::str to std::wstring, use param.cast<std::wstring>() directly.
Only use encode/decode for bytes/bytearray or when explicit encoding is needed.

2. LOB/Variable-Length Data Handling
For columns with unknown or huge size, the code falls back to streaming, but the allocation logic for buffers (e.g., 2GB for LONGVARCHAR) is risky.
Always prefer streaming for LOBs.
Avoid allocating large vectors based on column size alone; use a reasonable cap and document the fallback.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -317,16 +491,75 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                     bufferLength = 0;
                 } else {
                     // Normal small-string case
-                    std::wstring* strParam =
-                        AllocateParamBuffer<std::wstring>(paramBuffers, param.cast<std::wstring>());
-                    LOG("SQL_C_WCHAR Parameter[{}]: Length={}, isDAE={}", paramIndex, strParam->size(), paramInfo.isDAE);
+                    std::wstring* strParam = nullptr;
+                    
+                    if (py::isinstance<py::str>(param)) {
+                        // For Python strings, convert to wstring using EncodeString
```
gargsaumya: I’m concerned there may be a performance issue in the current implementation due to an unnecessary encode/decode cycle for SQL_C_WCHAR parameters.
Right now, the data goes through five conversions:
1. Python str → UTF-8 bytes (param.cast<std::string>())
2. UTF-8 bytes → UTF-16LE bytes (EncodeString)
3. UTF-16LE bytes → Python str (codecs.decode)
4. Python str → wstring (decoded.cast<std::wstring>())
5. wstring → SQLWCHAR buffer (WStringToSQLWCHAR)

However, since Python 3 strings are already Unicode internally, we can simplify this to just two conversions, like so:

if (py::isinstance<py::str>(param)) {
    // Direct conversion - pybind11 handles Python str → wstring efficiently
    std::wstring wstr = param.cast<std::wstring>();
    
    // Validation
    if (wstr.size() > paramInfo.columnSize) {
        // ... error handling ...
    }
    
    strParam = AllocateParamBuffer<std::wstring>(paramBuffers, wstr);
}

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -257,15 +387,49 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
                     *strLenOrIndPtr = SQL_LEN_DATA_AT_EXEC(0);
                     bufferLength = 0;
                 } else {
-                    std::string* strParam =
-                        AllocateParamBuffer<std::string>(paramBuffers, param.cast<std::string>());
+                    // Use the specified encoding when converting to string
+                    std::string* strParam = nullptr;
+                    if (py::isinstance<py::str>(param)) {
+                        // Use the EncodeString function to handle encoding properly
+                        std::string text_to_encode = param.cast<std::string>();
```
gargsaumya: Looks like we’re doing a double encoding here. 
1. First to UTF-8 (param.cast<std::string>())
2. Then again to the target encoding in EncodeString. 

We should skip the intermediate UTF-8 step and encode directly from Python str to the target encoding.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,134 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+        return py::none();
+    }
+
+    try {
+        if (isWideChar) {
+            // SQL Server always returns UTF-16LE for wide char columns
+            // Use PyUnicode_DecodeUTF16 directly for best performance
+            // Note: SQLWCHAR is always 2 bytes (UTF-16LE) on all platforms for SQL Server
+            int byteorder = -1;
+            PyObject* unicode = PyUnicode_DecodeUTF16(
```
gargsaumya, line 212: This usage is correct. The `py::reinterpret_steal` immediately wraps the PyObject*, ensuring automatic cleanup. This is the pattern that should be used throughout.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -167,6 +167,134 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+    }
+}
+
+// EncodeString: Efficiently encode Python str to bytes using CPython APIs where possible
+py::bytes EncodeString(const std::string& text, const std::string& encoding, bool toWideChar) {
+    try {
+        if (toWideChar) {
+            // Encode directly to UTF-16LE using CPython API
+            // First, create a Unicode object from the input string
+            PyObject* unicode = PyUnicode_FromStringAndSize(text.data(), static_cast<Py_ssize_t>(text.size()));
```
gargsaumya: If `PyUnicode_AsEncodedString` throws an exception before `Py_DECREF(unicode)` is called, the `unicode` object will leak.
You can fix this by using RAII for automatic cleanup, I’ve highlighted above where it’s used correctly.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Fixed! 

Changes made:

- Added byte boundary validation dataLen % sizeof(SQLWCHAR) != 0
- Surrogate pair protection (detect 0xD800-0xDBFF ranges)
- Consistent length passing to DecodeString
- Applied to all WCHAR fetch paths (SQLGetData, FetchBatch, FetchLob)

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Confirmed - `FetchLobColumnData` correctly passes the ODBC-reported byte length to `DecodeString`:

Uses actual ODBC byte length: `actualRead` from `SQLGetData_ptr()` (not buffer capacity)
Preserves byte length for wide chars: Accumulates actual bytes read, validates even boundaries for WCHAR without converting to character counts
Passes bytes to DecodeString: Final call uses `buffer.size()` (total bytes) as required by the function signature
The wide character handling maintains byte-level precision throughout the entire fetch-to-decode pipeline.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: RESOLVED - This optimization is already implemented and working correctly.

For Python `str` parameters, we are indeed avoiding legacy round-trip conversions:

Wide char (UTF-16LE): Direct single encode from Python Unicode → UTF-16LE bytes using `PyUnicode_AsEncodedString()`
Narrow char: Direct single encode from Python Unicode → target encoding bytes using CPython APIs for common encodings (UTF-8, ASCII) or codecs module for others
No decode→encode→decode→encode cycles: The implementation directly encodes the Python `str` (which is already Unicode) to the target encoding in one operation

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: RESOLVED - Static caching has been removed entirely.

The get_codec_cache() function with static local variables no longer exists. It has been replaced with a simple get_codecs_module() function that performs fresh imports each time

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: FIXED: Using RAII Solution
All manual decrements replaced with RAII wrappers
Both main encoding path and fallback error-handling paths are now memory-safe

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: FIXED: Using RAII Solution
All manual decrements replaced with RAII wrappers
Both main encoding path and fallback error-handling paths are now memory-safe

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: FIXED: Using RAII Solution
All manual decrements replaced with RAII wrappers
Both main encoding path and fallback error-handling paths are now memory-safe

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: FIXED: Using RAII Solution
All manual decrements replaced with RAII wrappers
Both main encoding path and fallback error-handling paths are now memory-safe

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: The double conversion overhead has been completely eliminated - both py::str and bytes/bytearray parameters now use direct, optimized conversion paths to std::wstring for WCHAR parameter binding.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Resolved:
The code now uses optimized direct conversion paths that eliminate the inefficient 5-step conversion chain you identified

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Resolved this.

jahnvi480:
Closing this PR created a new PR#301

PR #271, opened 2025-10-01
FIX: Static UUID variable causes pytest crash

bewithgaurav:

> [AB#39063](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39063)

## Problem
Pytest test suite (335+ tests) was crashing on exit with:
```
Fatal Python error: _Py_GetConfig: the function must be called with the GIL held, 
after Python initialization and before Python finalization, but the GIL is released
```

The crash occurred **after all tests passed**, during Python shutdown on Windows x64 and macOS.

## Root Cause
Static `pybind11::module_` and `pybind11::object` instances in `BindParameterArray()` function (lines 2072-2073) were being destroyed during C++ static destruction phase, which happens **after** Python has released the GIL and begun finalization.

When the destructors called `Py_XDECREF`, Python's `_Py_GetConfig()` was invoked without the GIL, causing the fatal error.

```cpp
// OLD CODE - Problematic static objects
static py::module_ uuid_mod = py::module_::import("uuid");
static py::object uuid_class = uuid_mod.attr("UUID");
```

## Solution
Moved UUID class caching to module initialization level using a helper function. This ensures the Python object lifecycle is managed within pybind11's module context, avoiding destructor calls during finalization.

```cpp
// NEW CODE - Module-level cached helper
py::object uuid_class = py::module_::import("mssql_python.ddbc_bindings").attr("_get_uuid_class")();
```

The `_get_uuid_class()` helper is defined at module initialization and maintains a static cache that lives for the module's lifetime, properly integrated with pybind11's lifecycle management.

## Testing
- ✅ Full test suite (335+ tests) now completes without crash on macOS
- ✅ Exit code 0 after pytest completion
- ✅ No functionality changes - UUID parameter binding works identically

## Impact
- **Platforms affected**: Windows x64, macOS (all platforms with the issue)
- **Risk**: Low - surgical fix, no behavioral changes
- **Files changed**: 1 file, ~15 lines added/modified

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -2069,8 +2069,10 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
                     SQLGUID* guidArray = AllocateParamBufferArray<SQLGUID>(tempBuffers, paramSetSize);
                     strLenOrIndArray = AllocateParamBufferArray<SQLLEN>(tempBuffers, paramSetSize);
 
-                    static py::module_ uuid_mod = py::module_::import("uuid");
-                    static py::object uuid_class = uuid_mod.attr("UUID");
+                    // Get cached UUID class from module-level helper
+                    // This avoids static object destruction issues during Python finalization
+                    py::object uuid_class = py::module_::import("mssql_python.ddbc_bindings").attr("_get_uuid_class")();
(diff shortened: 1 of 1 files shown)


bewithgaurav:
Adding this to the release branch

Issue #213, opened 2025-08-29
Handle Type -155


lstorme:
### Describe the bug
Need to handle MsSQL type -155

E       RuntimeError: Unsupported data type for column - 0x7ffee88bb7c0, Type - -155, column ID - 7
```
self = <mssql_python.cursor.Cursor object at 0x7f31b52f7980>

    def fetchone(self) -> Union[None, Row]:
        """
        Fetch the next row of a query result set.

        Returns:
            Single Row object or None if no more data is available.
        """
        self._check_closed()  # Check if the cursor is closed

        # Fetch raw data
        row_data = []
>       ret = ddbc_bindings.DDBCSQLFetchOne(self.hstmt, row_data)
E       RuntimeError: Unsupported data type for column - 0x7ffee88bb7c0, Type - -155, column ID - 7

../../../../py312/lib/python3.12/site-packages/mssql_python/cursor.py:724: RuntimeError

```

### To reproduce
Try to select from table containing datetimeoffset or datetime

### Expected behavior
Return record

### Further technical details
Python version: 3.12
SQL Server version: (e.g. SQL Server 2022)
Operating system: RedHat 8

**Additional context**
table:
    [ID] [bigint] IDENTITY(1,1) NOT NULL,
    [RecCreateDateTime] [datetime] default getutcdate(),
    [FileID] [varchar](10) NOT NULL,  -- merchant affected by the task
    [task] [varchar](16) NOT NULL,    -- type of the task BATCH, S3, ...
    [dateTimeUTC] [datetimeoffset] NULL,       -- date and time of the task
    [guid] [uniqueidentifier],

gargsaumya:
Thanks for reporting this! The error occurs because datetimeoffset (ODBC type -155) isn’t supported yet in mssql-python. We’re actively working on adding support, and it will be available shortly.

We really appreciate you trying out the driver, please keep sharing your feedback and inputs!

gargsaumya:
Hi @lstorme,
Wanted to let you know that the support for this ODBC type is now added and the related pull request (https://github.com/microsoft/mssql-python/pull/236 ) has been merged. You can expect this to be included in the next release, which is planned for Friday.

Once the release is out, please give it a try and let us know if it works as expected. Your feedback is really valuable.
Thanks again for contributing to improving mssql-python, and please keep raising any issues or suggestions you encounter. :)

gargsaumya:
[...]

PR #281, opened 2025-10-07
FIX: Retain original timezone in Python datetime objects

gargsaumya:

> AB#<WORK_ITEM_ID>

> GitHub Issue: #213 

### Summary   
  

This pull request updates how `datetimeoffset` values are handled when reading from SQL Server in the Python bindings. The main change is to preserve the original timezone information in returned Python `datetime` objects, instead of always converting them to UTC. Correspondingly, the test suite has been updated to compare datetimes with their original timezone rather than converting to UTC for assertions.

**Datetimeoffset handling improvements:**

* Removed forced conversion of `datetimeoffset` values to UTC in `SQLGetData_wrap` and `FetchBatchData`, so Python datetime objects retain their original timezone info. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2808) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3321)

**Test suite updates:**

* Updated all relevant tests in `tests/test_004_cursor.py` to compare datetimes directly, preserving timezone information, instead of converting to UTC for equality checks. This affects tests for read/write, max/min offsets, DST transitions, executemany, and extreme offsets. [[1]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L7890-R7890) [[2]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L7929-R7924) [[3]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L7989-R7979) [[4]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L8071-R8056) [[5]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L8147-R8122)

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -2805,7 +2805,6 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
                         microseconds,
                         tzinfo
                     );
-                    py_dt = py_dt.attr("astimezone")(datetime.attr("timezone").attr("utc"));
                     row.append(py_dt);
                 } else {
                     LOG("Error fetching DATETIMEOFFSET for column {}, ret={}", i, ret);
@@ -3318,7 +3317,6 @@ SQLRETURN FetchBatchData(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& colum
                             dtoValue.fraction / 1000,  // ns → µs
                             tzinfo
                         );
-                        py_dt = py_dt.attr("astimezone")(datetime.attr("timezone").attr("utc"));
                         row.append(py_dt);
                     } else {
                         row.append(py::none());
--- a/tests/test_004_cursor.py
+++ b/tests/test_004_cursor.py
@@ -7887,12 +7887,7 @@ def test_datetimeoffset_read_write(cursor, db_connection):
             assert row is not None
             fetched_id, fetched_dt = row
             assert fetched_dt.tzinfo is not None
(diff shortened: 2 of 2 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2805,7 +2805,6 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
                         microseconds,
                         tzinfo
                     );
-                    py_dt = py_dt.attr("astimezone")(datetime.attr("timezone").attr("utc"));
```
subrata-ms, line 2808: As you removing the UTC conversion here, how does the output looks like for the below query?

In [6]: cur.execute("SELECT TOP 1 Systime FROM cdp.EventLogs WHERE Recno=162243");cur.fetchall()
Out[6]: [(datetime.datetime(2025, 5, 14, 11, 35, 52, 501000, tzinfo=datetime.timezone.utc),)]

gargsaumya: @subrata-ms - that was user specific table. We have verified the handling using a similar test on a temporary table, inserting and fetching a known value. The fetched value preserves both the timestamp and the tzinfo-
 
Inserted: 2025-05-14 12:35:52.501000+01:00
Fetched : 2025-05-14 12:35:52.501000+01:00
tzinfo  : UTC+01:00

PR #282, opened 2025-10-07
FEAT: Support for Native_UUID Attribute

jahnvi480:

> AB#39059

### Summary   
This pull request introduces a new global setting, `native_uuid`, to the `mssql_python` package, allowing users to control whether UUID values are returned as `uuid.UUID` objects or as strings. The implementation includes updates to the package initialization, row processing logic, and a comprehensive set of tests to verify the new behavior and ensure backward compatibility.

**UUID Handling Improvements:**

* Added a `native_uuid` setting to the global configuration in `mssql_python/__init__.py`, defaulting to `False` for backward compatibility. This setting controls whether UUIDs are returned as `uuid.UUID` objects or as strings. 
* Updated the `Row` class in `mssql_python/row.py` to check the `native_uuid` setting and convert UUID values to strings when `native_uuid` is `False`, ensuring consistent output based on configuration.

**Testing Enhancements:**

* Updated and extended tests in `tests/test_004_cursor.py` to verify correct UUID handling for both `native_uuid=True` and `native_uuid=False`, including new tests for the setting and resetting of the `native_uuid` option. 

**Internal Refactoring:**

* Removed unused UUID mapping logic in `mssql_python/cursor.py` that is now handled via the new setting and row processing logic.
* Minor import and code organization cleanups in affected modules for clarity and maintainability.

These changes provide greater flexibility and control over how UUIDs are handled in query results, improving the usability of the package in different application contexts.

### DIFF
--- a/mssql_python/__init__.py
+++ b/mssql_python/__init__.py
@@ -5,6 +5,7 @@ This module initializes the mssql_python package.
 """
 import threading
 import locale
+import sys
 
 # Exceptions
 # https://www.python.org/dev/peps/pep-0249/#exceptions
@@ -31,6 +32,7 @@ class Settings:
         self.lowercase = False
         # Use the pre-determined separator - no locale access here
         self.decimal_separator = _DEFAULT_DECIMAL_SEPARATOR
+        self.native_uuid = False  # Default to False for backwards compatibility
 
 # Global settings instance
 _settings = Settings()
@@ -40,9 +42,11 @@ def get_settings():
     """Return the global settings object"""
     with _settings_lock:
         _settings.lowercase = lowercase
+        _settings.native_uuid = native_uuid
         return _settings
 
 lowercase = _settings.lowercase  # Default is False
+native_uuid = _settings.native_uuid  # Default is False
 
 # Set the initial decimal separator in C++
 from .ddbc_bindings import DDBCSetDecimalSeparator
@@ -166,7 +170,6 @@ def pooling(max_size=100, idle_timeout=600, enabled=True):
     else:
         PoolingManager.enable(max_size, idle_timeout)
 
-import sys
 _original_module_setattr = sys.modules[__name__].__setattr__
 
 def _custom_setattr(name, value):
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -363,19 +363,6 @@ class Cursor:
                     False,
                 )
                 
-            try:
-                val = uuid.UUID(param)
-                parameters_list[i] = val.bytes_le
-                return (
-                    ddbc_sql_const.SQL_GUID.value,
-                    ddbc_sql_const.SQL_C_GUID.value,
-                    16,
-                    0,
-                    False
-                )
-            except ValueError:
-                pass
-
             # String mapping logic here
             is_unicode = self._is_unicode_string(param)
 
--- a/mssql_python/row.py
+++ b/mssql_python/row.py
@@ -1,3 +1,6 @@
+from mssql_python import get_settings
+import uuid
+
 class Row:
     """
     A row of data from a cursor fetch operation. Provides both tuple-like indexing
@@ -30,7 +33,18 @@ class Row:
         if hasattr(cursor.connection, '_output_converters') and cursor.connection._output_converters:
             self._values = self._apply_output_converters(values)
         else:
-            self._values = values
+            # Check for native_uuid setting
+            if not get_settings().native_uuid:
+                # Convert UUID objects to strings when native_uuid is False
+                processed_values = []
+                for i, value in enumerate(values):
+                    if i < len(description) and description[i] and isinstance(value, uuid.UUID):
+                        processed_values.append(str(value))
+                    else:
+                        processed_values.append(value)
+                self._values = processed_values
+            else:
+                self._values = values
         
         # TODO: ADO task - Optimize memory usage by sharing column map across rows
         # Instead of storing the full cursor_description in each Row object:
--- a/tests/test_004_cursor.py
+++ b/tests/test_004_cursor.py
@@ -16,7 +16,6 @@ from contextlib import closing
 import mssql_python
 import uuid
 
-
 # Setup test table
 TEST_TABLE = """
 CREATE TABLE #pytest_all_data_types (
@@ -7187,8 +7186,16 @@ def test_uuid_insert_and_select_none(cursor, db_connection):
 
 def test_insert_multiple_uuids(cursor, db_connection):
     """Test inserting multiple UUIDs and verifying retrieval."""
-    table_name = "#pytest_uuid_multiple"
+    import uuid
+    
+    # Save original setting
+    original_value = mssql_python.native_uuid
+    
     try:
+        # Set native_uuid to True for this test
+        mssql_python.native_uuid = True
+        
+        table_name = "#pytest_uuid_multiple"
         cursor.execute(f"DROP TABLE IF EXISTS {table_name}")
         cursor.execute(f"""
             CREATE TABLE {table_name} (
@@ -7203,7 +7210,7 @@ def test_insert_multiple_uuids(cursor, db_connection):
 
         # Insert UUIDs and descriptions
         for desc, uid in uuids_to_insert.items():
-            cursor.execute(f"INSERT INTO {table_name} (id, description) VALUES (?, ?)", [uid, desc])
+            cursor.execute(f"INSERT INTO {table_name} (id, description) VALUES (?, ?)", [uid, desc]) 
         db_connection.commit()
 
         # Fetch all rows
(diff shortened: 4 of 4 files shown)


mssql_python/__init__.py
```diff
@@ -31,6 +32,7 @@ def __init__(self):
         self.lowercase = False
         # Use the pre-determined separator - no locale access here
         self.decimal_separator = _DEFAULT_DECIMAL_SEPARATOR
+        self.native_uuid = False  # Default to False for backwards compatibility
```
sumitmsft, line 36: `native_uuid` is a global flag. Row objects consult `get_settings().native_uuid` during construction (row.py lines 38–47). If one thread toggles `native_uuid` while another is fetching, the same result set can yield mixed types (some UUID columns as str, others as uuid.UUID), causing subtle downstream issues.

mssql_python/__init__.py
```diff
@@ -40,9 +42,11 @@ def get_settings():
     """Return the global settings object"""
     with _settings_lock:
         _settings.lowercase = lowercase
+        _settings.native_uuid = native_uuid
```
sumitmsft: The `global native_uuid flag` is not updated under the `_settings_lock`, unlike `lowercase`. The `_custom_setattr` only synchronizes `lowercase`; assigning `mssql_python.native_uuid = True` updates the module attribute but leaves `_settings.native_uuid` stale until the next `get_settings()` call. This creates a window where the two representations diverge, and future code that reads `_settings.native_uuid` directly (or caches settings early) could observe outdated state.

I think we can:

1. Make native_uuid thread-safe. Just like you already protect lowercase with a lock, do the same for native_uuid. (Or create a helper function set_native_uuid(flag) that:
>>checks the value is a real True/False (boolean)
>>uses the lock
>>updates both the internal settings object and the module variable.)

2. Simplify get_settings(). Right now it copies values back and forth. That’s unnecessary and can cause them to get briefly out of sync. Instead:
>>Keep _settings as the single “real” place where the values live.
>>get_settings() should just return _settings without rewriting anything.

Why this helps.
- Removes tiny timing windows where values disagree.
- Lowers the chance someone introduces a subtle bug later.

mssql_python/__init__.py
sumitmsft, line 36: Because the driver declares `threadsafety = 1`, users will typically assume that keeping separate connections per thread isolates behavior. 

Introducing a mutable global (`native_uuid`) breaks that assumption: one thread can flip the flag and silently change result types for another thread’s independent connection. 

This is similar to the behavior we have in PYODBC. However, this is quite risky. Now we have two options:

- To avoid surprising cross-thread effects, please consider: (a) making `native_uuid` a per-connection (or per-cursor) option captured at creation time, or 
- clearly documenting that it is a process-wide setting not safe to toggle concurrently. 

IMO - option 1 is much better and safer but it will break pyodbc based apps if the customers migrate to mssql-python. We can keep this behavior as is for now. Please clearly display warning in the WIKIs about this behavior and add a doc string in the code mentioning about this behavior.

sumitmsft, line 36: Please add a boolean-only, locked setter for native_uuid (reject non-bool inputs)

mssql_python/row.py
```diff
@@ -30,7 +33,21 @@ def __init__(self, cursor, description, values, column_map=None):
         if hasattr(cursor.connection, '_output_converters') and cursor.connection._output_converters:
             self._values = self._apply_output_converters(values)
         else:
-            self._values = values
+            self._values = self._process_uuid_values(values, description)
+
+    def _process_uuid_values(self, values, description):
```
sumitmsft, line 56: Row construction reads the global setting each time a row is materialized, so if `mssql_python.native_uuid` is toggled while a cursor is mid‑fetch, earlier rows in the same result set may have UUID columns as str, while later rows appear as `uuid.UUID` (or vice versa). This creates data shape instability inside a single logical query. No snapshot of the configuration is taken in Cursor.execute() or at the beginning of a result set.

Specifically in fetchone, fetchmany, and fetchall when we build Row objects repeatedly; each Row independently queries the current global setting.

In my opinion: We should capture the value of `native_uuid` once per result set (at `execute()` time or immediately after initializing description) and store it on the cursor (e.g. `self._native_uuid_snapshot`). Pass that snapshot into every Row so that mid‑stream changes to the global flag do not alter the representation of subsequent rows.

mssql_python/row.py
```diff
@@ -30,7 +33,21 @@ def __init__(self, cursor, description, values, column_map=None):
...
+        """
+        Convert UUID objects to strings if native_uuid setting is False.
+        """
+        if get_settings().native_uuid:
+            return values
+        processed_values = []
+        for i, value in enumerate(values):
```
sumitmsft: In row.py, each Row construction loops over every column and calls `isinstance(value, uuid.UUID)` even when only a small subset (often zero or one) of columns can ever be UUIDs. For large or wide result sets this adds avoidable overhead (millions of redundant `isinstance` checks across a typical workload).

Why this matters: Row materialization is on the critical path; Python per‑element overhead accumulates rapidly. The set of UUID columns is invariant for a result set—yet we re-discover that fact for every cell of every row.

Better approach is:
- At (or right after) query execution, inspect `cursor.description` once.
- Identify the indices whose SQL type maps to GUID / uniqueidentifier.
- Store that small list (maybe 0–2 indices typically) on the cursor (e.g., self._uuid_indices).
- During row construction, only examine those indices rather than scanning all columns.
- If native_uuid is True, skip everything entirely.


Additional Tests:
Benchmark or assert behavior on wide result set (e.g., 30 non-UUID columns + 1 UUID) to ensure conversion still correct.
Row with NULL in a UUID column (no error, remains None).
Confirm identical output vs current implementation for both native_uuid=False and True.

jahnvi480: Made the required changes

mssql_python/row.py
jahnvi480: Made the required changes

mssql_python/__init__.py
jahnvi480: Made the required changes

jahnvi480: Made the required changes

mssql_python/cursor.py
```diff
@@ -1005,24 +1005,26 @@ def execute(
...
+            # Capture settings snapshot for this result set
+            settings = get_settings()
+            self._settings_snapshot = {
+                'lowercase': settings.lowercase,
+                'native_uuid': settings.native_uuid
+            }
+            # Identify UUID columns (SQL_GUID = -11)
+            self._uuid_indices = []
+            for i, desc in enumerate(self.description):
+                if desc and desc[1] == uuid.UUID:  # Column type code at index 1
```
sumitmsft, line 1057: The UUID column detection uses if desc and desc[1] == uuid.UUID, relying on description tuples storing Python types. Elsewhere UUID handling appears to test for the integer SQL_GUID (-11). This inconsistency will prevent reliable precomputation of self._uuid_indices, causing unnecessary per-row scanning for UUID conversion.

Can we standardize on either integer SQL type codes (e.g., define SQL_GUID = ddbc_sql_const.SQL_GUID.value and store that in description) or Python type objects and update all UUID detection logic accordingly. If choosing integer codes, line 1020 should become if desc and desc[1] == SQL_GUID:

mssql_python/cursor.py
```diff
@@ -1005,24 +1005,26 @@ def execute(
         except Exception as e:
             # If describe fails, it's likely there are no results (e.g., for INSERT)
             self.description = None
-        
+
```
sumitmsft, line 1044: Lines 999–1007 perform the single metadata initialization after execute(). Did you remove a previously redundant second initialization here?  

Please confirm (or document) that the driver always exposes complete column metadata immediately after DDBCSQLExecute for all result-producing statements (including stored procedures, catalog functions). 
If there are edge cases where metadata becomes fully available only after the first fetch or SQLMoreResults, the removal could be a regression (UUID indices and type-dependent logic would be impacted). 
If guarantees exist, consider adding a brief comment above this block stating that one describe pass is sufficient. Optionally add a lightweight invariant check (e.g., all tuples have length 7; GUID columns mapped as expected) to fail fast in unusual scenarios.

If you verify that this is satisfied, you can safely proceed to just make it explicit so future contributors don’t reintroduce a second describe call or wonder why there is only one.

mssql_python/row.py
```diff
@@ -83,17 +142,22 @@ def _apply_output_converters(self, values):
...
                     # If value is already a Python type (str, int, etc.), 
-                    # we need to convert it to bytes for our converters
+                    # we need to handle it appropriately
                     if isinstance(value, str):
                         # Encode as UTF-16LE for string values (SQL_WVARCHAR format)
                         value_bytes = value.encode('utf-16-le')
                         converted_values[i] = converter(value_bytes)
+                    elif isinstance(value, int):
+                        # For integers, we'll convert to bytes
+                        value_bytes = value.to_bytes(8, byteorder='little')
```
sumitmsft: Integers are always converted to 8 little-endian bytes via value.to_bytes(8, byteorder='little'). This introduces multiple issues: 
1. If the value doesn’t fit in 8 bytes, to_bytes(8, ...) raises OverflowError. This is silently swallowed by the broad except block and the value is left unconverted which will be hard to diagnose.
2. Wrong widths for SQL integer types:
TINYINT → 1 byte
SMALLINT → 2 bytes
INTEGER → 4 bytes
BIGINT → 8 bytes
Padding everything to 8 bytes can break converters expecting the exact width.
3. Negative values need `signed=True`; the current call omits it and may fail or misrepresent data.
4. The except block logs and keeps the original value, hiding format incompatibilities and causing inconsistent behavior (some ints converted, others not).

Can you pls make a note of this problem and fix it?

mssql_python/cursor.py
jahnvi480: desc[1] is returning python object which we have already mapped 
description.append((
                column_name,                           # name 
                self._map_data_type(col["DataType"]),  # type_code
                None,                                  # display_size
                col["ColumnSize"],                     # internal_size
                col["ColumnSize"],                     # precision - should match ColumnSize
                col["DecimalDigits"],                  # scale
                col["Nullable"] == ddbc_sql_const.SQL_NULLABLE.value, # null_ok
            ))
            
 Thus we will be getting uuid.UUID instead of SQL_GUID

bewithgaurav (approved):

PR #293, opened 2025-10-14
FEAT: Complex Datatype support-XML

gargsaumya:

> [AB#38821](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/38821)

### Summary   
  

This pull request adds comprehensive support for the SQL Server `XML` data type to the Python MSSQL driver, ensuring proper handling for insertion, retrieval, batching, and streaming of XML data. It also introduces a suite of tests to verify correct XML behavior, including edge cases like empty, large, and malformed XML values.

### MSSQL Driver Enhancements

* Added support for the `SQL_SS_XML` data type throughout the driver, including binding, fetching, and row size calculations, so that XML columns are handled correctly during data operations. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R24) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R2529-R2534) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R2992) [[4]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3195-R3198) [[5]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3412)
* Updated logic in `FetchMany_wrap` and `FetchAll_wrap` to treat XML columns as LOBs, enabling efficient streaming for large XML values. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3504-R3517) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3626-R3639)

### Test Coverage for XML Support

* Added multiple tests in `tests/test_004_cursor.py` to verify XML handling, including basic insert/fetch, empty/null values, large XML streaming, batch inserts, and error handling for malformed XML input.

### DIFF
--- a/PyPI_Description.md
+++ b/PyPI_Description.md
@@ -39,12 +39,12 @@ PyBind11 provides:
 
 We are currently in **Public Preview**.
 
-## What's new in v0.13.0
+## What's new in v0.13.1
 
-- **Enhanced Batch Operations:** Complete support for UNIQUEIDENTIFIER and DATETIMEOFFSET in `executemany()` operations with automatic type inference, enabling efficient bulk inserts of complex data types including UUIDs and timezone-aware datetimes.
-- **Streaming Large Values:** Robust handling of large objects (NVARCHAR/VARCHAR/VARBINARY(MAX)) in `executemany()` with automatic Data-At-Execution detection and fallback, supporting streaming inserts and fetches for massive datasets.
-- **Improved Cursor Reliability:** Enhanced `cursor.rowcount` accuracy across all fetch operations, including proper handling of empty result sets and consistent behavior for SELECT, INSERT, and UPDATE operations.
-- **Critical Stability Fixes:** Resolved memory leaks with secure token buffer handling, fixed resource cleanup to prevent segmentation faults during Python shutdown, and corrected type inference bugs in batch operations.
+- **Authentication Reliability:** Fixed token handling for Microsoft Entra ID authentication to ensure stable and reliable connections.
+- **Timezone Preservation:** Removed forced UTC conversion for `datetimeoffset` values, preserving original timezone information in Python `datetime` objects for accurate cross-timezone data handling.
+- **Connection Pooling Stability:** Enhanced pool shutdown mechanism with proper tracking to prevent resource leaks and ensure cleanup operations execute reliably.
+- **Predictable Type Handling:** Refined UUID string parameter handling to prevent automatic type coercion, ensuring strings are processed as intended.
 
 For more information, please visit the project link on Github: https://github.com/microsoft/mssql-python
 
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -173,7 +173,7 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
     LOG("Setting SQL attribute");
     SQLPOINTER ptr = nullptr;
     SQLINTEGER length = 0;
-    std::string buffer; // to hold sensitive data temporarily
+    static std::string buffer; // to hold sensitive data temporarily
 
     if (py::isinstance<py::int_>(value)) {
         int intValue = value.cast<int>();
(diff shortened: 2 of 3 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2525,6 +2526,12 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
 				}
                 break;
             }
+            case SQL_SS_XML:
+            {
```
sumitmsft, line 2530: Speaking strictly from documentation point of view. The PR adds support for storing and retrieving XML data as a SQL Server XML datatype, treating it as a Large Object (LOB). This means:

- If a user wishes to insert the contents of an XML file into an XML column, they must first read the file contents into a string (in Python) and then provide that string as a parameter to the driver.
- The driver itself does not directly handle file objects or file streams as input for XML columns. It expects the XML data as a string or bytes-like object.
- There is no evidence in the PR or the broader codebase that the driver will automatically read or stream data directly from a file object for XML columns.

This must also be documentation explicitly in the public wikis

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -21,6 +21,7 @@
 #define SQL_SS_TIMESTAMPOFFSET (-155)
 #define SQL_C_SS_TIMESTAMPOFFSET (0x4001)
 #define MAX_DIGITS_IN_NUMERIC 64
+#define SQL_SS_XML (-152)
```
subrata-ms, line 24: what is this magic number 152 denotes here?
Pls add a comment if possible.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2525,6 +2526,12 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
 				}
                 break;
             }
+            case SQL_SS_XML:
+            {
+                LOG("Streaming XML for column {}", i);
+                row.append(FetchLobColumnData(hStmt, i, SQL_C_WCHAR, true, false));
```
subrata-ms, line 2532: Don't we need to handle (columnSize == SQL_NO_TOTAL || columnSize > 4000) clause like SQL_WLONGVARCHAR case?

gargsaumya: ### Why XML is Always Treated as LOB (No Size-Based Optimization)
XML columns require different handling than regular NVARCHAR columns due to several technical constraints:

1. **Unreliable Size Reporting**

    - **XML columns often return 0, [SQL_NO_TOTAL], or inconsistent values for [columnSize] via SQLDescribeCol from SQL Server**
    - Unlike NVARCHAR(n) with defined max length, XML is **inherently variable-length** with unpredictable sizes

2. Data Characteristics

    - XML is typically large due to markup overhead 
    - A simple element like < data > 123 </ data> is 16 characters, but realistic XML is much larger.

3.  False optimization: The "optimization" for small XML would rarely trigger because:

    - Most XML > 4000 characters
    - **[columnSize]is unreliable for XML**
    - Fallback to streaming happens frequently anyway. So, attempting buffer fetch first, then falling back to streaming is slower than direct streaming
    
cc: @sumitmsft

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: value of 152 is a reserved identifier used in Microsoft SQL Server to represent the XML data type, when working with ODBC drivers

subrata-ms (approved):

sumitmsft (approved):

PR #301, opened 2025-10-23
FIX: Encoding Decoding Github issue

jahnvi480:

> [AB#39049](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39049)

> GitHub Issue: #250 

### Summary   
This pull request introduces robust and secure handling of encoding settings throughout the `mssql_python` codebase, with a particular focus on validating encoding names, enforcing UTF-16 restrictions for certain SQL types, and ensuring safe propagation of encoding parameters between Python and C++ layers. The changes improve both security (by preventing injection attacks) and correctness (by enforcing SQL Server encoding requirements), while also adding better error handling and logging for encoding-related issues.

**Encoding validation and enforcement**

* Enhanced the `_validate_encoding` function in `connection.py` to strictly validate encoding names for security (length and allowed characters) and correctness (must be a valid Python codec), rejecting unsafe or invalid encodings.
* Enforced that only UTF-16 encodings are allowed for `SQL_WCHAR` and `SQL_WMETADATA` types in both `setencoding` and `setdecoding`, with warnings and automatic fallback to `'utf-16le'` if an invalid encoding is provided. 

**Propagation of encoding settings**

* Added helper methods in `cursor.py` (`_get_encoding_settings` and `_get_decoding_settings`) to retrieve encoding and decoding settings from the connection, with error handling and sensible defaults. These settings are now passed to statement execution and data fetching functions.

**C++ layer encoding safety**

* Implemented new helper functions in `ddbc_bindings.cpp` for encoding/decoding strings (`EncodingString`, `DecodingString`), validating encoding names (`is_valid_encoding`), error modes, and extracting encoding settings safely from Python dictionaries. These changes ensure only safe and valid encodings are used in the native layer and provide detailed logging for failures.

**Interface changes for parameter binding**

* Updated the parameter binding interface in the C++ layer to accept encoding settings, enabling correct handling of encoded data during SQL operations.

**Error handling improvements**

* Expanded exception handling in `cursor.py` to include `OperationalError` and `DatabaseError` when retrieving encoding settings, ensuring that only database-related errors are caught and logged, and that safe defaults are used if retrieval fails.

Let me know if you want to walk through how encoding settings flow from Python to C++ or discuss the impact of these changes on query execution and data retrieval.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -1,17 +1,26 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-// INFO|TODO - Note that is file is Windows specific right now. Making it arch agnostic will be
+// INFO|TODO - Note that is file is Windows specific right now. Making it arch
+// agnostic will be
 //             taken up in beta release
 #include "ddbc_bindings.h"
-#include "connection/connection.h"
-#include "connection/connection_pool.h"
 
+#include <algorithm>
 #include <cstdint>
+#include <cstdio>
+#include <filesystem>  // NOLINT(build/c++17)
 #include <iomanip>  // std::setw, std::setfill
 #include <iostream>
+#include <limits>
+#include <memory>
+#include <string>
+#include <unordered_set>
 #include <utility>  // std::forward
-#include <filesystem>
+#include <vector>
+
+#include "connection/connection.h"
+#include "connection/connection_pool.h"
 //-------------------------------------------------------------------------------------------------
 // Macro definitions
 //-------------------------------------------------------------------------------------------------
@@ -35,7 +44,8 @@
 
 // Architecture-specific defines
 #ifndef ARCHITECTURE
-#define ARCHITECTURE "win64"  // Default to win64 if not defined during compilation
+#define ARCHITECTURE \
+    "win64"  // Default to win64 if not defined during compilation
 #endif
 #define DAE_CHUNK_SIZE 8192
 #define SQL_MAX_LOB_SIZE 8000
@@ -53,7 +63,7 @@ struct ParamInfo {
     SQLSMALLINT decimalDigits;
     SQLLEN strLenOrInd = 0;  // Required for DAE
     bool isDAE = false;      // Indicates if we need to stream
-    py::object dataPtr; 
+    py::object dataPtr;
 };
 
 // Mirrors the SQL_NUMERIC_STRUCT. But redefined to replace val char array
@@ -62,17 +72,24 @@ struct ParamInfo {
 struct NumericData {
     SQLCHAR precision;
     SQLSCHAR scale;
-    SQLCHAR sign;  // 1=pos, 0=neg
-    std::string val; // 123.45 -> 12345
+    SQLCHAR sign;     // 1=pos, 0=neg
+    std::string val;  // 123.45 -> 12345
 
-    NumericData() : precision(0), scale(0), sign(0), val(SQL_MAX_NUMERIC_LEN, '\0') {}
+    NumericData()
+        : precision(0), scale(0), sign(0), val(SQL_MAX_NUMERIC_LEN, '\0') {}
 
-    NumericData(SQLCHAR precision, SQLSCHAR scale, SQLCHAR sign, const std::string& valueBytes)
-        : precision(precision), scale(scale), sign(sign), val(SQL_MAX_NUMERIC_LEN, '\0') {
+    NumericData(SQLCHAR precision, SQLSCHAR scale, SQLCHAR sign,
+                const std::string& valueBytes)
+        : precision(precision),
+          scale(scale),
+          sign(sign),
+          val(SQL_MAX_NUMERIC_LEN, '\0') {
         if (valueBytes.size() > SQL_MAX_NUMERIC_LEN) {
-            throw std::runtime_error("NumericData valueBytes size exceeds SQL_MAX_NUMERIC_LEN (16)");
+            throw std::runtime_error(
+                "NumericData valueBytes size exceeds SQL_MAX_NUMERIC_LEN (16)");
         }
-        // Secure copy: bounds already validated, but using std::copy_n for safety
+        // Secure copy: bounds already validated, but using std::copy_n for
+        // safety
         if (valueBytes.size() > 0) {
             std::copy_n(valueBytes.data(), valueBytes.size(), &val[0]);
         }
@@ -80,17 +97,16 @@ struct NumericData {
 };
 
 // Struct to hold the DateTimeOffset structure
-struct DateTimeOffset
-{
-    SQLSMALLINT    year;
-    SQLUSMALLINT   month;
-    SQLUSMALLINT   day;
-    SQLUSMALLINT   hour;
-    SQLUSMALLINT   minute;
-    SQLUSMALLINT   second;
-    SQLUINTEGER    fraction;        // Nanoseconds
-    SQLSMALLINT    timezone_hour;   // Offset hours from UTC
-    SQLSMALLINT    timezone_minute; // Offset minutes from UTC
+struct DateTimeOffset {
+    SQLSMALLINT year;
+    SQLUSMALLINT month;
+    SQLUSMALLINT day;
+    SQLUSMALLINT hour;
+    SQLUSMALLINT minute;
+    SQLUSMALLINT second;
+    SQLUINTEGER fraction;         // Nanoseconds
+    SQLSMALLINT timezone_hour;    // Offset hours from UTC
+    SQLSMALLINT timezone_minute;  // Offset minutes from UTC
 };
 
 // Struct to hold data buffers and indicators for each column
@@ -182,65 +198,62 @@ SQLTablesFunc SQLTables_ptr = nullptr;
 
 SQLDescribeParamFunc SQLDescribeParam_ptr = nullptr;
 
-
-// Encoding String
-static py::bytes EncodingString(const std::string& text, 
-                                const std::string& encoding, 
+// Encoding function with fallback strategy
+static py::bytes EncodingString(const std::string& text,
+                                const std::string& encoding,
                                 const std::string& errors = "strict") {
     try {
         py::gil_scoped_acquire gil;
         py::str unicode_str = py::str(text);
-        
+
         // Direct encoding - let Python handle errors strictly
         py::bytes encoded = unicode_str.attr("encode")(encoding, errors);
         return encoded;
-        
     } catch (const py::error_already_set& e) {
         // Re-raise Python exceptions (UnicodeEncodeError, etc.)
         throw std::runtime_error("Encoding failed: " + std::string(e.what()));
     }
 }
 
-// Decoding String
 static py::str DecodingString(const char* data, size_t length,
-                              const std::string& encoding, 
+                              const std::string& encoding,
                               const std::string& errors = "strict") {
     try {
         py::gil_scoped_acquire gil;
         py::bytes byte_data = py::bytes(data, length);
-        
+
         // Direct decoding - let Python handle errors strictly
         py::str decoded = byte_data.attr("decode")(encoding, errors);
         return decoded;
-        
     } catch (const py::error_already_set& e) {
         // Re-raise Python exceptions (UnicodeDecodeError, etc.)
         throw std::runtime_error("Decoding failed: " + std::string(e.what()));
     }
 }
 
-// Helper function to validate that an encoding string is a legitimate Python codec
-// This prevents injection attacks while allowing all valid encodings
+// Helper function to validate that an encoding string is a legitimate Python
+// codec This prevents injection attacks while allowing all valid encodings
 static bool is_valid_encoding(const std::string& enc) {
     if (enc.empty() || enc.length() > 100) {  // Reasonable length limit
         return false;
     }
-    
-    // Check for potentially dangerous characters that shouldn't be in codec names
+
+    // Check for potentially dangerous characters that shouldn't be in codec
+    // names
     for (char c : enc) {
         if (!std::isalnum(c) && c != '-' && c != '_' && c != '.') {
             return false;  // Reject suspicious characters
         }
     }
-    
+
     // Verify it's a valid Python codec by attempting a test lookup
     try {
         py::gil_scoped_acquire gil;
         py::module_ codecs = py::module_::import("codecs");
-        
+
         // This will raise LookupError if the codec doesn't exist
         codecs.attr("lookup")(enc);
-        
+
         return true;  // Codec exists and is valid
     } catch (const py::error_already_set& e) {
         // Expected: LookupError for invalid codec names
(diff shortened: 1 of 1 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -175,6 +180,108 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+        if (encoding == "utf-8" && errors == "strict") {
+            try {
+                // Try UTF-8 first
+                py::bytes encoded = unicode_str.attr("encode")(encoding, "strict");
+                return encoded;
+            } catch (const py::error_already_set&) {
+                // UTF-8 failed, try latin-1 for Western European characters
+                try {
+                    py::bytes encoded = unicode_str.attr("encode")("latin-1", "strict");
+                    LOG("EncodingString: UTF-8 failed, successfully encoded with latin-1 fallback for text: {}", text.substr(0, 50));
```
Ramudaykumar: Need to ensure that while encoding PII data, we should not log plain text of PII.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -175,6 +180,108 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+        if (encoding == "utf-8" && errors == "strict") {
+            try {
+                // Try UTF-8 first
+                py::str decoded = byte_data.attr("decode")(encoding, "strict");
+                return decoded;
+            } catch (const py::error_already_set&) {
+                // UTF-8 failed, try latin-1 for Western European characters
+                try {
+                    py::str decoded = byte_data.attr("decode")("latin-1", "strict");
+                    LOG("DecodingString: UTF-8 failed, successfully decoded with latin-1 fallback for {} bytes", length);
```
Ramudaykumar: Same here for PII.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -175,6 +180,108 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+
+// Helper function to safely extract encoding settings from Python dict
+static std::pair<std::string, std::string> extract_encoding_settings(const py::dict& settings) {
+    try {
+        std::string encoding = "utf-8";  // Default
+        std::string errors = "strict";   // Default
+        
+        if (settings.contains("encoding") && !settings["encoding"].is_none()) {
+            encoding = settings["encoding"].cast<std::string>();
+        }
```
Ramudaykumar, line 303: This is having potential injection attack scope, as we are not validating encoding against an allowlist. Malicious input could trigger unexpected behavior.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -175,6 +180,108 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+                try {
+                    py::bytes encoded = unicode_str.attr("encode")("latin-1", "strict");
+                    LOG("EncodingString: UTF-8 failed, successfully encoded with latin-1 fallback for text: {}", text.substr(0, 50));
+                    return encoded;
+                } catch (const py::error_already_set&) {
+                    // Both failed, use original approach with error handling
+                    py::bytes encoded = unicode_str.attr("encode")(encoding, errors);
+                    return encoded;
+                }
+            }
```
Ramudaykumar: Nested try-catch blocks for UTF-8 → latin-1 → original encoding can be expensive for large strings. Consider a single fallback path or configurable retry limit.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -175,6 +180,108 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+                // UTF-8 failed, try latin-1 for Western European characters
+                try {
+                    py::str decoded = byte_data.attr("decode")("latin-1", "strict");
+                    LOG("DecodingString: UTF-8 failed, successfully decoded with latin-1 fallback for {} bytes", length);
+                    return decoded;
+                } catch (const py::error_already_set&) {
+                    // Both failed, use original approach with error handling
+                    py::str decoded = byte_data.attr("decode")(encoding, errors);
+                    return decoded;
+                }
```
Ramudaykumar: Nested try-catch blocks for UTF-8 → latin-1 → original encoding can be expensive for large strings. Consider a single fallback path or configurable retry limit.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -175,6 +180,108 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+        if (settings.contains("encoding") && !settings["encoding"].is_none()) {
+            encoding = settings["encoding"].cast<std::string>();
+        }
+        
+        if (settings.contains("errors") && !settings["errors"].is_none()) {
+            errors = settings["errors"].cast<std::string>();
+        }
+        
+        return std::make_pair(encoding, errors);
+    } catch (...) {
```
Ramudaykumar, line 334: Catch-all catch (...) hides root cause. Log the error or rethrow after setting defaults.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -175,6 +180,108 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+        throw std::runtime_error("Encoding error: " + std::string(e.what()));
+    }
+}
+
+static py::str DecodingString(const char* data, size_t length, const std::string& encoding, const std::string& errors = "strict") {
+    try {
+        py::gil_scoped_acquire gil;
+        
+        // Create bytes object from input data
+        py::bytes byte_data = py::bytes(std::string(data, length));
```
Ramudaykumar: py::bytes(std::string(data, length)) copies data unnecessarily. Use py::bytes(data, length) directly to avoid extra allocation.

mssql_python/pybind/ddbc_bindings.cpp
Ramudaykumar: Use this before applying settings["encoding"].

> `
> static bool is_valid_encoding(const std::string& enc) {
>     static const std::unordered_set<std::string> allowed = {"utf-8", "latin-1", "ascii"};
>     return allowed.find(enc) != allowed.end();
> }
> `

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -261,24 +369,56 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
-                    strLenOrIndPtr = AllocateParamBuffer<SQLLEN>(paramBuffers);
-                    *strLenOrIndPtr = SQL_LEN_DATA_AT_EXEC(0);
-                    bufferLength = 0;
+                
+                std::string strValue;
+                
+                // Check if we have encoding settings and this is SQL_C_CHAR (not SQL_C_WCHAR)
+                if (encoding_settings && !encoding_settings.is_none() && 
+                    encoding_settings.contains("ctype") && 
+                    encoding_settings.contains("encoding")) {
```
Ramudaykumar: Malicious input could cause unexpected behavior or crashes. Validate ctype and encoding against an allowlist.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -261,24 +369,56 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
+
+                // Allocate buffer and copy string data
+                size_t bufferSize = strValue.length() + 1;  // +1 for null terminator
+                char* buffer = AllocateParamBufferArray<char>(paramBuffers, bufferSize);
+                
+                if (!buffer) {
+                    ThrowStdException("Failed to allocate buffer for SQL_C_CHAR parameter at index " + std::to_string(paramIndex));
                 }
+                
+                std::memcpy(buffer, strValue.c_str(), strValue.length());
```
Ramudaykumar: this can corrupt memory. Use std::copy_n or memcpy_s (secure alternative) and validate sizes before copying.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1858,9 +2000,41 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
...
+                            } else {
+                                // No encoding settings or SQL_C_BINARY - use default behavior
+                                str = columnValues[i].cast<std::string>();
+                            }
+                            
+                            if (str.size() > info.columnSize) {
                                 ThrowStdException("Input exceeds column size at index " + std::to_string(i));
+                            }
+                            
                             std::memcpy(charArray + i * (info.columnSize + 1), str.c_str(), str.size());
```
Ramudaykumar: Consider using safe memcpy.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: The DecodingString function is actually already secure - it only logs the byte length (length), not the actual decoded data content.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Changed this to only log the length of data returned instead of actual data.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Created helper functions to make sure that there is no scope for injection attack.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Removing the fallback and directly throwing error the way we have in pyodbc

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Removing the fallback and directly throwing error the way we have in pyodbc

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Made the required changes.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Added more checks

mssql_python/pybind/ddbc_bindings.cpp

gargsaumya (dismissed):

mssql_python/connection.py
```diff
@@ -424,6 +441,12 @@ def setencoding(
                 ),
             )
 
+        # Enforce UTF-16 encoding restriction for SQL_WCHAR
+        if ctype == ConstantsDDBC.SQL_WCHAR.value and encoding not in UTF16_ENCODINGS:
+            log('warning', "SQL_WCHAR only supports UTF-16 encodings. Attempted encoding '%s' is not allowed. Using default 'utf-16le' instead.", 
+                sanitize_user_input(encoding))
+            encoding = 'utf-16le'
```
sumitmsft: Instead of rejecting an invalid (or incompatible) encoding/ctype combination, the code logs a warning and silently forces 'utf-16le'. It hides configuration mistakes. A user may think their chosen encoding (e.g. utf-8 with SQL_WCHAR) is in effect. Second, silent mutation can create inconsistencies between logs, user expectations, and downstream behavior (especially if someone inspects `self._encoding_settings` later and sees a different value than what they passed). It complicates debugging subtle data corruption or performance anomalies.

IMO, can we fail fast: raise a ProgrammingError if the provided encoding is incompatible with the chosen ctype or sqltype?

If you wish to keep a permissive mode, make it explicit via a flag (e.g., strict_encoding=True) and only fallback when disabled.

mssql_python/connection.py
```diff
@@ -543,13 +566,26 @@ def setdecoding(
         # Normalize encoding to lowercase for consistency
         encoding = encoding.lower()
 
+        # Enforce UTF-16 encoding restriction for SQL_WCHAR and SQL_WMETADATA
+        if (sqltype == ConstantsDDBC.SQL_WCHAR.value or sqltype == SQL_WMETADATA) and encoding not in UTF16_ENCODINGS:
+            sqltype_name = "SQL_WCHAR" if sqltype == ConstantsDDBC.SQL_WCHAR.value else "SQL_WMETADATA"
+            log('warning', "%s only supports UTF-16 encodings. Attempted encoding '%s' is not allowed. Using default 'utf-16le' instead.", 
+                sqltype_name, sanitize_user_input(encoding))
+            encoding = 'utf-16le'
```
sumitmsft: silent force of 'utf-16le' is happening here as well..

mssql_python/connection.py
```diff
@@ -543,13 +566,26 @@ def setdecoding(
...
             if encoding in UTF16_ENCODINGS:
                 ctype = ConstantsDDBC.SQL_WCHAR.value
             else:
                 ctype = ConstantsDDBC.SQL_CHAR.value
 
+        # Additional validation: if user explicitly sets ctype to SQL_WCHAR but encoding is not UTF-16
+        if ctype == ConstantsDDBC.SQL_WCHAR.value and encoding not in UTF16_ENCODINGS:
+            log('warning', "SQL_WCHAR ctype only supports UTF-16 encodings. Attempted encoding '%s' is not compatible. Using default 'utf-16le' instead.", 
+                sanitize_user_input(encoding))
+            encoding = 'utf-16le'
```
sumitmsft: silent force of 'utf-16le' is happening here too

mssql_python/connection.py
sumitmsft: With a little bit of research using Copilot what I found was:

In Python, the name 'utf-16' is a special, BOM‑aware alias. What that means:

When you encode text with 'utf-16', Python includes a Byte Order Mark (BOM) at the start of the byte sequence (usually FF FE for little‑endian or FE FF for big‑endian).
When you decode bytes with 'utf-16', Python looks at the first two bytes (the BOM) to decide whether the data is little‑endian or big‑endian. If there is no BOM, behavior can vary depending on the implementation (it may assume one endianness or fail depending on context).
By contrast, 'utf-16le' and 'utf-16be' are explicit: no BOM is automatically added, and decoding doesn’t try to auto‑detect—it uses the specified endianness directly.

Treating 'utf-16' as interchangeable with 'utf-16le' (or 'utf-16be') can introduce ambiguity:

1. Downstream native (C++ / ODBC) code often assumes a fixed width of 2 bytes per code unit in a specific endianness (commonly little‑endian).
2. If data is encoded with 'utf-16', a BOM will be prepended. Your length calculations, buffer allocations, or offset logic may not account for that extra initial code unit.
3. When decoding, if you rely on Python’s auto‑detection, you’re delegating correctness to the BOM presence. If upstream bytes arrive without a BOM (e.g., produced by 'utf-16le' encoder but labeled 'utf-16'), the decoder might infer incorrectly or treat the first two bytes as BOM, shifting text.
4. Any mismatch can cause:
- Off‑by‑one truncation (dropping the first character or misinterpreting BOM bytes as data).
- Incorrect size estimation for parameter binding (e.g., allocating too few or too many bytes).
- Subtle corruption when surrogate pairs are processed assuming a different endianness.

Suggestion: Remove 'utf-16' from the set and either (a) require explicit 'utf-16le' / 'utf-16be', or (b) canonicalize 'utf-16' to 'utf-16le' with a clear log message. This makes encoding behavior deterministic and reduces risk of silent corruption.

@bewithgaurav @jahnvi480 Can you confirm this please? Because this could be risky

mssql_python/connection.py
jahnvi480: Removed all the silent fallbacks and have added fast failure instead.

mssql_python/connection.py
jahnvi480: Removed all the silent fallbacks and have added fast failure instead.

mssql_python/connection.py
jahnvi480: Removed all the silent fallbacks and have added fast failure instead.

mssql_python/connection.py
jahnvi480: You are absolutely correct. This is a real risk that could cause data corruption.

Issue Confirmed

`'utf-16'` in Python is **NOT safe** for use with SQL_WCHAR because:

1. **BOM is prepended**: Python's `'utf-16'` codec adds a 2-byte Byte Order Mark (U+FEFF or FF FE)
2. **Buffer size mismatch**: ODBC's `SQL_WCHAR` expects exact UTF-16LE data without BOM
3. **Data corruption**: The BOM could be:
   - Interpreted as actual data (ZERO WIDTH NO-BREAK SPACE character)
   - Cause off-by-one errors in string length calculations
   - Break parameter binding size checks
   - Corrupt surrogate pair handling

I have removed utf-16 from the list

sumitmsft (approved):

PR #304, opened 2025-10-28
FEAT: Perf Improvement

gargsaumya:

> [AB#39793](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39793)

### Summary   
  

1. Large result sets (100K+ rows) improved by up to ~2×
2. Very large result sets (~1.2M rows) improved by 1.4× to 1.7×
3. Complex joins and aggregation workloads improved by ~40–45%
4. General query workloads now operate at parity or better compared to pyodbc

This pull request introduces several performance and correctness improvements to the MSSQL Python driver, focusing on efficient row conversion, cursor behavior, and pybind object handling. The most significant changes include caching output converters and column maps for rows, enforcing forward-only cursor semantics, and optimizing Python type/class lookups in the C++ extension layer.

**Row conversion and cursor improvements:**

* Added caching of column name maps and output converter maps in the `Cursor` class, so that row and column conversions are computed only once per result set, improving performance for large queries. This affects `fetchone`, `fetchmany`, and `fetchall` methods, as well as execution logic. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L124-R131) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L826-R848) [[3]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R1160-R1167) [[4]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L1961-R1994) [[5]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R2038-R2046) [[6]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R2080-R2088)
* Changed the cursor's scroll behavior to explicitly reject absolute positioning for forward-only cursors, and implemented relative scrolling using repeated fetches to match pyodbc's behavior.

**C++ extension (pybind) optimizations:**

* Introduced a `PythonObjectCache` namespace to cache commonly used Python classes (datetime, date, time, decimal, uuid), reducing repeated module imports and attribute lookups throughout the C++ codebase. All parameter binding and result conversion logic now uses this cache. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R37-R98) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L461-R523) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L478-R540) [[4]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L491-R553) [[5]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L535-R597) [[6]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2005-R2067) [[7]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2083-R2146)
* Added caching of the decimal separator in the result set conversion logic to avoid repeated system calls during data retrieval.

**Cursor semantics and configuration:**

* Updated statement execution logic in the C++ layer to always configure cursors as forward-only, ensuring consistent behavior and compatibility with the Python cursor implementation. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1422-R1488) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1559-R1625)

These changes collectively improve performance and reliability when fetching large result sets, ensure correct cursor behavior, and make the C++ extension more efficient in its interaction with Python objects.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -23,7 +23,7 @@ Standalone script that tests real-world queries against AdventureWorks2022 datab
    - Ensure you have a running SQL Server instance.
    - Set the `DB_CONNECTION_STRING` environment variable with the connection string to your database. For example:
      ```bash
-     export DB_CONNECTION_STRING="Server=your_server;Database=your_database;UID=your_user;PWD=your_password;"
+     export DB_CONNECTION_STRING="Server=your_server;Database=AdventureWorks2022;UID=your_user;PWD=your_password;"
      ```
 
 2. **Install Richbench - Benchmarking Tool**:
--- a/benchmarks/perf-benchmarking.py
+++ b/benchmarks/perf-benchmarking.py
@@ -29,6 +29,10 @@ from mssql_python import connect
 # Configuration
 CONN_STR = os.getenv("DB_CONNECTION_STRING")
 
+if not CONN_STR:
+    print("Error: The environment variable DB_CONNECTION_STRING is not set. Please set it to a valid SQL Server connection string and try again.")
+    sys.exit(1)
+
 # Ensure pyodbc connection string has ODBC driver specified
 if CONN_STR and 'Driver=' not in CONN_STR:
     CONN_STR = f"Driver={{ODBC Driver 18 for SQL Server}};{CONN_STR}"


mssql_python/cursor.py
```diff
@@ -1957,11 +1984,14 @@ def fetchone(self) -> Union[None, Row]:
...
 
             self.rowcount = self._next_row_index
-
-            # Create and return a Row object, passing column name map if available
-            column_map = getattr(self, "_column_name_map", None)
-            return Row(self, self.description, row_data, column_map)
-        except Exception as e:  # pylint: disable=broad-exception-caught
+            
+            # Build column map once and cache it
+            if self._cached_column_map is None and self.description:
```
sumitmsft: While the caching of `column_map` and `converter_map` improves performance for large result sets, but we need to guarantee they are invalidated on every transition to a new result set. If the cursor executes a different query or moves to the next result set without clearing `self._cached_column_map` / `self._cached_converter_map`, rows could be built using stale column indices or converters, leading to silent data corruption. 

Please validate this in all the code paths wherever caching\mapping logic is used.

- Reset both caches (set to None) in execute(), executemany(), each metadata initializer, and nextset() when it returns a new set.
- Add tests: (a) same cursor executes two queries with different column orders/types; 
- (b) stored procedure returning multiple result sets; 
- (c) metadata call followed by a normal SELECT.
- Confirm wrapper fetch methods are restored or rewrapped per result set so specialized mappings don’t leak across queries.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1419,11 +1481,11 @@ SQLRETURN SQLExecDirect_wrap(SqlHandlePtr StatementHandle, const std::wstring& Q
         DriverLoader::getInstance().loadDriver();  // Load the driver
     }
 
-    // Ensure statement is scrollable BEFORE executing
+    // Configure forward-only cursor
     if (SQLSetStmtAttr_ptr && StatementHandle && StatementHandle->get()) {
         SQLSetStmtAttr_ptr(StatementHandle->get(),
                            SQL_ATTR_CURSOR_TYPE,
-                           (SQLPOINTER)SQL_CURSOR_STATIC,
+                           (SQLPOINTER)SQL_CURSOR_FORWARD_ONLY,
```
sumitmsft, line 1488: - Let's document in Wikis about this change as this can cause regression for customers using scrollable cursors. 
-

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2080,9 +2142,8 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt,
                     SQLGUID* guidArray = AllocateParamBufferArray<SQLGUID>(tempBuffers, paramSetSize);
                     strLenOrIndArray = AllocateParamBufferArray<SQLLEN>(tempBuffers, paramSetSize);
 
-                    // Get cached UUID class from module-level helper
-                    // This avoids static object destruction issues during Python finalization
-                    py::object uuid_class = py::module_::import("mssql_python.ddbc_bindings").attr("_get_uuid_class")();
+                    // Get cached UUID class
+                    py::object uuid_class = PythonObjectCache::get_uuid_class();
```
sumitmsft, line 2146: There are multiple places where `PythonObjectCache` is used. Can you validate `PythonObjectCache` it doesn't lack explicit thread-safety guarantees, which could lead to race conditions in multi-threaded applications?

Race Conditions During Initialization: Multiple threads could simultaneously attempt to initialize the cache
which could lead to multiple imports of the same module or corrupted cache state

Concurrent Access Patterns: Cached Python objects (datetime, date, time, decimal, uuid classes) are accessed from multiple result conversion threads, while Python's GIL provides some protection, it's not sufficient for all scenarios:
- GIL can be released during C++ operations
- Cache access might occur outside GIL protection
- Reference counting operations need to be atomic

Object Lifetime Issues: Cached Python objects could be deallocated while other threads are accessing them since there may not be reference counting protection in the cache layer. There could also be a risk of accessing freed memory in high-concurrency scenarios

Can you validate these scenarios?

mssql_python/cursor.py
sumitmsft: There are two addtional items to validate:
- cache cleanup in close() method
- Implement proper __del__ method with cache cleanup

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -34,6 +34,68 @@
...
+
+namespace PythonObjectCache {
+    static py::object datetime_class;
+    static py::object date_class; 
+    static py::object time_class;
+    static py::object decimal_class;
+    static py::object uuid_class;
+    static bool cache_initialized = false;
+    
+    void initialize() {
```
sumitmsft, line 46: I do not see PythonObjectCache implementation having robust error handling for module imports and attribute lookups during cache initialization. This could lead to undefined behavior, NULL pointer dereferences, or silent failures in production environments.

sumitmsft: Problems which may be surfaced:
- No try-catch for module imports (datetime, decimal, uuid)
- No validation of attribute lookups (.attr() calls)
- No NULL checks before using cached objects
- Silent failures could cause segfaults in production

bewithgaurav (dismissed):

sumitmsft (dismissed):

bewithgaurav (approved):

PR #307, opened 2025-10-29
FEAT: Improved Connection String handling in Python

### LINKED ISSUE #306: More usable handling of connection string for mssql-python
### Is your feature request related to a problem? Please describe.
Today mssql-python passes the connection string to the underlying implementation i.e. the ODBC driver for SQL server. 

While doing so, the driver exposes the problems with connection string with ODBC.

1. Permissive connection string handling. Unsupported / bad keywords like Thisdoesntexist=FooBar, is accepted and ignored. The problem is that using a connection string parameter which is not supported by ODBC and not seeing it being effective is hard to debug.
2. First instance wins: For long complex connection strings, if the keys are repeated then the first occurrence of the key and its value are honored.
3. Special keywords like Driver / App are a no-op with mssql-python. App is always overridden. 

### Describe the solution you'd like
1. Error out on duplicate keys.
2. Error out if the keys are not supported.
3. Disallow setting of reserved connection string keywords like Driver / App if they are going to be overridden.
4. Allow ODBC style escaping.
5. Don't error out on the first violation, e.g. if a connection string has repeated keys and unsupported keys, then don't error out with a duplicate key error, fixing which would lead to unsupported keys error. If possible, provide a list of all the errors in one go. 
6. The parameters which haven't been tested with the python experience, should not be allowed.

saurabh500:

> [AB#39896](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39896)

> GitHub Issue: #306

### Summary   
Fixes #306 by introducing a connection string parser for mssql-python.

### PR Title Guide

FEAT: Improve connection string parsing, honor the allow list, and improve error messaging.

# Connection String Allowlist Feature - Review Guide

## Overview

This PR introduces a comprehensive connection string validation and allowlist system that validates all ODBC connection parameters before passing them to the driver, improving usability and providing better error messages.

---

## What This Changes

**Before:** Connection strings were passed directly to the ODBC driver with minimal or no validation. Unknown parameters were silently ignored by ODBC, malformed strings caused cryptic ODBC errors.

**After:** All connection strings are parsed, validated against an allowlist of ODBC Driver 18 parameters, and reconstructed with proper escaping. Clear error messages are provided for any issues.

---

## Data Flow

```
User Input (connection string + kwargs)
    ↓
┌─────────────────────────────────────────────┐
│ Connection.__init__()                        │
│ _construct_connection_string()              │
└─────────────────────────────────────────────┘
    ↓
┌─────────────────────────────────────────────┐
│ Step 1: Parse Connection String             │
│ _ConnectionStringParser.parse()             │
│   - Tokenizes key=value pairs               │
│   - Handles braced values: {val}            │
│   - Processes escape sequences: }}-> }      │
│   - Detects syntax errors                   │
│   Output: Dict[str, str]                    │
└─────────────────────────────────────────────┘
    ↓
┌─────────────────────────────────────────────┐
│ Step 2: Validate Against Allowlist          │
│ ConnectionStringAllowList.normalize_keys()   │
│   - Checks unknown parameters               │
│   - Normalizes synonyms (host→Server)       │
│   - Blocks reserved params (Driver, APP)    │
│   - Warns about rejected params             │
│   Output: Dict[str, str] (filtered)         │
└─────────────────────────────────────────────┘
    ↓
┌─────────────────────────────────────────────┐
│ Step 3: Process kwargs                      │
│   - Normalize each kwarg key                │
│   - Block reserved parameters               │
│   - Merge into filtered params              │
│   - kwargs override connection string       │
└─────────────────────────────────────────────┘
    ↓
┌─────────────────────────────────────────────┐
│ Step 4: Build Connection String             │
│ _ConnectionStringBuilder.build()            │
│   - Add Driver (hardcoded)                  │
│   - Add APP=MSSQL-Python                    │
│   - Escape special characters               │
│   - Format: key1=value1;key2=value2;        │
│   Output: str (final connection string)     │
└─────────────────────────────────────────────┘
    ↓
ODBC Driver (validated, safe connection string)
```

---

### **Recommended Review Order**

Review in this sequence to understand the architecture:

#### 1. **Core Components** (understand the building blocks)

1. **`mssql_python/connection_string_parser.py`**  Start here
   - **Purpose:** Parses ODBC connection strings per MS-ODBCSTR spec
   - **Key Classes:** 
     - `ConnectionStringParseError` - Custom exception
     - `_ConnectionStringParser` - Main parser class
   - **Key Methods:**
     - `parse(connection_str)` - Main entry point
     - `_parse_value(str, pos)` - Handles braced values & escaping
   - **What to Look For:**
     - Proper handling of escape sequences (`{{`, `}}`)
     - Error collection and batch reporting
     - Edge cases: empty strings, whitespace, special chars

2. **`mssql_python/connection_string_allowlist.py`** Critical
   - **Purpose:** Validates parameters against ODBC Driver 18 keywords
   - **Key Classes:**
     - `ConnectionStringAllowList` - Singleton allowlist manager
   - **Key Methods:**
     - `normalize_key(key)` - Maps synonyms to canonical names
     - `filter_params(params)` - Validates and filters parameters
   - **What to Look For:**
     - Completeness of allowlist (compare with ODBC docs)
     - Synonym mappings (host→Server, user→UID, etc.)
     - Reserved parameter handling (Driver, APP)

3. **`mssql_python/connection_string_builder.py`**
   - **Purpose:** Safely constructs connection strings with escaping
   - **Key Classes:**
     - `_ConnectionStringBuilder` - Builder with escape logic
   - **Key Methods:**
     - `add_param(key, value)` - Adds a parameter
     - `_needs_braces(value)` - Determines if braces needed
     - `_escape_value(value)` - Escapes special characters
     - `build()` - Constructs final string
   - **What to Look For:**
     - Correct escaping logic (`;`, `{`, `}`, `=`)
     - Proper brace placement
     - Semicolon formatting

#### 2. **Integration** (see how it fits together)

4. **`mssql_python/connection.py`**  Integration point
   - **Modified Method:** `_construct_connection_string()`
   - **What Changed:**
     - Lines 241-303: New implementation replacing old concat logic
     - Now uses parser-> filter-> builder pipeline
     - Handles kwargs with allowlist validation
     - Raises ValueError for reserved parameters
   - **What to Look For:**
     - Error handling for parse failures
     - kwargs override behavior
     - Reserved parameter rejection
     - Logging of warnings

5. **`mssql_python/__init__.py`**
   - **What Changed:**
     - Added export: `ConnectionStringParseError`
   - **What to Look For:**
     - Proper exception exposure for users

#### 3. **Tests** (validate functionality)

6. **`tests/test_010_connection_string_parser.py`**  Parser tests
   - **Coverage:** ~40 test cases
   - **Test Categories:**
     - Valid parsing scenarios
     - Braced value handling
     - Escape sequence processing
     - Error detection and reporting
     - Edge cases (empty, whitespace, unicode)
   - **What to Look For:**
     - Test coverage of MS-ODBCSTR spec
     - Error message clarity
     - Edge case handling

7. **`tests/test_011_connection_string_allowlist.py`**  Allowlist tests
   - **Coverage:** ~25 test cases
   - **Test Categories:**
     - Key normalization (synonyms)
     - Parameter filtering
     - Reserved parameter blocking
     - Unknown parameter detection
   - **What to Look For:**
     - All ODBC parameters tested
     - Synonym mappings validated
     - Security (Driver/APP blocking)

8. **`tests/test_012_connection_string_integration.py`**  Integration tests
   - **Coverage:** ~20 test cases
   - **Test Categories:**
     - End-to-end parsing-> filtering-> building
     - Real connection string scenarios
     - Error propagation from connect() API
     - kwargs override behavior
   - **What to Look For:**
     - Real-world usage patterns
     - Error messages match user expectations
     - Backward compatibility where possible

9. **`tests/test_003_connection.py`** (Updated)
   - **What Changed:**
     - Updated assertions in `test_construct_connection_string()`
     - Updated assertions in `test_connection_string_with_attrs_before()`
     - Updated assertions in `test_connection_string_with_odbc_param()`
   - **What to Look For:**
     - Assertions match new builder output format
     - No semicolons in middle of assertions (builder handles formatting)

10. **`tests/test_006_exceptions.py`** (Updated)
    - **What Changed:**
      - `test_connection_error()` now expects `ConnectionStringParseError`
      - Updated error message assertions
    - **What to Look For:**
      - Proper exception type changes
      - Error messages are helpful

#### 4. **Documentation** (understand design decisions)

11. **`docs/connection_string_allow_list_design.md`**  Read this
    - **Content:**
      - Design rationale and motivation
      - Architecture decisions
      - Security considerations
      - Future enhancements
    - **What to Look For:**
      - Justification for approach
      - Trade-offs discussed
      - Security implications understood

12. **`docs/parser_state_machine.md`**
    - **Content:**
      - Detailed state machine for parser
      - Character-by-character processing flow
      - Error handling states
      - Escape sequence examples
    - **What to Look For:**
      - Parser logic is well-documented
      - Edge cases covered in examples

---

### 1. **Security**
- **Reserved Parameters:** Verify Driver and APP cannot be set by users
- **Allowlist Completeness:** Check all ODBC Driver 18 params are included
- **Escape Handling:** Ensure no injection via special characters

### 2. **Error Handling** 
- **Error Messages:** Are they clear and actionable?
- **Error Collection:** Multiple errors reported together?
- **Backward Compatibility:** Do meaningful errors replace cryptic ODBC errors?

### 3. **Performance** 
- **Parsing Overhead:** There is no string split being used during parsing. Hence the cost of multiple allocation of strings is avoided.
- **No Regex in Hot Path:** Parser uses character-by-character processing. Regex have been known to cause problems and its advisable to stay away from them. 
- **Allowlist Lookup:** O(1) dict lookups, minimal overhead

### 4. **Correctness** 
- **Synonym Handling:** All common aliases map correctly
- **Case Insensitivity:** Keys normalized consistently

---

### Test Coverage Map

| Component | Test File | Key Scenarios |
|-----------|-----------|---------------|
| Parser | `test_010_connection_string_parser.py` | Syntax, escaping, errors |
| Allowlist | `test_011_connection_string_allowlist.py` | Validation, normalization |
| Builder | `test_012_connection_string_integration.py` | Escaping, formatting |
| Integration | `test_012_connection_string_integration.py` | End-to-end flows |
| Connection | `test_003_connection.py` | Connection string construction |
| Exceptions | `test_006_exceptions.py` | Error propagation |

## Behavior Changes

Should be aware of these behavioral changes:

### 1. **Unknown Parameters Now Raise Errors**
**Before:**
```python
connect("Server=localhost;FakeParam=value")  # Silently ignored
```

**After:**
```python
connect("Server=localhost;FakeParam=value")
# Raises: ConnectionStringParseError: Unknown keyword 'FakeParam' is not recognized
```

### 2. **Malformed Strings Caught Early**
**Before:**
```python
connect("ServerLocalhost")  # ODBC error later
```

**After:**
```python
connect("ServerLocalhost")
# Raises: ConnectionStringParseError: Incomplete specification: keyword 'ServerLocalhost' has no value
```

### 3. **Reserved Parameters Blocked**
**Before:**
```python
connect("Server=localhost;Driver={SQL Server}")  # Maybe ignored
```

**After:**
```python
connect("Server=localhost;Driver={SQL Server}")
# Raises: ValueError: Connection parameter 'Driver' is reserved and controlled by the driver
```

---

### ODBC Connection String Format
```
key1=value1;key2=value2;key3={val;ue}
```

### Braced Values
Used when value contains semicolons or special characters:
```
PWD={my;pass;word} -> Password is "my;pass;word"
```

### Escape Sequences
- `{{`-> `{` (escaped left brace)
- `}}`-> `}` (escaped right brace)

Example:
```
PWD={a}}b{{c} -> Password is "a}b{c"
```

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -8,6 +8,7 @@
 #include "connection/connection_pool.h"
 
 #include <cstdint>
+#include <cstring>  // For std::memcpy
 #include <iomanip>  // std::setw, std::setfill
 #include <iostream>
 #include <utility>  // std::forward
@@ -2463,10 +2464,11 @@ static py::object FetchLobColumnData(SQLHSTMT hStmt,
             } else {
                 // Wide characters
                 size_t wcharSize = sizeof(SQLWCHAR);
-                if (bytesRead >= wcharSize) {
-                    auto sqlwBuf = reinterpret_cast<const SQLWCHAR*>(chunk.data());
+                if (bytesRead >= wcharSize && (bytesRead % wcharSize == 0)) {
                     size_t wcharCount = bytesRead / wcharSize;
-                    while (wcharCount > 0 && sqlwBuf[wcharCount - 1] == 0) {
+                    std::vector<SQLWCHAR> alignedBuf(wcharCount);
+                    std::memcpy(alignedBuf.data(), chunk.data(), bytesRead);
+                    while (wcharCount > 0 && alignedBuf[wcharCount - 1] == 0) {
                         --wcharCount;
                         bytesRead -= wcharSize;
                     }
@@ -2495,14 +2497,18 @@ static py::object FetchLobColumnData(SQLHSTMT hStmt,
     }
     if (isWideChar) {
 #if defined(_WIN32)
-        std::wstring wstr(reinterpret_cast<const wchar_t*>(buffer.data()), buffer.size() / sizeof(wchar_t));
+        size_t wcharCount = buffer.size() / sizeof(wchar_t);
+        std::vector<wchar_t> alignedBuf(wcharCount);
+        std::memcpy(alignedBuf.data(), buffer.data(), buffer.size());
+        std::wstring wstr(alignedBuf.data(), wcharCount);
         std::string utf8str = WideToUTF8(wstr);
         return py::str(utf8str);
 #else
         // Linux/macOS handling
         size_t wcharCount = buffer.size() / sizeof(SQLWCHAR);
-        const SQLWCHAR* sqlwBuf = reinterpret_cast<const SQLWCHAR*>(buffer.data());
-        std::wstring wstr = SQLWCHARToWString(sqlwBuf, wcharCount);
+        std::vector<SQLWCHAR> alignedBuf(wcharCount);
+        std::memcpy(alignedBuf.data(), buffer.data(), buffer.size());
+        std::wstring wstr = SQLWCHARToWString(alignedBuf.data(), wcharCount);
         std::string utf8str = WideToUTF8(wstr);
         return py::str(utf8str);
 #endif
@@ -2623,8 +2629,7 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
                             uint64_t numCharsInData = dataLen / sizeof(SQLWCHAR);
                             if (numCharsInData < dataBuffer.size()) {
 #if defined(__APPLE__) || defined(__linux__)
-                                const SQLWCHAR* sqlwBuf = reinterpret_cast<const SQLWCHAR*>(dataBuffer.data());
-                                std::wstring wstr = SQLWCHARToWString(sqlwBuf, numCharsInData);
+                                std::wstring wstr = SQLWCHARToWString(dataBuffer.data(), numCharsInData);
                                 std::string utf8str = WideToUTF8(wstr);
                                 row.append(py::str(utf8str));
 #else


mssql_python/constants.py
```diff
@@ -443,3 +444,164 @@ def get_attribute_set_timing(attribute):
...
+
+
+# Import RESERVED_PARAMETERS from parser module to maintain single source of truth
+def _get_reserved_parameters():
+    """Lazy import to avoid circular dependency."""
+    from mssql_python.connection_string_parser import RESERVED_PARAMETERS
+    return RESERVED_PARAMETERS
+
+
+class _ConnectionStringAllowList:
```
bewithgaurav: I see that the ConnectionStringAllowList is now a part of constants.py, the below comment is in continuation of the same in lines of re-structuring:
Currently, `_ConnectionStringAllowList` is implemented as a class with only class methods and static methods, and no instance state. In Python, when a class doesn't need instance state or inheritance, it's generally more idiomatic to use module-level functions and constants instead.
Also, it'll be better to move `RESERVED_PARAMETERS` in this file to remove the lazy import circular dependency block, making constants.py the ultimate source of truth for all type constants

**Current structure:**
```python
class _ConnectionStringAllowList:
    ALLOWED_PARAMS = {...}
    
    @classmethod
    def normalize_key(cls, key: str) -> Optional[str]:
        return cls.ALLOWED_PARAMS.get(key_lower)
    
    @staticmethod
    def _normalize_params(params: Dict[str, str], ...):
        # ...
```

**Suggested refactoring:**

Consider separating **data** (constants) from **logic** (validation functions) for better separation of concerns:

**Step 1: Keep only constants inside `constants.py` (data only):**
```python
# mssql_python/constants.py
# Pure data/configuration - no logic

# Internal connection string constants (underscore prefix = internal use only)
_RESERVED_CONNECTION_STRING_PARAMS = ('Driver', 'APP')

_ALLOWED_CONNECTION_STRING_PARAMS = {
    'server': 'Server',
    'address': 'Server',
    'uid': 'UID',
    'pwd': 'PWD',
    # ... etc (same dictionary content, just not inside a class)
}
```

**Step 2: Move validation logic to `connection_string_parser.py` (where it's used):**
```python
# mssql_python/connection_string_parser.py
# All connection string parsing and validation logic together

from mssql_python.constants import (
    _RESERVED_CONNECTION_STRING_PARAMS,
    _ALLOWED_CONNECTION_STRING_PARAMS
)

def _normalize_connection_string_key(key: str) -> Optional[str]:
    """Normalize a parameter key to its canonical form. (Internal use only)"""
    key_lower = key.lower().strip()
    return _ALLOWED_CONNECTION_STRING_PARAMS.get(key_lower)

def _validate_and_normalize_params(params: Dict[str, str], warn_rejected: bool = True) -> Dict[str, str]:
    """Validate and normalize parameters against the allow-list. (Internal use only)"""
    # ... implementation (moved from _ConnectionStringAllowList._normalize_params)
    filtered = {}
    for key, value in params.items():
        normalized_key = _normalize_connection_string_key(key)
        if normalized_key and normalized_key not in _RESERVED_CONNECTION_STRING_PARAMS:
            filtered[normalized_key] = value
    return filtered

class _ConnectionStringParser:
    """Parser for ODBC connection strings."""
    # ... existing parser implementation
```
why I made this suggestion:
- all connection string validation lives in one place (`connection_string_parser.py`)
- constants.py contains just data, no logic
- no more lazy import

mssql_python/constants.py
```diff
@@ -443,3 +444,164 @@ def get_attribute_set_timing(attribute):
         AttributeSetTime: When the attribute can be set
     """
     return ATTRIBUTE_SET_TIMING.get(attribute, AttributeSetTime.AFTER_ONLY)
+
+
+# Import RESERVED_PARAMETERS from parser module to maintain single source of truth
```
bewithgaurav: As suggested in the other comment, keeping `RESERVED_PARAMETERS` in this file itself would be a better choice, since parser module is technically not a single source of truth as of now.  Whenever lazy imports are needed to avoid circular dependencies, it often indicates that dependents should be co-located.

tests/test_012_connection_string_integration.py
```diff
@@ -0,0 +1,646 @@
...
+        # Multiple empty values
+        with pytest.raises(ConnectionStringParseError) as exc_info:
+            connect("Server=;Database=mydb;PWD=")
+        
+        errors = exc_info.value.errors
+        assert len(errors) >= 2
+        assert any("Empty value for keyword 'server'" in err for err in errors)
+        assert any("Empty value for keyword 'pwd'" in err for err in errors)
+
+
```
bewithgaurav, line 638: just checked for OCD purposes, adding these tests should cover up the remaining lines reported in codecoverage comment:
- `"Server=   localhost;Database=    mydb"` (excessive whitespace)
- `"Server=\t\tlocalhost;PWD=\t{pass}"`

mssql_python/constants.py
```diff
@@ -443,3 +444,164 @@ def get_attribute_set_timing(attribute):
...
+            Driver and APP parameters are filtered here but will be set by
+            the driver in _construct_connection_string to maintain control.
+        """
+        # Import here to avoid circular dependency issues
+        try:
+            from mssql_python.logging_config import get_logger
+            from mssql_python.helpers import sanitize_user_input
+            logger = get_logger()
+        except ImportError:
+            logger = None
```
bewithgaurav: As suggested in the restructuring comment, this whole method would be moved - and hence we wouldn't be needing logger at all in `constants.py` - and logger can be used readily in `connection_string_parser.py`

If not, this might be a problem since we should not set logger as none in case of circular dependencies. We should remove the try: except: block in imports and fix circular dependency imports, if any  - by streamlining modules.
defensive block in this case would render the logger of no use, which might be a problem in future - siliently skipping failures.

bewithgaurav (changes requested):
Added some restructuring comments which might streamline circular imports etc.
Logic lgtm and test coverage for all scenarios is super nice!

tests/test_012_connection_string_integration.py
saurabh500: Haha. Nice!!

mssql_python/__init__.py
```diff
@@ -25,6 +25,9 @@
     NotSupportedError,
 )
 
+# Connection string parser exceptions
```
jahnvi480: Instead of adding this new line can you please append in the above list on line 15

jahnvi480: Also please add the new files and classes that you have added in this folder

mssql_python/connection.py
```diff
@@ -242,39 +242,66 @@ def _construct_connection_string(
...
 
         Returns:
-            str: The constructed connection string.
+            str: The constructed and validated connection string.
         """
-        # Add the driver attribute to the connection string
-        conn_str = add_driver_to_connection_str(connection_str)
-
-        # Add additional key-value pairs to the connection string
+        from mssql_python.connection_string_parser import _ConnectionStringParser, RESERVED_PARAMETERS
```
jahnvi480: Please add all the import statements at the start of the file

mssql_python/connection_string_builder.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    
+    def __init__(self, initial_params: Optional[Dict[str, str]] = None):
+        """
+        Initialize the builder with optional initial parameters.
+        
+        Args:
+            initial_params: Dictionary of initial connection parameters
+        """
+        self._params: Dict[str, str] = initial_params.copy() if initial_params else {}
```
jahnvi480, line 29: Why are we creating a copy instead of directly assigning the value of initial_params?

mssql_python/connection_string_builder.py
```diff
@@ -0,0 +1,113 @@
...
+            
+        Note:
+            - Driver parameter is placed first
+            - Other parameters are sorted for consistency
+            - Values are escaped if they contain special characters
+        """
+        parts = []
+        
+        # Build in specific order: Driver first, then others
+        if 'Driver' in self._params:
```
jahnvi480: What if we find 'driver' or 'DRIVER' or 'dRiver', this if statement will become false.
Can we make all keys to smallercase or uppercase before comparing?

mssql_python/connection_string_parser.py
```diff
@@ -0,0 +1,306 @@
...
+    
+    Validates connection strings and raises errors for:
+    - Unknown/unrecognized keywords
+    - Duplicate keywords
+    - Incomplete specifications (keyword with no value)
+    
+    Reference: https://learn.microsoft.com/en-us/openspecs/sql_server_protocols/ms-odbcstr/55953f0e-2d30-4ad4-8e56-b4207e491409
+    """
+    
+    def __init__(self, allowlist=None):
```
jahnvi480: Please make sure that we have fully typed methods

mssql_python/connection_string_parser.py
```diff
@@ -0,0 +1,306 @@
...
+        seen_keys = {}  # Maps normalized key -> first occurrence position
+        
+        # Track current position in the string
+        current_pos = 0
+        str_len = len(connection_str)
+        
+        # Main parsing loop
+        while current_pos < str_len:
+            # Skip leading whitespace and semicolons
+            while current_pos < str_len and connection_str[current_pos] in ' \t;':
```
jahnvi480, line 176: Can we use .strip() to remove leading whitespaces

jahnvi480 (changes requested):
Left few comments, overall the logic looks good to me.

mssql_python/connection_string_builder.py
saurabh500: No, this wont happen. The builder is called after the mssql-python driver adds the "Driver" keyword to the dictionary. 
The "Driver" with casing is case sensitive. 

While case-sensitivity is not important, since "Driver" is an internal keyword, I will add it to constants.py and make sure the same constant string is referenced everywhere.

mssql_python/connection_string_parser.py
saurabh500: In the parser, I have avoided usign string library functions as much as possible, since any of these function invocation, ends up creating a new string. strip() will create a new string with stripped whitespaces.

mssql_python/connection_string_parser.py
saurabh500: Thanks. Makes sense.

mssql_python/__init__.py
saurabh500: Makese sense.

mssql_python/connection_string_builder.py
saurabh500:
```
# Step 5: Add Driver and APP parameters (always controlled by the driver)
# These maintain existing behavior: Driver is always hardcoded, APP is always MSSQL-Python
builder.add_param('Driver', 'ODBC Driver 18 for SQL Server')
builder.add_param('APP', 'MSSQL-Python')

# Step 6: Build final string
conn_str = builder.build()
```

mssql_python/connection_string_builder.py
saurabh500: The rest of the APIs on the builder mutate the dictionary. Hence the copy was created. 
If there is a perf problem due to this single copy, we can revisit this change. Connection String builders typically end up copying the input (like in C# ) so tha they can independently work on them.

mssql_python/connection_string_parser.py
```diff
@@ -0,0 +1,388 @@
...
+            Driver and APP parameters are filtered here but will be set by
+            the driver in _construct_connection_string to maintain control.
+        """
+        # Import here to avoid circular dependency issues
+        try:
+            from mssql_python.logging_config import get_logger
+            from mssql_python.helpers import sanitize_user_input
+            logger = get_logger()
+        except ImportError:
+            logger = None
```
bewithgaurav: Is the circular dependency actually happening here? tried it myself - imports work just fine since helpers and logging_config are independent of connection string modules.
we should remove this block and move the imports to top of the file, since keeping logger as None is not recommended.

mssql_python/connection_string_parser.py
```diff
@@ -0,0 +1,388 @@
...
+                    
+                # Parameter is allowed
+                filtered[normalized_key] = value
+            else:
+                # Parameter is not in allow-list
+                # Note: In normal flow, this should be empty since parser validates first
+                rejected.append(key)
+        
+        # Log all rejected parameters together if any were found
+        if rejected and warn_rejected and logger:
```
bewithgaurav: can skip None checking for logger, since it is always declared by default, and not used when logging is not turned on by the user

bewithgaurav (changes requested):
one minor import flow change in the connection string parser file, else lgtm

mssql_python/connection_string_parser.py
saurabh500: This comment makes sense. 
When I had initially written this, normalize params was a member function, and it was in ConnectionStringAllowedParams class.
As I refactored, I didn't realize that there isn't really any circular dependency issue anymore. 

Thanks for pointing out.

saurabh500: I copy pasted the function around, without looking into any semantic change to its content. :D

tests/test_012_connection_string_integration.py
bewithgaurav: btw in this we got one part covered, last test recommendation is 
`"Server=localhost;BadKeyword more_content;Database=mydb"`
this should seal the deal for coverage

mssql_python/connection_string_parser.py
bewithgaurav: happens a lot! 
in my experience GH copilot usually gets EXTRA careful of circular imports most of the times - requires manual intervention.

mssql_python/connection_string_parser.py
saurabh500: @bewithgaurav but the get_logger() can always return a None, isn't it? 
I see similar checks in other parts of the code in the repo. i.e. get_logger, check if not None, then log.

Shouldn't I simply be using `helpers.log('warning', f"Connection string parameters not in allow-list and will be ignored: {', '.join(safe_keys)}")`
The helper has the log method which takes the log level as well as the string to log. It checks if the logger is None or not, and then uses the logger.

bewithgaurav: yes - you're correct, I hadn't properly taken a look at the logger line.
As you said, you can import `log` from helpers and use it as is across the code.

saurabh500: @bewithgaurav  took care of the suggestions. I had to change tests to mock the logger in helper module. 
The resultant code is now simpler.

PR #312, opened 2025-10-31
FEAT: Comprehensive Logging Framework with Python-C++ Bridge & Instrumentation

bewithgaurav:

> [AB#39180](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39180)
> [AB#40126](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/40126)

### Summary   
  
# Logging Framework Implementation

## Overview
Implements a comprehensive logging system for mssql-python with both Python and C++ components, replacing the previous fragmented approach with a unified, performance-optimized solution.

### 🎯 Core Logging System
- **Unified Python Logger** (`mssql_python/logging.py`): Single DEBUG-level logger with file/stdout/both output modes
- **C++ Logger Bridge** (`mssql_python/pybind/logger_bridge.cpp`): High-performance bridge from C++ to Python logging with zero overhead when disabled
- **Simplified API**: `setup_logging(output='file'|'stdout'|'both')` - all-or-nothing DEBUG logging for troubleshooting

### 🔧 Technical Improvements
- **Atomic level checks**: Lock-free reads in C++ for minimal performance impact
- **Cached Python logger**: Import once, reuse everywhere
- **Lazy message formatting**: Only formats strings when logging is enabled
- **GIL-aware**: Proper Python GIL management in C++ bridge
- **Thread-safe**: Concurrent access supported via mutex protection

### 📝 Logging Coverage
- Replaced all `warnings.warn()` calls with proper logging
- Added strategic LOG statements at architectural boundaries (connections, transactions, driver loading)
- ~150 diagnostic LOG statements across C++ codebase for production debugging

### 🧪 Testing & Quality
- Updated all tests to use new logging system
- Fixed test bug: `rownumber` assertion (was expecting `None`, should be `-1`)
- DevSkim security: Added ignore directives for required vsnprintf usage
- Type safety: Replaced snprintf with std::ostringstream where appropriate

### 📚 Documentation
- Comprehensive learnings documentation in `learnings/` directory
- Migration guide for coverage logging (deferred to future PR due to performance)
- Philosophy: Logging is for troubleshooting, not production monitoring

## Performance Impact
- **When disabled** (default): Zero overhead via early returns
- **When enabled**: Massive overhead for comprehensive DEBUG logging (Not recommended to use in prod)
- Designed for temporary enablement during issue investigation

## Breaking Changes
- ⚠️ Removed old `logging_config.py` and `get_logger()` API
- ✅ New API: `setup_logging(output='file')` for DEBUG logging
- ✅ All functionality preserved, just different interface

## Testing
- ✅ All existing tests pass with new logging system
- ✅ Coverage maintained at 77%+ (LOG statements excluded, see comment below)
- ✅ Validated on Windows, macOS, Linux, RHEL, Alpine (x86_64 + ARM64)

## Related Work
- Addresses feedback from security review about logging practices
- Foundation for future observability improvements
- Documented roadmap for coverage improvements in separate PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,32 +86,6 @@ The driver offers a suite of Pythonic enhancements that streamline database inte
 ### Connection Pooling
  
 The Microsoft mssql_python driver provides built-in support for connection pooling, which helps improve performance and scalability by reusing active database connections instead of creating a new connection for every request. This feature is enabled by default. For more information, refer [Connection Pooling Wiki](https://github.com/microsoft/mssql-python/wiki/Connection#connection-pooling).
-
-### Logging
-
-The driver includes a comprehensive logging system with JDBC-style custom log levels (FINEST, FINER, FINE) for detailed SQL diagnostics. Features include:
-
-- **Custom Log Levels**: FINEST (most detailed), FINER, FINE for granular SQL operation logging
-- **Automatic Sanitization**: Passwords and sensitive data automatically redacted in logs
-- **Trace IDs**: Unique identifiers for tracking related operations
-- **File Rotation**: Automatic log file rotation to prevent disk space issues
-- **Thread-Safe**: Safe for multi-threaded applications
-- **Flexible Logging**: Custom log file paths, rotation, and output control
-
-Quick example:
-
-```python
-from mssql_python import logger, FINE
-
-# Enable detailed SQL logging
-logger.setLevel(FINE)
-logger.enable_file_logging(log_dir='./logs')
-
-# Use the driver - all SQL operations will be logged
-conn = mssql_python.connect(connection_string)
-```
-
-For complete logging documentation, see [LOGGING.md](LOGGING.md).
  
 ## Getting Started Examples
 Connect to SQL Server and execute a simple query:


mssql_python/logging.py
```diff
@@ -0,0 +1,517 @@
...
+                    cls._instance = super(MSSQLLogger, cls).__new__(cls)
+        return cls._instance
+    
+    def __init__(self):
+        """Initialize the logger (only once)"""
+        # Skip if already initialized
+        if hasattr(self, '_initialized'):
+            return
+        
+        self._initialized = True
```
sumitmsft: `__init__` is called EVERY time `MSSQLLogger()` is invoked
The check `hasattr(self, '_initialized')` and assignment are NOT atomic
If two threads see `_initialized` as False simultaneously - would it cause double initialization and hence race condition? Pls verify

mssql_python/logging.py
```diff
@@ -0,0 +1,517 @@
...
+    def _setup_handlers(self):
+        """
+        Setup handlers based on output mode.
+        Creates file handler and/or stdout handler as needed.
+        """
+        # Clear any existing handlers
+        if self._logger.handlers:
+            for handler in self._logger.handlers[:]:
+                handler.close()
+                self._logger.removeHandler(handler)
```
sumitmsft: `_setup_handlers()` closes and removes handlers while other threads may be logging
If Thread A logs while Thread B removes the handler - than there is a potential for race condition. 
It can cause lost log messages, exceptions, file handle leaks

pls validate

mssql_python/logging.py
```diff
@@ -0,0 +1,517 @@
...
+        if mode not in (FILE, STDOUT, BOTH):
+            raise ValueError(
+                f"Invalid output mode: {mode}. "
+                f"Must be one of: {FILE}, {STDOUT}, {BOTH}"
+            )
+        self._output_mode = mode
+        
+        # Only reconfigure if handlers were already initialized
+        if self._handlers_initialized:
+            self._reconfigure_handlers()
```
sumitmsft, line 539: Changing output mode above calls `_reconfigure_handlers() ` which closes handlers
Do we have synchronization with threads actively logging? Guess no. 
We need to validate if it can cause crashes when thread logs to a closed handler

mssql_python/logging.py
```diff
@@ -0,0 +1,517 @@
...
+                timestamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+                pid = os.getpid()
+                self._log_file = os.path.join(
+                    log_dir,
+                    f"mssql_python_trace_{timestamp}_{pid}.log"
+                )
+            
+            # Create rotating file handler (512MB, 5 backups)
+            # Use UTF-8 encoding for unicode support on all platforms
+            self._file_handler = RotatingFileHandler(
```
sumitmsft, line 203: `RotatingFileHandler` automatically rotates files at 512MB. If rotation happens during concurrent writes, files may not close properly. Do we have any mechanism to cleanup these file handles?

mssql_python/logging.py
```diff
@@ -0,0 +1,517 @@
...
+        # Add prefix if requested (only after level check)
+        if add_prefix:
+            msg = f"[Python] {msg}"
+        
+        # Format message with args if provided
+        if args:
+            msg = msg % args
+        
+        # Log the message (no args since already formatted)
+        self._logger.log(level, msg, **kwargs)
```
sumitmsft: If `msg % args` fails (wrong number of args, wrong type), exception propagates. If disk is full, `_logger.log()` raises `IOError`. In these cases will Logger crashes the application instead of failing silently? Ideally Logging should NEVER crash the app. Can we validate this?

sumitmsft: Probably write to stderr as last resort and silently pass.

mssql_python/logging.py
```diff
@@ -0,0 +1,517 @@
...
+            os_info = platform.platform()
+            
+            # Build header comment line
+            header_line = f"# MSSQL-Python Driver Log | Script: {script_name} | PID: {pid} | Log Level: DEBUG | Python: {python_version} | Driver: {driver_version} | Start: {start_time} | OS: {os_info}\n"
+            
+            # CSV column headers
+            csv_header = "Timestamp, ThreadID, Level, Location, Source, Message\n"
+            
+            # Write directly to file (bypass formatter)
+            with open(self._log_file, 'a') as f:
```
sumitmsft, line 322: What if header write fails (disk full, permissions, file deleted)? We should let the Failure pass silently.
In this case let user know why logs have no header. We can log to stderr.

mssql_python/logging.py
bewithgaurav: you're right, the non-atomic `hasattr(self, '_initialized')` check will cause potential double initialization when multiple threads invoke `MSSQLLogger()` simultaneously.

**Fix to be applied** (in logging.py itself):
- Adding separate `_init_lock` (class-level Lock)
- Wrapping the `hasattr` check and assignment in atomic block
- Prevents simultaneous initialization by multiple threads

**Validating**: `test_concurrent_initialization_no_double_init` passes - 10 threads cannot cause double initialization.

mssql_python/logging.py
bewithgaurav: valid problem, adding a fix and testcase for this

mssql_python/logging.py
bewithgaurav: this will get covered as a part of above two fixes and detailed tests will be added

mssql_python/logging.py
bewithgaurav: adding an atexit block to clean this up

mssql_python/logging.py
bewithgaurav: yes, will add a silent pass with stderr output for these scenarios

mssql_python/logging.py
bewithgaurav: will be adding this as a part of logging write failure exceptions

mssql_python/pybind/logger_bridge.cpp
```diff
@@ -0,0 +1,210 @@
...
+    va_end(args);
+    
+    // Extract filename from path
+    const char* filename = extractFilename(file);
+    
+    // Format the complete log message with [DDBC] prefix for CSV parsing
+    // File and line number are handled by the Python formatter (in Location column)
+    // std::snprintf is safe: always null-terminates, never overflows buffer
+    // DevSkim warning is false positive - this is the recommended safe alternative
+    char complete_message[4096];
```
sumitmsft: If `message.c_str()` is longer than 4090 bytes, it's silently truncated.. I don't see any warning to the user that log data was lost. In troubleshooting scenarios, critical diagnostic info may be missing. Can we atleast warn the customer?

mssql_python/pybind/logger_bridge.cpp
```diff
@@ -0,0 +1,210 @@
...
+        py_logger.attr("handle")(record);
+        
+    } catch (const py::error_already_set& e) {
+        // Python error during logging - ignore to prevent cascading failures
+        // (Logging errors should not crash the application)
+        (void)e;  // Suppress unused variable warning
+    } catch (const std::exception& e) {
+        // Other error - ignore
+        (void)e;
+    }
```
sumitmsft, line 222: are we missing catch here for unknown exceptions?

mssql_python/pybind/logger_bridge.cpp
```diff
@@ -0,0 +1,210 @@
...
+    std::lock_guard<std::mutex> lock(mutex_);
+    
+    // Skip if already initialized (check inside lock to prevent TOCTOU race)
+    if (initialized_) {
+        return;
+    }
+    
+    try {
+        // Acquire GIL for Python API calls
+        py::gil_scoped_acquire gil;
```
sumitmsft, line 35: Can this be a deadlock scenario?
If the calling thread already holds the GIL, calling py::gil_scoped_acquire can deadlock? 

Python thread calls C++ >> C++ logs >> tries to re-acquire GIL >> DEADLOCK?

bewithgaurav: actually not, py::gil_scoped_acquire is reentrant-safe. It checks if the current thread already holds the GIL:

If GIL is already held by this thread → it's a no-op (or increments a counter)
If GIL is held by another thread → it blocks and waits
If GIL is not held → it acquires it

mssql_python/pybind/logger_bridge.cpp
bewithgaurav: added

mssql_python/pybind/logger_bridge.cpp
bewithgaurav: added

mssql_python/logging.py
bewithgaurav: done, resolving

mssql_python/logging.py
bewithgaurav: done, resolving

mssql_python/logging.py
bewithgaurav: done, resolving

mssql_python/logging.py
bewithgaurav: done, resolving

mssql_python/logging.py
bewithgaurav: done, resolving

mssql_python/logging.py
bewithgaurav: done, resolving

sumitmsft (dismissed):

subrata-ms (approved):

PR #313, opened 2025-10-31
FEAT: Add support for decimal value in scientific notation

gargsaumya:

> [AB#40022](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/40022)

### Summary   
  

This pull request improves the handling of decimal values, especially those in scientific notation, when converting to SQL `VARCHAR` types. The changes ensure that decimals are consistently formatted as fixed-point strings rather than potentially problematic scientific notation, preventing SQL Server conversion errors. Additionally, new tests have been added to verify correct behavior for a variety of decimal values.

**Decimal formatting and conversion improvements:**

* Updated `_map_sql_type` in `mssql_python/cursor.py` to use `format(param, 'f')` instead of `str(param)` for `MONEY` and `SMALLMONEY` types, ensuring decimals are always converted to fixed-point strings. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L402-R402) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L412-R411)
* Modified `executemany` in `mssql_python/cursor.py` to format decimal values as fixed-point strings before insertion when mapped to `SQL_VARCHAR`.

**Test suite enhancements:**

* Added `test_decimal_scientific_notation_to_varchar` in `tests/test_004_cursor.py` to verify that decimals (including those with scientific notation) are correctly converted and stored as `VARCHAR` without causing conversion errors.

**Test cleanup and simplification:**

* Removed exception handling logic in `test_numeric_leading_zeros_precision_loss` and `test_numeric_extreme_exponents_precision_loss` that previously skipped tests on conversion errors, as these errors should no longer occur with the improved formatting. [[1]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L13944-L13957) [[2]](diffhunk://#diff-82594712308ff34afa8b067af67db231e9a1372ef474da3db121e14e4d418f69L14005-L14030)


mssql_python/cursor.py
```diff
@@ -399,8 +399,7 @@ def _map_sql_type(  # pylint: disable=too-many-arguments,too-many-positional-arg
 
             # Detect MONEY / SMALLMONEY range
             if SMALLMONEY_MIN <= param <= SMALLMONEY_MAX:
-                # smallmoney
-                parameters_list[i] = str(param)
+                parameters_list[i] = format(param, 'f')
```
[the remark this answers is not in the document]
sumitmsft: agree to this.. @gargsaumya

sumitmsft, line 398: The branches handling `SMALLMONEY` and `MONEY` use `format(param, 'f')` directly on the Decimal without enforcing the mandatory scale of 4 fractional digits. If the input Decimal has more than 4 fractional digits (e.g., 123.456789), the generated string preserves all digits. SQL Server expects MONEY/SMALLMONEY at scale 4; extra digits can lead to silent truncation, implicit rounding, or conversion errors—each a financial data correctness risk. Very small scientific-notation values (e.g., 1E-20) will also expand into long zero-padded strings unnecessarily.

Risk:

Data integrity: Values may be rounded/truncated differently than intended.
Inconsistent behavior across environments (implicit conversion vs. driver enforcement).
Potential performance/memory waste for tiny magnitudes due to long fixed-point expansion.

Recommendation:

- Quantize to Decimal('0.0001') with a deterministic rounding mode (ROUND_HALF_EVEN).
- Format with exactly four fractional digits: f"{param:.4f}".
- Consider rejecting or guarding extremely small magnitudes producing long strings.
- Ensure non-Decimal numeric types are converted safely to Decimal before quantization.

sumitmsft: Following tests will be useful:
Input with >4 fractional digits (e.g., 1.234567 → 1.2346 if HALF_EVEN).
Boundary MONEY values near MONEY_MAX / MONEY_MIN with extra fractional digits.
Very small values (e.g., 0.00000001) to ensure guard or expected formatting outcome.

gargsaumya: Thanks @sumitmsft. This PR fixes numeric precision issues; MONEY/SMALLMONEY formatting needs scale-4 handling, so I opened an issue (40169) to handle that separately and avoid expanding scope here since we are short on time.

bewithgaurav, line 398: if we're not handling MONEY/SMALLMONEY in this PR, can consider removing their changes since precision loss can anyways cause modified data or possibly error out
My suggestion is to just keep numeric in this PR since the original issue is handling numerics only

sumitmsft, line 398: Here we convert SMALLMONEY (and similarly MONEY later on) using `format(param, 'f')` and map to SQL_VARCHAR. 

In `executemany` there is a separate block that again formats Decimal values to fixed-point for SQL_VARCHAR. This duplicates logic and risks divergence.

Can we have a shared helper, then call it here and in the batch-processing loop before binding parameters?

sumitmsft, line 398: Before formatting MONEY/SMALLMONEY values with `format(param, 'f')`, we should explicitly reject non-finite decimals (Decimal('NaN'), Decimal('Infinity'), Decimal('-Infinity')). 

Relying on format() is fragile: it may raise a low-level exception or produce driver-dependent behavior. This is both a correctness and defensive security concern (unvalidated special numerics reaching lower layers).

Place this check before the formatting for BOTH branches (SMALLMONEY and MONEY), and also in the batch path (executemany) where Decimal values are converted to fixed-point strings. Centralizing this in a helper (e.g. _format_decimal_for_sql) would avoid future drift.

sumitmsft, line 398: We currently branch on `SMALLMONEY_MIN <= param <= SMALLMONEY_MAX` at line 401 and `MONEY_MIN <= param <= MONEY_MAX` at line 410 and, when in-range, map the value to SQL_VARCHAR after `format(param, 'f')`. 

Two concerns I see here:

- If the user intends to insert into a MONEY/SMALLMONEY column, values outside the respective ranges silently fall through to generic NUMERIC binding. This can mask logical mistakes (e.g., attempting to store an out-of-range monetary amount that will later be CAST to MONEY and overflow/round unexpectedly.
- For in-range values, we do not enforce scale = 4 (MONEY/SMALLMONEY fixed fractional precision). A value like Decimal('1.123456') will be stringified with all fractional digits; SQL Server will round/truncate implicitly if later CASTed to MONEY. That’s silent behavioral drift. (This I know you're going to take care post GA) Please note this down...

tests/test_004_cursor.py
```diff
@@ -13985,6 +13952,72 @@ def test_numeric_beyond_38_digit_precision_negative(
     ), f"Expected SQL Server precision limit message for {description}, got: {error_msg}"
 
 
+@pytest.mark.parametrize(
```
sumitmsft, line 13629: Recommended tests:

MONEY / SMALLMONEY rounding/truncation cases (e.g. Decimal('123.123456'))
Extreme negative exponents (e.g. Decimal('1E-20')) after your formatting
Very large positive exponents that produce large integer part (e.g. Decimal('9.99E+20'))
Boundary values (max/min for MONEY/SMALLMONEY)
Non-finite decimals rejection (NaN/Infinity)

Decimal('0E-10') (should result in '0.0000000000' or '0' depending on strategy)
Negative zero Decimal('-0.0') (ensure consistent canonical '0.0' or '0')
Precision just beyond MONEY range boundaries (overflow detection)
Very long integer part (e.g. Decimal('123456789012345678901234567890.12')) stored as VARCHAR—ensure truncation/parameter length handling does not break.

gargsaumya:
As discussed offline, I’ve created a work item (40169) for correct handling of Decimal types (Numeric, Money, and SmallMoney) based on proper precision and scale.

Issue #272, opened 2025-10-02
Insert Decimal with executemany throw an error


Tweentyy:
### Describe the bug
When attempting to add a Decimal to the database with an `executemany`, an error is returned.
```
Exception message: Exception: RuntimeError: Input string UTF-16 length exceeds allowed column size at parameter index 0. UTF-16 length: 20, Column size: 13
Stack trace:
[2025-10-02T10:57:39.569Z]   File "/Users/tweenty/Desktop/XXX/test.py", line 31, in insert_data
[2025-10-02T10:57:39.570Z]     cursor.executemany("INSERT INTO TestSQL (DecimalTest) VALUES (?)", data)
[2025-10-02T10:57:39.570Z]   File "/Users/tweenty/Desktop/XXX/.venv/lib/python3.12/site-packages/mssql_python/cursor.py", line 1682, in executemany
[2025-10-02T10:57:39.570Z]     ret = ddbc_bindings.SQLExecuteMany(
[2025-10-02T10:57:39.570Z] .
```

### To reproduce
Create a table with Decimal:
```sql
CREATE TABLE TestSQL (
	DecimalTest DECIMAL(30, 20)
)
```

```python
from decimal import Decimal

data = [(Decimal('35.1128407822'),), (Decimal('40000.56405640654065'),)]

cursor.execute("INSERT INTO TestSQL (DecimalTest) VALUES (?)", data[0]) # Works
cursor.executemany("INSERT INTO TestSQL (DecimalTest) VALUES (?)", data) # Doesn't work
```

Workaround found - Convert Decimal to str :
```python
from decimal import Decimal

data = [(str(Decimal('35.1128407822')),), (str(Decimal('40000.56405640654065')),)]

cursor.execute("INSERT INTO TestSQL (DecimalTest) VALUES (?)", data[0]) # Works
cursor.executemany("INSERT INTO TestSQL (DecimalTest) VALUES (?)", data) # Works
```

### Expected behavior
Insertion of lines into database.

### Further technical details
Python version: 3.12.10
SQL Server version: SQL Server 2022
Operating system: macOS 26.0.1
mssql_python version: 0.13.0

gargsaumya:
Thanks for trying out our driver and for sharing the details. We’ll take a look into this issue and update here once we have more information.
Please keep trying things out and raising issues. :)

subrata-ms:
[...]

PR #322, opened 2025-11-11
FIX: Insert Decimal with executemany throw an exception

subrata-ms:

> [AB#39795](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39795)

> GitHub Issue: #272

### Summary   
  

This pull request improves type inference for column values in the MSSQL Python cursor and adds a new test for handling lists of `Decimal` values in executemany operations. The main changes are focused on enhancing how column types are computed, especially for decimal values, and increasing test coverage for edge cases.

Enhancements to type inference logic:

* Improved the `_compute_column_type` method in `cursor.py` to better handle objects with length (like strings and bytes) and `decimal.Decimal` values, preferring longer objects and higher-precision decimals for sample value selection. This also ensures that decimals are preferred over non-decimals for more accurate type inference.

Test coverage improvements:

* Added a new test `test_executemany_Decimal_list` in `test_004_cursor.py` to verify correct behavior when using `executemany` with lists of `decimal.Decimal` values. This test ensures that the operation handles decimal data types properly and cleans up the test table afterward.


mssql_python/cursor.py
```diff
@@ -1717,9 +1717,19 @@ def _compute_column_type(self, column):
...
+            if not sample_value:
+                sample_value = v
+            elif hasattr(v, "__len__") and hasattr(sample_value, "__len__") and len(v) > len(sample_value):
+                # For objects with length (strings, bytes, etc.), prefer the longer one
+                sample_value = v
+            elif isinstance(v, decimal.Decimal) and isinstance(sample_value, decimal.Decimal):
+                # For Decimal objects, prefer the one with higher precision (more digits)
+                v_tuple = v.as_tuple()
+                sample_tuple = sample_value.as_tuple()
+                if len(v_tuple.digits) > len(sample_tuple.digits):
```
sumitmsft: The current logic for choosing a representative Decimal compares only the count of digits `(len(v.as_tuple().digits))`, ignoring scale (`-exponent` when `exponent < 0`). This can under-estimate required buffer size for parameter binding.

Two values with identical digit counts but different scale produce different string lengths (e.g. Decimal('12345') vs Decimal('123.45')). Binding logic often needs both precision and scale; choosing a lower-scale Decimal may cause truncated allocation and the runtime error you observed.

Please update decimal representative selection to use both precision and scale (from `Decimal.as_tuple()`), not just digit count, otherwise buffer sizing can be underestimated for values like 123.45 vs 12345

mssql_python/cursor.py
sumitmsft, line 1749: I was just reviewing this function and found something peculiar. 

The current logic returns immediately if it finds any `int`. In practice, the “column” here represents the batched parameter values for a single placeholder across executemany, not necessarily a single homogeneous SQL column source. It’s common for such a list to mix int and Decimal:

- Whole numbers entered without a fractional part remain int.
- Values with fractional digits (e.g. "2.50") are parsed/converted to Decimal.
- Mixed ingestion from CSV/JSON/Pandas, or business logic that only wraps fractional values in Decimal.

Returning early on integers masks the presence of later Decimal values, causing us to:
1. Lose scale information.
2. Under-size parameter buffers.
3. Potentially hit the observed runtime error (“Input string exceeds allowed column size”).

Please scan entire batch, prioritize decimals when present, and compare (precision, scale).

mssql_python/cursor.py
```diff
@@ -1717,9 +1717,19 @@ def _compute_column_type(self, column):
 
         sample_value = None
         for v in non_nulls:
-            if not sample_value or (
-                hasattr(v, "__len__") and len(v) > len(sample_value)
-            ):
+            if not sample_value:
+                sample_value = v
+            elif hasattr(v, "__len__") and hasattr(sample_value, "__len__") and len(v) > len(sample_value):
```
sumitmsft: The function `_compute_column_type` calls `len(v)` after only checking `hasattr(v, "__len__")`. That’s not always safe.

Why this is a problem:

- A custom object’s `__len__` can raise an exception.
- Computing length might be expensive (lazy or remote data).
- Bad implementations may return invalid values (causing `TypeError`).
- One failing `len()` breaks the whole type inference and can lead to wrong buffer sizing later.

Recommended Fixes:
- Wrap the `len()` call in `try/except`; if it fails, treat length as unavailable (e.g. `l = -1`).
- Cache `len(sample_value)` instead of recomputing it repeatedly.

tests/test_004_cursor.py
```diff
@@ -1663,6 +1663,22 @@ def test_executemany_empty_parameter_list(cursor, db_connection):
         db_connection.commit()
 
 
+def test_executemany_Decimal_list(cursor, db_connection):
+    """Test executemany with an empty parameter list."""
```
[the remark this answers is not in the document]
sumitmsft: Pls see if this can be changed @subrata-ms

tests/test_004_cursor.py
```diff
@@ -1663,6 +1663,22 @@ def test_executemany_empty_parameter_list(cursor, db_connection):
...
+    try:
+        cursor.execute("CREATE TABLE #pytest_empty_params (val DECIMAL(30, 20))")
+        data = [(decimal.Decimal('35.1128407822'),), (decimal.Decimal('40000.5640564065406'),)]
+        cursor.executemany("INSERT INTO #pytest_empty_params VALUES (?)", data)
+        db_connection.commit()
+
+        cursor.execute("SELECT COUNT(*) FROM #pytest_empty_params")
+        count = cursor.fetchone()[0]
+        assert count == 0
```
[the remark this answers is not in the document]
sumitmsft: Please validate this as well

tests/test_004_cursor.py
```diff
@@ -1663,6 +1663,22 @@ def test_executemany_empty_parameter_list(cursor, db_connection):
         db_connection.commit()
 
 
+def test_executemany_Decimal_list(cursor, db_connection):
```
sumitmsft, line 1666: Please see if we can add following test cases:

1. Ensure scale is considered (not just digit count). Data: [Decimal('12345'), Decimal('123.45')] Expect: Representative = Decimal('123.45') (higher scale with same precision).

2. Verify removal of early integer short‑circuit; decimals take precedence. Data: [1, 2, Decimal('3.50')] Expect: Representative is the Decimal; min/max ints are not returned.

3. Stress buffer sizing at upper bound. Data: [Decimal('9' * 38), Decimal('1.' + '2' * 37)] (e.g. precision 38, scale 37) Expect: Representative = second value if comparison is (precision, scale); no overflow/truncation in binding.

4. Ensure scale counts fractional zeros (impacts length). Data: [Decimal('123.4500'), Decimal('123.45'), Decimal('123.4560')] Expect: Representative = Decimal('123.4560') (highest (precision, scale) or longest effective fractional part).

5. Verify that a batch with varying scales binds without error. SQL Column: DECIMAL(10,4) Batch: [Decimal('1.2345'), Decimal('9999.0000')] Expect: executemany succeeds; rowcount = 2.

tests/test_004_cursor.py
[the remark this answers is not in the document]
subrata-ms: part of next commit.

tests/test_004_cursor.py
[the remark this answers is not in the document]
subrata-ms: Part of next commit

subrata-ms:
@microsoft-github-policy-service agree company="Microsoft"

mssql_python/cursor.py
subrata-ms: Thanks for pointing this. Fixed in next commit.

mssql_python/cursor.py
subrata-ms: Made the changes. Should be available in next commit.

tests/test_004_cursor.py
subrata-ms: Thanks for the detailed review. I have added test cases for 1,4, 5 and additional string of decimal test cases.

mssql_python/cursor.py
subrata-ms: As discusses, i will create a separate github bug for this scenario. This is not working currently and throwing an exception in ddbc layer.

Issue #273, opened 2025-10-02
Insert None with executemany is not supported


Tweentyy:
### Describe the bug
When inserting a column that accepts NULL values with the value None, an error is returned.
```
Exception message: Exception: RuntimeError: BindParameterArray: Unsupported C type: 99
Stack trace:
[2025-10-02T10:57:39.569Z]   File "/Users/tweenty/Desktop/XXX/test.py", line 31, in insert_data
[2025-10-02T10:57:39.570Z]     cursor.executemany("INSERT INTO TestSQL (TestValue) VALUES (?)", data)
[2025-10-02T10:57:39.570Z]   File "/Users/tweenty/Desktop/XXX/.venv/lib/python3.12/site-packages/mssql_python/cursor.py", line 1682, in executemany
[2025-10-02T10:57:39.570Z]     ret = ddbc_bindings.SQLExecuteMany(
[2025-10-02T10:57:39.570Z] .
```

### To reproduce
Create a table with Nullable column:
```sql
CREATE TABLE TestSQL (
      TestValue VARCHAR(50)
)
```

```python
data = [(None,), (None,)]

cursor.execute("INSERT INTO TestSQL (TestValue) VALUES (?)", data[0]) # Works
cursor.executemany("INSERT INTO TestSQL (TestValue) VALUES (?)", data) # Doesn't work
```

### Expected behavior
Insertion of lines into database.
I have tested with a Decimal column, it's not working so maybe it's a general problem.

### Further technical details
Python version: 3.12.10
SQL Server version: SQL Server 2022
Operating system: macOS 26.0.1
mssql_python version: 0.13.0

PR #332, opened 2025-11-17
FIX: Invalid data type for None

subrata-ms:
> [AB#39794](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39794)

> GitHub Issue: #273

### Summary   
  
This pull request adds support for binding arrays of NULL parameters in SQL statements and introduces a corresponding test to ensure correct behavior when executing multiple inserts with only NULL values. The main focus is on handling the `SQL_C_DEFAULT` type for parameter arrays, ensuring all values are NULL, and verifying this functionality through unit testing.

**Parameter binding improvements:**

* Added logic to `BindParameterArray` in `ddbc_bindings.cpp` to handle the `SQL_C_DEFAULT` type, ensuring that arrays of NULL values are properly bound and validated. If any non-NULL value is detected, an exception is thrown.

**Testing enhancements:**

* Added a new test `test_executemany_NONE_parameter_list` in `test_004_cursor.py` to verify that `executemany` correctly inserts rows with all NULL values and that the count matches the expected number of inserts.

### DIFF
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -1301,21 +1301,21 @@ class Connection:
                     # Make sure we use the correct amount of data based on length
                     actual_data = data[:length]
 
-                    # Now decode the string data
-                    try:
-                        return actual_data.decode("utf-8").rstrip("\0")
-                    except UnicodeDecodeError:
+                    # SQLGetInfoW returns UTF-16LE encoded strings (wide-character ODBC API)
+                    # Try UTF-16LE first (expected), then UTF-8 as fallback
+                    for encoding in ("utf-16-le", "utf-8"):
                         try:
-                            return actual_data.decode("latin1").rstrip("\0")
-                        except Exception as e:
-                            logger.debug(
-                                "error",
-                                "Failed to decode string in getinfo: %s. "
-                                "Returning None to avoid silent corruption.",
-                                e,
-                            )
-                            # Explicitly return None to signal decoding failure
-                            return None
+                            return actual_data.decode(encoding).rstrip("\0")
+                        except UnicodeDecodeError:
+                            continue
+
+                    # All decodings failed
+                    logger.debug(
+                        "Failed to decode string in getinfo (info_type=%d) with supported encodings. "
+                        "Returning None to avoid silent corruption.",
+                        info_type,
+                    )
+                    return None
                 else:
                     # If it's not bytes, return as is
                     return data
--- a/tests/test_003_connection.py
+++ b/tests/test_003_connection.py
@@ -2187,6 +2187,132 @@ def test_getinfo_basic_driver_info(db_connection):
         pytest.fail(f"getinfo failed for basic driver info: {e}")
 
 
+def test_getinfo_string_encoding_utf16(db_connection):
+    """Test that string values from getinfo are properly decoded from UTF-16."""
+
+    # Test string info types that should not contain null bytes
+    string_info_types = [
+        ("SQL_DRIVER_VER", sql_const.SQL_DRIVER_VER.value),
+        ("SQL_DRIVER_NAME", sql_const.SQL_DRIVER_NAME.value),
+        ("SQL_DRIVER_ODBC_VER", sql_const.SQL_DRIVER_ODBC_VER.value),
+        ("SQL_SERVER_NAME", sql_const.SQL_SERVER_NAME.value),
+    ]
+
+    for name, info_type in string_info_types:
+        result = db_connection.getinfo(info_type)
+
+        if result is not None:
+            # Verify it's a string
+            assert isinstance(result, str), f"{name}: Expected str, got {type(result).__name__}"
+
+            # Verify no null bytes (indicates UTF-16 decoded as UTF-8 bug)
+            assert (
+                "\x00" not in result
+            ), f"{name} contains null bytes, likely UTF-16/UTF-8 encoding mismatch: {repr(result)}"
+
+            # Verify it's not empty (optional, but good sanity check)
+            assert len(result) > 0, f"{name} returned empty string"
+
+
+def test_getinfo_string_decoding_utf8_fallback(db_connection):
+    """Test that getinfo falls back to UTF-8 when UTF-16LE decoding fails.
+
+    This test verifies the fallback path in the encoding loop where
+    UTF-16LE fails but UTF-8 succeeds.
+    """
+    from unittest.mock import MagicMock
+
+    # UTF-8 encoded "Hello" - this is valid UTF-8 but NOT valid UTF-16LE
+    # (odd number of bytes would fail UTF-16LE decode)
+    utf8_data = "Hello".encode("utf-8")  # b'Hello' - 5 bytes, odd length
+
+    mock_result = {"data": utf8_data, "length": len(utf8_data)}
+
+    # Use a string-type info_type (SQL_DRIVER_NAME = 6 is in string_type_constants)
+    info_type = sql_const.SQL_DRIVER_NAME.value
+
+    # Save the original _conn and replace with a mock
+    original_conn = db_connection._conn
+    try:
+        mock_conn = MagicMock()
+        mock_conn.get_info.return_value = mock_result
+        db_connection._conn = mock_conn
+
+        result = db_connection.getinfo(info_type)
+
+        assert result == "Hello", f"Expected 'Hello', got {repr(result)}"
+        assert isinstance(result, str), f"Expected str, got {type(result).__name__}"
+    finally:
+        # Restore the original connection
+        db_connection._conn = original_conn
+
+
+def test_getinfo_string_decoding_all_fail_returns_none(db_connection):
+    """Test that getinfo returns None when all decoding attempts fail.
+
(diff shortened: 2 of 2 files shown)


mssql_python/cursor.py
```diff
@@ -307,7 +307,7 @@ def _map_sql_type(  # pylint: disable=too-many-arguments,too-many-positional-arg
             logger.debug('_map_sql_type: NULL parameter - index=%d', i)
             return (
                 ddbc_sql_const.SQL_VARCHAR.value,
-                ddbc_sql_const.SQL_C_DEFAULT.value,
+                ddbc_sql_const.SQL_C_CHAR.value,
```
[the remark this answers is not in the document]
subrata-ms: reverted the changes

gargsaumya (approved):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2335,6 +2335,33 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt, const py::list& columnwise_params,
                     bufferLength = sizeof(SQLGUID);
                     break;
                 }
+                case SQL_C_DEFAULT: {
+                    // Handle NULL parameters - all values in this column should be NULL
+                    LOG("BindParameterArray: Binding SQL_C_DEFAULT (NULL) array - param_index=%d, count=%zu", paramIndex, paramSetSize);
+                    
+                    // Verify all values are indeed NULL
+                    for (size_t i = 0; i < paramSetSize; ++i) {
```
sumitmsft: The NULL validation loop occurs AFTER the C type has already been determined as SQL_C_DEFAULT. This creates a timing issue where:
- Type determination logic has already selected SQL_C_DEFAULT
- Code enters this case block
- Only then does validation check if all values are actually NULL
- If non-NULL values exist, an exception is thrown

The validation logic needs to be moved earlier in the BindParameterArray function flow - specifically, to wherever the C type (SQL_C_DEFAULT) is initially determined for the parameter array, before entering this switch/case block.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2335,6 +2335,33 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt, const py::list& columnwise_params,
...
+                        if (!columnValues[i].is_none()) {
+                            LOG("BindParameterArray: SQL_C_DEFAULT non-NULL value detected - param_index=%d, row=%zu", paramIndex, i);
+                            ThrowStdException("SQL_C_DEFAULT (99) should only be used for NULL parameters at index " + std::to_string(paramIndex));
```
sumitmsft: If we're inside the case SQL_C_DEFAULT: block, it means the type determination algorithm already decided this parameter should use SQL_C_DEFAULT
Finding non-NULL values here means the type detection logic failed upstream
This exception is a band-aid that catches the failure after the fact rather than preventing it

Wouldn't this lead to:
- Masking the underlying bugs - the real bug is in the type detection algorithm, not here
- Data corruption risk - Other code paths may have already made assumptions based on the incorrect type
- Inconsistent behavior: Why is SQL_C_DEFAULT being selected for mixed/non-NULL data in the first place?

So, my question is:

Where in the code is the C type determined for each parameter column?
What logic decides to use SQL_C_DEFAULT vs other types?
Why isn't the type detection checking for all-NULL before selecting SQL_C_DEFAULT?

I feel that the fix belongs in the type detection logic, not here. This validation should be an assert() in debug builds (since it should never happen), not runtime error handling.

mssql_python/pybind/ddbc_bindings.cpp
sumitmsft: If the validation loop throws an exception at line 2346 (`ThrowStdException...`), execution jumps out immediately
Buffer allocation at lines 2352-2353 never happens
BUT: Looking at the code structure, if any allocations happened before entering this case block in the broader function context, they may not be tracked in tempBuffers yet, leading to leaks during exception unwinding.

basically, this may lead to memory leaks under error conditions - every failed validation leaks memory
In high-throughput database operations, this accumulates rapidly
Production servers could experience memory exhaustion

tests/test_004_cursor.py
```diff
@@ -1598,6 +1598,38 @@ def test_executemany_empty_parameter_list(cursor, db_connection):
         db_connection.commit()
 
 
+def test_executemany_NONE_parameter_list(cursor, db_connection):
```
sumitmsft: Critical Missing Test Cases:

Large arrays of NULL values (e.g., 10,000+ rows) to test memory allocation limits
Multi-column NULL arrays - test only covers single column
NULL values with different SQL types (INT, DATE, BINARY, etc.)
Concurrent executemany calls with NULL parameters (thread safety)
Error recovery: What happens after the exception in line 2345 is thrown? Is the connection still usable?

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: Thanks @sumitmsft for the detail review. I have recheck the logic and looks fine to me. The code flow is as below step-wise,
**Step 1:** Python: Scan entire column → and returns None only if ALL column values are None. Below is the code snippet,

**Cursor.py:: _compute_column_type**

        Returns:
            sample_value: Representative value for type inference and modified_row.
            min_val: Minimum for integers (None otherwise).
            max_val: Maximum for integers (None otherwise).
        """
        non_nulls = [v for v in column if v is not None]
        if not non_nulls:
            return None, None, None

**Step 2:** Python: After step-1, python process calls _create_parameter_types_list where sample_value is passed. This value will be only None when all column values are None ( Step-1). Hence a column can have a sample_value as None only when all columns across all rows is None.

**Step 3:** C++: Python process is now call ddbc_bindings.SQLExecuteMany with data prepared as part of Step-1 & Step-2. Under BindParameterArray, SQL_C_DEFAULT will be selected if column value is "None" across all rows (as per step-1 and Step-2). Under SQL_C_DEFAULT (switch case), we are just doing a extra check to validate all the columns are really None or not. So, your observation is accurate and the C++ validation is a band-aid that verifies the upstream Python logic worked correctly. It's a defensive assertion that ensures consistency between Python's type detection and C++'s parameter binding. This actually protects potential crash incase upstream python logic does not work properly.

Although, the logic protects us from a potential crash due to irregularities in upstream processing, but it would be rarest incidence. Other hand it will have a perf impact in large data set. Hence it is batter to remove this "NULL" checking loop. I am fixing this in latest commit.

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: You're absolutely right. The defensive assertion loops through every row in the column just to validate something that should never happen if the upstream logic is correct. For large datasets, this is wasteful. This check also protects us from a potential crash (due to memory allocation) in-case of irregularities in python upstream logic. However, it will have potentially perf impact on large dataset for all the positive cases. Considering the perf impact, I am removing this check in my latest commit.

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: Removing this loop (validation for None) due to perf impact for positive scenario. Provided more details on earlier comments.

tests/test_004_cursor.py
subrata-ms: Added following test cases to check the logic thoroughly.

1. test_executemany_concurrent_null_parameters --> Thread safety check for ExecuteMany. 
2. test_executemany_multi_column_null_arrays  --> Check for multi column null arrays with different data type
3. test_executemany_mixed_null_and_typed_values  --> 50 rows * 10 column mixed data type check.

I haven't added any large data set like 10,000 rows because this will slow down the PR testing pipeline. Currently it takes 45min+ to validate a PR across windows, mac and linux. I also remove the loop that was doing double validation for Null column. Hence we do not need to handle the error recovery for that scenario.

gargsaumya (approved):

tests/test_004_cursor.py
```diff
@@ -1598,6 +1598,260 @@ def test_executemany_empty_parameter_list(cursor, db_connection):
...
+
+        cursor.execute("SELECT COUNT(*) FROM #pytest_empty_params")
+        count = cursor.fetchone()[0]
+        assert count == 3
+    finally:
+        cursor.execute("DROP TABLE IF EXISTS #pytest_empty_params")
+        db_connection.commit()
+
+
+@pytest.mark.skip(reason="Skipping due to commit reliability issues with executemany")
```
gargsaumya, line 1749: Adding a comment based on our discussion - please create an ADO task to fix this issue if we decide to keep it.

gargsaumya (commented):
Approved, with a comment added that needs to be addressed.

tests/test_004_cursor.py
subrata-ms: This test is not stable on windows but works perfectly fine in Linux.I will do more investigation on this and fix it with my next PR. Worst case, i may need to remove this test altogether.

Issue #333, opened 2025-11-17
cursor.execute() with INSERT statement containing OUTPUT inserted + multiple VALUES entries does not raise IntegrityError


zzzeek:
### Describe the bug

a UNIQUE constraint error is not appropriately triggered for a multi-row INSERT statement that uses OUTPUT inserted

this is a much more specific case than it first looked like, so I've run a full comparison to pyodbc to see what the best case behavior is.    it looks like based on what pyodbc and pymssql do, it's not apparent there's an integrity violation until the fetchall() is called.  so this suggests there is a class of error conditions in SQL server's protocol that does not become apparent until rows are fetched, which may be missing in mssql-python.

### To reproduce

```py
from mssql_python import connect
import mssql_python

conn_str = "UID=scott;PWD=tiger^5HHH;Server=mssql2022;Database=test;Encrypt=No"
conn = connect(conn_str)

cursor = conn.cursor()

cursor.execute("""DROP TABLE IF EXISTS uniq_cons""")

cursor.execute("""
CREATE TABLE uniq_cons (
	id INTEGER NOT NULL IDENTITY,
	data VARCHAR(50) NULL,
	PRIMARY KEY (id),
	UNIQUE (data)
)

""")

# insert a row.  works
cursor.execute("INSERT INTO uniq_cons (data) OUTPUT inserted.id VALUES (?)", ('the data 1', ))
# this was OUTPUT inserted, so fetchall to make sure operation is complete
cursor.fetchall()

# insert a dupe row.  raises as expected
try:
    cursor.execute("INSERT INTO uniq_cons (data) OUTPUT inserted.id VALUES (?)", ('the data 1', ))
    # this was OUTPUT inserted, so fetchall to make sure
    cursor.fetchall()
except mssql_python.IntegrityError:
    print("raised as expected")
else:
    print("INCORRECT: should have raised")

# insert two rows in one statement.   that works
cursor.execute("INSERT INTO uniq_cons (data) OUTPUT inserted.id VALUES (?), (?)", ('the data 2', 'the data 3'))
# this was OUTPUT inserted, so fetchall to make sure
cursor.fetchall()




# assert what's in the table.   seems correct (we lost one IDENTITY value due
# to the exception, that's fine)
cursor.execute("SELECT * from uniq_cons ORDER BY id")
rows = cursor.fetchall()

# we can't compare mssql_python.row.Row to a tuple, might be another issue
# or not, not that important right now
[...]

PR #338, opened 2025-11-20
FIX: GH Issue #333 "cursor.execute() with INSERT statement containing OUTPUT inserted + multiple VALUES entries does not raise IntegrityError"

gargsaumya:

> [AB#40543](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/40543)

> GitHub Issue: #333 

### Summary   
  
This pull request introduces an error handling improvement to the `fetchall` method in `mssql_python/cursor.py`. The change ensures that errors from the fetch operation are properly checked and handled.

Error handling enhancement:

* Added a call to `check_error` after fetching data in the `fetchall` method to verify and handle any errors returned by `DDBCSQLFetchAll`.

### DIFF
--- a/mssql_python/pybind/CMakeLists.txt
+++ b/mssql_python/pybind/CMakeLists.txt
@@ -8,6 +8,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # Enable verbose output to see actual compiler/linker commands
 set(CMAKE_VERBOSE_MAKEFILE ON CACHE BOOL "Verbose output" FORCE)
 
+# Treat CMake warnings as errors
+set(CMAKE_ERROR_DEPRECATED TRUE)
+set(CMAKE_WARN_DEPRECATED TRUE)
+
 if (MSVC)
     # Security compiler options for OneBranch compliance
     message(STATUS "Applying MSVC security compiler options for OneBranch compliance")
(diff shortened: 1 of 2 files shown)


mssql_python/cursor.py
```diff
@@ -2167,7 +2167,10 @@ def fetchall(self) -> List[Row]:
         # Fetch raw data
         rows_data = []
         try:
-            _ = ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
+            ret = ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
+
+            # Check for errors
+            check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
```
subrata-ms, line 2173: Can we also LOG the error message in exception block?

saurabh500, line 2173: Does this error check need to be added to other fetch* api calls (hinted by Sumit in the issue) or does the check already exist?

gargsaumya: Yes, that’s correct.
We’ve identified an internal issue with fetchone and fetchmany, already root-caused it. The fix will go in right after this PR. The return-code checks for those APIs will also be part of that follow-up PR.

(Linking the ADO task for fetchone and fetchmany here- AB#40686)

subrata-ms (approved):

mssql_python/cursor.py
gargsaumya: yes, we are already logging the details before raising exception inside check_error function.

Issue #318, opened 2025-11-06
Connection.getinfo() returns undecoded utf_16_le


gordthompson:
### Describe the bug
String values returned by `Connection.getinfo()` are string representations of `utf_16_le` bytes.

### To reproduce

```
>>> import mssql_python
>>> cnxn = mssql_python.connect(
...     "SERVER=192.168.0.199;"
...         "DATABASE=test;"
...             "UID=scott;PWD=tiger^5HHH;"
...                 "TrustServerCertificate=yes;"
...                 )
>>> driver_ver = cnxn.getinfo(mssql_python.SQL_DRIVER_VER)
>>> type(driver_ver)
<class 'str'>
>>> driver_ver
'1\x008\x00.\x000\x005\x00.\x000\x000\x000\x001'
```

### Expected behavior
Return a normal Unicode string.

### Further technical details
Python version: 3.13
SQL Server version: SQL Server 2019
Operating system: Client - Windows 11; Server - Docker container

PR #340, opened 2025-11-21
FIX: Decode Raw UTF-16 data from Conn.getinfo()

bewithgaurav:

> AB#<WORK_ITEM_ID>

> GitHub Issue: #318 

### Summary   
  
This pull request introduces improvements to the handling of string encoding in the `getinfo` method for SQL Server connections, adds support for profiling builds in the Windows build script, and enhances test coverage for string decoding. The most important changes are grouped below:

### String Decoding Improvements

* The `getinfo` method in `connection.py` now attempts to decode string results from SQL Server using multiple encodings in order: UTF-16LE (Windows default), UTF-8, and Latin-1. This improves robustness when handling driver responses and avoids silent data corruption by returning `None` if all decoding attempts fail.

### Test Coverage

* Added a new test `test_getinfo_string_encoding_utf16` in `test_003_connection.py` to verify that string values returned by `getinfo` are properly decoded from UTF-16, contain no null bytes, and are non-empty, helping catch encoding mismatches early.

### Build Script Cleanup

* Removed redundant logic from `build.bat` related to copying the `msvcp140.dll` redistributable, simplifying the post-build process.

### DIFF
--- a/OneBranchPipelines/github-ado-sync.yml
+++ b/OneBranchPipelines/github-ado-sync.yml
@@ -1,5 +1,18 @@
 # GitHub-to-ADO Sync Pipeline
 # Syncs main branch from public GitHub to internal Azure DevOps daily at 5pm IST
+#
+# SYNC STRATEGY RATIONALE:
+# This pipeline uses a "replace-all" approach rather than traditional git merge/rebase because:
+# 1. DIVERGENT HISTORY: ADO repository contains commits from early development that don't exist
+#    in GitHub. These historical commits were made before GitHub became the source of truth.
+# 2. AVOIDING CONFLICTS: Standard git operations (merge, rebase, reset --hard) fail when
+#    repositories have divergent commit histories. Attempting to merge results in conflicts
+#    that cannot be automatically resolved.
+# 3. IMPLEMENTATION: We use 'git fetch + git rm + git checkout' to completely replace ADO's
+#    working tree with GitHub's files without attempting to reconcile git history. This creates
+#    a clean sync commit that updates all files to match GitHub exactly.
+# 4. CHANGE DETECTION: The pipeline checks if any files actually differ before creating PRs,
+#    avoiding unnecessary sync operations when repositories are already aligned.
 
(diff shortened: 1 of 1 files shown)


mssql_python/connection.py
```diff
@@ -1198,21 +1198,21 @@ def getinfo(self, info_type: int) -> Union[str, int, bool, None]:
                     # Make sure we use the correct amount of data based on length
                     actual_data = data[:length]
 
-                    # Now decode the string data
-                    try:
-                        return actual_data.decode("utf-8").rstrip("\0")
-                    except UnicodeDecodeError:
+                    # SQLGetInfoW returns UTF-16LE encoded strings
+                    # Try encodings in order: UTF-16LE (Windows), UTF-8, Latin-1
+                    for encoding in ("utf-16-le", "utf-8", "latin1"):
```
saurabh500: Why latin1 ? :) Asking out of curiosity. IIUC, utf-16-le and utf-8 might be sufficient. If you have other reasons to add latin1, it would be good to know

bewithgaurav: You're right, I had included this in code as some kind of fallback strategy - however the expected encoding is UTF-16LE, and UTF-8 is a reasonable fallback for edge cases where data might come pre-decoded
Better to remove it. I'll make this change. Thanks for the suggestion!

mssql_python/connection.py
```diff
@@ -1301,21 +1301,22 @@ def getinfo(self, info_type: int) -> Union[str, int, bool, None]:
...
+                        except UnicodeDecodeError:
+                            continue
+
+                    # All decodings failed
+                    logger.debug(
+                        "error",
+                        "Failed to decode string in getinfo (info_type=%d) with supported encodings. "
+                        "Returning None to avoid silent corruption.",
+                        info_type,
+                    )
```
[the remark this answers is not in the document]
bewithgaurav: correct, done

jahnvi480 (dismissed):

jahnvi480 (approved):

PR #342, opened 2025-11-24
FIX: Encoding Decoding

jahnvi480:

> [AB#39049](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/39049)

> GitHub Issue: #250 

### Summary   
This pull request introduces significant improvements to encoding and decoding handling in the MSSQL Python driver, focusing on thread safety, security, and robustness. The main changes include stricter validation and enforcement of encoding rules for SQL_WCHAR types, making encoding/decoding settings thread-safe, and updating cursor methods to consistently use these settings. This ensures correct handling of Unicode data, prevents ambiguous encoding scenarios, and improves reliability in multi-threaded environments.

**Encoding and Decoding Validation & Enforcement**
* Enforced strict validation so that only `'utf-16le'` and `'utf-16be'` encodings are accepted for SQL_WCHAR, explicitly rejecting `'utf-16'` with BOM due to byte order ambiguity. Programming errors are raised if invalid encodings are used, both in `setencoding` and `setdecoding` methods.
* Added validation to ensure encoding names only contain safe characters and are of reasonable length, preventing security issues and denial-of-service attacks.

**Thread Safety**
* Introduced a re-entrant lock (`_encoding_lock`) to protect encoding and decoding settings, making `setencoding`, `setdecoding`, `getencoding`, and `getdecoding` thread-safe and preventing race conditions. 

**Cursor Integration**
* Updated cursor methods (`execute`, `executemany`, `fetchone`, `fetchmany`, `fetchall`) to retrieve encoding and decoding settings from the connection and pass them to low-level bindings, ensuring consistent Unicode handling throughout query execution and result fetching

**Error Handling and Logging**
* Improved error handling in cursor encoding/decoding retrieval, logging warnings if settings cannot be accessed due to database errors and falling back to safe defaults.

**Bindings Interface Update**
* Updated the C++ binding for parameter encoding to accept an explicit encoding argument, supporting the new encoding flow from Python.

### DIFF
--- a/eng/pipelines/pr-validation-pipeline.yml
+++ b/eng/pipelines/pr-validation-pipeline.yml
@@ -1395,12 +1395,14 @@ jobs:
 
   - script: |
       # Create a Docker container for testing on x86_64
+      # TODO(AB#40901): Temporary pin to 3.22 due to msodbcsql ARM64 package arch mismatch
+      # Revert to alpine:latest once ODBC team releases fixed ARM64 package
       docker run -d --name test-container-alpine \
         --platform linux/amd64 \
         -v $(Build.SourcesDirectory):/workspace \
         -w /workspace \
         --network bridge \
-        alpine:latest \
+        alpine:3.22 \
         tail -f /dev/null
     displayName: 'Create Alpine x86_64 container'
 
@@ -1637,12 +1639,14 @@ jobs:
 
   - script: |
       # Create a Docker container for testing on ARM64
+      # TODO(AB#40901): Temporary pin to 3.22 due to msodbcsql ARM64 package arch mismatch
+      # Revert to alpine:latest once ODBC team releases fixed ARM64 package
       docker run -d --name test-container-alpine-arm64 \
         --platform linux/arm64 \
         -v $(Build.SourcesDirectory):/workspace \
         -w /workspace \
         --network bridge \
-        alpine:latest \
+        alpine:3.22 \
         tail -f /dev/null
     displayName: 'Create Alpine ARM64 container'
 


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -279,7 +279,8 @@ std::string DescribeChar(unsigned char ch) {
 // each of them with appropriate arguments
 SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
                          std::vector<ParamInfo>& paramInfos,
-                         std::vector<std::shared_ptr<void>>& paramBuffers) {
+                         std::vector<std::shared_ptr<void>>& paramBuffers,
+                         const std::string& charEncoding = "utf-8") {
```
saurabh500, line 293: Shouldn't the default be UTF-16 LE?

saurabh500: Is this the target encoding to be used for sending data to ODBC or is it the intended encoding on the connection

mssql_python/connection.py
```diff
@@ -227,6 +243,11 @@ def __init__(
         self._output_converters = {}
         self._converters_lock = threading.Lock()
 
+        # Initialize encoding/decoding settings lock for thread safety
+        # This lock protects both _encoding_settings and _decoding_settings dictionaries
+        # to prevent race conditions when multiple threads are reading/writing encoding settings
+        self._encoding_lock = threading.RLock()  # RLock allows recursive locking
```
saurabh500: This seems like a use case of multiple readers and single writer. Is Rlock the right synchronization primitive for such a purpose ?

saurabh500: Rlock would allow a single reader at a time, which can be detrimental to performance, assuming that we expect many concurrent readers. IF we dont expect concurrent opearations, then is locking useful?

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: For SQL_CHAR our default encoding we are using is UTF-8 and for SQL_WCHAR we have UTF-16 LE

mssql_python/connection.py
jahnvi480: Thank you for sharing your review, you are correct about this.
Solution that we can implement to remove Rlock and still not have race condition will be to have our own custom implementation by replacing with a ReadWriteLock that allows:
- Multiple concurrent readers getencoding
- Exclusive writer access setencoding/setdecoding
Should I make the changes or have it for next PR?
@sumitmsft @bewithgaurav @saurabh500 Please let me know your views

mssql_python/connection.py
```diff
@@ -430,6 +451,20 @@ def setencoding(self, encoding: Optional[str] = None, ctype: Optional[int] = Non
         encoding = encoding.casefold()
         logger.debug("setencoding: Encoding normalized to %s", encoding)
 
+        # Reject 'utf-16' with BOM for SQL_WCHAR (ambiguous byte order)
+        if encoding == "utf-16" and ctype == ConstantsDDBC.SQL_WCHAR.value:
+            logger.debug(
```
sumitmsft: Can we change this to:
logger.warning("utf-16 with BOM rejected for SQL_WCHAR") ? 

Or if it is truly debug then:
logger.debug("utf-16 with BOM rejected for SQL_WCHAR")

This will make sure not to confuse us when it says "warning" but is logged at DEBUG level.

There are bunch of other places where we need to make similar changes.

mssql_python/connection.py
```diff
@@ -54,7 +54,13 @@
...
 
 # UTF-16 encoding variants that should use SQL_WCHAR by default
-UTF16_ENCODINGS: frozenset[str] = frozenset(["utf-16", "utf-16le", "utf-16be"])
+# Note: "utf-16" with BOM is NOT included as it's problematic for SQL_WCHAR
+UTF16_ENCODINGS: frozenset[str] = frozenset(["utf-16le", "utf-16be"])
+
+# Valid encoding characters (alphanumeric, dash, underscore only)
+import string
+
+VALID_ENCODING_CHARS: frozenset[str] = frozenset(string.ascii_letters + string.digits + "-_")
```
sumitmsft: If we do want to keep a strict whitelist of allowed encodings, then it would be good to:
- Document explicitly which encodings are supported by the driver, and
- Add tests to cover common aliases (utf_8, latin_1, cp1252, etc.) and ensure they continue to work.
This will help avoid surprising ProgrammingErrors in existing applications that rely on Python’s broader codec support.


It's ok to sanitize the user input and adding a length limit, but this changes the behavior of setencoding/setdecoding in a way that may reject encodings which Python itself considers valid:
- Python’s codec registry is the real source of truth for valid encodings. If a codec or alias ever includes characters outside [A–Z a–z 0–9 - _] (e.g., via a custom codec, or in a future Python version), codecs.lookup() would succeed but _validate_encoding will reject it.
- That means some encodings that previously worked with this driver (as long as codecs.lookup accepted them) will now start throwing ProgrammingError, which can be a breaking behavior change for users.

Could we consider relaxing this to rely primarily on codecs.lookup and only use the character filter as a secondary check on the normalized codec name?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2403,10 +2462,18 @@ SQLRETURN SQLExecuteMany_wrap(const SqlHandlePtr statementHandle, const std::wst
...
+    std::string charEncoding = "utf-8";  // default
+    if (encodingSettings.contains("encoding")) {
+        charEncoding = encodingSettings["encoding"].cast<std::string>();
+    }
+
     if (!hasDAE) {
         LOG("SQLExecuteMany: Using array binding (non-DAE) - calling "
             "BindParameterArray");
         std::vector<std::shared_ptr<void>> paramBuffers;
+        // TODO: Pass charEncoding to BindParameterArray when it's updated to support encoding
```
sumitmsft: Does this TODO mean BindParameterArray does not receive charEncoding? Would that in turn mean that Single-row execute uses the new path, but executemany may still use old encoding behavior for array binding?

I am a bit concerned because this may cause different behavior between execute and executemany for the same query and parameters.

mssql_python/cursor.py
```diff
@@ -285,6 +291,55 @@ def _get_numeric_data(self, param: decimal.Decimal) -> Any:
...
+                return self._connection.getencoding()
+            except (OperationalError, DatabaseError) as db_error:
+                # Only catch database-related errors, not programming errors
+                from mssql_python.helpers import log
+
+                log(
+                    "warning",
+                    f"Failed to get encoding settings from connection due to database error: {db_error}",
+                )
+                return {"encoding": "utf-16le", "ctype": ddbc_sql_const.SQL_WCHAR.value}
```
sumitmsft: If getencoding/getdecoding raises OperationalError or DatabaseError, you silently log a warning and then return defaults.

I see two issues;
- If the connection’s encoding/decoding configuration is broken, this quietly changes behavior to a default rather than failing fast.
- Users will see inconsistent results (e.g., data mis-decoded) instead of a clear error, making debugging very difficult.

Can we re-raise after logging for unexpected database errors, especially if they indicate broken connection state.

mssql_python/connection.py
```diff
@@ -456,20 +488,43 @@ def setencoding(self, encoding: Optional[str] = None, ctype: Optional[int] = Non
...
+                        "Cannot use 'utf-16' encoding with SQL_WCHAR due to Byte Order Mark ambiguity. "
+                        "Use 'utf-16le' or 'utf-16be' instead for explicit byte order."
+                    ),
+                )
+            elif encoding not in UTF16_ENCODINGS:
+                logger.warning(
+                    "Non-UTF-16 encoding %s attempted with SQL_WCHAR ctype",
+                    sanitize_user_input(encoding),
+                )
+                raise ProgrammingError(
```
sumitmsft: For setencoding:

If ctype == SQL_WCHAR and encoding == "utf-16" >>  ProgrammingError.
If ctype == SQL_WCHAR and encoding not in UTF16_ENCODINGS >> ProgrammingError.

Similarly for setdecoding:

If sqltype == SQL_WCHAR and encoding == "utf-16" >> ProgrammingError.
If sqltype == SQL_WCHAR and encoding not in UTF16_ENCODINGS>> ProgrammingError.
Similarly when ctype == SQL_WCHAR.

This seems like a breaking change in behavior for any user who previously used "utf-16" with SQL_WCHAR.

The validation appears duplicated (you check for "utf-16" both before and inside the ctype == SQL_WCHAR branch).

Behavioral breaking changes in encoding/decoding can be sensitive and can regress existing applications. Even if the earlier behavior was ambiguous, users might rely on it. 

We need to document this clearly in release notes and on wikis very clearly. 

Please also have test coverage:

- setencoding("utf-16", SQL_WCHAR) should cause ProgrammingError.
- setencoding("utf-16le", SQL_WCHAR) should be success.
Do the same tests for setdecoding.

tests/test_013_encoding_decoding.py
```diff
@@ -0,0 +1,5261 @@
+"""
+Comprehensive Encoding/Decoding Test Suite
+
+This consolidated module provides complete testing for encoding/decoding functionality
+in mssql-python, ensuring pyodbc compatibility, thread safety, and connection pooling support.
```
sumitmsft: Missing / important edge cases to ensure are covered
Connection setencoding("utf-8") vs setdecoding(SQL_CHAR, "latin-1"); confirm behavior is correct and consistent.
Attempt to use "utf-16" or invalid encodings with SQL_WCHAR: assert that ProgrammingError is raised and messages are meaningful.
Multiple threads calling setencoding / getencoding concurrently to ensure no race or crash.
Simulate OperationalError/DatabaseError in _get_encoding_settings / _get_decoding_settings and assert that the log message is emitted.
The default encoding is used as intended.

mssql_python/connection.py
jahnvi480: I have removed this strict checking

tests/test_013_encoding_decoding.py
jahnvi480: Added more testcases

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: This comment was left during my previous iteration and I forgot to remove this comment later. 

This has been executed

mssql_python/cursor.py
jahnvi480: Now the code will raise exception instead of just logging

mssql_python/connection.py
jahnvi480: I have removed the duplicate validation and have noted down about the breaking change, will update the docs and readme file

mssql_python/pybind/ddbc_bindings.cpp
sumitmsft: Ok, so my question was do we have different behavior between execute and executemany for the same query and parameters?

jahnvi480: No both will have same behavior

mssql_python/connection.py
sumitmsft: @jahnvi480 IMO we need to honor concurrent readers from performance perspectives. I think Saurabh has a good point, Rlock will be detrimental to perf. I would rather with a primitive where multiple readers can concurrently access the resource. I'd rather do that change here in this PR.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2797,18 +2910,33 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                                    row.append(decoded);
+                                    LOG("SQLGetData: CHAR column %d decoded with '%s', %zu bytes "
+                                        "-> %zu chars",
+                                        i, charEncoding.c_str(), (size_t)dataLen, py::len(decoded));
+                                } catch (const py::error_already_set& e) {
+                                    LOG_ERROR(
+                                        "SQLGetData: Failed to decode CHAR column %d with '%s': %s",
+                                        i, charEncoding.c_str(), e.what());
+                                    // Return raw bytes as fallback
+                                    py::bytes raw_bytes(reinterpret_cast<char*>(dataBuffer.data()),
```
gargsaumya: When decoding fails, py::bytes is allocated twice from the same buffer - once in the try block (failed decode), and again in the catch block (fallback).
I would suggest create py::bytes once before the try block and reuse it in both success and failure paths.

mssql_python/connection.py
```diff
@@ -54,7 +54,57 @@
 INFO_TYPE_STRING_THRESHOLD: int = 10000
 
 # UTF-16 encoding variants that should use SQL_WCHAR by default
-UTF16_ENCODINGS: frozenset[str] = frozenset(["utf-16", "utf-16le", "utf-16be"])
+# Note: "utf-16" with BOM is NOT included as it's problematic for SQL_WCHAR
+UTF16_ENCODINGS: frozenset[str] = frozenset(["utf-16le", "utf-16be"])
+
+
+def _validate_utf16_wchar_compatibility(
```
gargsaumya, line 61: Please document that SQL_WCHAR is always using UTF-16LE (ODBC specification requirement)

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Removed the duplicate lines

mssql_python/connection.py
jahnvi480: Instead of RLock I'm using normal threading.Lock this will solve the purpose and won't do recursive locking

Issue #337, opened 2025-11-19
built-in pooling does not reset session state for isolation level


zzzeek:
### Describe the bug
setting the transaction isolation level on a connection carries over into new connections due to implicit connection pooling which seems to lack a means of establishing reset behaviors

### To reproduce
```py
import mssql_python

MSSQLPYTHON = (
    "Server=mssql2022,1433;Database=test;UID=scott;PWD=tiger^5HHH;Encrypt=No"
)


def get_isolation_level(dbapi_connection):
    cursor = dbapi_connection.cursor()
    view_name = "sys.system_views"
    cursor.execute(
        (
            "SELECT name FROM {} WHERE name IN "
            "('dm_exec_sessions', 'dm_pdw_nodes_exec_sessions')"
        ).format(view_name)
    )
    row = cursor.fetchone()

    view_name = f"sys.{row[0]}"

    cursor.execute(
        """
            SELECT CASE transaction_isolation_level
            WHEN 0 THEN NULL
            WHEN 1 THEN 'READ UNCOMMITTED'
            WHEN 2 THEN 'READ COMMITTED'
            WHEN 3 THEN 'REPEATABLE READ'
            WHEN 4 THEN 'SERIALIZABLE'
            WHEN 5 THEN 'SNAPSHOT' END
            AS TRANSACTION_ISOLATION_LEVEL
            FROM {}
            where session_id = @@SPID
        """.format(
            view_name
        )
    )
    row = cursor.fetchone()

    cursor.close()

    return row[0].upper()


connection = mssql_python.connect(MSSQLPYTHON)

print(
    f"New connection {connection}: "
    f"isolation level: {get_isolation_level(connection)}"
)

cursor = connection.cursor()
cursor.execute(f"SET TRANSACTION ISOLATION LEVEL SERIALIZABLE")
cursor.close()

print(
    f"Changed isolation level for {connection}: "
    f"now it's: {get_isolation_level(connection)}"
)

connection.close()

connection = mssql_python.connect(MSSQLPYTHON)
print(
    f"Made another new connection {connection}: "
    f"isolation level: {get_isolation_level(connection)}"
)
connection.close()


mssql_python.pooling(enabled=False)
connection = mssql_python.connect(MSSQLPYTHON)
print(
    f"Disabled pooling and connected again: {connection}: "
    f"isolation level: {get_isolation_level(connection)}"
)

```

this outputs:

```
[...]

PR #343, opened 2025-11-24
FIX: Fix for isolation level for in-built connection pool

subrata-ms:
> [AB#40573](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/40573)

> GitHub Issue: #337 

### Summary   
  
This pull request addresses a critical issue in connection pooling for the `mssql_python` library by ensuring that the transaction isolation level is properly reset when a pooled connection is reused. Previously, the isolation level could leak between usages, potentially causing unexpected behavior. The changes include an explicit reset of the isolation level in the connection reset logic and a new test to verify this behavior.

**Bug fix for connection pooling session state:**

* Explicitly reset the transaction isolation level to `READ COMMITTED` in the `Connection::reset()` method to prevent isolation level settings from leaking between pooled connection usages. This addresses the limitation of `SQL_ATTR_RESET_CONNECTION`, which does not reset the isolation level. (`mssql_python/pybind/connection/connection.cpp`)

**Testing improvements:**

* Added a new test `test_connection_pooling_isolation_level_reset` in `tests/test_009_pooling.py` to verify that the isolation level is reset to the default when a connection is reused from the pool, preventing session state leakage.
* Imported the `mssql_python` module in `tests/test_009_pooling.py` to support new test functionality.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -2167,7 +2167,10 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         # Fetch raw data
         rows_data = []
         try:
-            _ = ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
+            ret = ddbc_bindings.DDBCSQLFetchAll(self.hstmt, rows_data)
+
+            # Check for errors
+            check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
 
             if self.hstmt:
                 self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
--- a/mssql_python/pybind/CMakeLists.txt
+++ b/mssql_python/pybind/CMakeLists.txt
@@ -8,6 +8,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # Enable verbose output to see actual compiler/linker commands
 set(CMAKE_VERBOSE_MAKEFILE ON CACHE BOOL "Verbose output" FORCE)
 
+# Treat CMake warnings as errors
+set(CMAKE_ERROR_DEPRECATED TRUE)
+set(CMAKE_WARN_DEPRECATED TRUE)
+
 if (MSVC)
     # Security compiler options for OneBranch compliance
     message(STATUS "Applying MSVC security compiler options for OneBranch compliance")
@@ -300,6 +304,21 @@ target_compile_definitions(ddbc_bindings PRIVATE
 # Add warning level flags for MSVC
 if(MSVC)
     target_compile_options(ddbc_bindings PRIVATE /W4 /WX)
+endif()
+
+# Add warning flags for GCC/Clang on Linux and macOS
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    target_compile_options(ddbc_bindings PRIVATE 
+        -Werror        # Treat warnings as errors
+        -Wattributes   # Enable attribute warnings (cross-compiler)
+    )
+    
+    # GCC-specific warning flags
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        target_compile_options(ddbc_bindings PRIVATE 
+            -Wint-to-pointer-cast  # GCC-specific warning for integer-to-pointer casts
+        )
+    endif()
 endif()
 
 # Add macOS-specific string conversion fix
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -114,6 +114,13 @@ py::object get_uuid_class() {
 
 // Struct to hold parameter information for binding. Used by SQLBindParameter.
 // This struct is shared between C++ & Python code.
+// Suppress -Wattributes warning for ParamInfo struct
+// The warning is triggered because pybind11 handles visibility attributes automatically,
+// and having additional attributes on the struct can cause conflicts on Linux with GCC
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wattributes"
+#endif
 struct ParamInfo {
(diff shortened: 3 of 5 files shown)


gargsaumya (approved):

mssql_python/pybind/connection/connection.cpp
```diff
@@ -308,6 +308,19 @@ bool Connection::reset() {
         disconnect();
         return false;
     }
+    
+    // SQL_ATTR_RESET_CONNECTION does NOT reset the transaction isolation level.
+    // Explicitly reset it to the default (SQL_TXN_READ_COMMITTED) to prevent
+    // isolation level settings from leaking between pooled connection usages.
+    LOG("Resetting transaction isolation level to READ COMMITTED");
+    ret = SQLSetConnectAttr_ptr(_dbcHandle->get(), SQL_ATTR_TXN_ISOLATION,
```
saurabh500, line 316: What happens for cases where the database option has been changed with alter DB
```
ALTER DATABASE CURRENT
SET READ_COMMITTED_SNAPSHOT ON; 

```

In these cases, I think the snapshot iso level by default changes. If Python sets the ISO to READ COMMITTED, then that likely violates the modified default? (can you verify?)

mssql_python/pybind/connection/connection.cpp
```diff
@@ -308,6 +308,19 @@ bool Connection::reset() {
...
+                                (SQLPOINTER)SQL_TXN_READ_COMMITTED, SQL_IS_INTEGER);
```
saurabh500, line 317: When does this set attr on connection get honored? Will this cause an additional roundtrip to SQL server during connection close/open ?

mssql_python/pybind/connection/connection.cpp
sumitmsft, line 316: What if two threads access the connection pool simultaneously? One could grab this connection while reset() is still executing here, causing race conditions and potential crashes.  

Would it make sense if the entire reset() method should be protected by the connection pool's mutex to prevent concurrent access during reset operations?

sumitmsft: Agreed! Setting isolation level hard coded to Read Committed is not correct. Isolation Level is always a DB property and can be changed by Alter Database... command by a DBA. But at a session level, isolation level can be changed and supersede DB level property - only for that session. Even though the session is finished, DB property will remain as is and should be shown as the default isolation level.

sumitmsft: Having said that, this is the standard default for 99% of SQL Server installations. I have rarely seen this setting being changed frequently. 
DBAs rarely change the server-wide default isolation level - It's an unusual configuration

sumitmsft: I took a dig into this once more. Apparently, there are two things which we need to remember:
1. A DBA can change the isolation level _ONLY_ from READ_COMMITTED to READ_COMMITTED_SNAPSHOT
2. The default isolation level on SQL Server database will always be shown\logged in system metadata as READ_COMMITTED (even though the isolation level is set to READ_COMMITTED_SNAPSHOT). There is another column in system tables (is_snapshot_isolation_enabled) which will be toggled from 0 to 1 if READ_COMMITTED_SNAPSHOT is enabled. 

So, essentially, the isolation level of the DB will also remain READ_COMMITTED. So, even if we hardcode it this will not create any problem. I am ok to go ahead with current changes @subrata-ms

subrata-ms: Thanks @sumitmsft for clarifying this scenario in detail.

mssql_python/pybind/connection/connection.cpp
subrata-ms: This issue is only reproducible when we enable the connection pooling in ddbc. The Connection::reset() method will get called when we release the connection to the connection pool. Hence, we need to explicitly reset the isolation level to default ( Read_Committed) otherwise it will hold the previous reset isolation level. Yes, this involves a additional roundtrip to SQL Server.

mssql_python/pybind/connection/connection.cpp
subrata-ms: Until Connection::reset() completes it's execution, it will not return to the pool. Hence no other python/query thread get the same connection handle/object. Hence, it is already thread safe through object level. Alternately, the container that holds all the healthy connection is already protected by mutex lock under ConnectionPool::acquire function. Hence current design looks proper to me in this aspect.

saurabh500: @sumitmsft  and @subrata-ms the rare behavior is a non-zero chance. If the feature exists to change the default ISO of a DB, then someone is clearly using it. 

In this case our guidance to folks using non READ_COMMITTED ISO should be to open a connection and explicitly set Isolation level to READ_COMMITTED_SNAPSHOT. Otherwise, a very valid bug saying that the connection pool doesn't honor defaults where the default for DB is READ_COMMITTED_SNAPSHOT can be easily filed. 

We can even do better, have a connection string keyword to specify the isolation level for the connection for customers in this rare scenario. By default we override to READ_COMMITTED and provide a path forward to customers on READ_COMMITTED_SNAPSHOT default databases with connection string override. 

This can be a logical next step to cover the edge cases with this change being brought in. Not a requirement for this PR

tests/test_009_pooling.py
```diff
@@ -229,6 +309,9 @@ def test_pool_idle_timeout_removes_connections(conn_str):
 # =============================================================================
 
 
+@pytest.mark.skip(
```
saurabh500, line 312: Why skip the test and not delete it?
This seems like a bug which needs to eventually be fixed? 
If yes, it would be good to add a bug number on GH in the skip comment.

saurabh500 (approved):
Provided some feedback. For its current scope the PR looks good.

tests/test_009_pooling.py
subrata-ms: There are couple of test that is consistently failing. Had a look and seems it is not proper. I will take this up as part of below task @saurabh500 . 
https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/40818

Issue #345, opened 2025-11-25
FetchMany(number of rows) ignores batch size when table contains an LOB


dlevy-msft-sql:
### Describe the bug
FetchMany(number of rows) ignores batch size when table contains an LOB

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
Exception message:
Stack trace:
```

### To reproduce
Do an executemany(1) on a table that has an lob column and more than 1 row

### Expected behavior
The requested number of rows would be returned

### Further technical details
Python version: any
SQL Server version: (e.g. SQL Server 2022) any
Operating system: (e.g. Windows Server 2022, Ubuntu 24.04, macOS 14.7.1, Docker container) any

**Additional context**

PR #346, opened 2025-11-25
FIX: FetchMany(number of rows) ignores batch size with LOB columns

dlevy-msft-sql:

> GitHub Issue: #345 

### Summary   
Fixes FetchMany(number of rows) ignores batch size when table contains an LOB

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -845,34 +845,25 @@ std::string GetLastErrorMessage();
 
 // TODO: Move this to Python
 std::string GetModuleDirectory() {
+    namespace fs = std::filesystem;
     py::object module = py::module::import("mssql_python");
     py::object module_path = module.attr("__file__");
     std::string module_file = module_path.cast<std::string>();
 
-#ifdef _WIN32
-    // Windows-specific path handling
-    char path[MAX_PATH];
-    errno_t err = strncpy_s(path, MAX_PATH, module_file.c_str(), module_file.length());
-    if (err != 0) {
-        LOG("GetModuleDirectory: strncpy_s failed copying path - "
-            "error_code=%d, path_length=%zu",
-            err, module_file.length());
-        return {};
-    }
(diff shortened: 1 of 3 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3870,22 +3870,23 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
         }
     }
 
+    SQLULEN numRowsFetched = 0;
```
sumitmsft, line 4038: The `numRowsFetched` variable is used differently in the two code paths:

LOB path: Requires = 0 initialization (used as loop counter)
Non-LOB path: Initial value ignored (overwritten by `SQL_ATTR_ROWS_FETCHED_PTR`)

Adding a comment to clarify this dual purpose would be appropariate:
This will help future maintainers understand the design decision.


```C++
// Initialized to 0 for LOB path counter; overwritten by ODBC in non-LOB path
SQLULEN numRowsFetched = 0;

File: tests/test_004_cursor.py [hunk 2]
```diff
@@ -947,12 +973,32 @@ def test_fetchmany(cursor):
...
 def test_fetchall(cursor):
+    """Test fetching all rows"""
+    cursor.execute(
+        "SELECT id, bit_column, tinyint_column, smallint_column, bigint_column, integer_column, float_column, wvarchar_column, time_column, datetime_column, date_column, real_column FROM #pytest_all_data_types"
+    )
+    rows = cursor.fetchall()
+    assert isinstance(rows, list), "fetchall should return a list"
+    assert len(rows) == len(PARAM_TEST_DATA), "Incorrect number of rows returned"
+
+
```
sumitmsft, line 1212: Add some additional test cases to cover edge cases:

1. Test fetchmany with size=0 for LOB columns
2. Test fetchmany requesting more rows than exist with LOB columns
3. Test fetchmany on empty result set with LOB columns
4. Test fetchmany with very large LOB column data (large_data = "x" * 10000)

Issue #291, opened 2025-10-13
request timeout not working


lstorme:
### Describe the bug

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
10
Done:
```

### To reproduce
Include a complete code listing (or project/solution) that we can run to reproduce the issue.

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to reproduce the issue.

Use step from https://github.com/microsoft/mssql-python/wiki/Connection

```python3.12

    conn = mssql_python.connect(create_dsn(), timeout=10)
    print(conn.timeout)  # 10

    # Change timeout dynamically
    conn.timeout = 10

    cursor = conn.cursor()
    try:
        # Query that might run longer than 10 seconds
        cursor.execute("WAITFOR DELAY '00:00:15'")
    except mssql_python.OperationalError as e:
        print("Query timed out:", e)
    print("Done:")
```

### Expected behavior
-> should see Query timed out...
-> no exception raised

### Further technical details
Python version: 3.12
SQL Server version: (e.g. SQL Server 2022)
Operating system: RedHat 8

sumitmsft:
Hello @lstorme - thanks for reporting this issue. We will triage this issue and let you know our next plan of action on this. Please allow time to revert. If this is extremely critical for your project please do let us know, so that we take it up on priority. 

Thanks again for reporting this issue.

jahnvi480:
Thank you for reporting this issue! I've confirmed the bug and implemented a complete fix. The timeout feature was not working due to missing constants and incorrect type casting in the C++ bindings.

Verification
Tested with your exact scenario:
```python
conn = mssql_python.connect(conn_str, timeout=2)
cursor = conn.cursor()
cursor.execute("WAITFOR DELAY '00:00:05'")  # 5 seconds
```

Result: ✅ Query now properly times out after ~2 seconds with `OperationalError`

The fix has been implemented in PR #348 and will be available once merged.

jahnvi480:
@lstorme PR is merged now and this will start working from next release.

PR #348, opened 2025-11-25
FIX: Timeout during cursor creation and not execute

jahnvi480:

> [AB#40635](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/40635)

> GitHub Issue: #291 

### Summary   
This pull request refactors how query timeouts are set for statement handles in the `mssql_python/cursor.py` module. The main improvement is moving the logic for setting the query timeout from the `execute` method to the cursor initialization process, ensuring the timeout is consistently applied whenever the statement handle is allocated or reset.

Statement handle timeout management:

* Introduced a new `_set_timeout` method to set the query timeout attribute on the statement handle during cursor initialization, following best practices for performance. (`mssql_python/cursor.py`)
* Removed redundant timeout-setting logic from the `execute` method, centralizing timeout management in the cursor lifecycle. (`mssql_python/cursor.py`)

### DIFF
--- a/eng/pipelines/pr-validation-pipeline.yml
+++ b/eng/pipelines/pr-validation-pipeline.yml
@@ -1395,14 +1395,12 @@ jobs:
 
   - script: |
       # Create a Docker container for testing on x86_64
-      # TODO(AB#40901): Temporary pin to 3.22 due to msodbcsql ARM64 package arch mismatch
-      # Revert to alpine:latest once ODBC team releases fixed ARM64 package
       docker run -d --name test-container-alpine \
         --platform linux/amd64 \
         -v $(Build.SourcesDirectory):/workspace \
         -w /workspace \
         --network bridge \
-        alpine:3.22 \
+        alpine:latest \
         tail -f /dev/null
     displayName: 'Create Alpine x86_64 container'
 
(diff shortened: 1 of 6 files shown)


mssql_python/cursor.py
```diff
@@ -656,13 +656,34 @@ def _initialize_cursor(self) -> None:
...
         self._allocate_statement_handle()
+        self._set_timeout()
 
     def _allocate_statement_handle(self) -> None:
         """
         Allocate the DDBC statement handle.
         """
         self.hstmt = self._connection._conn.alloc_statement_handle()
 
+    def _set_timeout(self) -> None:
```
sumitmsft, line 692: I see missing tests..
1. A test that runs a long‑running query with a small timeout (e.g., 1–2 seconds) and asserts a timeout error.
2. Ensure that creating a cursor with _timeout set and calling execute once behaves correctly.
3. Execute multiple times with same cursor and verify:
The timeout applies consistently.
No unexpected change in behavior from previous versions (if there was a known contract).
4. If _reset_cursor reallocates or reuses statement handles, ensure _set_timeout is called as intended and tested.

Issue #130, opened 2025-07-09
Arrow Support


aersam:
### Is your feature request related to a problem? Please describe.
Python lists / arrays are pretty slow and memory-intense, therefore many libraries allow reading things as Arrow Data. Examples are [`fastexcel`](https://github.com/ToucanToco/fastexcel) or [`arrow-odbc`](https://github.com/pacman82/arrow-odbc). Many Data libraries support Arrow out of the box, like DuckDB/Polars/Spark/Pandas/DeltaLake.  Would also be a great source for Bulk Import. 

### Describe the solution you'd like
Best would be an API similar to duckdb which has methods for Arrow Tables and RecordBatchReader: https://duckdb.org/docs/stable/guides/python/export_arrow.html#export-as-a-recordbatchreader This can also be done without a Dependency on `PyArrow`, since arrow supports PyCapsules (eg, see [arro3](https://github.com/kylebarron/arro3) )

### Describe alternatives you've considered
Well, I can also live with not-so fast queries :)

dlevy-msft-sql:
> I'm excited to see work ongoing for `arrow` read support! 🚀
> 
> When the time comes to implement bulk-import it would also be great to have a `pyarrow.Table` as a supported data source.

I've got good news - BCP takes an iterator. It should already just work.

gargsaumya:
Closing this issue as Apache Arrow support has already been implemented. Support for Apache Arrow in Bulk Copy is being tracked separately under #570 
Thanks for the feedback and discussion.

PR #354, opened 2025-11-30
FEAT: Add arrow fetch support

ffelixg:

> GitHub Issue: #130 

### Summary   
  

Hey, you mentioned in issue #130 that you were willing to consider community contributions for adding Apache Arrow support, so here you go. I have focused only on fetching data into Arrow structures from the Database.

The Function signatures I chose are:

*   **`arrow_batch(chunk_size=10000)`**: Fetch a single `pyarrow.RecordBatch`, base for the other two methods.
*   **`arrow(chunk_size=10000)`**: Fetches the entire result set as a single `pyarrow.Table`.
*   **`arrow_reader(chunk_size=10000)`**: Returns a `pyarrow.RecordBatchReader` for streaming results without loading the entire dataset into RAM.

Using `fetch_arrow...` instead of just `arrow...` could also be a good option, but I think the terse version is not too ambiguous. 

### Technical details

I am not very familiar with C++, but I did have some prior practice for this task from implementing my [own ODBC driver in Zig](https://github.com/ffelixg/zodbc_py) (a very good language for projects like this!). The implementation is written almost entirely in C++ in the `FetchArrowBatch_wrap` function, which produces PyCapsules that are then consumed by `arrow_batch` and turned into actual arrow objects.

The function itself is very large. I'm sure it could be factored in a better way, even sharing some code with the other methods of fetching, but my goal was to keep the whole thing as straight forward as possible.

I have also implemented my own loop for SQLGetData for Lob-Columns. Unlike with the python fetch methods, I don't use the result directly, but instead copy it into the same buffer I would use for the case with bound columns. Maybe that's an abstraction that would make sense for that case as well.

### Notes on data types

I noticed that you use SQL_C_TYPE_TIME for time(x) columns. The arrow fetch does the same, but I think it would be better to use SQL_C_SS_TIME2, since that supports fractional seconds.

Datetimeoffset is a bit tricky, since SQL Server stores timezone information alongside each cell, while arrow tables expect a fixed timezone for the entire column. I don't really see any solution other than converting everything to UTC and returning a UTC column, so that's what I did.

SQL_C_CHAR columns get copied directly into arrow utf8 arrays. Maybe some encoding options would be useful.

### Performance

I think the main performance win to be gained is not interacting with any Python data structures in the hot path. That is satisfied. Further optimizations, which I did not make are:

*   Releasing the GIL for the entire fetch loop
*   Sharing the bound fetch buffer across repeated fetch calls
*   Improve the hot loop switching

Instead of looping over rows and columns and then switching on the data type for each cell, you could

*   Put the row loop inside each switch case (fastest I think, but would bloat the code a lot more)
*   Use function pointers like you recently did for python fetching (has overhead because of the indirect function call I think, also code is more scattered)
*   Replace both loops and the switch with computed gotos. That's what I opted for in my ODBC driver (the Zig equivalent is a labeled switch) and I am quite happy with how it came out. Performance seems very good and it allows you to abstract the fetching process on a row by row basis. I don't know how well that would translate to C++.

Overall the arrow performance seems not too far off from what I achieved with zodbc.


mssql_python/cursor.py
```diff
@@ -2198,6 +2201,89 @@ def fetchall(self) -> List[Row]:
...
+
+        batches: list["pyarrow.RecordBatch"] = []
+        while True:
+            batch = self.arrow_batch(batch_size)
+            if batch.num_rows < batch_size or batch_size <= 0:
+                if not batches or batch.num_rows > 0:
+                    batches.append(batch)
+                break
+            batches.append(batch)
+        return pyarrow.Table.from_batches(batches, schema=batches[0].schema)
```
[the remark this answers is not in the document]
ffelixg: Not true, `if not batches or batch.num_rows > 0` ensures there is always at least one batch. There is also `test_arrow_empty_result_set`, which tests against the suggested counterexample

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4003,956 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+    tm_date.tm_mon = month - 1;    // tm_mon is 0-11
+    tm_date.tm_mday = day;
+
+    std::time_t time_since_epoch = std::mktime(&tm_date);
+    if (time_since_epoch == -1) {
+        LOG("Failed to convert SQL_DATE_STRUCT to time_t");
+        ThrowStdException("Date conversion error");
+    }
+    // Calculate days since epoch
+    return time_since_epoch / 86400;
```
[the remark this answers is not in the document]
ffelixg: I'm not sure if this is true. The tests don't indicate such an issue.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4003,956 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+    auto arrow_array_batch = new ArrowArray({
+        .length = static_cast<int64_t>(idxRowArrow),
+        .n_buffers = 1,
+        .n_children = numCols,
+        .buffers = arrow_array_batch_buffers,
+        .children = new ArrowArray* [numCols],
+        .release = ArrowArray_release,
+    });
+    // Necessary dummy buffer
+    arrow_array_batch->buffers[1] = new int[1];
```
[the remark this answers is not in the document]
ffelixg: Solved by using private_data & unique pointers instead of free in release callback

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4003,956 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+            case SQL_WLONGVARCHAR:
+            case SQL_GUID:
+            case SQL_BINARY:
+            case SQL_VARBINARY:
+            case SQL_LONGVARBINARY: {
+                assert(buffersArrow.var[col][0] == 0);
+                // length of string at index i is the difference between values at i and i+1
+                // so total length is value at index idxRowArrow
+                auto data_buf_len_total = buffersArrow.var[col][idxRowArrow];
+                uint8_t* dataBuffer = new uint8_t[data_buf_len_total];
```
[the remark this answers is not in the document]
ffelixg: Solved by using private_data & unique pointers instead of free in release callback

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4003,960 @@
...
+            .format = columnFormats[i].release(),
+            .name = columnNamesCStr[i].release(),
+            .flags = columnNullable[i] ? 2 : 0, // ARROW_FLAG_NULLABLE
+            .release = ArrowSchema_release,
+        });
+        batch_children[i] = arrow_schema;
+    }
+
+    auto arrow_schema_batch = new ArrowSchema({
+        .format = []{ char* f = new char[3]; std::strcpy(f, "+s"); return f; }(),
```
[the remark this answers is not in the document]
ffelixg: Removed

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4003,960 @@
...
+            case SQL_GUID:
+            case SQL_BINARY:
+            case SQL_VARBINARY:
+            case SQL_LONGVARBINARY: {
+                assert(buffersArrow.var[col][0] == 0);
+                // length of string at index i is the difference between values at i and i+1
+                // so total length is value at index idxRowArrow
+                auto data_buf_len_total = buffersArrow.var[col][idxRowArrow];
+                uint8_t* dataBuffer = new uint8_t[data_buf_len_total];
+                std::memcpy(dataBuffer, buffersArrow.var_data[col].data(), data_buf_len_total);
```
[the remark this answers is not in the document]
ffelixg: I can't think of a way to eliminate any of these memcpys

ffelixg:
Hello @sumitmsft,

I'm happy to hear that.

1. I don't have any design document beyond what I wrote in the PR description. Are there any areas in particular you would like me to provide more information on?
2. I assume the motivation is mostly in line with what most arrow users like about arrow. Mainly I believe that arrow is the correct format for anything that is using batches of data and has C-Extensions for both producer and consumer. For example arrow gives you great interop with things like duckdb, polars, pandas on the analytics/ML side. Also I want python to be the obvious one stop shop for ETL workloads and for that, plain python types don't work well both for performance and reliability. You still have plenty of situations though where you want to fetch one result set with python types and the next with arrow types, so it has to be in the same driver as well.
3. Yes, for sure. I have sent you an Email.

Regards,
Felix

bewithgaurav (changes requested):
@ffelixg - Thanks for the contribution! :)
Before we get started on this PR - there are a few dev build workflows we need to fix.
Could you please take a look at the Azure DevOps Workflows which are failing? (goes by the check [MSSQL-Python-PR-Validation](https://github.com/microsoft/mssql-python/pull/354/checks?check_run_id=57430108121)):

- Build issues on windows ([ref](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=132737&view=logs&j=102d2815-99f5-500b-5c4e-56465a96d55f&t=bbc97363-a679-59a0-323f-f9dd63558a62)) 
- A few tests failing on MacOS ([ref](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=132737&view=logs&j=48594931-6250-55fa-3f5d-cfde1f0326c3&t=632523ab-df87-5aff-a272-1650b6b0b41b))

mssql_python/pybind/ddbc_bindings.cpp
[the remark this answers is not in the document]
ffelixg: Ended up going this route because of apparent cross platform issues with std::mktime

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4002,1037 @@
...
+                hasLobColumns = true;
+                if (fetchSize > 1) {
+                    fetchSize = 1; // LOBs require row-by-row fetch
+                }
+        }
+
+        std::string columnName = colMeta["ColumnName"].cast<std::string>();
+        size_t nameLen = columnName.length() + 1;
+        arrowSchemaPrivateData[i]->name = std::make_unique<char[]>(nameLen);
+        std::memcpy(arrowSchemaPrivateData[i]->name.get(), columnName.c_str(), nameLen);
```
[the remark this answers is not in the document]
ffelixg: Not avoidable I think

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4002,1037 @@
...
+                arrowColumnProducer->ptrValueBuffer = arrowColumnProducer->float64Val.get();
+                break;
+            case SQL_DECIMAL:
+            case SQL_NUMERIC: {
+                std::ostringstream formatStream;
+                formatStream << "d:" << columnSize << "," << colMeta["DecimalDigits"].cast<SQLSMALLINT>();
+                std::string formatStr = formatStream.str();
+                size_t formatLen = formatStr.length() + 1;
+                arrowSchemaPrivateData[i]->format = std::make_unique<char[]>(formatLen);
+                std::memcpy(arrowSchemaPrivateData[i]->format.get(), formatStr.c_str(), formatLen);
```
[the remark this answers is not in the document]
ffelixg: Not avoidable I think

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4002,1037 @@
...
+                ThrowStdException(errorString.str());
+                break;
+        }
+        
+        // Store format string if not already stored.
+        // For non-decimal types, format is now a static string.
+        if (!arrowSchemaPrivateData[i]->format) {
+            size_t formatLen = format.length() + 1;
+            arrowSchemaPrivateData[i]->format = std::make_unique<char[]>(formatLen);
+            std::memcpy(arrowSchemaPrivateData[i]->format.get(), format.c_str(), formatLen);
```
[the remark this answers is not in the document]
ffelixg: Not avoidable I think

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4002,1037 @@
...
+                    case SQL_VARBINARY:
+                    case SQL_LONGVARBINARY: {
+                        uint64_t fetchBufferSize = columnSize /* bytes are not null terminated */;
+                        auto target_vec = &arrowColumnProducer->varData;
+                        auto start = arrowColumnProducer->varVal[idxRowArrow];
+                        while (target_vec->size() < start + dataLen) {
+                            target_vec->resize(target_vec->size() * 2);
+                        }
+
+                        std::memcpy(&(*target_vec)[start], &buffers.charBuffers[idxCol][idxRowSql * fetchBufferSize], dataLen);
```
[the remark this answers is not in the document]
ffelixg: Not avoidable I think

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4002,1037 @@
...
+                    case SQL_VARCHAR:
+                    case SQL_LONGVARCHAR: {
+                        uint64_t fetchBufferSize = columnSize + 1 /* null-termination */;
+                        auto target_vec = &arrowColumnProducer->varData;
+                        auto start = arrowColumnProducer->varVal[idxRowArrow];
+                        while (target_vec->size() < start + dataLen) {
+                            target_vec->resize(target_vec->size() * 2);
+                        }
+
+                        std::memcpy(&(*target_vec)[start], &buffers.charBuffers[idxCol][idxRowSql * fetchBufferSize], dataLen);
```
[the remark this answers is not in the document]
ffelixg: Not avoidable I think

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3926,6 +4002,1037 @@
...
+                        }
+                        WideCharToMultiByte(CP_UTF8, 0, wcharSource, dataLenW, &(*target_vec)[start], dataLenConverted, NULL, NULL);
+                        arrowColumnProducer->varVal[idxRowArrow + 1] = start + dataLenConverted;
+#else
+                        // On Unix, use the SQLWCHARToWString utility and then convert to UTF-8
+                        std::string utf8str = WideToUTF8(SQLWCHARToWString(wcharSource, dataLenW));
+                        while (target_vec->size() < start + utf8str.size()) {
+                            target_vec->resize(target_vec->size() * 2);
+                        }
+                        std::memcpy(&(*target_vec)[start], utf8str.data(), utf8str.size());
```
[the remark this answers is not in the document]
ffelixg: Not avoidable I think

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4082,6 +4202,1035 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+                    case SQL_SMALLINT:
+                        arrowColumnProducer->int16Val[idxRowArrow] = buffers.smallIntBuffers[idxCol][idxRowSql];
+                        break;
+                    case SQL_INTEGER:
+                        arrowColumnProducer->int32Val[idxRowArrow] = buffers.intBuffers[idxCol][idxRowSql];
+                        break;
+                    case SQL_BIGINT:
+                        arrowColumnProducer->int64Val[idxRowArrow] = buffers.bigIntBuffers[idxCol][idxRowSql];
+                        break;
+                    case SQL_REAL:
```
gargsaumya, line 5164: In the LOB fetch path (line 4660), SQL_REAL is correctly fetched into buffers.realBuffers as SQLREAL (4-byte [float]). But here SQL_REAL falls through with SQL_FLOAT/SQL_DOUBLE and reads from buffers.doubleBuffers which is an 8-byte double buffer that was never populated for this column. This might produce garbage values for any real column when a LOB column is also present in the result set.
Probably we can give SQL_REAL its own case.

mssql_python/cursor.py
```diff
@@ -2397,6 +2400,89 @@ def fetchall(self) -> List[Row]:
...
+        query result set.
+
+        Args:
+            batch_size: Maximum number of rows to fetch in the Record Batch.
+
+        Returns:
+            A pyarrow RecordBatch object containing up to batch_size rows.
+        """
+        self._check_closed()  # Check if the cursor is closed
+        if not self._has_result_set and self.description:
```
gargsaumya, line 2549: arrow_batch checks self._has_result_set and calls self._reset_rownumber() when it's False, but never sets it to True afterward. This means every call in a loop (e.g., from arrow()) hits _reset_rownumber() repeatedly.
Other fetch methods (fetchone, fetchmany, etc.) set self._has_result_set = True after the first successful fetch. Should this do the same?

mssql_python/cursor.py
```diff
@@ -2397,6 +2400,89 @@ def fetchall(self) -> List[Row]:
...
+        Fetch the entire result as a pyarrow Table.
+
+        Args:
+            batch_size: Size of the Record Batches which make up the Table.
+
+        Returns:
+            A pyarrow Table containing all remaining rows from the result set.
+        """
+        try:
+            import pyarrow
```
gargsaumya: arrow_batch() calls self._check_closed() at the top, but arrow() and arrow_reader() don't. If pyarrow isn't installed and the cursor is closed, the user may get ImportError instead of the more helpful "cursor is closed" error.
Maybe we should add self._check_closed() before the import pyarrow in both methods?

requirements.txt
```diff
@@ -4,6 +4,7 @@ pytest-cov
 coverage
 unittest-xml-reporting
 psutil
+pyarrow
```
gargsaumya, line 7: The Python code already handles ImportError gracefully with a helpful message. Should this be under an optional extras group in setup.py (e.g., pip install mssql-python[arrow]) instead of being a hard requirement for all users?
For the test file it's fine, but this requirements.txt drives the project's general dependencies, not just test deps.

@sumitmsft @bewithgaurav thoughts?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -157,6 +157,126 @@ struct NumericData {
...
+struct ArrowArrayPrivateData {
+    std::unique_ptr<uint8_t[]> valid;
+
+    std::unique_ptr<uint8_t[]> uint8Val;
+    std::unique_ptr<int16_t[]> int16Val;
+    std::unique_ptr<int32_t[]> int32Val;
+    std::unique_ptr<int64_t[]> int64Val;
+    std::unique_ptr<double[]> float64Val;
+    std::unique_ptr<uint8_t[]> bitVal;
+    std::unique_ptr<uint32_t[]> varVal;
```
gargsaumya: varVal is uint32_t[], making the offset buffer a 32-bit offset array. As per the Arrow spec, this means variable-length columns are capped at ~4 GB of data per batch. That's probably fine for typical use, but it could overflow silently with large batch_size + large varchar(max) data.
For example, if someone fetches 8,192 rows and each row contains a 1 MB varchar, that results in ~8 GB of data. In that case, the uint32 offsets would wrap around silently, and pyarrow would end up reading corrupted data.

Would it make sense to add a runtime check to ensure the total accumulated byte size does not exceed UINT32_MAX? Please correct me if I’m misunderstanding anything here.

gargsaumya (commented):
This is a really solid contribution. The Arrow C Data Interface implementation is well structured, the PyCapsule ownership chain is handled correctly with RAII and proper try/catch guards, and the test coverage is comprehensive across a wide range of SQL types.
I’ve added a few non-blocking suggestions for improvement.

Thanks, @ffelixg , for the great work on this! 🎉

ffelixg:
Thank you for the kind words and for the careful review @gargsaumya! I think they are all great points and I have made changes to address each one.

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: Good catch, I totally missed this case somehow. I've mapped it to arrow's float32 and added a test case for it.

mssql_python/cursor.py
ffelixg: `_reset_rownumber` actually takes care of setting it to True, but the rownumber wasn't being updated correctly by arrow_batch. I've included the logic from fetchmany to take care of that.

mssql_python/cursor.py
ffelixg: I don't have a preference either way, but I can see how this would be more consistent behavior. I've made the change.

requirements.txt
ffelixg: Yes, the idea is 100% that pyarrow is not a required dependency for most of mssql_python. From my understanding, the requirements.txt is actually only used by CI and the actual dependencies are given by the `install_requires` parameter for `setup`, correct?. PyTest etc are also listed in requirements.txt. Supporting the mssql-python[arrow] syntax is a good thing though, I've added it under the `extras_require` parameter.

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: Your understanding is totally correct and I can see how it could be an annoying limitation if you put it like that. There is another option as well, which is to use the large_string/large_binary arrow types, which are indexed by uint64_t's. This can potentially double memory usage if there are a lot of nulls, but probably safe is better than sorry. I've opted for that route for now.

A possible optimization could be to use the smaller type if the column size is sufficiently small.

mssql_python/cursor.py
```diff
@@ -2397,6 +2400,110 @@ def fetchall(self) -> List[Row]:
...
+        Args:
+            batch_size: Maximum number of rows to fetch in the Record Batch.
+
+        Returns:
+            A pyarrow RecordBatch object containing up to batch_size rows.
+        """
+        self._check_closed()  # Check if the cursor is closed
+
+        try:
+            import pyarrow
```
sumitmsft: There are repeated `import pyarrow` / `except ImportError` pattern across all three methods (`arrow_batch`, `arrow`, `arrow_reader`). 

Can we extract a helper method `_ensure_pyarrow()` that performs the import check once and caches the module, or rely on the top-level conditional import already present.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4097,6 +4218,1041 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+                    case SQL_VARBINARY:
+                    case SQL_LONGVARBINARY: {
+                        uint64_t fetchBufferSize = columnSize /* bytes are not null terminated */;
+                        auto target_vec = &arrowColumnProducer->varData;
+                        auto start = arrowColumnProducer->varVal[idxRowArrow];
+                        while (target_vec->size() < start + dataLen) {
+                            target_vec->resize(target_vec->size() * 2);
+                        }
+
+                        std::memcpy(&(*target_vec)[start], &buffers.charBuffers[idxCol][idxRowSql * fetchBufferSize], dataLen);
```
sumitmsft, line 5073: There are multiple memcpy calls that were flagged by the code scanning tools. While you noted these are unavoidable for this type of data manipulation, several of the memcpy calls copy data from ODBC driver buffers without explicit pre-validation of the dataLen indicator against the destination buffer capacity. 

Can we add assertions or checks before memcpy calls to validate that dataLen does not exceed the allocated buffer size?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4097,6 +4218,1041 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+    SQLSetStmtAttr_ptr(hStmt, SQL_ATTR_ROW_ARRAY_SIZE, (SQLPOINTER)(intptr_t)fetchSize, 0);
+    SQLSetStmtAttr_ptr(hStmt, SQL_ATTR_ROWS_FETCHED_PTR, &numRowsFetched, 0);
+
+    size_t idxRowArrow = 0;
+    // arrowBatchSize % fetchSize == 0 ensures that any followup (even non-arrow) fetches
+    // start with a fresh batch
+    assert(fetchSize == 0 || arrowBatchSize % fetchSize == 0);
+    assert(fetchSize <= arrowBatchSize);
+
+    while (idxRowArrow < arrowBatchSize) {
```
sumitmsft, line 4745: `numRowsFetched` is a stack-local variable whose address is handed to the ODBC driver. The cleanup that resets `SQL_ATTR_ROWS_FETCHED_PTR` to `NULL` and `SQL_ATTR_ROW_ARRAY_SIZE` back to 1 only runs on the normal exit path. All the early `return ret` statements inside the fetch loop skip this cleanup entirely.

After an early exit, `numRowsFetched` is destroyed, but the driver still holds its address. The next `SQLFetch` on the same `hStmt` writes to invalid stack memory, undefined behavior (silent corruption, random crashes). `SQL_ATTR_ROW_ARRAY_SIZE` also stays elevated, breaking subsequent non-Arrow fetches.

I suggest using RAII guard: to ensure `stmt` attributes are always reset, even on early return or exception. 

Then the manual cleanup block near the end (// Reset attributes before returning...) can be removed, the destructor handles all exit paths automatically.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4097,6 +4218,1041 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+                        case SQL_WLONGVARCHAR: {
+                            ret = GetDataVar(
+                                hStmt,
+                                idxCol + 1,
+                                SQL_C_WCHAR,
+                                buffers.wcharBuffers[idxCol],
+                                buffers.indicators[idxCol].data()
+                            );
+                            if (!SQL_SUCCEEDED(ret)) {
+                                LOG("Error fetching binary data for column %d", idxCol + 1);
```
sumitmsft: The error message for WCHAR LOB fetch failure - So, the error message must be: `Error fetching WCHAR data for column %d`

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4097,6 +4218,1041 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+            arrowSchemaPrivateData[i]->format = std::make_unique<char[]>(formatLen);
+            std::memcpy(arrowSchemaPrivateData[i]->format.get(), format.c_str(), formatLen);
+        }
+
+        arrowColumnProducer->valid = std::make_unique<uint8_t[]>((arrowBatchSize + 7) / 8);
+        // Initialize validity bitmap to all valid
+        std::memset(arrowColumnProducer->valid.get(), 0xFF, (arrowBatchSize + 7) / 8);
+    }
+
+    if (fetchSize > 1) {
```
sumitmsft: The fetch size selection loop iterates from min(64, arrowBatchSize) down to 1, looking for a divisor of arrowBatchSize. While correct, if arrowBatchSize is prime and > 64, fetchSize will always be 1, leading to extremely slow row-by-row fetching. For example, arrowBatchSize=8191 (prime) would result in fetchSize=1.

Can we take different approach, say, find the largest divisor ≤ 64 of arrowBatchSize, but if none exists > 1, use a reasonable non-divisor size and handle the remainder differently, or document this behavior so users pick good batch sizes.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4097,6 +4218,1041 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
...
+                            break;
+                        default:
+                            break;
+                    }
+
+                    nullCounts[idxCol] += 1;
+                    continue;
+                } else if (indicator < 0) {
+                    // Negative value is unexpected, log column index, SQL type & raise exception
+                    LOG("Unexpected negative data length. Column ID - {}, SQL Type - {}, Data Length - {}", idxCol + 1, dataType, indicator);
```
sumitmsft: This may produce garbled log output since the LOG macro uses printf-style formatting.
Instead, we may use: 
`LOG("Unexpected negative data length. Column ID - %d, SQL Type - %d, Data Length - %lld", idxCol + 1, dataType, (long long)indicator);`

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: In the example you quoted (and a few other places), the assertion I would write would be:

```c
assert(target_vec->size() >= start + dataLen);
```

The while loop in the line above already has
```c
while (target_vec->size() < start + dataLen)
```
as the loop condition, so we already ensured that there is space for `memcpy`. If you think the assertion helps readability, I can add it.

For name and format string, we know the exact string length before allocating and copying, so the buffer fits exactly and is allocated in the preceding line. Maybe there is a more elegant way to do it, but the bounds checks would be a bit redundant there as well.

mssql_python/cursor.py
ffelixg: I've implemented `self._ensure_pyarrow()` In terms of caching, we can rely on the [builtin module cache](https://docs.python.org/3/reference/import.html#the-module-cache) I think. The top-level pyarrow import only counts for type checking, so it's not actually executed at runtime.

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: I agree with what you're saying and I've implemented the RAII guard. The same reasoning applies to `SQLFreeStmt_ptr(hStmt, SQL_UNBIND)`, so I've included that as well. As of now, all the fetch functions freshly configure these variables when they're called, so I don't think this could have lead to an actual issue.

Note that I've simply copied that part from the other fetch functions like fetchmany, so maybe we should have the same update there?

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: Thanks, I put in WCHAR LOB and also adjusted the messages in the other two GetDataVar cases accordingly.

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: I've made the change to where the fetch size is potentially updated for the final call. I originally wanted to avoid that for simplicity, but it actually turned out simpler than the gcd calculation I think. The call to update this size should be trivial performance wise.

In an optimal world, I think we would have a shared buffer for the entire cursor, which is used by all fetch functions. That way we could always fetch the same sized batch and - if one fetch call doesn't fully consume it - the next fetch call will start where the previous one left off. For fetchone this would also open the door to using bound columns, resulting in big performance gains. For bigger batch sizes it would still simplify logic.

The difference in performance between using SQLGetData and SQLBindCol with size 1 is bigger than between SQLBindCol with size 1 and SQLBindCol with larger sizes from what I can tell.

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: Thanks, fixed

gargsaumya (approved):

sumitmsft (approved):

bewithgaurav (approved):

gargsaumya:
Hi @ffelixg , 
We have merged this PR, thank you for the substantial work you’ve put into this feature, it’s much appreciated.

Could you share some performance comparisons for this change? It would be helpful to see metrics (e.g., before vs after numbers, benchmarks, or any relevant stats) to better understand the impact this feature is delivering.

Issue #341, opened 2025-11-21
Segmentation Fault in libmsodbcsql-18.5 during SQLFreeHandle()


zzzeek:
Hi mssql-python team, apologies that I used Claude to analyze this segfault and prepare the bug report. This is a segfault I can reproduce intermittently with the SQLAlchemy test suite within some tests where we are having connections raise errors and then throwing them away, but I can't easily work out a simple reproducer. Hopefully the core analysis here is enough to find the issue however I can work out the full steps to install and run SQLAlchemy's test suite if needed.

- mike (a human)

claude content follows below:

# Segmentation Fault in libmsodbcsql-18.5 during SQLFreeHandle()

## Summary

The Microsoft ODBC Driver for SQL Server 18.5 is crashing with a segmentation fault during `SQLFreeHandle()` when called from Python garbage collection finalizers. This is an intermittent crash that occurs during cleanup of `SqlHandle` objects.

## Environment

- **OS**: Fedora Linux 41 (Kernel 6.17.7-100.fc41.x86_64)
- **Python**: 3.14.0
- **mssql_python**: 1.0.0
- **ODBC Driver**: Microsoft ODBC Driver for SQL Server 18.5
- **Driver Library**: libmsodbcsql-18.5.so.1.1

## Crash Location

**Module**: `libmsodbcsql-18.5.so.1.1`
**Function**: Inside `SQLFreeHandle()` at offset `+0x5a4bc`
**Signal**: SIGSEGV (Segmentation Fault)

## Reproduction

The crash occurs intermittently when running:
```bash
pytest test/engine/test_reconnect.py --db mssql_python --disable-asyncio
```

This is being tested in the SQLAlchemy test suite with connection pooling and reconnection logic.

## Root Cause Analysis

From the core dump analysis, the crash follows this pattern:

1. Python's garbage collector runs (`gc_collect_region`)
2. Finalizer is invoked on a `SqlHandle` object (`slot_tp_finalize`)
3. The finalizer calls `SqlHandle::free()` via pybind11
4. This calls ODBC's `SQLFreeHandle()`
5. **SEGFAULT occurs inside `SQLFreeHandle()` at offset `0x5a4bc`**

## Stack Trace (Key Frames)

```
#0  0x00007f2aa6d097d4 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007f2aa6cb039e in raise () from /lib64/libc.so.6
#2  <signal handler called>
[...]

PR #361, opened 2025-12-04
FIX: Fixing segmentation fault issue due to double free on SqlHandle::free in Linux

subrata-ms:
> GitHub Issue: #341

### Summary   

**Problem:** Segmentation fault occurs during Python's garbage collection when freeing ODBC handles.

**Stack Trace Analysis:**

0-2:  Signal handler (SIGSEGV)
3:    libmsodbcsql-18.5.so.1.1 - CRASH LOCATION
4:    SQLFreeHandle() from ODBC driver
5:    SqlHandle::free() from ddbc_bindings
6-11: pybind11 call stack
12:   Python method call
20:   slot_tp_finalize during __del__
21-27: Python GC finalizing objects

**Root Cause:**  The crash occurs when;
1. Python's garbage collector runs during finalization
2. Objects with [__del__] methods are being cleaned up
3. The [__del__] calls [close()] which internally calls SQLFreeHandle()
4. The ODBC driver crashes because:
-     Handle is already freed (double-free)
-     Handle is in invalid state
-     Connection handle freed before cursor statement handle
-     Wrong finalization order due to circular references

**Fix Details:**
This pull request introduces a critical fix to the handle cleanup logic in the SQL bindings for Python, specifically addressing potential segfaults during interpreter shutdown. The change ensures that both statement and database connection handles are not freed if Python is shutting down, preventing invalid memory access.

Handle cleanup logic improvements:

* Updated the `SqlHandle::free()` method in `mssql_python/pybind/ddbc_bindings.cpp` to skip freeing both statement (`SQL_HANDLE_STMT`) and database connection (`SQL_HANDLE_DBC`) handles during Python shutdown, rather than only statement handles. This prevents segfaults caused by freeing handles in the wrong order when their parent resources may have already been released.

### DIFF
--- a/OneBranchPipelines/github-ado-sync.yml
+++ b/OneBranchPipelines/github-ado-sync.yml
@@ -1,0 +1,127 @@
+# GitHub-to-ADO Sync Pipeline
+# Syncs main branch from public GitHub to internal Azure DevOps daily at 5pm IST
+
+name: GitHub-Sync-$(Date:yyyyMMdd)$(Rev:.r)
+
+schedules:
+  - cron: "30 11 * * *"
+    displayName: "Daily sync at 5pm IST"
+    branches:
+      include:
+        - main
+    always: true
+
+trigger: none
+pr: none
+
+jobs:
+- job: SyncFromGitHub
+  displayName: 'Sync main branch from GitHub'
+  pool:
+    vmImage: 'windows-latest'
+  
+  steps:
+  - checkout: none
+
+  - task: CmdLine@2
+    displayName: 'Clone GitHub repo'
+    inputs:
+      script: git clone https://github.com/microsoft/mssql-python.git repo-dir -b main
+      workingDirectory: $(Agent.TempDirectory)
+
+  - task: CmdLine@2
+    displayName: 'Add Azure DevOps remote'
+    inputs:
+      script: git remote add azdo-mirror https://$(System.AccessToken)@sqlclientdrivers.visualstudio.com/mssql-python/_git/mssql-python
+      workingDirectory: $(Agent.TempDirectory)/repo-dir
+
+  - task: CmdLine@2
+    displayName: 'Fetch ADO repo'
+    inputs:
+      script: git fetch azdo-mirror
+      workingDirectory: $(Agent.TempDirectory)/repo-dir
+
+  - task: CmdLine@2
+    displayName: 'Create timestamped sync branch'
+    inputs:
+      script: |
+        echo Getting current timestamp...
+        powershell -Command "Get-Date -Format 'yyyyMMdd-HHmmss'" > timestamp.txt
+        set /p TIMESTAMP=<timestamp.txt
+        set SYNC_BRANCH=github-sync-%TIMESTAMP%
+        echo %SYNC_BRANCH% > branchname.txt
+        echo Creating sync branch: %SYNC_BRANCH%
+        git fetch azdo-mirror
+        git show-ref --verify --quiet refs/remotes/azdo-mirror/main
+        if %ERRORLEVEL% EQU 0 (
+          git checkout -b %SYNC_BRANCH% -t azdo-mirror/main
+        ) else (
+          echo azdo-mirror/main does not exist. Exiting.
+          exit /b 1
+        )
+        echo ##vso[task.setvariable variable=SYNC_BRANCH;isOutput=true]%SYNC_BRANCH%
+      workingDirectory: $(Agent.TempDirectory)/repo-dir
+
+  - task: CmdLine@2
+    displayName: 'Reset branch to match GitHub main exactly'
+    inputs:
+      script: |
+        git -c user.email="sync@microsoft.com" -c user.name="ADO Sync Bot" reset --hard origin/main
+      workingDirectory: $(Agent.TempDirectory)/repo-dir
+
+  - task: CmdLine@2
+    displayName: 'Push branch to Azure DevOps'
+    inputs:
+      script: |
+        set /p SYNC_BRANCH=<branchname.txt
+        echo Pushing branch: %SYNC_BRANCH%
+        
+        git config user.email "sync@microsoft.com"
+        git config user.name "ADO Sync Bot"
+        
+        git push azdo-mirror %SYNC_BRANCH% --set-upstream
+        
+        if %ERRORLEVEL% EQU 0 (
+          echo Branch pushed successfully!
+        ) else (
+          echo ERROR: Push failed!
+          exit /b 1
+        )
+      workingDirectory: $(Agent.TempDirectory)/repo-dir
+
+  - task: CmdLine@2
+    displayName: 'Create pull request'
+    inputs:
+      script: |
+        echo Installing Azure DevOps extension...
+        call az extension add --name azure-devops --only-show-errors
+        
+        echo Setting up authentication...
+        set AZURE_DEVOPS_EXT_PAT=$(System.AccessToken)
+        
+        echo Configuring Azure DevOps defaults...
+        call az devops configure --defaults organization=https://sqlclientdrivers.visualstudio.com project=mssql-python
+        
+        set /p SYNC_BRANCH=<branchname.txt
+        echo Creating PR from branch: %SYNC_BRANCH% to main
+        
+        call az repos pr create --source-branch %SYNC_BRANCH% --target-branch main --title "ADO-GitHub Sync - %date%" --description "Automated sync from GitHub main branch" --auto-complete true --squash true --delete-source-branch true --output table
+        
+        if %ERRORLEVEL% EQU 0 (
+          echo PR created successfully!
+          echo Adding reviewers...
+          call az repos pr list --source-branch %SYNC_BRANCH% --target-branch main --status active --output tsv --query "[0].pullRequestId" > pr_id.txt
+          set /p PR_ID=<pr_id.txt
+          call az repos pr reviewer add --id %PR_ID% --reviewers gargsaumya@microsoft.com sharmag@microsoft.com sumit.sarabhai@microsoft.com jathakkar@microsoft.com spaitandi@microsoft.com --output table
+          if %ERRORLEVEL% EQU 0 (
+            echo Reviewers added successfully!
+          )
+          del pr_id.txt
+        ) else (
+          echo ERROR: Failed to create PR
+          exit /b 1
+        )
+        
+        if exist timestamp.txt del timestamp.txt
+        if exist branchname.txt del branchname.txt
+      workingDirectory: $(Agent.TempDirectory)/repo-dir


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1127,8 +1127,10 @@ void SqlHandle::free() {
         // CRITICAL FIX: During Python shutdown, don't free STMT handles as
         // their parent DBC may already be freed This prevents segfault when
         // handles are freed in wrong order during interpreter shutdown Type 3 =
```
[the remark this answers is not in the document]
sumitmsft: This make sense

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1127,8 +1127,10 @@ void SqlHandle::free() {
...
-        // SQL_HANDLE_STMT, Type 2 = SQL_HANDLE_DBC, Type 1 = SQL_HANDLE_ENV
-        if (pythonShuttingDown && _type == 3) {
+        // Type 3 = SQL_HANDLE_STMT (parent: DBC)
+        // Type 2 = SQL_HANDLE_DBC (parent: ENV, which is static and may destruct first)
+        // Type 1 = SQL_HANDLE_ENV (no parent)
+        if (pythonShuttingDown && (_type == 3 || _type == 2)) {
```
sumitmsft: Can you confirm if there could be a resource/memory leaks during interpreter shutdown?

So, here we skipped cleanup at shutdown which means any remaining ODBC handles that weren’t explicitly closed (via `.close()` or similar) may not be released properly. This can lead to memory or connection leaks, especially in long-lived or embedded scenarios or stress situations.

sumitmsft: Please add documentation explaining that explicit calls to `.close()` are strongly recommended for all connections/cursors, as relying solely on `__del__` may no longer guarantee cleanup under all scenarios. This will help downstream users avoid unexpected resource retention in their scripts, especially in long-running or embedded environments.

mssql_python/pybind/ddbc_bindings.cpp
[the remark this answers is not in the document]
subrata-ms: done

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: I have addressed this in latest commit.

subrata-ms: This is now handled via "atexit" (Latest commit) during python process shutdown explicitly. All ODBC handles will be explicitly and automatically cleaned up during python process shutdown ( atexit). In Other scenarios ( while python is not shutting down), client/user need to maintain the delete handles in proper sequence i.e                              STMT( type3)->DBC(type2)->Env(type1)

tests/test_013_SqlHandle_free_shutdown.py
```diff
@@ -0,0 +1,762 @@
...
+        assert "=== Comprehensive Handle Test ===" in result.stdout
+        assert "Scenario 1: Normal cleanup completed" in result.stdout
+        assert "Scenario 2: Cursor closed, connection left open" in result.stdout
+        assert "Scenario 3: Both cursor and connection left open" in result.stdout
+        assert "Scenario 4: Multiple cursors per connection left open" in result.stdout
+        assert "=== Exiting ===" in result.stdout
+        print(f"PASS: Comprehensive all handle types test")
+
+
+if __name__ == "__main__":
```
bewithgaurav: this block isn't needed since we have a conftest.py in pytests which take care of setup stuff
if you just run `python -m pytest -v tests\tes_013_....py` you would be able to run this file on its own
I believe we can remove the `__main__` calling of tests.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1124,11 +1124,22 @@ void SqlHandle::free() {
...
+        //
+        // RESOURCE LEAK MITIGATION:
+        // When handles are skipped during shutdown, they are not freed, which could
+        // cause resource leaks. However, this is mitigated by:
+        // 1. Python-side atexit cleanup (in __init__.py) that explicitly closes all
+        //    connections before shutdown, ensuring handles are freed in correct order
+        // 2. OS-level cleanup at process termination recovers any remaining resources
+        // 3. This tradeoff prioritizes crash prevention over resource cleanup, which
+        //    is appropriate since we're already in shutdown sequence
+        if (pythonShuttingDown && (_type == 3 || _type == 2)) {
```
sumitmsft: Instead of using 3 and 2: can we define them as contants?

const int SQL_HANDLE_STMT = 3;
const int SQL_HANDLE_DBC = 2;
const int SQL_HANDLE_ENV = 1;

Wouldn't that be a good option?

mssql_python/__init__.py
```diff
@@ -67,6 +69,40 @@
 # Pooling
 from .pooling import PoolingManager
 
+# Global registry for tracking active connections (using weak references)
+_active_connections = weakref.WeakSet()
```
sumitmsft, line 74: As per my understanding, using `WeakSet` allows connections to be garbage collected before `atexit` cleanup runs, which means the segfault still occurs during GC? 
The `atexit` cleanup runs too late and finds nothing?

I can imagine following scenario:
```py
def query():
    conn = connect(...)                           # Added to WeakSet
    return conn.cursor(). fetchall()
    # conn goes out of scope >>  GC collects it
    # SqlHandle::free() called >>  SEGFAULT!
# atexit runs later, finds nothing (weakref died)
```

mssql_python/__init__.py
```diff
@@ -67,6 +69,40 @@
...
+
+def _cleanup_connections():
+    """
+    Cleanup function called by atexit to close all active connections.
+
+    This prevents resource leaks during interpreter shutdown by ensuring
+    all ODBC handles are freed in the correct order before Python finalizes.
+    """
+    # Make a copy of the connections to avoid modification during iteration
+    connections_to_close = list(_active_connections)
```
sumitmsft: Seems that `_active_connections` is being modified from multiple threads without synchronization which may cause: RuntimeError or clean up issues or race conditions between registration and cleanup. 

Can you validate it?

mssql_python/__init__.py
```diff
@@ -67,6 +69,40 @@
...
+
+    for conn in connections_to_close:
+        try:
+            # Check if connection is still valid and not closed
+            if hasattr(conn, "_closed") and not conn._closed:
+                # Close will handle both cursors and the connection
+                conn.close()
+        except Exception:
```
sumitmsft: Are we purposefully silently ignoring errors during shutdown cleanup? If any issues happen here we will not have any visibility to those issues. Do you not anticipate any issue here?

mssql_python/connection.py
```diff
@@ -239,6 +239,17 @@ def __init__(
         )
         self.setautocommit(autocommit)
 
+        # Register this connection for cleanup before Python shutdown
+        # This ensures ODBC handles are freed in correct order, preventing leaks
+        try:
+            import mssql_python
```
sumitmsft: This is called for every connection. Isn't this inefficient?

mssql_python/connection.py
```diff
@@ -239,6 +239,17 @@ def __init__(
...
+
+            if hasattr(mssql_python, "_register_connection"):
+                mssql_python._register_connection(self)
+        except (ImportError, AttributeError):
```
sumitmsft: Silent failure without logging. 

Can we do something like:
```
except (ImportError, AttributeError) as e:
    import warnings ####  Can be on top of connection.py
    warnings.warn(f"Connection registration failed: {e}. Cleanup will rely on __del__", RuntimeWarning)
```

mssql_python/__init__.py
```diff
@@ -67,6 +69,40 @@
...
+    This prevents resource leaks during interpreter shutdown by ensuring
+    all ODBC handles are freed in the correct order before Python finalizes.
+    """
+    # Make a copy of the connections to avoid modification during iteration
+    connections_to_close = list(_active_connections)
+
+    for conn in connections_to_close:
+        try:
+            # Check if connection is still valid and not closed
+            if hasattr(conn, "_closed") and not conn._closed:
```
sumitmsft, line 98: This logic assumed that all connections have `_closed` attribute. Is there any possibility that future connection types don't have this attribute? They'll be silently skipped in that case?

mssql_python/pybind/ddbc_bindings.cpp
sumitmsft: Here we assumed python-side atexit cleanup will catch everything (but it might not if connections are created during shutdown). Also, OS will cleanup (true for processes, but not for embedded Python). 

In both these cases, there will definitely be some edge cases where Handle Leak may happen. 

What can be done to tackle these leaked handles?

tests/test_013_SqlHandle_free_shutdown.py
```diff
@@ -0,0 +1,1138 @@
+"""
```
sumitmsft, line 1: Can we add a few more test from robustness perspective:

1. Test thread safety when connections are created while shutdown starts. For ex: Thread 1-5 are creating connections, Main thread: Triggers shutdown
Verify: No race conditions, all connections tracked or cleaned

2. Test that _cleanup_connections() can be called multiple times safely. For ex: Create connections, Call _cleanup_connections() multiple times
Verify: No errors, connections cleaned only once

3. Stress Test: Test cleanup performance with many connections. For ex: Create 1000+ connections, Measure cleanup time, Verify: Cleanup completes within reasonable time (shouldn't take forever or even 20 seconds). It should be fairly fast

mssql_python/__init__.py
bewithgaurav: This case should not be a concern, since GC doesn't run when things go out of scope

The scenario describes normal operation (not shutdown), where GC runs
During normal operation, `pythonShuttingDown` is False, so handles are freed properly anyways
The atexit mechanism only matters during actual interpreter shutdown
However, the WeakSet design means connections collected BEFORE shutdown won't be in the set
So, during:

**Normal GC (during runtime):** Works fine, handles freed normally
**Shutdown GC:** WeakSet may miss some connections, but those would be caught by the C++ shutdown protection anyway

@subrata-ms - please add if my analysis here is incorrect
Please also add a small description in comment of why we're using WeakRefs here - and what purpose are they serving (similar to whats added in ddbc bindings)

mssql_python/__init__.py
bewithgaurav: This is a valid concern
Currently, `_active_connections` is a `WeakSet` that can be accessed from multiple threads without any locks:

**Thread 1** (creating a connection):
```python
# In connection.py
mssql_python._register_connection(self)  
  ↓
# In __init__.py
_active_connections.add(conn)  # ← No lock here!
```

**Thread 2** (also creating a connection at the same time):
```python
_active_connections.add(another_conn)  # ← Same WeakSet, no lock!
```

**Thread 3** (cleanup during shutdown):
```python
# In _cleanup_connections()
connections_to_close = list(_active_connections)  # ← Reading while others write!
```

**Why This Is Dangerous**

`WeakSet` operations in Python are **not thread-safe**. When multiple threads access it simultaneously:

1. **Race condition during `add()`**: Internal data structure corruption
2. **Race condition between iteration and modification**: You might iterate over a set while another thread is adding to it
3. **Potential crashes or data loss**: Missing connections in cleanup, or worse, Python internals corruption

**Real-World Scenario**

```python
# Thread A (web request 1)
conn1 = connect(...)  # Calls _register_connection(conn1)

# Thread B (web request 2) - happens at EXACT same time
conn2 = connect(...)  # Calls _register_connection(conn2)

# Thread C (shutdown signal received)
_cleanup_connections()  # Tries to iterate _active_connections

# Without locks:
# - Connections might be lost
# - Set might be corrupted  
# - Iteration might crash
```

**Potential Solution**

Add a lock to protect all access to `_active_connections`:

```python
import threading

_connections_lock = threading.Lock()
_active_connections = weakref.WeakSet()

def _register_connection(conn):
    """Register a connection for cleanup before shutdown."""
    with _connections_lock:  # ← Only one thread at a time
        _active_connections.add(conn)

def _cleanup_connections():
    """Cleanup function called by atexit to close all active connections."""
    with _connections_lock:  # ← Only one thread at a time
        connections_to_close = list(_active_connections)
    
    # Release lock before cleanup (don't hold lock during slow operations)
    for conn in connections_to_close:
        # ... cleanup code ...
```

Without the lock, even creating connections in different threads could corrupt the tracking set.

**Please also add testcases for the above multi-threaded scenarios**

mssql_python/__init__.py
```diff
@@ -4,8 +4,10 @@
 This module initializes the mssql_python package.
 """
 
+import atexit
```
bewithgaurav, line 7: **Move connection lifecycle management to connection.py**

The current implementation spreads connection lifecycle logic across two files (`__init__.py` and connection.py), requiring a circular import workaround with the lazy `import mssql_python` pattern inside `Connection.__init__()`.

**Recommend:** Move `_active_connections`, `_register_connection()`, and `_cleanup_connections()` to connection.py as module-level code. This would:
- Eliminate the circular dependency pattern
- Keep all connection lifecycle logic in one place
- Reduce bloat in `__init__.py` 
- Make the code more maintainable and easier to understand

The `__init__.py` should only handle module initialization and exports, not resource cleanup logic.

mssql_python/connection.py
bewithgaurav: +1, This is done to bypass circular import using lazy import and should be avoided.
This will be eliminated as a part of moving connection lifecycle management from init to this file

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: fixed.

mssql_python/__init__.py
subrata-ms: Ya this is a valid scenario. I have fixed the code and added a test for thread safe access to the list _active_connections

mssql_python/__init__.py
subrata-ms: fixed

mssql_python/connection.py
subrata-ms: fixed. Moved the import statement to module level

tests/test_013_SqlHandle_free_shutdown.py
subrata-ms: fixed.

mssql_python/connection.py
subrata-ms: fixed

mssql_python/__init__.py
subrata-ms: In Connection class (python), "self._closed" is the only attribute that currently denotes the state of the connection (true/false). The whole connection class is currently implemented based on this attribute. Removal of this attribute from this class triggers refactoring of respective functionality under connection class. The change looks good based on current architecture.

tests/test_013_SqlHandle_free_shutdown.py
subrata-ms: Added test cases for 1 & 2. For 3, I need to run stress test for connection/connection pool to identify any anomaly.

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: Good point. Our scenario is to protect the double free on ODBC handle. If python process is shutting down and there is connection handle getting created in ODBC and didn't get freed up by "atexit", then it should get cleaned up by garbage collector/parent process cleanup. We will discuss in detail and will cover if any possible edge case need to be added.

mssql_python/__init__.py
gargsaumya: We should avoid calling this as “_active_connections” in naming, because what we are actually tracking here are active ODBC handles that remain open until Python shutdown.

Also, if we move this into connection.py, it gives the impression that the list is tied to individual Connection instances, which isn’t accurate. These are module-level driver resources. With clearer naming (e.g., _active_odbc_handles or _open_driver_handles), it may make more sense to keep this in __init__.py.

@bewithgaurav what do you think of this?

gargsaumya (changes requested):

mssql_python/__init__.py
subrata-ms: Thanks @gargsaumya and @bewithgaurav for the detail review on the code component. However, the design and naming convention seems to be proper and intact. We will have detail discussion on this and will address the outcome with a follow-up PR if needed. My inputs on this would be as below - 
**1.** _active_connections is a global registry/container to hold reference of all the active connections using weak references. Hence we cannot move this inside connection object itself because this container does not belongs to any connection object. Naming convention also seems to be correct as it holds active connections references.

**2.** _register_connection(), and _cleanup_connections() are introduced to operate (kind of getter and setter) on the _active_connections container (list) which is a global entity. Hence it has to be outside of connection class.

mssql_python/__init__.py
subrata-ms: sounds good. Additionally, below is the sequence for python shutdown. Hence if any connections missed in step-1 (current implementation) then it potentially getting take care at step 5.
**Python’s shutdown sequence:**
1. Run atexit callbacks
2. Finalize modules
3. Release GIL
4. Shutdown interpreter
5. Then C++ destructors run

mssql_python/pybind/ddbc_bindings.cpp
sumitmsft: Where was this changed made @subrata-ms ?

subrata-ms: This already fixed. Somehow it is not reflecting here. Please review it in the ddbc_bindings.cpp and you should see the change as below.
        // RESOURCE LEAK MITIGATION:
        // When handles are skipped during shutdown, they are not freed, which could
        // cause resource leaks. However, this is mitigated by:
        // 1. Python-side atexit cleanup (in __init__.py) that explicitly closes all
        //    connections before shutdown, ensuring handles are freed in correct order
        // 2. OS-level cleanup at process termination recovers any remaining resources
        // 3. This tradeoff prioritizes crash prevention over resource cleanup, which
        //    is appropriate since we're already in shutdown sequence
        **if (pythonShuttingDown && (_type == SQL_HANDLE_STMT || _type == SQL_HANDLE_DBC)) {**
            _handle = nullptr;  // Mark as freed to prevent double-free attempts
            return;
        }

sumitmsft: Somehow, I was unable to see the change\commit. Now, its visible to me

mssql_python/__init__.py
bewithgaurav: @subrata-ms - understood your points, lets have a detailed discussion on this tomorrow and finalize this part

sumitmsft (approved):

bewithgaurav (approved):

Issue #295, opened 2025-10-15
setDecimalSeparator() causes DECIMAL values to become None


bewithgaurav:
### Describe the bug

When using `setDecimalSeparator()` to change the decimal separator from the default `.` (period) to any other character (e.g., `,` for European locale), DECIMAL values retrieved via **bulk fetch operations** (`fetchall()`, `fetchmany()` with large result sets) become `None`, resulting in data loss.

**Critical Finding:** The bug only affects the **bulk/columnar fetch path** in the C++ layer. Single-row fetches via `fetchone()` work correctly because they use a different code path that doesn't apply the separator during parsing.

The decimal separator setting is intended to affect **display formatting only** (in `Row.__str__()`), but the bulk fetch implementation incorrectly applies it during **parsing** of numeric values from the database.

**Exception/Error:**
No exception is raised. DECIMAL values silently become `None` in bulk fetch operations.

```
Expected: Decimal('1234.56')
Got:      None
```

### To reproduce

**Bug appears ONLY with bulk fetch operations** (`fetchall()`, `fetchmany()`). Single-row `fetchone()` works correctly.

Complete reproduction code:

```python
from mssql_python import connect, setDecimalSeparator, getDecimalSeparator
import os
from decimal import Decimal

conn_str = os.getenv("DB_CONNECTION_STRING")
conn = connect(conn_str)
cursor = conn.cursor()

# Create temp table with multiple rows
cursor.execute("DROP TABLE IF EXISTS #test_decimal")
cursor.execute("""
    CREATE TABLE #test_decimal (
        id INT,
        price DECIMAL(10, 2)
    )
""")

# Insert test data
for i in range(10):
    cursor.execute(f"INSERT INTO #test_decimal VALUES ({i}, {1234.56 + i})")

# Test 1: Bulk fetch with default separator works
print("Test 1: Bulk fetch with default separator")
setDecimalSeparator('.')
cursor.execute("SELECT id, price FROM #test_decimal ORDER BY id")
rows = cursor.fetchall()
print(f"Fetched {len(rows)} rows")
print(f"First row: {rows[0][1]}")  # Output: Decimal('1234.56') ✅
print(f"Type: {type(rows[0][1])}")

# Test 2: Bulk fetch with comma separator - BUG!
print("\nTest 2: Bulk fetch with comma separator")
setDecimalSeparator(',')
cursor.execute("SELECT id, price FROM #test_decimal ORDER BY id")
rows = cursor.fetchall()
[...]

PR #384, opened 2025-12-21
FIX: Remove unused decimal separator code and add regression tests

Chetan-Kansal:

GitHub Issue: #295

### Summary   
Removes misleading unused decimal-separator code from the bulk fetch path and adds a regression test to ensure DECIMAL values are parsed correctly regardless of setDecimalSeparator().

### DIFF
--- a/mssql_python/logging.py
+++ b/mssql_python/logging.py
@@ -17,7 +17,6 @@ import platform
 import atexit
 from typing import Optional
 
-
 # Single DEBUG level - all or nothing philosophy
 # If you need logging, you need to see everything
 DEBUG = logging.DEBUG  # 10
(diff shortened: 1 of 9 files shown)


tests/test_decimal_separator.py
```diff
@@ -0,0 +1,59 @@
+import os
+from decimal import Decimal
+
+import pytest
+import mssql_python
+
+
+@pytest.mark.skipif(not os.getenv("DB_CONNECTION_STRING"), reason="Requires DB_CONNECTION_STRING")
+def test_decimal_separator_bug(db_connection):
```
bewithgaurav: Great job on adding this focused regression test!
Suggestion: Instead of creating a new file `test_decimal_separator.py`, could we add these tests to `test_004_cursor.py`?

There are already decimal separator tests there (around lines 8370-8730), so it would keep all the related tests together. Plus, these are really testing cursor fetch behavior with the decimal separator setting, which fits perfectly in that file.

Issue #20, opened 2025-03-11
Support params dict


amachanic:
Instead of forcing ordered params, can you also allow a dict? This is a fairly common convention in Python DB libraries (see psycopg and pymssql for example) and is much nicer than passing a tuple.

Example:
```
cursor.execute(
    'select * from tbl where col1 = @a and col2 = @b',
    {'@a': 123, '@b': 234}
)
```

Note on this example: Many libraries (psycopg in particular) instead use a format like the following. I think that format would be useful too, but as SQL Server's variables are much more flexible it would be great to support them natively as in the above.

```
cursor.execute(
    'select * from tbl where col1 = %(a)s and col2 = %(b)s',
    {'a': 123, 'b': 234}
)
```

Thank you!

PR #385, opened 2025-12-22
FEAT: Param as Dict

jahnvi480:

> [AB#40995](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/40995)

> GitHub Issue: #20 

### Summary   
This pull request introduces support for both `qmark` (`?`) and `pyformat` (`%(name)s`) parameter styles in SQL queries, improving compatibility and usability for users of the `mssql_python` package. The default `paramstyle` is now set to `pyformat`, and both `execute` and `executemany` methods have been updated to automatically detect and convert parameter styles as needed. A new utility module, `parameter_helper.py`, has been added to handle parameter style parsing and conversion.

**Parameter style support and conversion:**

* Changed the global `paramstyle` in `mssql_python/__init__.py` from `"qmark"` to `"pyformat"`, making `pyformat` the default parameter style.
* Added a new module `mssql_python/parameter_helper.py` containing helper functions to parse pyformat parameters, convert pyformat SQL to qmark style, and auto-detect/convert parameter styles in queries.

**Enhancements to parameter handling in cursor methods:**

* Updated the `execute` method in `mssql_python/cursor.py` to auto-detect and convert parameter styles, supporting both single values and various parameter formats, and to use the new helper functions for conversion.
* Refactored parameter flattening logic in `execute` to rely on the new auto-detection and conversion, removing the old manual flattening.
* Enhanced the `executemany` method in `mssql_python/cursor.py` to auto-detect parameter style, convert pyformat to qmark for all rows if needed, and wrap single parameters for backward compatibility.

**Tests and validation:**

* Updated the `test_paramstyle` test in `tests/test_001_globals.py` to expect the new default `paramstyle` value of `"pyformat"`.

### DIFF
--- a/.github/actions/post-coverage-comment/action.yml
+++ b/.github/actions/post-coverage-comment/action.yml
@@ -1,0 +1,105 @@
+name: Post Coverage Comment
+description: Posts a standardized code coverage comment on a pull request
+
+inputs:
+  pr_number:
+    description: 'Pull request number'
+    required: true
+  coverage_percentage:
+    description: 'Overall coverage percentage'
+    required: true
+  covered_lines:
+    description: 'Number of covered lines'
+    required: true
+  total_lines:
+    description: 'Total number of lines'
+    required: true
+  patch_coverage_pct:
+    description: 'Patch/diff coverage percentage'
+    required: true
+  low_coverage_files:
+    description: 'Files with lowest coverage (multiline)'
+    required: true
+  patch_coverage_summary:
+    description: 'Patch coverage summary markdown (multiline)'
+    required: true
+  ado_url:
+    description: 'Azure DevOps build URL'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Post coverage comment
+      uses: marocchino/sticky-pull-request-comment@v2
+      with:
+        header: Code Coverage Report
+        number: ${{ inputs.pr_number }}
+        message: |
+          # 📊 Code Coverage Report
+          
+          <table>
+          <tr>
+          <td align="center" width="200">
+          
+          ### 🔥 Diff Coverage
+          ### **${{ inputs.patch_coverage_pct }}**
+          <br>
+          </td>
+          <td align="center" width="200">
+          
+          ### 🎯 Overall Coverage
+          ### **${{ inputs.coverage_percentage }}**
+          <br>
+          </td>
+          <td>
+          
+          **📈 Total Lines Covered:** `${{ inputs.covered_lines }}` out of `${{ inputs.total_lines }}`  
+          **📁 Project:** `mssql-python`  
+          
+          </td>
+          </tr>
+          </table>
+          
+          ---
+          
+          ${{ inputs.patch_coverage_summary }}
+          
+          ---
+          ### 📋 Files Needing Attention
+          
+          <details>
+          <summary>📉 <strong>Files with overall lowest coverage</strong> (click to expand)</summary>
+          <br>
+          
+          ```diff
+          ${{ inputs.low_coverage_files }}
+          ```
+          
+          </details>
+          
+          ---
+          ### 🔗 Quick Links
+
+          <table>
+          <tr>
+          <td align="left" width="200">
+          <b>⚙️ Build Summary</b>
+          </td>
+          <td align="left">
+          <b>📋 Coverage Details</b>
+          </td>
+          </tr>
+          <tr>
+          <td align="left" width="200">
+
+          [View Azure DevOps Build](${{ inputs.ado_url }})
+
+          </td>
+          <td align="left">
+
+          [Browse Full Coverage Report](${{ inputs.ado_url }}&view=codecoverage-tab) 
+
+          </td>
+          </tr>
+          </table>
--- a/.github/workflows/forked-pr-coverage.yml
+++ b/.github/workflows/forked-pr-coverage.yml
@@ -1,0 +1,111 @@
+name: Post Coverage Comment
+
+# This workflow handles posting coverage comments for FORKED PRs.
+#
+# Why a separate workflow?
+# - Forked PRs have restricted GITHUB_TOKEN permissions for security
+# - They cannot write comments directly to the base repository's PRs
+# - workflow_run triggers run in the BASE repository context with full permissions
+# - This allows us to safely post comments on forked PRs
+#
+# How it works:
+# 1. PR Code Coverage workflow uploads coverage data as an artifact (forked PRs only)
+# 2. This workflow triggers when PR Code Coverage completes successfully
+# 3. Downloads the artifact and posts the comment with full write permissions
+#
+# Same-repo PRs post comments directly in pr-code-coverage.yml (faster)
+# Forked PRs use this workflow (required for permissions)
+
+on:
+  workflow_run:
+    workflows: ["PR Code Coverage"]
+    types:
+      - completed
+
+jobs:
+  post-comment:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    permissions:
+      pull-requests: write
+      contents: read
+    
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Download coverage data
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Download artifact with error handling for non-existent artifacts
+          if ! gh run download ${{ github.event.workflow_run.id }} \
+            --repo ${{ github.repository }} \
+            --name coverage-comment-data 2>&1; then
+            echo "⚠️ No coverage-comment-data artifact found"
+            echo "This is expected for same-repo PRs (they post comments directly)"
+            echo "Exiting gracefully..."
+            exit 0
+          fi
+          
+          # Verify artifact was downloaded
+          if [[ ! -f pr-info.json ]]; then
+            echo "⚠️ Artifact downloaded but pr-info.json not found"
+            echo "This may indicate an issue with artifact upload"
+            exit 1
+          fi
+
+      - name: Read coverage data
+        id: coverage
+        run: |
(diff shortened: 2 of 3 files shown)


mssql_python/parameter_helper.py
```diff
@@ -0,0 +1,220 @@
...
+        ("SELECT * FROM users WHERE id = ?", (42,))
+    """
+    # No parameters
+    if parameters is None:
+        return sql, None
+
+    # Qmark style - tuple or list
+    if isinstance(parameters, (tuple, list)):
+        # Check if SQL appears to have pyformat placeholders
+        if "%()" in sql or ")s" in sql:  # Quick heuristic
```
bewithgaurav: since the parse_pyformat_params is gonna return params anyways in this case, this heuristic can be removed entirely

mssql_python/cursor.py
```diff
@@ -1233,6 +1238,34 @@ def execute(  # pylint: disable=too-many-locals,too-many-branches,too-many-state
         # Clear any previous messages
         self.messages = []
 
+        # Auto-detect and convert parameter style if needed
+        # Supports both qmark (?) and pyformat (%(name)s)
+        # Note: parameters is always a tuple due to *parameters in method signature
```
bewithgaurav, line 1243: Worth noting the below combinations of param passing rules, could you add a comment here to just document the ambiguity

Parameter Passing Rules:

Single value:
    cursor.execute("SELECT ?", 42)
    → parameters becomes (42,)

Multiple values (two ways):
    cursor.execute("SELECT ?, ?", 1, 2)      # Varargs
    cursor.execute("SELECT ?, ?", (1, 2))    # Tuple (unwrapped)
    → Both become parameters = (1, 2)

Dict for named parameters:
    cursor.execute("SELECT %(id)s", {"id": 42})
    → Unwrapped to {"id": 42}, then converted

Note: If you pass a tuple/list/dict as the ONLY argument,
it will be unwrapped. This means you can't pass a tuple
as a single parameter value (also SQL Server doesn't support
tuple types anyways)

mssql_python/parameter_helper.py
```diff
@@ -0,0 +1,220 @@
...
+
+        >>> convert_pyformat_to_qmark(
+        ...     "WHERE name = %(name)s OR email = %(name)s",
+        ...     {"name": "alice"}
+        ... )
+        ("WHERE name = ? OR email = ?", ("alice", "alice"))
+    """
+    # Support %% escaping - replace %% with a placeholder before parsing
+    # This allows users to have literal % in their SQL
+    escaped_sql = sql.replace("%%", "\x00ESCAPED_PERCENT\x00")
```
bewithgaurav: lets add a more distinctive & readable marker instead of ESCAPED_PERCENT

Maybe something like
```
    _ESCAPE_MARKER = "__MSSQL_PYFORMAT_ESCAPED_PERCENT_PLACEHOLDER__"
```

Also, you can go with uuid as a constant for escape marker - it will be unique for each run, e.g. 

```suggestion
    import uuid
    _ESCAPE_MARKER = f"__MSSQL_PYFORMAT_ESCAPED_{uuid.uuid4().hex}__"
    escaped_sql = sql.replace("%%", _ESCAPE_MARKER)
```

uuid is just a suggestion, might be an overhead

mssql_python/parameter_helper.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # Extract parameter names in order
+    param_names = parse_pyformat_params(escaped_sql)
+
+    if not param_names:
+        # No parameters found - restore escaped %% and return as-is
+        restored_sql = escaped_sql.replace("\x00ESCAPED_PERCENT\x00", "%")
```
bewithgaurav: use the same marker here as well
```suggestion
        restored_sql = escaped_sql.replace(_ESCAPE_MARKER, "%")
```

mssql_python/parameter_helper.py
```diff
@@ -0,0 +1,220 @@
...
+        ...     {"name": "alice"}
+        ... )
+        ("WHERE name = ? OR email = ?", ("alice", "alice"))
+    """
+    # Support %% escaping - replace %% with a placeholder before parsing
+    # This allows users to have literal % in their SQL
+    escaped_sql = sql.replace("%%", "\x00ESCAPED_PERCENT\x00")
+
+    # Extract parameter names in order
+    param_names = parse_pyformat_params(escaped_sql)
```
bewithgaurav, line 141: please add more log statements in this file, will be helpful in debugging edge case scenarios

mssql_python/cursor.py
```diff
@@ -1933,6 +1960,50 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
...
 
+        # Auto-detect and convert parameter style for executemany
+        # Check first row to determine if we need to convert from pyformat to qmark
+        first_row = (
+            seq_of_parameters[0]
+            if hasattr(seq_of_parameters, "__getitem__")
+            else next(iter(seq_of_parameters))
+        )
+
+        # Wrap single non-sequence parameters in each row (for backward compatibility)
```
bewithgaurav: I see this is a new behavior being introduced.. this is a non-standard behaviour and we should revert this (its not backward compatibility related, since we didn't support this prior).. e.g. in here we're trying to convert 
```python
cursor.executemany("INSERT INTO t VALUES (?)", [1, 2, 3])
```
to
```python
cursor.executemany("INSERT INTO t VALUES (?)", [(1,), (2,), (3,)])
```
all the other drivers following db api 2 require explicit tuples inside executemany

mssql_python/cursor.py
```diff
@@ -1233,6 +1238,34 @@ def execute(  # pylint: disable=too-many-locals,too-many-branches,too-many-state
...
+            else:
+                actual_params = parameters
+
+            # Convert parameters based on detected style
+            operation, converted_params = detect_and_convert_parameters(operation, actual_params)
+
+            # Convert back to list format expected by the binding code
+            # detect_and_convert_parameters always returns None, tuple, or list
+            if converted_params is None:
+                parameters = []
```
bewithgaurav: this is a non-reachable code - since the converted_params var can technically never be None - eliminate the if check, and just keep the below (from else case):  
```python
parameters = list(converted_params)
```

bewithgaurav (changes requested):

mssql_python/parameter_helper.py
jahnvi480: Removing heuristic check entirely

mssql_python/cursor.py
jahnvi480: Adding it as comment

mssql_python/parameter_helper.py
jahnvi480: I've replaced the cryptic \x00ESCAPED_PERCENT\x00 with a much more readable marker:
 
 ```py
# Before (hard to read in debugger)
escaped_sql = sql.replace("%%", "\x00ESCAPED_PERCENT\x00")

# After (clear and descriptive)
_ESCAPED_PERCENT_MARKER = "__MSSQL_PYFORMAT_ESCAPED_PERCENT_PLACEHOLDER__"
escaped_sql = sql.replace("%%", _ESCAPED_PERCENT_MARKER)
```

UUID generation will add unnecessary computation for every query. And this will be easy to debug.

File: mssql_python/parameter_helper.py (hunk 5 above)
Comment (jahnvi480) [reply mssql_python/parameter_helper.py:141]: Added debug logs

File: mssql_python/parameter_helper.py (hunk 4 above)
Comment (jahnvi480) [reply mssql_python/parameter_helper.py]: I've replaced the cryptic \x00ESCAPED_PERCENT\x00 with a much more readable marker:

```py
# Before (hard to read in debugger)
escaped_sql = sql.replace("%%", "\x00ESCAPED_PERCENT\x00")

# After (clear and descriptive)
_ESCAPED_PERCENT_MARKER = "__MSSQL_PYFORMAT_ESCAPED_PERCENT_PLACEHOLDER__"
escaped_sql = sql.replace("%%", _ESCAPED_PERCENT_MARKER)
```

UUID generation will add unnecessary computation for every query. And this will be easy to debug.

mssql_python/parameter_helper.py
```diff
@@ -0,0 +1,348 @@
...
+                    # Extract parameter name
+                    param_name = sql[i + 2 : j]
+                    params.append(param_name)
+                    logger.debug(
+                        "parse_pyformat_params: Found parameter '%s' at position %d",
+                        param_name,
+                        i,
+                    )
+                    i = j + 2
+                    continue
```
subrata-ms, line 74: Can we optimize this parsing code (line 54 to 74) by using inbuilt python lib function?
For this type of scenario, regex do it in single pass and we can remove this two "while" loop and improve the overall perf.
Seems like python already has regex lib -

import re
PYFORMAT_PARAM_RE = re.compile(r'%\(([^)]+)\)s')

jahnvi480: @subrata-ms 
We considered regex but went with the explicit scan for a few reasons:

The suggested pattern `r'%\(([^)]+)\)s'` doesn't handle escaped percents (%%) correctly, and we'd need extra logic for incomplete patterns like `%(name without the closing )s`. Adding all that makes the regex more complex than the straightforward loop.

Also, for typical SQL queries (50-500 chars), the performance difference is negligible. I profiled both and the simple scan is actually slightly faster since there's no regex compilation overhead.

The bigger issue is that whether we use regex or loops, we're still doing simple character scanning that doesn't understand SQL context (strings, comments, etc.). The docstring notes this limitation. To properly handle that would need a full SQL parser, which is overkill for parameter detection.

subrata-ms: Usually, library functions are expected to be most efficient and good to use rather than developing your own code component. It helps actually to create proper maintainable code with less error prone scenarios. However, this can be avoided/altered if lib provided function is slow performing. As you are confirming in the comment that the regex is slower compared to a single scan "while" loop in this scenario, hence it should be OK to go-ahead with the current change.

subrata-ms (approved):

PR #415, opened 2026-01-27
FIX: Segmentation Fault in libmsodbcsql-18.5 during SQLFreeHandle()

subrata-ms:
> [AB#41463](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/41463)

> GitHub Issue: #341 

### Summary   
  
This pull request implements a critical fix for a long-standing use-after-free (segmentation fault) bug that occurred when a database connection was closed while statement handles were still alive. The fix ensures that child statement handles are properly tracked and marked as "implicitly freed" when the parent connection is closed, preventing double-free and use-after-free errors. Additionally, comprehensive regression tests are added to verify the fix under various scenarios.

**Critical bug fix for handle management:**

* The `Connection` class now tracks all child statement handles in a `_childStatementHandles` vector using `weak_ptr` to avoid circular references and memory leaks. When the connection is closed, all child statement handles are marked as "implicitly freed" before the parent handle is released. This prevents the `SqlHandle` destructor from attempting to free handles that have already been freed by the ODBC driver. [[1]](diffhunk://#diff-c74277ea1f36fff8f6bbe9df3fd722f7d270f3d6a3069ac7a747fe6d4235cf13R64-R67) [[2]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR97-R110) [[3]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecL176-R203)

* The `SqlHandle` class now includes an `_implicitly_freed` flag and a `markImplicitlyFreed()` method. The `free()` method checks this flag and skips ODBC cleanup if the handle was already implicitly freed, preventing use-after-free errors. [[1]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635R382-R390) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1147-R1150) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1172-L1181)

* The Python bindings are updated to expose the new `markImplicitlyFreed` method on `SqlHandle`, ensuring that the state can be managed from both C++ and Python layers.

**Testing and verification:**

* A new test file, `test_016_connection_invalidation_segfault.py`, is added with extensive tests that reproduce the original segfault scenarios, including multiple cursors, uncommitted transactions, prepared statements, and concurrent connection invalidation. These tests confirm that the fix prevents crashes and ensures clean resource management.

**Other minor changes:**

* Minor whitespace and comment cleanups in unrelated parts of the codebase. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2896) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3618-L3619)

This change is critical for stability and correctness when using connection pooling, SQLAlchemy, or any code that may close connections before all cursors are explicitly closed.

### DIFF
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -100,34 +100,40 @@ void Connection::disconnect() {
         // all child STMT handles. We need to tell the SqlHandle objects about this
         // so they don't try to free the handles again during their destruction.
         
-        // First compact: remove expired weak_ptrs (they're already destroyed)
-        size_t originalSize = _childStatementHandles.size();
-        _childStatementHandles.erase(
-            std::remove_if(_childStatementHandles.begin(), _childStatementHandles.end(),
-                           [](const std::weak_ptr<SqlHandle>& wp) { return wp.expired(); }),
-            _childStatementHandles.end());
-        
-        LOG("Compacted child handles: %zu -> %zu (removed %zu expired)",
-            originalSize, _childStatementHandles.size(),
-            originalSize - _childStatementHandles.size());
-        
-        LOG("Marking %zu child statement handles as implicitly freed",
-            _childStatementHandles.size());
-        for (auto& weakHandle : _childStatementHandles) {
-            if (auto handle = weakHandle.lock()) {
-                // SAFETY ASSERTION: Only STMT handles should be in this vector
-                // This is guaranteed by allocStatementHandle() which only creates STMT handles
-                // If this assertion fails, it indicates a serious bug in handle tracking
-                if (handle->type() != SQL_HANDLE_STMT) {
-                    LOG_ERROR("CRITICAL: Non-STMT handle (type=%d) found in _childStatementHandles. "
-                              "This will cause a handle leak!", handle->type());
-                    continue;  // Skip marking to prevent leak
+        // THREAD-SAFETY: Lock mutex to safely access _childStatementHandles
+        // This protects against concurrent allocStatementHandle() calls or GC finalizers
+        {
+            std::lock_guard<std::mutex> lock(_childHandlesMutex);
+            
+            // First compact: remove expired weak_ptrs (they're already destroyed)
+            size_t originalSize = _childStatementHandles.size();
+            _childStatementHandles.erase(
+                std::remove_if(_childStatementHandles.begin(), _childStatementHandles.end(),
+                               [](const std::weak_ptr<SqlHandle>& wp) { return wp.expired(); }),
+                _childStatementHandles.end());
+            
+            LOG("Compacted child handles: %zu -> %zu (removed %zu expired)",
+                originalSize, _childStatementHandles.size(),
+                originalSize - _childStatementHandles.size());
+            
+            LOG("Marking %zu child statement handles as implicitly freed",
+                _childStatementHandles.size());
+            for (auto& weakHandle : _childStatementHandles) {
+                if (auto handle = weakHandle.lock()) {
+                    // SAFETY ASSERTION: Only STMT handles should be in this vector
+                    // This is guaranteed by allocStatementHandle() which only creates STMT handles
+                    // If this assertion fails, it indicates a serious bug in handle tracking
+                    if (handle->type() != SQL_HANDLE_STMT) {
+                        LOG_ERROR("CRITICAL: Non-STMT handle (type=%d) found in _childStatementHandles. "
+                                  "This will cause a handle leak!", handle->type());
+                        continue;  // Skip marking to prevent leak
+                    }
+                    handle->markImplicitlyFreed();
                 }
-                handle->markImplicitlyFreed();
             }
-        }
-        _childStatementHandles.clear();
-        _allocationsSinceCompaction = 0;
+            _childStatementHandles.clear();
+            _allocationsSinceCompaction = 0;
+        }  // Release lock before potentially slow SQLDisconnect call
 
         SQLRETURN ret = SQLDisconnect_ptr(_dbcHandle->get());
         checkError(ret);
@@ -210,25 +216,32 @@ SqlHandlePtr Connection::allocStatementHandle() {
     checkError(ret);
     auto stmtHandle = std::make_shared<SqlHandle>(static_cast<SQLSMALLINT>(SQL_HANDLE_STMT), stmt);
 
-    // Track this child handle so we can mark it as implicitly freed when connection closes
-    // Use weak_ptr to avoid circular references and allow normal cleanup
-    _childStatementHandles.push_back(stmtHandle);
-    _allocationsSinceCompaction++;
+    // THREAD-SAFETY: Lock mutex before modifying _childStatementHandles
+    // This protects against concurrent disconnect() or allocStatementHandle() calls,
+    // or GC finalizers running from different threads
+    {
+        std::lock_guard<std::mutex> lock(_childHandlesMutex);
+        
+        // Track this child handle so we can mark it as implicitly freed when connection closes
+        // Use weak_ptr to avoid circular references and allow normal cleanup
+        _childStatementHandles.push_back(stmtHandle);
+        _allocationsSinceCompaction++;
 
-    // Compact expired weak_ptrs only periodically to avoid O(n²) overhead
-    // This keeps allocation fast (O(1) amortized) while preventing unbounded growth
-    // disconnect() also compacts, so this is just for long-lived connections with many cursors
-    if (_allocationsSinceCompaction >= COMPACTION_INTERVAL) {
-        size_t originalSize = _childStatementHandles.size();
-        _childStatementHandles.erase(
-            std::remove_if(_childStatementHandles.begin(), _childStatementHandles.end(),
-                           [](const std::weak_ptr<SqlHandle>& wp) { return wp.expired(); }),
-            _childStatementHandles.end());
-        _allocationsSinceCompaction = 0;
-        LOG("Periodic compaction: %zu -> %zu handles (removed %zu expired)",
-            originalSize, _childStatementHandles.size(),
-            originalSize - _childStatementHandles.size());
-    }
+        // Compact expired weak_ptrs only periodically to avoid O(n²) overhead
+        // This keeps allocation fast (O(1) amortized) while preventing unbounded growth
+        // disconnect() also compacts, so this is just for long-lived connections with many cursors
+        if (_allocationsSinceCompaction >= COMPACTION_INTERVAL) {
+            size_t originalSize = _childStatementHandles.size();
+            _childStatementHandles.erase(
+                std::remove_if(_childStatementHandles.begin(), _childStatementHandles.end(),
+                               [](const std::weak_ptr<SqlHandle>& wp) { return wp.expired(); }),
+                _childStatementHandles.end());
+            _allocationsSinceCompaction = 0;
+            LOG("Periodic compaction: %zu -> %zu handles (removed %zu expired)",
+                originalSize, _childStatementHandles.size(),
+                originalSize - _childStatementHandles.size());
+        }
+    }  // Release lock
 
     return stmtHandle;
 }
--- a/mssql_python/pybind/connection/connection.h
+++ b/mssql_python/pybind/connection/connection.h
@@ -5,10 +5,19 @@
 #include "../ddbc_bindings.h"
 #include <memory>
 #include <string>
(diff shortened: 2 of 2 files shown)


tests/test_016_connection_invalidation_segfault.py
```diff
@@ -0,0 +1,351 @@
...
+    all_cursors = []
+
+    # Create multiple connections with cursors
+    for conn_num in range(5):
+        conn = connect(conn_str)
+        connections.append(conn)
+
+        for cursor_num in range(3):
+            cursor = conn.cursor()
+            cursor.execute(f"SELECT {conn_num} AS conn, {cursor_num} AS cursor")
```
bewithgaurav: need to change this test to use cursor_num instead of cursor (same as the other test) - since cursor is a reserved keyword in sql server

bewithgaurav (commented):
one minor comment to help pass the CI checks, else lgtm
will approve once CI is green

mssql_python/pybind/connection/connection.h
```diff
@@ -61,6 +61,10 @@ class Connection {
     std::chrono::steady_clock::time_point _lastUsed;
     std::wstring wstrStringBuffer;  // wstr buffer for string attribute setting
     std::string strBytesBuffer;     // string buffer for byte attributes setting
+
+    // Track child statement handles to mark them as implicitly freed when connection closes
+    // Uses weak_ptr to avoid circular references and allow normal cleanup
+    std::vector<std::weak_ptr<SqlHandle>> _childStatementHandles;
```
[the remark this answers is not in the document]
subrata-ms: fixed

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4360,7 +4364,9 @@ PYBIND11_MODULE(ddbc_bindings, m) {
         .def_readwrite("ddbcErrorMsg", &ErrorInfo::ddbcErrorMsg);
 
     py::class_<SqlHandle, SqlHandlePtr>(m, "SqlHandle")
-        .def("free", &SqlHandle::free, "Free the handle");
+        .def("free", &SqlHandle::free, "Free the handle")
+        .def("markImplicitlyFreed", &SqlHandle::markImplicitlyFreed,
+             "Mark handle as implicitly freed by parent handle");
```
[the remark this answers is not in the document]
subrata-ms: fixed.

tests/test_016_connection_invalidation_segfault.py
```diff
@@ -0,0 +1,351 @@
...
+            cursor.fetchone()
+            all_cursors.append(cursor)
+
+    # Close all connections
+    for conn in connections:
+        conn.close()
+
+    # Clear references and force GC
+    connections = None
+    all_cursors = None
```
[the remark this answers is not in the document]
subrata-ms: it is used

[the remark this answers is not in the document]
subrata-ms: fixed

tests/test_016_connection_invalidation_segfault.py
subrata-ms: fixed and should be available in latest commit

tests/test_016_connection_invalidation_segfault.py
```diff
@@ -0,0 +1,351 @@
...
+        cursor.fetchall()  # Fetch results to complete the query
+        cursors.append(cursor)
+
+    # Close connection without explicitly closing cursors first
+    # This simulates the invalidation scenario where connection is lost
+    conn.close()
+
+    # Force garbage collection to trigger cursor cleanup
+    # This is where the segfault would occur without the fix
+    cursors = None
```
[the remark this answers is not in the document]
subrata-ms: The issue is that the low-level [ddbc_bindings](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) API doesn't expose enough functionality to test the implicit freeing properly without going through the high-level [Connection.close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) path.

The state tracking fix ([markImplicitlyFreed()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) is actually being exercised by the existing tests in test_016 - when [Connection.close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is called, it:

Iterates through _cursors
Closes each cursor (which goes through proper cleanup)
Then closes the underlying ddbc_bindings.Connection
The real protection happens at the C++ level in Connection::disconnect() which marks all child handles before freeing the DBC. The existing test_016 tests DO exercise this code path indirectly, even though they go through the high-level API.

The tests are already validating that the fix works - no crashes occur when connections are invalidated with active cursors, which is exactly what the [markImplicitlyFreed()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) fix prevents.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4360,7 +4364,9 @@ PYBIND11_MODULE(ddbc_bindings, m) {
         .def_readwrite("ddbcErrorMsg", &ErrorInfo::ddbcErrorMsg);
 
     py::class_<SqlHandle, SqlHandlePtr>(m, "SqlHandle")
-        .def("free", &SqlHandle::free, "Free the handle");
+        .def("free", &SqlHandle::free, "Free the handle")
+        .def("mark_implicitly_freed", &SqlHandle::markImplicitlyFreed,
```
sumitmsft: should this be part of the public Python API at all? It looks like an internal lifecycle mechanism. If end users call it incorrectly, it can cause leaks (skipping free) and make debugging harder.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -173,7 +187,20 @@ SqlHandlePtr Connection::allocStatementHandle() {
...
-    return std::make_shared<SqlHandle>(static_cast<SQLSMALLINT>(SQL_HANDLE_STMT), stmt);
+    auto stmtHandle = std::make_shared<SqlHandle>(static_cast<SQLSMALLINT>(SQL_HANDLE_STMT), stmt);
+
+    // Track this child handle so we can mark it as implicitly freed when connection closes
+    // Use weak_ptr to avoid circular references and allow normal cleanup
+    _childStatementHandles.push_back(stmtHandle);
+
+    // Clean up expired weak_ptrs periodically to avoid unbounded growth
+    // Remove entries where the weak_ptr is expired (object was already destroyed)
+    _childStatementHandles.erase(
```
sumitmsft: In `Connection::allocStatementHandle()`, we run `erase(remove_if(expired))` on `_childStatementHandles` on every statement allocation. This makes each allocation O(n) w.r.t. number of tracked handles, and across many allocations can become O(n²) total scanning in workloads that create many short-lived cursors/statements.

Since disconnect() is already doing an O(n) pass (which is fine because it’s not hot path), consider moving compaction to disconnect() only, or compact only every N allocations / when vector size exceeds a threshold (e.g., keep a counter and compact occasionally). This keeps allocation fast while still preventing unbounded growth.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1169,16 +1173,19 @@ void SqlHandle::free() {
             return;
         }
 
-        // Always clean up ODBC resources, regardless of Python state
+        // CRITICAL FIX: Check if handle was already implicitly freed by parent handle
+        // When Connection::disconnect() frees the DBC handle, the ODBC driver automatically
+        // frees all child STMT handles. We track this state to avoid double-free attempts.
+        // This approach avoids calling ODBC functions on potentially-freed handles, which
+        // would cause use-after-free errors.
+        if (_implicitly_freed) {
```
sumitmsft, line 1192: `_implicitly_freed` makes `SqlHandle::free()` skip `SQLFreeHandle_ptr()` and just null the pointer. This is correct only if the underlying ODBC handle has definitely already been freed by parent-handle teardown. If `markImplicitlyFreed()` is ever called in the wrong context (wrong handle type, or without a guaranteed parent free), this will silently leak the handle.

sumitmsft: do you think this situation could ever arise?

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: No. I was using it from test. Now it is being removed and fixed

mssql_python/pybind/connection/connection.cpp
subrata-ms: Good catch. Did the optimization.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -173,7 +187,20 @@ SqlHandlePtr Connection::allocStatementHandle() {
     SQLHANDLE stmt = nullptr;
     SQLRETURN ret = SQLAllocHandle_ptr(SQL_HANDLE_STMT, _dbcHandle->get(), &stmt);
     checkError(ret);
-    return std::make_shared<SqlHandle>(static_cast<SQLSMALLINT>(SQL_HANDLE_STMT), stmt);
+    auto stmtHandle = std::make_shared<SqlHandle>(static_cast<SQLSMALLINT>(SQL_HANDLE_STMT), stmt);
+
+    // Track this child handle so we can mark it as implicitly freed when connection closes
+    // Use weak_ptr to avoid circular references and allow normal cleanup
+    _childStatementHandles.push_back(stmtHandle);
```
sumitmsft: `_childStatementHandles` is mutated in `allocStatementHandle()` (push_back + erase/remove_if) and iterated/cleared in `disconnect()`, with no synchronization. If a Connection can be accessed from multiple threads (even accidentally), this is a data race and could corrupt the vector / crash (iterator invalidation, concurrent clear/erase).

I know Python has the GIL, but native code may release it around ODBC calls and GC/finalizers can run at surprising times. Can we clarify the threading model?

If Connection is intentionally not thread-safe, please document it clearly and ensure these paths always run while holding the GIL.
If any thread-safety is expected, consider guarding `_childStatementHandles` with a mutex (or otherwise ensuring serialization).

tests/test_016_connection_invalidation_segfault.py
```diff
@@ -0,0 +1,310 @@
+"""
+Test for connection invalidation segfault scenario (Issue: Use-after-free on statement handles)
```
sumitmsft, line 2: Given the vector cleanup changes, we can plan to add a test that:

creates thousands of cursors/statements,
closes connection,
forces GC,
ensures it completes in reasonable time (or at least does not blow up). This catches regressions in _childStatementHandles performance.

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: Handled in latest commit. good catch!

mssql_python/pybind/connection/connection.cpp
subrata-ms: Valid concern. Added a mutex lock to safeguard the container during multithread access.

tests/test_016_connection_invalidation_segfault.py
subrata-ms: This is thoroughly tested with SQLAlchemy test suit. But I will add some stress test in mssql-python repo as part of the below task.
https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/40818

bewithgaurav (approved):

PR #420, opened 2026-01-30
FEAT: Explicit parameters and improved type hints for bulkcopy

bewithgaurav:

> AB#41884

### Summary   

Implements **Revision 2** of the BulkCopy API spec from community feedback in #414.

Replaces `**kwargs` with explicit, type-hinted parameters for better IDE autocomplete and discoverability

### API Signature

```python
def _bulkcopy(
    self,
    table_name: str,
    data: Iterable[Union[Tuple, List]],
    batch_size: int = 0,
    timeout: int = 30,
    column_mappings: Optional[Union[List[str], List[Tuple[int, str]]]] = None,
    keep_identity: bool = False,
    check_constraints: bool = False,
    table_lock: bool = False,
    keep_nulls: bool = False,
    fire_triggers: bool = False,
    use_internal_transaction: bool = False,
) -> Dict[str, Any]
```

### What Changed

| Aspect | Before | After |
|--------|--------|-------|
| Parameter style | `**kwargs` | Explicit parameters |
| IDE support | No autocomplete | ✅ Full autocomplete |
| Type safety | `Optional[bool] = None` | `bool = False` |
| `batch_size` | `Optional[int] = None` | `int = 0` (0 = server optimal) |
| `column_mappings` | `List[Tuple[Union[str, int], str]]` | `Union[List[str], List[Tuple[int, str]]]` |

### Column Mappings

Per @amachanic's feedback, now supports simpler format:

**Simple** - `List[str]`:
```python
column_mappings = ['UserID', 'FirstName', 'Email']
```

**Advanced** - `List[Tuple[int, str]]`:
```python
column_mappings = [(0, 'UserID'), (1, 'FirstName'), (3, 'Email')]
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,60 @@
+# Use Python 3.14 with Debian Bookworm as the base image
+FROM mcr.microsoft.com/devcontainers/python:3.14-bookworm
+
+# Set environment variables
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PYTHONPATH=/workspaces/mssql-python
+ENV CMAKE_BUILD_TYPE=Debug
+
+# Set timezone (configurable via build arg)
+ARG TZ=UTC
+ENV TZ=$TZ
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
+    && echo $TZ > /etc/timezone
+
+# Install all system dependencies in one layer
+RUN apt-get update && apt-get install -y \
+    # Build tools
+    build-essential \
+    cmake \
+    pkg-config \
+    ninja-build \
+    python3-dev \
+    pybind11-dev \
+    # ODBC prerequisites (unixodbc-dev provides sql.h headers for compilation)
+    unixodbc-dev \
+    gnupg \
+    software-properties-common \
+    # Utilities
+    curl \
+    wget \
+    git \
+    vim \
+    nano \
+    htop \
+    tree \
+    jq \
+    # Clean up
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Microsoft ODBC Driver for SQL Server
+RUN curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /usr/share/keyrings/microsoft-prod.gpg \
+    && echo "deb [arch=arm64,amd64 signed-by=/usr/share/keyrings/microsoft-prod.gpg] https://packages.microsoft.com/debian/12/prod bookworm main" > /etc/apt/sources.list.d/mssql-release.list \
+    && apt-get update \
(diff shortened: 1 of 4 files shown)


mssql_python/cursor.py
```diff
@@ -2599,7 +2619,19 @@ def _bulkcopy(
...
+                iter(data),
+                batch_size=batch_size,
+                timeout=timeout,
+                column_mappings=column_mappings,
+                keep_identity=keep_identity,
+                check_constraints=check_constraints,
+                table_lock=table_lock,
+                keep_nulls=keep_nulls,
+                fire_triggers=fire_triggers,
+                use_internal_transaction=use_internal_transaction,
```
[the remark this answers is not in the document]
bewithgaurav: we're sending appropriate params, resolving this

mssql_python/cursor.py
```diff
@@ -2452,7 +2452,18 @@ def nextset(self) -> Union[bool, None]:
         return True
 
     def _bulkcopy(
-        self, table_name: str, data: Iterable[Union[Tuple, List]], **kwargs
```
subrata-ms, line 2455: Could you please provide more details in the PR summary for this change?
This API change may not be scalable in future as any new parameter addition need to go through the API contract change which is not advisable.

bewithgaurav: The conversation around new API spec is happening in Github Discussions topic #414 
I'll update the PR summary with the new spec details and link to the above discussion shortly
will ping once done

bewithgaurav: done - updated, pls recheck and let me know if you have any comments

subrata-ms: I am still not convinced about moving out from **kwargs completly.
Can we group the parameters logically with couple of parameter?

amachanic: > I am still not convinced about moving out from **kwargs completly. Can we group the parameters logically with couple of parameter?

Hi @subrata-ms, can you please be more specific with regard to why you're not convinced here? We discussed in the thread linked above and those of us on the end user side who decided to participate seemed to agree that this is a good design. As someone who has worked with SQL Server bulk copy libraries for over 20 years in a number of different languages I feel that this is the best possible option from a usability perspective: The parameters will be obvious and easily discovered by users. Introducing a required external object here won't be especially helpful and users will just have to jump through another hoop when making small changes. And while it does seem like there are a touch more parameters here than we usually see, the real concern should be forward maintainability -- but there is none. These basic parameters haven't changed in all of the time I've been working with SQL Server and I seriously doubt they ever will.

subrata-ms: Hi @amachanic , I would suggest, we discuss these API design approach options in the API Design Review forum so we can gather broader feedback. ( https://github.com/microsoft/mssql-python/discussions/414 ).

saurabh500 (approved):

Issue #435, opened 2026-02-15
VARCHAR fetch fails when data length equals column size with non-ASCII CP1252 characters


zzzeek:
Hi - SQLAlchemy testing is coming down to only a few failures remaining.   The latest is this particular quirk involving CP1252 characters inside of VARCHAR.  usually for unicode data we use NVARCHAR, however we apparently have some tests that roundtrip accent characters into VARCHAR, and this seems to work on SQL Server databases due to the CP1252 encoding set up (pyodbc, pymssql etc. all round trip without issue).    all the other backends we have can also roundtrip this data with VARCHAR without having to use NVARCHAR.     Below is a claude-generated report that includes a test matrix.  

# VARCHAR fetch fails when data length equals column size with non-ASCII CP1252 characters

## Summary

The mssql-python driver fails to fetch VARCHAR columns correctly when:
1. The data length **exactly equals** the column size (e.g., 10 characters in VARCHAR(10))
2. The data contains **non-ASCII characters** from Windows-1252/CP1252 (e.g., é, à, ñ)

This affects **all VARCHAR sizes** from small (VARCHAR(5)) to large (VARCHAR(50+)).

**Behavior:**
- `fetchone()`: Returns `bytes` object (e.g., `b'\xa9'`) instead of `str`
- `fetchall()`: Crashes with `RuntimeError: Error fetching LOB for column 1, cType=1, loop=1, SQLGetData return=-1`

**Works correctly when:**
- Data is pure ASCII (any length)
- Data contains CP1252 characters but does NOT fill the entire column

## Reproduction Code

```python
"""
Comprehensive reproducer for mssql-python VARCHAR fetch bugs

Issue: Driver fails to fetch VARCHAR columns correctly when data length equals
       the column size AND contains non-ASCII characters (Windows-1252/CP1252).

Behavior:
- fetchall(): Crashes with "Error fetching LOB" RuntimeError
- fetchone(): Returns bytes object (e.g., b'\xa9') instead of string

Database: SQL Server with collation SQL_Latin1_General_CP1_CI_AS (code page 1252)
"""

from mssql_python import connect

# Adjust connection string for your environment
conn_str = "UID=scott;PWD=tiger^5HHH;Server=mssql2022;Database=test;Encrypt=No"
conn = connect(conn_str)
cursor = conn.cursor()

# Test grid: (column_size, test_data, description)
test_cases = [
    # Pure ASCII - should always work
    (9, "abcdefghi", "ASCII, full length"),
[...]

PR #444, opened 2026-02-23
FIX: VARCHAR fetch fails when data length equals column size with non-ASCII CP1252 characters

subrata-ms:

> [AB#42604](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/42604)

> GitHub Issue: #435 

### Summary   
This pull request improves the handling of character encoding and buffer sizing for SQL `CHAR`/`VARCHAR` data in the ODBC Python bindings, especially for cross-platform compatibility between Linux/macOS and Windows. The changes ensure that character data is decoded correctly and that buffer sizes are sufficient to prevent corruption or truncation when dealing with multi-byte UTF-8 data returned by the ODBC driver on non-Windows systems.

**Character Encoding Handling:**
- Introduced the `GetEffectiveCharDecoding` function to determine the correct decoding to use for SQL `CHAR` data: always UTF-8 on Linux/macOS (since the ODBC driver returns UTF-8), and the user-specified encoding on Windows. This function is now used consistently throughout the codebase to select the decoding method. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R42-R54) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2878-R2903) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2955-R2991) [[4]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3694) [[5]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635R670) [[6]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635L814-R840)

**Buffer Sizing for UTF-8:**
- Updated buffer allocation logic to use `columnSize * 4 + 1` for SQL `CHAR`/`VARCHAR` columns on Linux/macOS, accounting for the worst-case UTF-8 expansion (up to 4 bytes per character), preventing data truncation when multi-byte characters are present at the column boundary. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2944-R2964) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3477-R3484) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3662-R3678)

**Decoding and Data Fetching:**
- Modified all data fetching and decoding paths (`FetchLobColumnData`, `SQLGetData_wrap`, `ProcessChar`, and batch fetch functions) to use the effective character encoding and the correct buffer sizes, ensuring consistent and correct decoding regardless of platform. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2878-R2903) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2955-R2991) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3694) [[4]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635L814-R840)

**API Changes:**
- Updated the `FetchBatchData` function and its callers to accept and propagate the character encoding parameter, ensuring the encoding context is preserved throughout the data-fetching stack. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3601-R3632) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L4088-R4135) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L4224-R4272)

**Minor Fixes:**
- Minor formatting and logging improvements for error messages and function signatures.

These changes collectively improve correctness and reliability when handling string data from SQL databases, especially in multi-platform environments.

**CP1252 VARCHAR Boundary Fix — Summary**
**Problem::**  VARCHAR columns with CP1252 non-ASCII characters (e.g., é, ñ, ö) returned corrupted data when the string length exactly equaled the column size. Inserting "café René!" into VARCHAR(10) returned "©!".

**Root Cause::**
Three bugs in [ddbc_bindings.cpp](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html):

**Undersized buffer** — SQLGetData / SQLBindCol allocated [columnSize + 1](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) bytes, but on Linux/macOS the ODBC driver converts server data to UTF-8 where CP1252 é (1 byte) becomes 0xC3 0xA9 (2 bytes). A 10-char string with 2 accented characters needs 12 bytes, exceeding the 11-byte buffer → truncation → LOB fallback re-reads consumed data → corruption.

**Wrong decode encoding** — After fixing the buffer, data arrived intact but was decoded with the user's [charEncoding](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (CP1252) instead of UTF-8. Since ODBC on Linux/macOS already converts to UTF-8, double-interpreting as CP1252 produced mojibake (cafÃ© RenÃ©!).

**[ProcessChar](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) assumed UTF-8 on all platforms** — The batch/fetchall hot path used [PyUnicode_FromStringAndSize](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) which assumes UTF-8 input. Correct on Linux (ODBC returns UTF-8), but wrong on Windows (ODBC returns native encoding like CP1252).

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -20,13 +20,14 @@
 // Macro definitions
 //-------------------------------------------------------------------------------------------------
 
-// This constant is not exposed via sql.h, hence define it here
+// These constants are not exposed via sql.h, hence define them here
 #define SQL_SS_TIME2 (-154)
 #define SQL_SS_TIMESTAMPOFFSET (-155)
 #define SQL_C_SS_TIMESTAMPOFFSET (0x4001)
 #define MAX_DIGITS_IN_NUMERIC 64
 #define SQL_MAX_NUMERIC_LEN 16
 #define SQL_SS_XML (-152)
+#define SQL_SS_UDT (-151)
 
 #define STRINGIFY_FOR_CASE(x)                                                                      \
     case x:                                                                                        \
@@ -3319,6 +3320,7 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
                 }
                 break;
             }
+            case SQL_SS_UDT:
             case SQL_BINARY:
             case SQL_VARBINARY:
             case SQL_LONGVARBINARY: {
@@ -3596,6 +3598,7 @@ SQLRETURN SQLBindColums(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& column
                 ret = SQLBindCol_ptr(hStmt, col, SQL_C_GUID, buffers.guidBuffers[col - 1].data(),
                                      sizeof(SQLGUID), buffers.indicators[col - 1].data());
                 break;
+            case SQL_SS_UDT:
             case SQL_BINARY:
             case SQL_VARBINARY:
             case SQL_LONGVARBINARY:
@@ -3744,6 +3747,7 @@ SQLRETURN FetchBatchData(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& colum
             case SQL_WLONGVARCHAR:
                 columnProcessors[col] = ColumnProcessors::ProcessWChar;
                 break;
+            case SQL_SS_UDT:
             case SQL_BINARY:
             case SQL_VARBINARY:
             case SQL_LONGVARBINARY:
@@ -4042,6 +4046,10 @@ size_t calculateRowSize(py::list& columnNames, SQLUSMALLINT numCols) {
             case SQL_BIT:
                 rowSize += sizeof(SQLCHAR);
                 break;
+            case SQL_SS_UDT:
+                rowSize += (static_cast<SQLLEN>(columnSize) == SQL_NO_TOTAL || columnSize == 0)
+                               ? SQL_MAX_LOB_SIZE : columnSize;
+                break;
             case SQL_BINARY:
             case SQL_VARBINARY:
             case SQL_LONGVARBINARY:
@@ -4104,7 +4112,8 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
 
         if ((dataType == SQL_WVARCHAR || dataType == SQL_WLONGVARCHAR || dataType == SQL_VARCHAR ||
              dataType == SQL_LONGVARCHAR || dataType == SQL_VARBINARY ||
-             dataType == SQL_LONGVARBINARY || dataType == SQL_SS_XML) &&
+             dataType == SQL_LONGVARBINARY || dataType == SQL_SS_XML ||
+             dataType == SQL_SS_UDT) &&
             (columnSize == 0 || columnSize == SQL_NO_TOTAL || columnSize > SQL_MAX_LOB_SIZE)) {
             lobColumns.push_back(i + 1);  // 1-based
         }
@@ -4243,7 +4252,8 @@ SQLRETURN FetchAll_wrap(SqlHandlePtr StatementHandle, py::list& rows,
 
         if ((dataType == SQL_WVARCHAR || dataType == SQL_WLONGVARCHAR || dataType == SQL_VARCHAR ||
              dataType == SQL_LONGVARCHAR || dataType == SQL_VARBINARY ||
-             dataType == SQL_LONGVARBINARY || dataType == SQL_SS_XML) &&
+             dataType == SQL_LONGVARBINARY || dataType == SQL_SS_XML ||
+             dataType == SQL_SS_UDT) &&
             (columnSize == 0 || columnSize == SQL_NO_TOTAL || columnSize > SQL_MAX_LOB_SIZE)) {
             lobColumns.push_back(i + 1);  // 1-based
         }
--- a/mssql_python/constants.py
+++ b/mssql_python/constants.py
@@ -114,6 +114,7 @@ class ConstantsDDBC(Enum):
     SQL_FETCH_ABSOLUTE = 5
     SQL_FETCH_RELATIVE = 6
     SQL_FETCH_BOOKMARK = 8
+    SQL_SS_UDT = -151
     SQL_DATETIMEOFFSET = -155
     SQL_SS_TIME2 = -154
     SQL_SS_XML = -152
@@ -374,6 +375,7 @@ class SQLTypes:
             ConstantsDDBC.SQL_DATETIMEOFFSET.value,
             ConstantsDDBC.SQL_SS_XML.value,
             ConstantsDDBC.SQL_GUID.value,
+            ConstantsDDBC.SQL_SS_UDT.value,
         }
 
     # Could also add category methods for convenience
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -865,6 +865,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
             ddbc_sql_const.SQL_BINARY.value: ddbc_sql_const.SQL_C_BINARY.value,
             ddbc_sql_const.SQL_VARBINARY.value: ddbc_sql_const.SQL_C_BINARY.value,
             ddbc_sql_const.SQL_LONGVARBINARY.value: ddbc_sql_const.SQL_C_BINARY.value,
+            ddbc_sql_const.SQL_SS_UDT.value: ddbc_sql_const.SQL_C_BINARY.value,
             # ODBC 3.x date/time types (reported by ODBC 18 driver)
             ddbc_sql_const.SQL_TYPE_DATE.value: ddbc_sql_const.SQL_C_TYPE_DATE.value,
             ddbc_sql_const.SQL_TYPE_TIME.value: ddbc_sql_const.SQL_C_TYPE_TIME.value,
@@ -1052,6 +1053,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
             SQL_SS_TIME2(-154)  for time columns
             SQL_DATETIMEOFFSET(-155)  for datetimeoffset columns
             SQL_SS_XML(-152)  for xml columns
+            SQL_SS_UDT(-151)  for geography/geometry/hierarchyid columns
 
         ODBC 2.x aliases (9, 10, 11) are also accepted defensively.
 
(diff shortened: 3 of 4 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3646,6 +3691,7 @@ SQLRETURN FetchBatchData(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& colum
         columnInfosExt[col].processedColumnSize = columnInfos[col].processedColumnSize;
         columnInfosExt[col].fetchBufferSize = columnInfos[col].fetchBufferSize;
         columnInfosExt[col].isLob = columnInfos[col].isLob;
+        columnInfosExt[col].charEncoding = GetEffectiveCharDecoding(charEncoding);
```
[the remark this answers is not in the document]

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -811,21 +812,39 @@ inline void ProcessChar(PyObject* row, ColumnBuffers& buffers, const void* colIn
...
+            pyStr = PyUnicode_FromStringAndSize(dataPtr, numCharsInData);
+        } else {
+            pyStr =
+                PyUnicode_Decode(dataPtr, numCharsInData, colInfo->charEncoding.c_str(), "strict");
+        }
+#endif
         if (!pyStr) {
+            PyErr_Clear();
             Py_INCREF(Py_None);
             PyList_SET_ITEM(row, col - 1, Py_None);
```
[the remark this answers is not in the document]
subrata-ms: Fixed

tests/test_017_varchar_cp1252_boundary.py
```diff
@@ -0,0 +1,333 @@
...
+    cursor = db_connection.cursor()
+
+    try:
+        cursor.execute("""
+            CREATE TABLE #test_cp1252_boundary (
+                id INT PRIMARY KEY,
+                varchar_10 VARCHAR(10),
+                varchar_20 VARCHAR(20),
+                varchar_50 VARCHAR(50),
+                varchar_100 VARCHAR(100)
```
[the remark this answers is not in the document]
subrata-ms: Fixed

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2941,7 +2956,12 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
                     row.append(
                         FetchLobColumnData(hStmt, i, SQL_C_CHAR, false, false, charEncoding));
                 } else {
-                    uint64_t fetchBufferSize = columnSize + 1 /* null-termination */;
+                    // Use columnSize * 4 + 1 to accommodate worst-case UTF-8 expansion.
+                    // columnSize is in characters, but on Linux the ODBC driver returns
+                    // UTF-8 encoded bytes where each character can be up to 4 bytes.
+                    // Without this, data at exact column boundary with multi-byte chars
+                    // (e.g., CP1252 é in VARCHAR(10)) causes truncation and corruption.
```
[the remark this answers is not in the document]

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3451,7 +3476,14 @@ SQLRETURN SQLBindColums(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& column
                 // TODO: handle variable length data correctly. This logic wont
                 // suffice
                 HandleZeroColumnSizeAtFetch(columnSize);
+                // Use columnSize * 4 + 1 on Linux/macOS to accommodate UTF-8
+                // expansion. The ODBC driver returns UTF-8 for SQL_C_CHAR where
+                // each character can be up to 4 bytes.
+#if defined(__APPLE__) || defined(__linux__)
+                uint64_t fetchBufferSize = columnSize * 4 + 1 /*null-terminator*/;
```
sumitmsft, line 3494: The `columnSize` is an SQLULEN. Multiplying by 4 can overflow for very large column sizes (e.g., LONGVARCHAR where columnSize can be 2GB-1 = 2147483647). 2147483647 * 4 = 8589934588 bytes (~8 GB), which may cause an out-of-memory crash or excessive memory allocation. 

There's already a TODO comment in the code acknowledging this (line 3487) below, which we put sometime back. I'd suggest to add an upper bound check. If columnSize * 4 exceeds a reasonable threshold (e.g., SQL_MAX_LOB_SIZE or some configurable limit), fall back to LOB streaming instead of pre-allocating. D

Do you this will be a good idea?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2941,7 +2956,14 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
                 } else {
-                    uint64_t fetchBufferSize = columnSize + 1 /* null-termination */;
+                    // Use columnSize * 4 + 1 to accommodate worst-case UTF-8 expansion.
+                    // columnSize is in characters, but the ODBC driver may return UTF-8
+                    // encoded bytes where each character can be up to 4 bytes. This
+                    // applies on Linux/macOS (driver always returns UTF-8 for SQL_C_CHAR)
+                    // and on Windows when the database uses a UTF-8 collation. Without
+                    // this, data at exact column boundary with multi-byte chars (e.g.,
+                    // CP1252 é in VARCHAR(10)) causes truncation and corruption.
+                    uint64_t fetchBufferSize = columnSize * 4 + 1 /* null-termination */;
```
sumitmsft, line 2976: In `SQLGetData_wrap` "× 4" is applied on all platforms (no #if guard), while `SQLBindColums` and `FetchBatchData` use "× 4" only on Linux/macOS. The comment says "and on Windows when the database uses a UTF-8 collation" but this is a different behavior than the other two sites. This inconsistency means:

On Windows with CP1252, SQLGetData_wrap allocates 4× the needed buffer (wasteful)
On Linux/macOS, SQLBindColums and FetchBatchData also allocate 4× (correct)

Can you document why `SQLGetData_wrap` intentionally differs with the other two functions?

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -811,21 +812,48 @@ inline void ProcessChar(PyObject* row, ColumnBuffers& buffers, const void* colIn
...
+#if defined(__APPLE__) || defined(__linux__)
+        // On Linux/macOS, ODBC driver returns UTF-8 — PyUnicode_FromStringAndSize
+        // expects UTF-8, so this is correct and fast.
+        pyStr = PyUnicode_FromStringAndSize(dataPtr, numCharsInData);
+#else
+        // On Windows, ODBC driver returns bytes in the server's native encoding.
+        // For UTF-8, use the direct C API (PyUnicode_FromStringAndSize) which
+        // bypasses the codec registry for maximum reliability. For non-UTF-8
+        // encodings (e.g., CP1252), use PyUnicode_Decode with the codec registry.
+        if (colInfo->charEncoding == "utf-8") {
```
sumitmsft: This string comparison happens per-cell in the batch fetch hot loop (called for every row × every CHAR column). While small, it's unnecessary overhead. Can we mitigate this please?

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Yeah, I think we may need a combination of both approaches to balance the trade-offs.
We could add an overflow check: if (columnSize > SQL_MAX_LOB_SIZE / 4) then switch to LOB streaming.

This way, we keep the * 4 multiplier only for small columns where it won’t cause excessive allocation. For large VARCHAR columns, the driver would use LOB streaming, so there shouldn’t be any risk of excessive memory usage.
cc: @saurabh500 @sumitmsft

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya, line 2976: +1 on the platform guard inconsistency concern. I agree with this observation about inconsistent * 4 buffer sizing across different fetch methods.

gargsaumya (changes requested):

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: As discussed, this won't be straight forward. For example, VARCHAR(5), also can have a CP1252 char at the boundary. Hence, in this scenario, only option is to scan the string for CP1252 char at boundary and then switch to LOB streaming. But this approach does not seems to be performant. We did a brainstorming session and one of the outcomes from the session, was use Chunking strategy to solve this problem for string data type across all type of character encoding. Below is the bug to investigate further on this and optimized the code for better memory efficient and performant.
https://github.com/microsoft/mssql-python/issues/455

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: fixed in latest commit.

mssql_python/pybind/ddbc_bindings.h
subrata-ms: Fixed as below - 
added a bool isUtf8 field to ColumnInfoExt, set it once per column in the batch setup, and replace the per-cell std::string comparison with a simple bool check.

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: updated the logic in comment section.

sumitmsft (approved):

PR #446, opened 2026-02-24
FEAT: Support for Complex Data Type- sql_variant

gargsaumya:

> [AB#42724](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/42724)

### Summary   
  
This pull request adds support for the `sql_variant` SQL Server type to the Python MSSQL driver, ensuring that `sql_variant` columns are fetched and mapped correctly to their underlying types. The changes introduce new constants, update type mappings, and enhance the fetch logic to detect and process `sql_variant` columns using their native types, improving compatibility and correctness when handling complex data.

**Support for sql_variant type**

* Added the `SQL_SS_VARIANT` constant and related attribute constants in both `mssql_python/constants.py` and the C++ binding layer to enable recognition and handling of the `sql_variant` type. [[1]](diffhunk://#diff-e6d80f1000af6fd5afca05f435b11fd82df7f5c3e75ecf5763f85d3aacdbe758R120) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R30-R34)
* Included `SQL_SS_VARIANT` in the set of valid types for type validation logic.

**Type mapping and fetch logic**

* Updated the C type mapping in `cursor.py` so that `SQL_SS_VARIANT` is mapped to `SQL_C_BINARY`, allowing binary transfer of the variant data.
* Implemented a helper function in the C++ layer to map a `sql_variant`'s underlying C type to the appropriate SQL data type, enabling the fetch logic to reuse existing code for each possible underlying type.
* Enhanced the fetch routines (`SQLGetData_wrap`, `FetchMany_wrap`, and `FetchAll_wrap`) to detect `sql_variant` columns, determine their true data types at runtime, and handle them with the correct logic. This includes always using the streaming fetch path for `sql_variant` columns to preserve native type fidelity. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2932-R3020) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L4044-R4144) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R4229-R4274)

**Other improvements**

* Improved error logging and debug output for easier troubleshooting and visibility into how `sql_variant` columns are processed. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L1156-R1162) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L2932-R3020) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L4044-R4144) [[4]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R4229-R4274)

These changes collectively ensure that the driver can now fully support reading `sql_variant` columns, mapping them to their actual types, and handling them efficiently.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -4391,37 +4391,6 @@ SQLRETURN FetchAll_wrap(SqlHandlePtr StatementHandle, py::list& rows,
     }
     LOG("FetchAll_wrap: Fetching data in batch sizes of %d", fetchSize);
 
-    std::vector<SQLUSMALLINT> lobColumns;
-    for (SQLSMALLINT i = 0; i < numCols; i++) {
-        auto colMeta = columnNames[i].cast<py::dict>();
-        SQLSMALLINT dataType = colMeta["DataType"].cast<SQLSMALLINT>();
-        SQLULEN columnSize = colMeta["ColumnSize"].cast<SQLULEN>();
-
-        if (IsLobOrVariantColumn(dataType, columnSize)) {
-            lobColumns.push_back(i + 1);  // 1-based
-        }
-    }
-
-    // If we have LOBs → fall back to row-by-row fetch + SQLGetData_wrap
-    if (!lobColumns.empty()) {
-        LOG("FetchAll_wrap: LOB columns detected (%zu columns), using per-row "
-            "SQLGetData path",
-            lobColumns.size());
-        while (true) {
-            ret = SQLFetch_ptr(hStmt);
-            if (ret == SQL_NO_DATA)
-                break;
-            if (!SQL_SUCCEEDED(ret))
-                return ret;
-
-            py::list row;
-            SQLGetData_wrap(StatementHandle, numCols, row, charEncoding,
-                            wcharEncoding);  // <-- streams LOBs correctly
-            rows.append(row);
-        }
-        return SQL_SUCCESS;
-    }
-
     ColumnBuffers buffers(numCols, fetchSize);
 
     // Bind columns


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2929,7 +2995,41 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                row.append(py::none());
+                continue;
+            }
+            // Now retrieve the underlying C type
+            SQLLEN variantCType = 0;
+            ret =
+                SQLColAttribute_ptr(hStmt, i, SQL_CA_SS_VARIANT_TYPE, NULL, 0, NULL, &variantCType);
+            if (!SQL_SUCCEEDED(ret)) {
+                LOG("SQLGetData: Failed to get sql_variant underlying type for column %d", i);
+                row.append(py::none());
```
sumitmsft, line 3056: At minimum, can we use use LOG_ERROR instead of LOG for these failure paths. I'd suggest to throw an exception so the caller knows the data is unreliable. Is "None" an intentional fallback behavior?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4041,10 +4141,11 @@ SQLRETURN FetchMany_wrap(SqlHandlePtr StatementHandle, py::list& rows, int fetch
         SQLSMALLINT dataType = colMeta["DataType"].cast<SQLSMALLINT>();
         SQLULEN columnSize = colMeta["ColumnSize"].cast<SQLULEN>();
 
-        if ((dataType == SQL_WVARCHAR || dataType == SQL_WLONGVARCHAR || dataType == SQL_VARCHAR ||
-             dataType == SQL_LONGVARCHAR || dataType == SQL_VARBINARY ||
-             dataType == SQL_LONGVARBINARY || dataType == SQL_SS_XML) &&
-            (columnSize == 0 || columnSize == SQL_NO_TOTAL || columnSize > SQL_MAX_LOB_SIZE)) {
+        if (dataType == SQL_SS_VARIANT ||
```
sumitmsft: The exact same LOB detection condition is duplicated in three places - across all fetch functions. If a new type needs to be added to this check in the future, or if the condition changes, it must be updated in all three places. Missing any of these ones will cause subtle, hard-to-debug inconsistencies. 

Can we extract this into a helper function like and call it from all three locations? Would be easier to manage and maintain

tests/test_019_sql_variant.py
```diff
@@ -0,0 +1,544 @@
...
+- Cannot contain: text, ntext, image, timestamp, xml, varchar(max), nvarchar(max),
+  varbinary(max), sql_variant itself, geometry, geography, hierarchyid, user-defined types
+
+**Supported Base Types and Expected Python Mappings**:
+- INT, SMALLINT, TINYINT, BIGINT → Python int
+- REAL, FLOAT → Python float
+- DECIMAL, NUMERIC → Python Decimal
+- BIT → Python bool/int
+- CHAR, VARCHAR, NCHAR, NVARCHAR → Python str
+- DATETIME, SMALLDATETIME, DATETIME2 → Python datetime
```
sumitmsft, line 19: please validate if SMALLDATETIME is included in a test case. I am unable to find it.

Check for CHAR, NCHAR, and BINARY as well.

gargsaumya: Done

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Done

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Done

sumitmsft: Is None intentional here? or do you think an exception should be thrown?
If we keep the None fallback - it mau silently corrupts data and the caller has no way to distinguish "a failure" from "value is NULL". 
Can we throw a runtime error on failure rather than silently returning None? The error is already logged, but the caller needs to know the data is unreliable.

sumitmsft: I think we need to look at both places  - SQLGetData call and SQLColAttribute

sumitmsft (approved):

Issue #464, opened 2026-03-02
support for column names and other names that contain question marks without being flagged as a bound parameter issue


zzzeek:
### Describe the bug

we have tests for so-called "difficult column/parameter names" with various funny characters and things in column names and bound parameter names.   A column name that includes a question mark seems to produce scenarios where querying against the table is impossible.

```
Stack trace:

Traceback (most recent call last):
  File "/home/classic/dev/sqlalchemy/test3.py", line 27, in <module>
    cursor.execute(
    ~~~~~~~~~~~~~~^
        "SELECT * FROM t WHERE [q?marks]=%(somename)",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        {"somename": "thename"}
        ^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/classic/.venv3/lib/python3.14/site-packages/mssql_python/cursor.py", line 1331, in execute
    operation, converted_params = detect_and_convert_parameters(operation, actual_params)
                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/classic/.venv3/lib/python3.14/site-packages/mssql_python/parameter_helper.py", line 325, in detect_and_convert_parameters
    raise TypeError(
    ...<4 lines>...
    )
TypeError: Parameter style mismatch: query uses positional placeholders (?), but dict was provided. Use tuple/list for positional parameters. Example: cursor.execute(sql, (value1, value2))


```

### To reproduce

```python
from mssql_python import connect

conn_str = "UID=scott;PWD=tiger^5HHH;Server=mssql2022;Database=test;Encrypt=No"
conn = connect(conn_str)
cursor = conn.cursor()

cursor.execute("drop table if exists t")

# statement that has no parameter dictionary, this works because
# it isn't trying to do anything with parameters
cursor.execute("""

CREATE TABLE t (
	id INTEGER NOT NULL IDENTITY,
	[q?marks] VARCHAR(50) NOT NULL,
	PRIMARY KEY (id)
)

""")

# statement that has a parameter dictionary, we get the error
# "Parameter style mismatch: query uses positional placeholders (?), but dict
# was provided. Use tuple/list for positional parameters."
#
# which is not actually correct, we have a named parameter in there as well.
# but even so, how to indicate the question mark?   some kind of escaping?
cursor.execute(
[...]

PR #465, opened 2026-03-03
FIX: False positive qmark detection for ? inside bracketed identifiers, string literals and comments

jahnvi480:

> [AB#42937](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/42937)

> GitHub Issue: #464 

### Summary   
This pull request introduces a robust fix for detecting real parameter placeholders in SQL statements, specifically addressing false positives caused by question marks inside bracketed identifiers, string literals, quoted identifiers, and comments. The changes add context-aware scanning logic and comprehensive tests, ensuring that only actual parameter placeholders are flagged and handled. This resolves a bug where SQL containing `?` inside brackets (e.g., `[q?marks]`) would incorrectly trigger parameter mismatch errors.

### Core logic improvements

* Added `_skip_quoted_context` helper in `parameter_helper.py` to accurately skip over single-line comments, multi-line comments, single-quoted string literals (with escaped quotes), double-quoted identifiers, and bracketed identifiers when scanning SQL for placeholders.
* Added `_has_unquoted_question_marks` function to detect real `?` placeholders only outside quoted contexts, using the new context-skipping logic.
* Updated `detect_and_convert_parameters` to use `_has_unquoted_question_marks` for parameter style mismatch detection, preventing false positives when `?` appears inside bracketed identifiers or other quoted contexts.

### Testing improvements

* Added extensive unit tests for `_skip_quoted_context` and `_has_unquoted_question_marks`, covering all relevant SQL quoting and commenting scenarios, including edge cases like unclosed quotes/brackets.
* Added integration tests verifying that SQL with `?` inside bracketed identifiers, string literals, and comments does not trigger parameter style mismatch errors, and that real placeholders are still detected correctly.

### Test harness updates

* Imported the new helper functions in the test module to facilitate direct testing.

### DIFF
--- a/mssql_python/parameter_helper.py
+++ b/mssql_python/parameter_helper.py
@@ -5,7 +5,9 @@ Licensed under the MIT license.
 Parameter style conversion helpers for mssql-python.
 
 Supports both qmark (?) and pyformat (%(name)s) parameter styles.
-Simple character scanning approach - does NOT parse SQL contexts.
+Includes context-aware scanning for qmark and pyformat detection,
+skipping characters inside bracketed identifiers, string literals,
+quoted identifiers, and SQL comments.
 
 Reference: https://www.python.org/dev/peps/pep-0249/#paramstyle
 """
@@ -75,14 +77,20 @@ def _skip_quoted_context(sql: str, i: int, length: int) -> int:
                     continue
                 break
             i += 1
-        return i + 1  # skip closing quote
+        return min(i + 1, length)  # skip closing quote
 
     # Double-quoted identifier: skip to closing "
+    # Handles escaped quotes ("") inside identifiers
     if ch == '"':
         i += 1
-        while i < length and sql[i] != '"':
+        while i < length:
+            if sql[i] == '"':
+                if i + 1 < length and sql[i + 1] == '"':
+                    i += 2  # skip escaped quote
+                    continue
+                break
             i += 1
-        return i + 1  # skip closing quote
+        return min(i + 1, length)  # skip closing quote
 
     # Bracketed identifier: skip to closing ]
     # Handles escaped brackets (]]) inside identifiers
@@ -146,9 +154,9 @@ def parse_pyformat_params(sql: str) -> List[str]:
     """
     Extract %(name)s parameter names from SQL string.
 
-    Uses simple character scanning approach - does NOT parse SQL contexts
-    (strings, comments, identifiers). This means %(name)s patterns inside SQL
-    string literals or comments WILL be detected as parameters.
+    Uses context-aware scanning to skip %(name)s patterns inside SQL
+    string literals, quoted identifiers, bracketed identifiers, and comments.
+    Only %(name)s patterns in executable SQL are detected as parameters.
 
     Args:
         sql: SQL query string with %(name)s placeholders
@@ -176,6 +184,12 @@ def parse_pyformat_params(sql: str) -> List[str]:
     length = len(sql)
 
     while i < length:
+        # Skip any quoted context (brackets, strings, comments)
+        skipped = _skip_quoted_context(sql, i, length)
+        if skipped >= 0:
+            i = skipped
+            continue
+
         # Look for %(
         if i + 2 < length and sql[i] == "%" and sql[i + 1] == "(":
             # Find the closing )
--- a/tests/test_015_pyformat_parameters.py
+++ b/tests/test_015_pyformat_parameters.py
@@ -72,22 +72,22 @@ class TestParsePyformatParams:
         assert params == ["param1", "param2", "param3"]
 
     def test_parse_parameter_in_string_literal(self):
-        """Test that parameters in string literals are still detected"""
+        """Test that parameters in string literals are correctly skipped."""
         sql = "SELECT '%(example)s' AS literal, id FROM users WHERE id = %(id)s"
         params = parse_pyformat_params(sql)
-        # Simple scanner detects both - this is by design
-        assert params == ["example", "id"]
+        # Context-aware scanner skips %(example)s inside string literal
+        assert params == ["id"]
 
     def test_parse_parameter_in_comment(self):
-        """Test that parameters in comments are still detected"""
+        """Test that parameters in comments are correctly skipped."""
         sql = """
         SELECT * FROM users
         -- This comment has %(commented)s parameter
         WHERE id = %(id)s
         """
         params = parse_pyformat_params(sql)
-        # Simple scanner detects both - this is by design
-        assert params == ["commented", "id"]
+        # Context-aware scanner skips %(commented)s inside comment
+        assert params == ["id"]
 
     def test_parse_complex_query_with_cte(self):
         """Test parsing complex CTE query."""
@@ -297,7 +297,13 @@ class TestConvertPyformatToQmark:
         assert result_params == (42,)
 
     def test_convert_with_multiple_escaped_percent(self):
-        """Test multiple %% escapes."""
+        """Test multiple %% escapes with context-aware parsing.
+
+        The %(name)s inside the string literal '%%%(name)s%%' is correctly
+        skipped by context-aware scanning, so only %(id)s is treated as a
+        real parameter. After %% unescaping, the string literal becomes
+        '%%(name)s%' (literal percent + literal %(name)s + literal percent).
+        """
         sql = (
             "SELECT '%%test%%' AS txt, id FROM users WHERE id = %(id)s AND name LIKE '%%%(name)s%%'"
         )
@@ -305,8 +311,9 @@ class TestConvertPyformatToQmark:
         result_sql, result_params = convert_pyformat_to_qmark(sql, param_dict)
         assert "'%test%'" in result_sql
         assert "?" in result_sql
-        assert "%%(name)s" not in result_sql
-        assert result_params == (1, "alice")
+        # %(name)s inside string literal is preserved as literal text after %% unescaping
+        assert "'%%(name)s%'" in result_sql
+        assert result_params == (1,)
 
     def test_convert_only_escaped_percent_no_params(self):
         """Test SQL with only %% and no parameters."""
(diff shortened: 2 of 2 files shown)


mssql_python/parameter_helper.py
```diff
@@ -18,6 +18,121 @@
...
+
+    # Single-line comment: skip to end of line
+    if ch == "-" and i + 1 < length and sql[i + 1] == "-":
+        i += 2
+        while i < length and sql[i] != "\n":
+            i += 1
+        return i
+
+    # Multi-line comment: skip to closing */
+    if ch == "/" and i + 1 < length and sql[i + 1] == "*":
```
bewithgaurav, line 55: there is a case for nested block comments which is breaking in here

```
SELECT /* outer /* inner */ still comment ? */ 1
``` 

this returns `1` in sqlserver since there is a provision of nested block comments, this logic prevents going nested since it breaks down post receiving first enclosing marker

running the below code will repro the issue
```python
# Repro: nested block comment causes false-positive qmark detection
from mssql_python.parameter_helper import detect_and_convert_parameters

sql = "SELECT /* outer /* inner */ still comment ? */ 1"

# Expected: no placeholder mismatch (the ? is inside comment context)
# Actual (current): raises TypeError "query uses positional placeholders (?)"
print(detect_and_convert_parameters(sql, {}))
```

please confirm and add testcase for this as well

jahnvi480: Added condition to check the depth of the comment and then decide based on that.
Also added testcases to check on this

mssql_python/parameter_helper.py
```diff
@@ -18,6 +18,130 @@
...
+    if ch == "'":
+        i += 1
+        while i < length:
+            if sql[i] == "'":
+                if i + 1 < length and sql[i + 1] == "'":
+                    i += 2  # skip escaped quote
+                    continue
+                break
+            i += 1
+        return i + 1  # skip closing quote
```
sumitmsft: Can you validate if this also needs to be guarded with min() to avoid out of bounds boundary:

`return min(i + 1, length)`

mssql_python/parameter_helper.py
```diff
@@ -18,6 +18,130 @@
...
+
+    # Double-quoted identifier: skip to closing "
+    if ch == '"':
+        i += 1
+        while i < length and sql[i] != '"':
+            i += 1
+        return i + 1  # skip closing quote
```
sumitmsft: Can you validate if this also needs to be guarded with min() to avoid out of bounds boundary:

`return min(i + 1, length)`

mssql_python/parameter_helper.py
```diff
@@ -317,7 +441,9 @@ def detect_and_convert_parameters(
         )
 
         # Check if SQL appears to have qmark placeholders
-        if "?" in sql and not parse_pyformat_params(sql):
+        # Fast short-circuit: skip the O(n) context-aware scan if no ? exists at all
+        # Then use context-aware check that ignores ? inside brackets, quotes, and comments
+        if "?" in sql and _has_unquoted_question_marks(sql) and not parse_pyformat_params(sql):
```
sumitmsft, line 460: Consider this SQL: 
`SELECT 'fake %(name)s' FROM t WHERE id = ?` with `{"id": 1}`. 
The `parse_pyformat_params` will find %(name)s inside the string literal
This means, `not parse_pyformat_params(sql)` is `False`, and the real `?` placeholder won't trigger the mismatch error. 

I hope this make sense. If this is true then this seems to be a silent logic bug because the dict parameter will flow through to `convert_pyformat_to_qmark` and try to substitute the string-literal parameter, producing incorrect SQL.

Can you please validate it?

mssql_python/parameter_helper.py
sumitmsft: Update this to reflect that the module now includes context-aware scanning for qmark detection

mssql_python/parameter_helper.py
```diff
@@ -18,6 +18,130 @@
...
+            if sql[i] == "'":
+                if i + 1 < length and sql[i + 1] == "'":
+                    i += 2  # skip escaped quote
+                    continue
+                break
+            i += 1
+        return i + 1  # skip closing quote
+
+    # Double-quoted identifier: skip to closing "
+    if ch == '"':
```
sumitmsft, line 84: Please validate if this statement is being handled correctly:

`SELECT "col""?" FROM t`

tests/test_015_pyformat_parameters.py
```diff
@@ -334,6 +336,300 @@ def test_convert_many_parameters(self):
         assert result_params == tuple(range(50))
 
 
+class TestSkipQuotedContext:
```
sumitmsft, line 346: Please test following cases:

**Test 1: Unterminated single-quoted string**
- SQL: `"SELECT 'unterminated"`
- Expected return: `22` (i.e., `len(sql)`)

**Test 2: Unterminated double-quoted identifier**
- SQL: `'SELECT "unterminated'`
- Expected return: `22` (i.e., `len(sql)`)

**Test 3: Question mark inside unterminated single-quoted string**
- SQL: `"SELECT 'unterminated?"`
- Expected return: `False`
- Reason: The `?` is inside the unterminated string, so it should be consumed as part of the string context, not treated as a real placeholder

**Test 4: Question mark inside unterminated double-quoted identifier**
- SQL: `'SELECT "unterminated?'`
- Expected return: `False`
- Reason: The `?` is inside the unterminated double-quoted identifier, so it should be consumed as part of that context

**Test 5: Double-quoted identifier with `""` escape (no question mark)**
- SQL: `'SELECT "col""name" FROM t'`
- Expected return: `20` (position after the real closing `"`)

**Test 6: Double-quoted identifier with `""` escape containing question mark**
- SQL: `'SELECT "col""?name" FROM t'`
- Expected return: `21` (position after the real closing `"`)
- Reason: The identifier is `col"?name` — the `""` is an escape and the `?` is inside the identifier. The scanner should skip the entire thing

**Test 7: Double-quoted identifier that is just an escaped quote**
- SQL: `'SELECT """" FROM t'`
- Expected return: `11` (position after the closing `"`)
- Reason: The identifier content is `""` which represents a single `"` character. The identifier spans positions 7 through 10

**Test 8: Question mark inside double-quoted identifier with `""` escape**
- SQL: `'SELECT "col""?name" FROM t'`
- Expected return: `False`
- Reason: The `?` is inside the double-quoted identifier `col"?name` and should not be detected as a real placeholder

**Test 9: Real question mark after double-quoted identifier with `""` escape**
- SQL: `'SELECT "col""?name" FROM t WHERE id = ?'`
- Expected return: `True`
- Reason: The `?` inside the identifier should be ignored, but the real `?` at the end is an actual placeholder

**Test 10: Multiple `""` escapes with no question mark**
- SQL: `'SELECT "a""b""c" FROM t'`
- Expected return: `False`
- Reason: No `?` anywhere — multiple `""` escapes but no question mark at all

**Test 11: Question mark sandwiched between `""` escapes**
- SQL: `'SELECT "a""?""b" FROM t'`
- Expected return: `False`
- Reason: The `?` is inside the identifier `a"?"b` — surrounded by escaped double quotes but still within the identifier boundaries

**Test 12: Double-quoted `""` escape with pyformat params**
- SQL: `'SELECT "col""?" FROM t WHERE id = %(id)s'`
- Parameters: `{"id": 1}`
- Expected behavior: Should succeed without raising `TypeError`
- Expected `result_params`: `(1,)`
- Reason: The `?` is inside the double-quoted identifier and should not trigger a "Parameter style mismatch" error

**Test 13: Double-quoted `""` escape with real question mark and dict**
- SQL: `'SELECT "col""?" FROM t WHERE id = ?'`
- Parameters: `{"id": 1}`
- Expected behavior: Should raise `TypeError` with message containing `"positional placeholders"`
- Reason: The `?` inside the identifier is ignored, but the real `?` after `WHERE id =` is a genuine placeholder mismatch with dict params

**Test 14: Empty single-quoted string**
- SQL: `"SELECT '' FROM t"`
- Expected return: `9` (position after the second `'`)
- Reason: `''` is either an empty string literal or an escaped quote — the scanner should consume both characters

**Test 15: Empty double-quoted identifier**
- SQL: `'SELECT "" FROM t'`
- Expected return: `9` (position after the second `"`)
- Reason: `""` is an empty double-quoted identifier — the scanner should consume both characters

**Test 16: Empty bracketed identifier**
- SQL: `"SELECT [] FROM t"`
- Expected return: `9` (position after the `]`)
- Reason: `[]` is an empty bracketed identifier

## Tests for boundary/edge cases in `_has_unquoted_question_marks`

**Test 17: Question mark immediately after closing bracket**
- SQL: `"SELECT [col]? FROM t"`
- Expected return: `True`
- Reason: The `?` at position 13 is immediately after the closing `]` — it's outside the bracketed identifier and is a real unquoted question mark

**Test 18: Question mark immediately after closing single quote**
- SQL: `"SELECT 'str'? FROM t"`
- Expected return: `True`
- Reason: The `?` is immediately after the closing `'` — it's outside the string literal and is unquoted

**Test 19: Only whitespace characters**
- SQL: `"   \n\t  "`
- Expected return: `False`
- Reason: No `?` present at all

**Test 20: Question mark between two block comments**
- SQL: `"SELECT /* comment */ ? /* comment */"`
- Expected return: `True`
- Reason: The `?` between the two properly closed block comments is unquoted

## Tests for mixed `?` and `%(name)s` interaction

### `detect_and_convert_parameters` integration tests

**Test 21: Question mark in block comment with pyformat params**
- SQL: `"SELECT /* hint? */ * FROM t WHERE id = %(id)s"`
- Parameters: `{"id": 1}`
- Expected behavior: Should succeed without raising `TypeError`
- Expected `result_params`: `(1,)`
- Reason: The `?` inside the block comment should be recognized as non-placeholder, so no mismatch error

**Test 22: Question mark in string literal with pyformat params**
- SQL: `"SELECT '?' FROM t WHERE id = %(id)s"`
- Parameters: `{"id": 42}`
- Expected behavior: Should succeed without raising `TypeError`
- Expected `result_params`: `(42,)`
- Reason: The `?` inside the string literal is not a real placeholder

**Test 23: Question mark in single-line comment with pyformat params**
- SQL: `"SELECT -- comment?\nFROM t WHERE id = %(id)s"`
- Parameters: `{"id": 5}`
- Expected behavior: Should succeed without raising `TypeError`
- Expected `result_params`: `(5,)`
- Reason: The `?` inside the single-line comment is not a real placeholder

mssql_python/parameter_helper.py
jahnvi480: Fixed this

mssql_python/parameter_helper.py
jahnvi480: Fixed this

mssql_python/parameter_helper.py
jahnvi480: Fixed this

tests/test_015_pyformat_parameters.py
jahnvi480: Few tests were already covered.. added the remaining tests as well.

mssql_python/parameter_helper.py
jahnvi480: We already have test for this. and it is working fine

mssql_python/parameter_helper.py
jahnvi480: Addressed this bug and fixed it and also added testcase for the same.

Issue #458, opened 2026-02-27
VARBINARY conversion error when passing a NULL


amachanic:
### Describe the bug
When passing a NULL to a VARBINARY column for insert, an error occurs rather than the correct behavor.

### To reproduce
```python
import mssql_python

conn = mssql_python.connect([conn string], autocommit=True)
cur = conn.cursor()
cur.execute('create table bytestest (b varbinary(50));')

cur.execute('insert into bytestest values (?)', None)
```

### Exception message:
```
ProgrammingError: Driver Error: Syntax error or access violation; DDBC Error: [Microsoft][SQL Server]Implicit conversion from data type varchar to varbinary is not allowed. Use the CONVERT function to run this query.
```

### Expected behavior
The None should be inserted as a NULL and I should see no error about VARCHAR.

PR #466, opened 2026-03-03
FIX: NULL parameter type mapping for VARBINARY columns (#458)

gargsaumya:

> [AB#42894](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/42894)

> GitHub Issue: #458 

### Summary   
  
This pull request makes a targeted fix to the parameter type mapping logic for NULL values in the `mssql_python/cursor.py` file. The change improves compatibility and prevents implicit conversion errors when executing queries with NULL parameters.

Parameter mapping improvement:

* Changed the mapping for NULL parameters in the `_map_sql_type` function to use `SQL_UNKNOWN_TYPE` instead of `SQL_VARCHAR`, allowing the driver to correctly determine the column type and avoid conversion errors.

### DIFF
--- a/tests/test_004_cursor.py
+++ b/tests/test_004_cursor.py
@@ -184,15 +184,13 @@ def test_mixed_empty_and_null_values(cursor, db_connection):
     try:
         # Create test table
         drop_table_if_exists(cursor, "#pytest_empty_vs_null")
-        cursor.execute(
-            """
+        cursor.execute("""
             CREATE TABLE #pytest_empty_vs_null (
                 id INT,
                 text_col NVARCHAR(100),
                 binary_col VARBINARY(100)
             )
-        """
-        )
+        """)
         db_connection.commit()
 
         # Insert mix of empty and NULL values
@@ -907,15 +905,13 @@ def test_rowcount(cursor, db_connection):
         cursor.execute("INSERT INTO #pytest_test_rowcount (name) VALUES ('JohnDoe3');")
         assert cursor.rowcount == 1, "Rowcount should be 1 after third insert"
 
-        cursor.execute(
-            """
+        cursor.execute("""
             INSERT INTO #pytest_test_rowcount (name) 
             VALUES 
             ('JohnDoe4'), 
             ('JohnDoe5'), 
             ('JohnDoe6');
-        """
-        )
+        """)
         assert cursor.rowcount == 3, "Rowcount should be 3 after inserting multiple rows"
 
         cursor.execute("SELECT * FROM #pytest_test_rowcount;")
@@ -1011,14 +1007,12 @@ def test_fetchmany_size_zero_lob(cursor, db_connection):
     """Test fetchmany with size=0 for LOB columns"""
     try:
         cursor.execute("DROP TABLE IF EXISTS #test_fetchmany_lob")
-        cursor.execute(
-            """
+        cursor.execute("""
             CREATE TABLE #test_fetchmany_lob (
                 id INT PRIMARY KEY,
                 lob_data NVARCHAR(MAX)
             )
-        """
-        )
+        """)
 
         # Insert test data
         test_data = [(1, "First LOB data"), (2, "Second LOB data"), (3, "Third LOB data")]
@@ -1043,14 +1037,12 @@ def test_fetchmany_more_than_exist_lob(cursor, db_connection):
     """Test fetchmany requesting more rows than exist with LOB columns"""
     try:
         cursor.execute("DROP TABLE IF EXISTS #test_fetchmany_lob_more")
-        cursor.execute(
-            """
+        cursor.execute("""
             CREATE TABLE #test_fetchmany_lob_more (
                 id INT PRIMARY KEY,
                 lob_data NVARCHAR(MAX)
             )
-        """
-        )
+        """)
 
         # Insert only 3 rows
         test_data = [(1, "First LOB data"), (2, "Second LOB data"), (3, "Third LOB data")]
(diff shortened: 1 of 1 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -471,14 +471,19 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
                         LOG("BindParameters: SQLDescribeParam failed for "
-                            "param[%d] (NULL parameter) - SQLRETURN=%d",
+                            "param[%d] (NULL parameter) - SQLRETURN=%d, falling back to SQL_VARCHAR",
                             paramIndex, rc);
-                        return rc;
+                        sqlType = SQL_VARCHAR;
+                        columnSize = 1;
+                        decimalDigits = 0;
+                    } else {
+                        sqlType = describedType;
```
bewithgaurav, line 485: could you add some tests to cover the else scenario? seems uncovered atm
the fallback scenario does look covered

gargsaumya: The "else" scenario (when SQLDescribeParam succeeds) IS covered by existing tests:

Covered by:
1. [test_none] (line 2495) - Inserts None into NVARCHAR(255) column → SQLDescribeParam succeeds, returns NVARCHAR type
2. [test_executemany_ints_with_none] (line 1766) - Inserts None into INT column → SQLDescribeParam succeeds, returns INT type

mssql_python/cursor.py
```diff
@@ -396,7 +396,7 @@ def _map_sql_type(  # pylint: disable=too-many-arguments,too-many-positional-arg
         if param is None:
             logger.debug("_map_sql_type: NULL parameter - index=%d", i)
             return (
-                ddbc_sql_const.SQL_VARCHAR.value,
+                ddbc_sql_const.SQL_UNKNOWN_TYPE.value,
```
subrata-ms, line 399: I think we should not map NULL to a SQL_UNKNOWN_TYPE. In SQL , NULL is not a data type and hence mapping it to SQL_UNKNOWN_TYPE is wrong. Can you please check whether you can use SQL_NULL_DATA  for representing the NULL.

gargsaumya: Thanks for the review @subrata-ms . 
I understand your semantic concern, NULL in SQL represents the absence of a value, not an “unknown type.” However, SQL_NULL_DATA cannot be used here because it serves a completely different purpose within the ODBC API.

In ODBC, SQL_UNKNOWN_TYPE (value 0) is used as the SQL type in SQLBindParameter to indicate that the driver should determine the actual parameter type. In contrast, SQL_NULL_DATA (value -1) is not a type at all, it is an indicator used in the StrLen_or_IndPtr buffer to signal that the parameter value is NULL.

In the current implementation, this distinction is handled correctly:

1. The Python layer returns SQL_UNKNOWN_TYPE as the SQL type, which triggers the C++ layer to resolve the actual type.
2. The C++ layer calls SQLDescribeParam to discover the target column type (e.g., VARBINARY, INT, NVARCHAR).
3. The C++ layer sets *strLenOrIndPtr = SQL_NULL_DATA, which correctly communicates that the value is NULL.

So, SQL_NULL_DATA is already being used exactly where it should be, as an indicator of a NULL value. Meanwhile, SQL_UNKNOWN_TYPE is intentionally used to trigger type discovery.
If we attempted to pass SQL_NULL_DATA (-1) as the SQL type, it would not behave as intended. Instead, it would be interpreted as SQL_LONGVARCHAR (which also has the value -1), leading to incorrect type binding.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -471,14 +471,19 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
                         &describedSize, &describedDigits, &nullable);
                     if (!SQL_SUCCEEDED(rc)) {
+                        // SQLDescribeParam can fail for generic SELECT statements where
+                        // no table column is referenced. Fall back to SQL_VARCHAR as a safe default.
                         LOG("BindParameters: SQLDescribeParam failed for "
-                            "param[%d] (NULL parameter) - SQLRETURN=%d",
+                            "param[%d] (NULL parameter) - SQLRETURN=%d, falling back to SQL_VARCHAR",
                             paramIndex, rc);
-                        return rc;
+                        sqlType = SQL_VARCHAR;
```
sumitmsft, line 481: For `INSERT` into a non-VARCHAR column where `SQLDescribeParam` fails for an unexpected reason (driver bug, network issue, etc.), silently falling back to `SQL_VARCHAR` could mask the real error and cause a different, harder-to-debug type mismatch later. Can we consider logging at a WARN level (not just LOG) when falling back because silent fallback could be surprising

mssql_python/cursor.py
sumitmsft, line 399: Can you add a test similar to this:

```python
def test_execute_none_into_varbinary_column(db_connection):
    """Regression test for #458: None into VARBINARY should not raise varchar->varbinary error."""
    drop_table_if_exists(cursor, "#test_varbinary_null")

    try:
        cursor.execute("CREATE TABLE #test_varbinary_null (data VARBINARY(100))")
        cursor.execute("INSERT INTO #test_varbinary_null (data) VALUES (?)", None)
        cursor.execute("SELECT data FROM #test_varbinary_null")
        row = cursor.fetchone()
        assert row[0] is None
    finally:
        drop_table_if_exists(cursor, "#test_varbinary_null")
        cursor.close()
```

gargsaumya: Done

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Done

Issue #270, opened 2025-10-01
Export Row class in __init__.py


Tweentyy:
Hey!

### Is your feature request related to a problem? Please describe.
Unable to type a row as Row.

### Describe the solution you'd like
Add the Row class to `__init__.py`.

### Additional context
Since this [PR](https://github.com/microsoft/mssql-python/pull/216), the fetchall and fetchone functions are working correctly (Cursor class). For my part, I would also like to be able to type them so that I can pass them as parameters/function return types.

bewithgaurav:
As of now, we will need only Row to be exported - since this is a very minor task, marking this `up for grabs` for anyone to contribute!

PR #474, opened 2026-03-10
FEAT: Export Row class and refactor __init__.py (#270)

jahnvi480:

> [AB#43062](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/43062)

> GitHub Issue: #270

### Summary   
This pull request refactors how module-level constants and decimal separator configuration are handled in the `mssql_python` package. Constants are now dynamically exported from the `constants` module, and decimal separator functions are moved to a dedicated module for improved maintainability and clarity. The changes also add new exports for row objects and simplify imports.

**Module-level constant export and API improvements:**

* Constants such as SQL types and GetInfo constants are now dynamically exported from the `constants` module, allowing direct import from `mssql_python` and reducing manual duplication. The `get_info_constants` function is also moved and exported from `constants.py`.
* All manual constant exports and the `get_info_constants` function implementation are removed from `__init__.py` in favor of wildcard imports from `constants.py`. 

**Decimal separator configuration refactor:**

* The decimal separator getter/setter logic is moved to a new module `decimal_config.py`, which encapsulates validation and C++ binding logic. A factory function now creates bound getter/setter functions for use in `__init__.py`. 
* The previous implementation of decimal separator functions in `__init__.py` is removed and replaced with the new factory-based approach. 

**API and import enhancements:**

* The `Row` object is now exported at the top level for easier access. 
* Import statements are clarified and simplified, including the settings import and constant exports. 
These changes improve the maintainability, clarity, and usability of the package's public API and configuration options.

### DIFF
--- a/mssql_python/__init__.py
+++ b/mssql_python/__init__.py
@@ -346,6 +346,24 @@ class _MSSQLModule(types.ModuleType):
         with _settings_lock:
             _settings.lowercase = value
 
+    @property
+    def native_uuid(self) -> bool:
+        """Get the native_uuid setting.
+
+        Controls whether UNIQUEIDENTIFIER columns return uuid.UUID objects (True)
+        or str (False). Default is True.
+        Set to False to return str for pyodbc-compatible migration.
+        """
+        return _settings.native_uuid
+
+    @native_uuid.setter
+    def native_uuid(self, value: bool) -> None:
+        """Set the native_uuid setting."""
+        if not isinstance(value, bool):
+            raise ValueError("native_uuid must be a boolean value")
+        with _settings_lock:
+            _settings.native_uuid = value
+
 
 # Replace the current module with our custom module class
 old_module: types.ModuleType = sys.modules[__name__]
@@ -364,3 +382,4 @@ sys.modules[__name__] = new_module
 
 # Initialize property values
 lowercase: bool = _settings.lowercase
+native_uuid: bool = _settings.native_uuid
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -203,6 +203,7 @@ class Connection:
         autocommit: bool = False,
         attrs_before: Optional[Dict[int, Union[int, str, bytes]]] = None,
         timeout: int = 0,
+        native_uuid: Optional[bool] = None,
         **kwargs: Any,
     ) -> None:
         """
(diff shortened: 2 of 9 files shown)


mssql_python/__init__.py
```diff
@@ -119,97 +121,13 @@ def _cleanup_connections():
...
-SQL_TXN_READ_UNCOMMITTED: int = ConstantsDDBC.SQL_TXN_READ_UNCOMMITTED.value
-SQL_TXN_READ_COMMITTED: int = ConstantsDDBC.SQL_TXN_READ_COMMITTED.value
-SQL_TXN_REPEATABLE_READ: int = ConstantsDDBC.SQL_TXN_REPEATABLE_READ.value
-SQL_TXN_SERIALIZABLE: int = ConstantsDDBC.SQL_TXN_SERIALIZABLE.value
-
-# Access Mode Constants
-SQL_MODE_READ_WRITE: int = ConstantsDDBC.SQL_MODE_READ_WRITE.value
-SQL_MODE_READ_ONLY: int = ConstantsDDBC.SQL_MODE_READ_ONLY.value
+# Import all module-level constants from constants module
+from .constants import *  # noqa: F401, F403
```
sumitmsft: Using from `.constants import *` makes it impossible to determine what names `mssql_python` exports by reading `__init__.py` alone. 

A developer (or IDE/type checker) must chase through `constants.py` to `__all__` to the dynamic `_exported_names` construction to `_DDBC_PUBLIC_API` set to understand the public API.

The old approach of explicit assignments (`SQL_CHAR: int = ConstantsDDBC.SQL_CHAR.value`) was repetitive, but it made the public API immediately visible and gave static analysis tools (mypy, pyright, IDE autocompletion) a clear picture.

Can we do something to make sure we do not significantly hurt maintainability and developer experience when working with this module?

mssql_python/decimal_config.py
```diff
@@ -0,0 +1,114 @@
...
+This module provides functions for managing decimal separator configuration.
+"""
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover
+    from mssql_python.helpers import Settings
+
+
+def setDecimalSeparator(separator: str, settings: "Settings", set_in_cpp_func=None) -> None:
```
sumitmsft: The module-level `setDecimalSeparator` and `getDecimalSeparator` have public-looking names but are only used internally by the `create_decimal_separator_functions`. External consumers should never call these directly, they require `settings` and `set_in_cpp_func` parameters that are wired up by the factory.

This creates a mild naming/convention issue: a contributor or tool exploring this module might mistake these for directly usable public functions.

mssql_python/__init__.py
jahnvi480: Explicitly mentioning the constants that we are exposing for visibility

mssql_python/decimal_config.py
jahnvi480: Added _ for that it is not exposed to end user

Issue #468, opened 2026-03-06
Inconsistent retrieval of CP1252 encoded data in VARCHAR columns - Windows vs. Linux


sschaub:
### Describe the bug
When retrieving data from a VARCHAR column containing byte value 173 using default encoding settings, the Windows driver returns a bytes value, while the Linux driver returns a str value.

I suspect this is happening because, under Windows, the driver is attempting to decode the data in a VARCHAR column using utf-8 encoding, and a single byte value 173, which is a valid value in various code pages, including the default Windows CP-1252, does not map to a valid character under utf-8 encoding. When the decode fails, the driver passes the data through, unencoded, as a bytes value. Under Linux, some other decoding behavior is going on that produces a str.

### To reproduce

```python
con = mssql_python.connect(... connect string ...)
c = con.cursor()
c.autocommit = True
c.execute("create table tab(col varchar(10) collate SQL_Latin1_General_CP1_CI_AS)")
# Insert a value containing a byte 173
c.execute("insert into tab values('\xad')")
c.execute("select * from tab")
print(c.fetchall())
```

Windows output:
```
[(b'\xad',)]
```

Linux output:
```
[('\xad',)]
```

### Expected behavior
When run under Windows and Linux I would expect this program to produce the same results - ideally, a string containing '\xad'.

Interestingly, when the `setdecoding()` option is used to specify encoding "cp1252", the output under both Windows and Linux is the same:

```
con = mssql_python.connect(... connect string ...)
con.setdecoding(mssql_python.SQL_CHAR, "cp1252")
...
```

produces output:

```
[('\xad',)]
```

### Further technical details
Python version: 3.14
SQL Server version: SQL Server 2022
Operating system: Docker container

subrata-ms:
**Change Summary::**
This pull request updates the default handling of SQL CHAR/VARCHAR columns to use UTF-16 (wide character) encoding instead of UTF-8, primarily to address encoding mismatches on Windows and ensure consistent Unicode decoding. The changes span the connection, cursor, and C++ binding layers, and update related tests to reflect the new default behavior.

**Default Encoding and Decoding Changes:**

[...]

PR #478, opened 2026-03-19
FIX: Inconsistent retrieval of CP1252 encoded data in VARCHAR columns - Windows vs. Linux #468

subrata-ms:

> [AB#43177](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/43177)

> GitHub Issue: #468 

### Summary   
  
This pull request makes significant improvements to how VARCHAR (and related CHAR) columns are fetched and decoded, especially for Windows platforms, to ensure consistent Unicode string handling across all operating systems. The changes unify the decoding behavior for problematic CP1252 byte values, preventing platform-specific inconsistencies and ensuring that VARCHAR columns always return Python `str` objects, not `bytes`, regardless of the server collation or platform. The update also adds comprehensive tests to verify this behavior.

**Cross-platform VARCHAR decoding and buffer management:**

* On Windows, VARCHAR columns are now always fetched as `SQL_C_WCHAR`, ensuring the ODBC driver converts from the server's native encoding (such as CP1252) to UTF-16, matching the behavior of NVARCHAR columns and eliminating the need to guess the server code page. This change guarantees that Python always receives Unicode strings (`str`) for VARCHAR columns, even when they contain bytes that are invalid in UTF-8. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R2998-R3001) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3142-R3201) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3606-R3671) [[4]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3906-R3913)
* On Linux/macOS, the logic for buffer allocation and decoding remains unchanged, continuing to fetch as `SQL_C_CHAR` and decode as UTF-8, but the code is refactored for clarity and consistency with the Windows path. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3072-R3073) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3076-L3103) [[3]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3120-L3124) [[4]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1L3794-R3839) [[5]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R3848-R3851)

**Testing improvements:**

* Adds 23 new parameterized tests to verify that all problematic CP1252 byte values in VARCHAR columns are always decoded to `str` (not `bytes`) on all platforms and via all fetch methods (`fetchone`, `fetchall`, `fetchmany`).
* Adds tests to ensure that embedded problematic bytes and explicit CP1252 decoding are handled correctly, further validating the new cross-platform consistency.
* Updates the test count in the encoding/decoding test module to reflect the increased coverage.

These changes ensure robust, predictable Unicode handling for VARCHAR columns across all supported platforms, and the expanded test suite helps prevent regressions in encoding/decoding behavior.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -184,6 +184,127 @@ struct NumericData {
     }
 };
 
+struct Int128_t {
+    uint64_t low;
+    int64_t high;
+
+    Int128_t() : low(0), high(0) {}
+    Int128_t(uint64_t l, int64_t h) : low(l), high(h) {}
+
+    Int128_t multiply_by_10() const {
+        // value * 10 = (value * 8) + (value * 2)
+        Int128_t shift3 = *this << 3;
+        Int128_t shift1 = *this << 1;
+        return shift3 + shift1;
+    }
+
+    Int128_t operator<<(int shift) const {
+        // These would require special cases. We only shift by 1 and 3 for multiply_by_10.
+        assert(shift > 0);
+        assert(shift < 64);
+        uint64_t new_low = low << shift;
+        uint64_t new_high = (static_cast<uint64_t>(high) << shift) | (low >> (64 - shift));
+        return {new_low, static_cast<int64_t>(new_high)};
+    }
+
+    Int128_t operator+(const Int128_t& other) const {
+        uint64_t sum_low = low + other.low;
+        uint64_t carry = (sum_low < low) ? 1 : 0;
+        int64_t sum_high = high + other.high + carry;
+        return {sum_low, sum_high};
+    }
+
+    Int128_t operator+(uint64_t digit) const {
+        uint64_t sum_low = low + digit;
+        uint64_t carry = (sum_low < low) ? 1 : 0;
+        int64_t sum_high = high + carry;
+        return {sum_low, sum_high};
+    }
+
+    Int128_t operator-() const {
+        uint64_t new_low = ~low + 1;
+        uint64_t new_high = ~high + (new_low == 0 ? 1 : 0);
+        return {new_low, static_cast<int64_t>(new_high)};
+    }
+};
+
+struct ArrowArrayPrivateData {
+    std::unique_ptr<uint8_t[]> valid;
+
+    std::unique_ptr<uint8_t[]> uint8Val;
+    std::unique_ptr<int16_t[]> int16Val;
+    std::unique_ptr<int32_t[]> int32Val;
+    std::unique_ptr<int64_t[]> int64Val;
+    std::unique_ptr<double[]> float64Val;
+    std::unique_ptr<float[]> float32Val;
+    std::unique_ptr<uint8_t[]> bitVal;
+    std::unique_ptr<uint64_t[]> varVal;
+    std::unique_ptr<int32_t[]> dateVal;
+    std::unique_ptr<int64_t[]> tsMicroVal;
+    std::unique_ptr<int32_t[]> timeSecondVal;
+    std::unique_ptr<Int128_t[]> decimalVal;
+
+    std::vector<uint8_t> varData;
+
+    // first buffer will be the valid bitmap
+    // second buffer will be one of the value buffers above
+    // third buffer will be the varData buffer for variable length types
+    std::array<void*, 3> buffers;
+
+    // Points to one of the typed *Val buffers above. Since the buffer pointers
+    // don't change, this can be set once during batch initialization.
+    void* ptrValueBuffer;
+};
+
+struct ArrowSchemaPrivateData {
+    std::unique_ptr<char[]> name;
+    std::unique_ptr<char[]> format;
+};
+
+#ifndef ARROW_C_DATA_INTERFACE
+#define ARROW_C_DATA_INTERFACE
+
+#define ARROW_FLAG_DICTIONARY_ORDERED 1
+#define ARROW_FLAG_NULLABLE 2
+#define ARROW_FLAG_MAP_KEYS_SORTED 4
+
+struct ArrowSchema {
+  // Array type description
+  const char* format;
+  const char* name;
+  const char* metadata;
+  int64_t flags;
+  int64_t n_children;
+  struct ArrowSchema** children;
+  struct ArrowSchema* dictionary;
+
+  // Release callback
+  void (*release)(struct ArrowSchema*);
+  // Opaque producer-specific data
+  // Only our child-arrays will set this, so we can give it the correct type
+  ArrowSchemaPrivateData* private_data;
+};
+
+struct ArrowArray {
+  // Array data description
+  int64_t length;
+  int64_t null_count;
+  int64_t offset;
+  int64_t n_buffers;
+  int64_t n_children;
+  const void** buffers;
+  struct ArrowArray** children;
+  struct ArrowArray* dictionary;
+
+  // Release callback
+  void (*release)(struct ArrowArray*);
+  // Opaque producer-specific data
+  // Only our child-arrays will set this, so we can give it the correct type
+  ArrowArrayPrivateData* private_data;
+};
+
+#endif  // ARROW_C_DATA_INTERFACE
+
 //-------------------------------------------------------------------------------------------------
 // Function pointer initialization
 //-------------------------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2914,6 +2914,10 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
     // Note: wcharEncoding parameter is reserved for future use
     // Currently WCHAR data always uses UTF-16LE for Windows compatibility
     (void)wcharEncoding;  // Suppress unused parameter warning
+#if !defined(__APPLE__) && !defined(__linux__)
```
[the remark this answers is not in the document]
subrata-ms: done

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3037,6 +3023,66 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                            LOG("SQLGetData: Column %d is NULL (VARCHAR via WCHAR)", i);
+                            row.append(py::none());
+                        } else if (dataLen == 0) {
+                            row.append(py::str(""));
+                        } else if (dataLen == SQL_NO_TOTAL) {
+                            LOG("SQLGetData: Cannot determine data length "
+                                "(SQL_NO_TOTAL) for column %d (VARCHAR via WCHAR), "
+                                "returning NULL",
+                                i);
+                            row.append(py::none());
```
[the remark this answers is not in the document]

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3153,6 +3147,74 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                        (columnSize + 1) * sizeof(SQLWCHAR);  // +1 for null terminator
+                    std::vector<SQLWCHAR> dataBuffer(columnSize + 1);
+                    SQLLEN dataLen;
+                    ret = SQLGetData_ptr(hStmt, i, SQL_C_WCHAR, dataBuffer.data(), fetchBufferSize,
+                                         &dataLen);
+                    if (SQL_SUCCEEDED(ret)) {
+                        if (dataLen > 0) {
+                            uint64_t numCharsInData = dataLen / sizeof(SQLWCHAR);
+                            if (numCharsInData < dataBuffer.size()) {
+                                std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()));
```
sumitmsft, line 3293: The `std::wstring` constructor relies on finding a null terminator in the wide-char buffer. However, if the driver doesn't null-terminate exactly at the expected position (e.g., data fills the buffer boundary), this reads past the intended data or truncates. The variable numCharsInData is already computed from dataLen but is not used to bound the wstring construction. Do you think this as a problem?
Can we use the length-bounded constructor: to match the dataLen-derived character count and avoid reading garbage past the end of valid data?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2995,6 +2995,10 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
     // Note: wcharEncoding parameter is reserved for future use
     // Currently WCHAR data always uses UTF-16LE for Windows compatibility
     (void)wcharEncoding;  // Suppress unused parameter warning
+#if defined(_WIN32)
+    // On Windows, VARCHAR is fetched as SQL_C_WCHAR, so charEncoding is unused.
+    (void)charEncoding;
```
sumitmsft, line 3121: I consider this as a big behavioral (or potentially breaking) change, why? Because:
 
Consider a user with a server using a rare or custom collation (say a legacy Thai or Japanese single-byte encoding). Before this PR, they could call `setdecoding(SQL_CHAR, encoding="tis-620")` to control how bytes were interpreted. After this PR on Windows, that call is silently ignored, the driver always asks ODBC for UTF-16, which may or may not produce the correct result depending on the driver's conversion capabilities. The user gets no indication that their explicit configuration was bypassed.

What I get from this change is that the `charEncoding` parameter is explicitly (void) suppressed on Windows and all three fetch paths (`SQLGetData_wrap`, `SQLBindColums`, `FetchBatchData`) now unconditionally use `SQL_C_WCHAR` for VARCHAR columns. This means the user-facing `setdecoding(SQL_CHAR, encoding="cp1252", ctype=SQL_CHAR)` API call is accepted without error but has no effect on Windows.

The user's explicit decoding configuration is silently discarded. The test `test_cp1252_varchar_explicit_decoding` passes only because SQL_C_WCHAR handles CP1252 natively, not because the `setdecoding` call influenced anything.

sumitmsft: I am still thinking about what could be the possible solution for this to avoid this "potentially" big change.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Actually this is not a strict bug, the ODBC spec guarantees null-termination because the buffer is allocated as [columnSize + 1] WCHARs, so there is always room. The [numCharsInData < dataBuffer.size()] guard also confirms data didn't fill the entire buffer including the +1 slot. That said, the length-bounded constructor is still a better choice, it makes it consistent with the existing NVARCHAR path a few lines above, which already uses it. Worth fixing for clarity.

tests/test_013_encoding_decoding.py
```diff
@@ -7256,5 +7256,179 @@ def test_dae_encoding_large_string(db_connection):
...
+        else:
+            rows = cursor.fetchmany(1)
+
+        assert len(rows) == 1
+        value = rows[0][0]
+        assert isinstance(value, str), (
+            f"{fetch_method}: expected str but got {type(value).__name__}: "
+            f"{value!r} (platform={sys.platform})."
+        )
+        assert "\u00ad" in value, (
```
gargsaumya, line 7365: The fetchone parametrized tests all use strict equality (==). This batch test uses "in", so containment check would pass even if the driver incorrectly prepended/appended characters. Please change to assert value == "\u00ad" to match the stronger assertion pattern used in the rest of the test suite.

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: seems to be genuine issue and fixed under the next commit.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: There's actually a cleaner fix that avoids the breaking change entirely. `_decoding_settings` already stores both `encoding `and `ctype`, but only `encoding` is forwarded to C++ today and `ctype `is never passed down. So instead, as a fix:

- We can also pass `ctype` into` FetchOne/Many/All_wrap`

- In the `SQL_CHAR`/`SQL_VARCHAR` C++ branch: use `SQL_C_WCHAR` when `ctype == SQL_WCHAR` and use `SQL_C_CHAR `+ decode with `charEncoding `when `ctype == SQL_CHAR`

- Change the default `ctype` for `SQL_CHAR` from `SQL_CHAR` → `SQL_WCHAR`

This means default users get `SQL_C_WCHAR` automatically (fixed). Users who explicitly passed `ctype=SQL_CHAR` keep their existing behavior unchanged. No platform #if guards needed, and the Linux locale issue @ffelixg raised is also covered since the default becomes `SQL_C_WCHAR`.

subrata-ms: Thanks @gargsaumya for the details.
The solution that you are suggesting is almost same like what we have implemented in current PR considering memory consumption, performance and correctness. But above solution seems to be more robust approach as it removes multiple OS level checks (#-ifdef) in code and hence improves the code readability and maintainability. There is one problem with this approach and that is like we can't use SQL_WCHAR for some dibean linux version. It actually not supported. But we can make a twik in the code for not to consider SQL_WCHAR for those unsupported linux versions. Othere side with SQL_CHAR, we don't face this issue across all platform. 
I am abandoning this PR as this is difficult to rework (rollback the changes + arrow support also created some test failure) and equally difficult to review further. Below is the PR which we will be reviewing and following going forward.

https://github.com/microsoft/mssql-python/pull/495

Issue #388, opened 2025-12-23
Caching issues with EntraID credentials


j0nahm:
### Is your feature request related to a problem? Please describe.

Yes, I am experiencing ~1s latency per connection when using `Authentication=ActiveDirectoryDefault` with `AzureCliCredential` (in a FastAPI app).

The `mssql-python` driver instantiates a new `DefaultAzureCredential` object for every new `connect()` call. Since the Azure CLI credential in the `azure-identity` library does not provide token caching, this triggers a slow az subprocess call for every single database request, creating a massive performance bottleneck.

### Describe the solution you'd like

The driver should implement its own token caching mechanism.

### Describe alternatives you've considered

I've considered passing my own token using `attrs_before`, but this would require developing my own token cache. I think it makes more sense for the  DB driver to handle the caching.

### Additional context

The [redis-py-entraid](https://github.com/redis/redis-py-entraid) library avoids this issue by implementing its own token caching layer.

jahnvi480:
@j0nahm Thank you for raising this issue, the PR for this bug has been merged, this fix will be available in our next release.
Closing this GitHub issue.

PR #483, opened 2026-03-23
FIX: Credential instance cache

jahnvi480:

> [AB#43254](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/43254)

> GitHub Issue: #388

### Summary   
This pull request introduces credential instance caching for Azure Active Directory (AAD) authentication in the `mssql_python` package, improving performance by reusing credential objects instead of creating new ones for each token request. It also adds comprehensive tests for the caching logic and edge cases, and includes a new benchmark script to measure the performance impact of credential caching.

**Credential instance caching and logic changes:**

* Added a module-level credential cache (`_credential_cache`) and lock in `auth.py` to reuse Azure Identity credential instances per authentication type, enabling the SDK's in-memory token cache and reducing redundant token acquisitions. (`mssql_python/auth.py`)
* Modified `AADAuth.get_raw_token` to use the cached credential instance rather than creating a new one each call, ensuring efficient token reuse. (`mssql_python/auth.py` )

**Testing improvements:**

* Added a pytest fixture to clear the credential cache between tests, ensuring test isolation. (`tests/test_008_auth.py` )
* Introduced a new `TestCredentialInstanceCache` class with tests verifying credential reuse, separation by auth type, and cache state. (`tests/test_008_auth.py`)
* Added tests for error handling and edge cases, including missing Azure libraries, authentication errors, and unusual connection string parameter cases. (`tests/test_008_auth.py` )

**Benchmarking:**

* Added a new benchmark script (`benchmarks/bench_credential_cache.py`) to measure and compare the performance of credential caching versus the previous behavior of creating new credentials for every token request. (`benchmarks/bench_credential_cache.py`)


mssql_python/auth.py
```diff
@@ -6,11 +6,19 @@
...
 from typing import Tuple, Dict, Optional, List
 
 from mssql_python.logging import logger
 from mssql_python.constants import AuthType, ConstantsDDBC
 
+# Module-level credential instance cache.
+# Reusing credential objects allows the Azure Identity SDK's built-in
+# in-memory token cache to work, avoiding redundant token acquisitions.
+# See: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/TOKEN_CACHING.md
+_credential_cache: Dict[str, object] = {}
```
sumitmsft, line 19: Do we have any eviction policy? In long-running processes or scenarios where auth types are dynamic, stale credential objects will remain in memory indefinitely

tests/test_008_auth.py
```diff
@@ -403,6 +413,201 @@ def test_unsupported_auth(self):
...
+                raise RuntimeError("credential creation exploded")
+
+        try:
+            azure_identity.DefaultAzureCredential = FailingCredential
+            result = get_auth_token("default")
+            assert result is None
+        finally:
+            azure_identity.DefaultAzureCredential = original
+
+
```
sumitmsft, line 611: Add the following tests:

1. Verify thread-safe behavior, there is no test which validates that concurrent access creates only one credential instance per auth type.
2. `test_client_auth_error_in_acquire_token` verifies the error is raised but does not check whether the broken credential remains in `_credential_cache` afterward. This is critical for understanding the retry behavior.
3. `test_get_token_general_exception_handling_init_error` tests that a `RuntimeError` is raised when `credential_class()` fails, but does not verify whether a stale/partial entry was left in `_credential_cache`. Since the exception occurs during `_credential_cache[auth_type] = credential_class()`, the dict assignment should never complete but this should be explicitly asserted.
4. The test checks cache dict internal state (`_credential_cache["default"`] is first_instance) but never asserts on the returned token bytes from `AADAuth.get_token()`. A caching test should verify the primary output is correct on both cache-miss and cache-hit paths.
5. In `process_connection_string()` (auth.py), when `get_auth_token()` returns `None` (token acquisition failure), the function falls through to return `(conn_str, None, auth_type)`. No test covers this specific path.

mssql_python/auth.py
jahnvi480: Currently there's no explicit eviction policy, but the cache is naturally bounded, the keys are the auth_type values, which can only be one of three fixed strings: "default", "devicecode", or "interactive". So the cache holds at most 3 credential objects regardless of how long the process runs. Since auth types aren't dynamic, stale accumulation isn't a concern.

PR #493, opened 2026-03-31
FIX: Fixing test that does not work properly / Failing in connection pool

subrata-ms:

> [AB#40818](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/40818)

### Summary   
  
This pull request improves the robustness, reliability, and debuggability of the connection pooling and cleanup logic in the SQL Server Python driver. It introduces safer error handling in destructors, enhances logging for connection pool operations, and makes the pooling tests more reliable and meaningful by using unique connection identifiers and removing unnecessary skips.

**Error handling and resource management:**
- The `Connection` destructor now safely handles exceptions from `disconnect()` by logging and suppressing them, preventing potential crashes due to exceptions during stack unwinding.
- In `disconnect()`, errors from `SQLDisconnect` are logged instead of thrown, ensuring cleanup is always attempted and never throws from contexts like destructors or pool cleanup.

**Logging and diagnostics:**
- Added detailed logging in the connection pool's `acquire()` method to track pool size, idle timeout, and connection idle times, aiding in diagnosing pool behavior.
- Improved formatting and consistency of log messages in connection handle compaction routines. [[1]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecL115-R122) [[2]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecL240-R253)
- Enhanced error log formatting for non-statement handle detection in `disconnect()`.

**Connection pooling configuration:**
- Removed the unnecessary `std::once_flag` from `enable_pooling()`, allowing reconfiguration of the pool for each test run.

**Testing improvements:**
- Re-enabled and rewrote previously skipped pooling tests to be more reliable:
  - `test_pool_idle_timeout_removes_connections` now uses `connection_id` instead of `@@SPID` to uniquely identify physical connections and increases the idle wait to avoid flakiness in CI environments.
  - `test_pool_removes_invalid_connections` is now robust, directly simulates a dirty connection, and verifies that the pool discards invalid connections and creates new ones using `connection_id`.
  - Improved password replacement logic in `test_pool_recovery_after_failed_connection` for broader compatibility with different connection string formats.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4514,10 +4514,8 @@ SQLLEN SQLRowCount_wrap(SqlHandlePtr StatementHandle) {
     return rowCount;
 }
 
-static std::once_flag pooling_init_flag;
 void enable_pooling(int maxSize, int idleTimeout) {
-    std::call_once(pooling_init_flag,
```
sumitmsft, line 5836: Duee to this change and several other changes, please help understand what will happen during this scenario:

If a user calls `pooling(max_size=50)` at startup, then later calls `pooling(max_size=10)` expecting to shrink their pools but nothing will happen now. The Python guard returns early because _enabled is already True. The pools continue with max_size=50.

In ideal scenario, max_size should have been 10.

If they go through `disable()` >> `enable(max_size=10)`, the `_reset_for_testing()` (which resets `_pools_closed`) is not available as a public API, so `disable()` sets `_pools_closed = True` and `enable()` may or may not work depending on the internal state machine.

I guess this is a subtle bug introduced here. Please validate

mssql_python/pybind/connection/connection.cpp
```diff
@@ -51,7 +51,13 @@ Connection::Connection(const std::wstring& conn_str, bool use_pool)
 }
 
 Connection::~Connection() {
-    disconnect();  // fallback if user forgets to disconnect
+    try {
+        disconnect();  // fallback if user forgets to disconnect
+    } catch (...) {
```
sumitmsft: The change to make `disconnect()` never-throw is correct for destructor and pool-cleanup paths. 

However, `disconnect()` is also the code path for explicit user `conn.close()` on non-pooled connections (via `ConnectionHandle::close()` >> `_conn->disconnect()`).

After this change, if `SQLDisconnect` fails (e.g. due to an uncommitted transaction or a corrupt connection state), the user gets no error at all - `close()` returns successfully and the failure is only visible in debug logs. This breaks the expectation that explicit operations report their failures.

Can we split disconnect() into two methods:
1. `disconnect()` - throws on failure (used by `ConnectionHandle::close()`)
2. `disconnect_nothrow()` - swallows errors (used by `~Connection()`, `ConnectionPool::close()`, `Connection::reset()`)

This way, cleanup paths remain safe while user-facing operations continue to report errors.

tests/test_009_pooling.py
```diff
@@ -278,82 +270,92 @@ def try_overflow():
...
     # First, try to connect with a bad password (should fail)
-    if "Pwd=" in conn_str:
-        bad_conn_str = conn_str.replace("Pwd=", "Pwd=wrongpassword")
-    elif "Password=" in conn_str:
-        bad_conn_str = conn_str.replace("Password=", "Password=wrongpassword")
-    else:
+    import re
+
+    # Replace the value of the first Pwd/Password key-value pair with "wrongpassword"
+    pattern = re.compile(r"(?i)(Pwd|Password\s*=\s*)([^;]*)")
```
sumitmsft: Can you please validate this regex `Pwd|Password\s*=\s*`

Looks wrong, if you are testing it for wrong password.

sumitmsft (changes requested):

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: fixed and added additional test to verify this behavior

mssql_python/pybind/connection/connection.cpp
subrata-ms: good suggestion. Fixed in latest commit.

tests/test_009_pooling.py
subrata-ms: fixed

subrata-ms:
replacing with a new PR - https://github.com/microsoft/mssql-python/pull/564

PR #495, opened 2026-04-03
FIX: Inconsistent retrieval of CP1252 encoded data in VARCHAR columns - Windows vs. Linux #468

subrata-ms:

> [AB#43177](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/43177)

> GitHub Issue: #468 

### Summary   
  
This pull request updates the default handling of SQL `CHAR`/`VARCHAR` columns to use UTF-16 (wide character) encoding instead of UTF-8, primarily to address encoding mismatches on Windows and ensure consistent Unicode decoding. The changes span the connection, cursor, and C++ binding layers, and update related tests to reflect the new default behavior.

**Default Encoding and Decoding Changes:**

* The default decoding for SQL `CHAR` columns is now set to use `"utf-16le"` encoding and the `SQL_WCHAR` ctype, replacing the previous `"utf-8"`/`SQL_CHAR` defaults. This avoids issues where Windows ODBC drivers return raw bytes in the server's native code page, which may not decode as UTF-8. (`mssql_python/connection.py`, [mssql_python/connection.pyR264-R271](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R264-R271))

* All cursor fetch methods (`fetchone`, `fetchmany`, `fetchall`) are updated to request UTF-16 decoding and pass the correct ctype when fetching `CHAR` data, ensuring consistent behavior across platforms. (`mssql_python/cursor.py`, [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L2371-R2373) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L2437-R2440) [[3]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L2495-R2499)

**C++ Binding and Processing Updates:**

* The `ColumnInfoExt` struct now tracks whether wide character (UTF-16) fetching is used for a column, and the `ProcessChar` function is updated to handle both wide and narrow character paths, decoding appropriately based on the new setting. (`mssql_python/pybind/ddbc_bindings.h`, [[1]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635R671) [[2]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635R795-R798) [[3]](diffhunk://#diff-85167a2d59779df18704284ab7ce46220c3619408fbf22c631ffdf29f794d635R816-R846)

**Test Adjustments:**

* Tests are updated to expect `"utf-16le"` and `SQL_WCHAR` as the default decoding settings for `SQL_CHAR` columns, and to validate the new default behavior. (`tests/test_013_encoding_decoding.py`, [[1]](diffhunk://#diff-97f01c2139fb5a0dc283aacb2982e014f2f3cd8bb88079451eba91362e2fb3f1L607-R614) [[2]](diffhunk://#diff-97f01c2139fb5a0dc283aacb2982e014f2f3cd8bb88079451eba91362e2fb3f1L4924-R4927) [[3]](diffhunk://#diff-97f01c2139fb5a0dc283aacb2982e014f2f3cd8bb88079451eba91362e2fb3f1L5651-R5656)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -5070,7 +5103,9 @@
                             target_vec->resize(target_vec->size() * 2);
                         }
 
-                        std::memcpy(&(*target_vec)[start], &buffers.charBuffers[idxCol][idxRowSql * fetchBufferSize], dataLen);
+                        std::memcpy(&(*target_vec)[start],
+                                    &buffers.charBuffers[idxCol][idxRowSql * fetchBufferSize],
+                                    dataLen);
```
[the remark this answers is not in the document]
subrata-ms: fetchBufferSize and dataLen should take care of the memory allocation buffer size

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -5088,7 +5123,9 @@
...
```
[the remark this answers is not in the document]
subrata-ms: fetchBufferSize and dataLen should take care of the memory allocation buffer size

tests/test_013_encoding_decoding.py
```diff
@@ -7256,5 +7258,374 @@ def test_dae_encoding_large_string(db_connection):
...
+# those >= 0x80) are consistently returned as Python str regardless of platform.
+#
+# Previously (before fix): Windows + default UTF-8 decoding → bytes (fallback)
+# Now (after fix): Default SQL_C_WCHAR → str on all platforms
+#
+# Users can still explicitly set SQL_C_CHAR via setdecoding() for backward compat.
+# ====================================================================================
+
+
+@pytest.mark.skipif(
```
gargsaumya, line 7278: One gap worth considering: the default ctype for SQL_CHAR is now SQL_C_WCHAR / utf-16le on all platforms, not just Windows. On Linux/macOS, ProcessChar now takes the PyUnicode_DecodeUTF16 branch for every VARCHAR fetch instead of the previous PyUnicode_FromStringAndSize path. There are no platform-neutral tests guarding this new code path.

A small non-skipif test class with two cases would help:
  1. A plain ASCII VARCHAR value — the most common case, ensures the new default doesn't regress the happy path on Linux/macOS.
  2. A non-ASCII Unicode value in a VARCHAR column (e.g. a UTF-8 collation or an explicit CONVERT) — ensures multi-byte characters round-trip correctly through the new PyUnicode_DecodeUTF16 path on Linux/macOS.

Without this, a bug in the wide-char path on Linux/macOS (e.g. a byte-order or null-terminator issue) would not be caught by CI.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4729,14 +4817,15 @@ SQLRETURN FetchArrowBatch_wrap(
     ColumnBuffers buffers(numCols, fetchSize);
 
     if (!hasLobColumns && fetchSize > 0) {
-        // Bind columns
-        ret = SQLBindColums(hStmt, buffers, columnNames, numCols, fetchSize);
+        // Bind columns — Arrow always uses SQL_C_CHAR for VARCHAR because
+        // it processes raw byte buffers directly, not via Python codecs.
+        ret = SQLBindColums(hStmt, buffers, columnNames, numCols, fetchSize, SQL_C_CHAR);
```
gargsaumya, line 4822: This PR fixes issue #468 for fetchone/fetchmany/fetchall, but the identical bug remains for fetch_arrow_batch() on Windows. A user migrating to Arrow-based fetching to improve performance would still get garbled or raw-byte data for any CP-1252 high byte (≥ 0x80) in a VARCHAR column.

The comment says "not via Python codecs" — but the Arrow path already has the infrastructure to handle `SQL_C_WCHAR` for `SQL_WCHAR/NVARCHAR` columns: it converts the UTF-16 buffer to UTF-8 via `WideCharToMultiByte(CP_UTF8, ...)` on Windows (and `WideToUTF8(SQLWCHARToWString(...)`) on Linux/macOS) and writes the result into the Arrow varData buffer. The same approach would work for VARCHAR columns bound as `SQL_C_WCHAR` — the `SQL_CHAR/SQL_VARCHAR/SQL_LONGVARCHAR` case in the Arrow data loop could be extended with an #ifdef _WIN32 branch that reads from wcharBuffers and converts to UTF-8, identical to the existing `NVARCHAR` path.

If fixing this is out of scope for this PR, it would be helpful to document the limitation in code. We can then track this as a follow-up improvement.

cc: @ffelixg  would appreciate your thoughts here.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -808,6 +813,37 @@ inline void ProcessChar(PyObject* row, ColumnBuffers& buffers, const void* colIn
...
+                PyUnicode_DecodeUTF16(reinterpret_cast<const char*>(wcharData),
+                                      numCharsInData * sizeof(SQLWCHAR), nullptr, nullptr);
+#else
+            PyObject* pyStr =
+                PyUnicode_FromWideChar(reinterpret_cast<const wchar_t*>(wcharData), numCharsInData);
+#endif
+            if (!pyStr) {
+                PyErr_Clear();
+                Py_INCREF(Py_None);
+                PyList_SET_ITEM(row, col - 1, Py_None);
```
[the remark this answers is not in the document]
gargsaumya: +1 to Copilot's comment here. Worth adding a bit more context on why this matters.

The three decode-failure paths across closely related functions currently behave differently:

  ProcessChar narrow-char  → falls back to raw bytes  (data preserved)
  ProcessWChar             → falls back to empty str   (data lost but visible)
  ProcessChar wide-char    → stores None               (data lost, looks like SQL NULL)

None is the most harmful choice because it is indistinguishable from a genuine SQL NULL to the caller — there is no way to detect the difference. That said, a PyUnicode_FromWideChar / PyUnicode_DecodeUTF16 failure on data that the ODBC driver has already converted to valid UTF-16 almost exclusively means OOM, so this branch is very unlikely to fire in practice. Still, it seems worth aligning with the narrow-char fallback (PyBytes_FromStringAndSize over the raw SQLWCHAR buffer) so that if it ever does fire, no data is silently lost and the behavior is consistent across the function.

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: @gargsaumya @subrata-ms Fundamentally I would prefer if it was impossible for the user to configure the driver such that corrupt data is returned. If we want that, we can only allow two scenarios:

- Fetch as WCHAR & decode utf16
- Fetch as CHAR & decode utf8 IF locale is set to utf8 at connection time

Even if the locale is set to cp1252 and the decoding is also correctly set to cp1252, we may still return corrupt data if the column we are querying has any other locale. From what I can tell there isn't even any easy way for us to know the actual column collation so this corruption would be silent (I think the driver simply replaces characters with a questionmark)

I get that for fetchall etc we may want to preserve other decoding scenarios for backwards compatibility, but since the arrow fetch path is new, I think the best course of action would be to either throw exceptions if we can't guarantee correct data for the current configuration or silently use WCHAR to ensure correct data.

So the simple variant is what Saumya says: Always pass SQL_C_WCHAR to SQLBindColums and integrate the SQL_VARCHAR cases with the SQL_WVARCHAR cases. Optionally raise a NotImplementedError if the user deviated from the default utf16 configuration.

A more complex, but also potentially ~2x more performant option would be to keep the existing path, but only allow it if the decoding is set to utf8 and the connection-time locale is also utf8. The scope of this performance boost would be limited though, since the majority of SQL Server users already default to nvarchars for string data, so for now it's probably best to keep it safe and simple.

mssql_python/connection.py
```diff
@@ -261,10 +261,14 @@ def __init__(
...
+        # UTF-16 data for VARCHAR columns. This avoids encoding mismatches on
+        # Windows where the driver returns raw bytes in the server's native
+        # code page (e.g. CP-1252) that may fail to decode as UTF-8.
         self._decoding_settings = {
             ConstantsDDBC.SQL_CHAR.value: {
-                "encoding": "utf-8",
-                "ctype": ConstantsDDBC.SQL_CHAR.value,
+                "encoding": "utf-16le",
+                "ctype": ConstantsDDBC.SQL_WCHAR.value,
             },
```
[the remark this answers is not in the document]
subrata-ms: fixed

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3186,13 +3187,99 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                        if (dataLen > 0) {
+                            uint64_t numCharsInData = dataLen / sizeof(SQLWCHAR);
+                            if (numCharsInData < dataBuffer.size()) {
+#if defined(__APPLE__) || defined(__linux__)
+                                std::wstring wstr =
+                                    SQLWCHARToWString(dataBuffer.data(), numCharsInData);
+                                std::string utf8str = WideToUTF8(wstr);
+                                row.append(py::str(utf8str));
+#else
+                                std::wstring wstr(reinterpret_cast<wchar_t*>(dataBuffer.data()));
```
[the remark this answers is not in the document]

mssql_python/pybind/ddbc_bindings.h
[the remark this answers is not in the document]
subrata-ms: fixed.

tests/test_013_encoding_decoding.py
subrata-ms: The default scenario (test cases) is now incorporated in this PR. The issue also got reported by a customer https://github.com/microsoft/mssql-python/issues/531

tests/test_013_encoding_decoding.py
```diff
@@ -7256,5 +7258,374 @@ def test_dae_encoding_large_string(db_connection):
...
+        CP-1252 defines mappings for all byte values 0-255 (except 5 undefined
+        positions: 0x81, 0x8D, 0x8F, 0x90, 0x9D which Python's cp1252 codec
+        will raise on).
+        """
+        db_connection.setdecoding(SQL_CHAR, encoding="cp1252")
+
+        # CP-1252 defined high bytes (excluding 0x81, 0x8D, 0x8F, 0x90, 0x9D)
+        defined_bytes = [b for b in range(128, 256) if b not in (0x81, 0x8D, 0x8F, 0x90, 0x9D)]
+        for b in defined_bytes:
+            cursor.execute(f"INSERT INTO {self.TABLE_NAME} (id, data) VALUES ({b}, CHAR({b}))")
```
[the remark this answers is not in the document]
subrata-ms: fixed

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: @gargsaumya , We will address the CP1252 for Arrow support under this bug https://github.com/microsoft/mssql-python/issues/553 and task https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/44922

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3318,13 +3338,105 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                                    i);
+                                row.append(FetchLobColumnData(hStmt, i, SQL_C_WCHAR, true, false,
+                                                              "utf-16le"));
+                            }
+                        } else if (dataLen == SQL_NULL_DATA) {
+                            LOG("SQLGetData: Column %d is NULL (CHAR via WCHAR)", i);
+                            row.append(py::none());
+                        } else if (dataLen == 0) {
+                            row.append(py::str(""));
+                        } else if (dataLen == SQL_NO_TOTAL) {
```
sumitmsft, line 3427: In the new wide-char `SQLGetData` path, when `dataLen == SQL_NO_TOTAL` , the code logs a message and appends `py::none()`. This is indistinguishable from a genuine SQL NULL to the caller.

The existing narrow-char path (and the LOB check at the top of the case block) correctly handles `SQL_NO_TOTAL` by falling through to `FetchLobColumnData` for streaming. Shouldn't the wide-char path should do the same?

PS: `SQL_NO_TOTAL` means the driver can't determine the length upfront - it does not mean the data is absent. Returning `None` here could silently loses data.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3318,13 +3338,105 @@ SQLRETURN SQLGetData_wrap(SqlHandlePtr StatementHandle, SQLUSMALLINT colCount, p
...
+                            row.append(py::none());
+                        } else if (dataLen < 0) {
+                            LOG("SQLGetData: Unexpected negative data length "
+                                "for column %d - dataType=%d, dataLen=%ld",
+                                i, dataType, (long)dataLen);
+                            ThrowStdException("SQLGetData returned an unexpected negative "
+                                              "data length");
+                        }
+                    } else {
+                        LOG("SQLGetData: Error retrieving data for column %d "
```
sumitmsft: Here also, the user silently gets None with no way to know data retrieval failed. 

In the narrow-char path (which correctly returns the error SQLRETURN and lets the caller handle it), the wide-char path should be consistent with it. 

Or, if there's a specific reason to be lenient here, at minimum can we throw an exception so the error is observable? Silently converting errors to NULL values is a data integrity risk.

mssql_python/connection.py
```diff
@@ -261,10 +261,14 @@ def __init__(
...
         # Initialize decoding settings with Python 3 defaults
+        # SQL_CHAR default uses SQL_WCHAR ctype so the ODBC driver returns
+        # UTF-16 data for VARCHAR columns. This avoids encoding mismatches on
+        # Windows where the driver returns raw bytes in the server's native
+        # code page (e.g. CP-1252) that may fail to decode as UTF-8.
         self._decoding_settings = {
             ConstantsDDBC.SQL_CHAR.value: {
-                "encoding": "utf-8",
-                "ctype": ConstantsDDBC.SQL_CHAR.value,
+                "encoding": "utf-16le",
```
sumitmsft, line 270: This changes the default SQL_CHAR decoding from utf-8 + SQL_C_CHAR to utf-16le + SQL_C_WCHAR for all users on all platforms. While the fix is correct for the CP1252 issue, this is a silent behavioral change that affects every existing consumer. 

Add a changelog / release notes entry explicitly calling out this default change and the motivation (CP1252 mismatch).

sumitmsft (commented):
Left some minor comments. Otherwise, I am ok with this change

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: I got curious if requesting SQL_C_CHARs could actually be more performant, so [I integrated the SQLBindColumns changes from this PR with the simdutf PR branch](https://github.com/ffelixg/mssql-python/commit/219c27dcea96eef2864fdf3e88527bce47adefba). Note that thematically this doesn't fit into the simdutf branch, I just based it on that to make sure that suboptimal utf16 decoding didn't interfere with the performance numbers. [According to my results](https://gist.github.com/ffelixg/a0a95ac1d580476d7dc08a26ad685a94) there's no significant performance difference when requesting SQL_C_WCHAR instead of SQL_C_CHAR. Requesting more complicated characters like '€' didn't change the times either.

Interestingly, selecting an NCHAR did slow things down by 2x, even though exactly the same memory is allocated & code executed on our side. Possible explanation could be that the odbc driver converts to utf16 locally in the case of varchars and that network usage is still the limiting factor even with very good connectivity.

Either way, seems like there is no disadvantage to deleting the SQL_C_CHAR branches in favor of SQL_C_WCHAR, so let's just do that for arrow, I think. @gargsaumya @subrata-ms

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: agreed and fixed

mssql_python/connection.py
subrata-ms: we will add it in release note as needed

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: Ya. We can throw an exception. Made the same change in required places.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -68,6 +68,27 @@ inline std::string GetEffectiveCharDecoding(const std::string& userEncoding) {
...
+// SQL_C_CHAR + utf-8 decoding (e.g. setdecoding(SQL_CHAR, "utf-8", SQL_CHAR)),
+// the SQL Server ODBC driver on Windows returns VARCHAR data in the server's
+// ANSI code page (e.g. CP1252) regardless of the column's actual collation.
+// For UTF-8 collation columns or any non-ASCII data, this is lossy ('?'
+// substitution) and unrecoverable on the Python side. Internally upgrading
+// the fetch to SQL_C_WCHAR triggers the driver's lossless UTF-16 conversion,
+// which produces a correct Python Unicode string regardless of column
+// collation. On Linux/macOS the SQL_C_CHAR path already returns UTF-8 from
+// the driver, so this upgrade is a no-op there.
+inline int EffectiveCharCtypeForFetch(int charCtype, const std::string& charEncoding) {
```
jahnvi480, line 81: Minor suggestion (non-blocking): The `EffectiveCharCtypeForFetch` override silently upgrades `SQL_C_CHAR + utf-8` to `SQL_C_WCHAR` on Windows. While the output is correct (and better), the user's explicit `ctype` setting from `setdecoding(SQL_CHAR, encoding="utf-8", ctype=SQL_CHAR)` is overridden without any indication. Consider adding a LOG_INFO when this override triggers.
```cpp
#ifdef _WIN32
    if (charCtype == SQL_C_CHAR && charEncoding == "utf-8") {
        LOG_INFO("EffectiveCharCtypeForFetch: Upgrading SQL_C_CHAR + utf-8 to "
                 "SQL_C_WCHAR on Windows to avoid lossy ACP conversion");
        return SQL_C_WCHAR;
    }
#endif
```
This would help users debugging encoding behavior understand why their explicit `ctype=SQL_CHAR` setting isn't being honored.

subrata-ms: Yes. Log here would definitely enhance the clarity and helps users debugging encoding behavior.

jahnvi480 (dismissed):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -5929,16 +6069,19 @@ PYBIND11_MODULE(ddbc_bindings, m) {
...
           py::arg("StatementHandle"), py::arg("row"), py::arg("charEncoding") = "utf-8",
-          py::arg("wcharEncoding") = "utf-16le");
+          py::arg("wcharEncoding") = "utf-16le", py::arg("charCtype") = SQL_C_WCHAR);
     m.def("DDBCSQLFetchMany", &FetchMany_wrap, py::arg("StatementHandle"), py::arg("rows"),
           py::arg("fetchSize"), py::arg("charEncoding") = "utf-8",
-          py::arg("wcharEncoding") = "utf-16le", "Fetch many rows from the result set");
+          py::arg("wcharEncoding") = "utf-16le", py::arg("charCtype") = SQL_C_WCHAR,
+          "Fetch many rows from the result set");
     m.def("DDBCSQLFetchAll", &FetchAll_wrap, "Fetch all rows from the result set",
           py::arg("StatementHandle"), py::arg("rows"), py::arg("charEncoding") = "utf-8",
```
gargsaumya: The pybind11 default for charEncoding is still "utf-8" here, but this PR changed the connection-level default to "utf-16le" everywhere else (Connection.__init__, setdecoding(), all three cursor fallbacks in cursor.py). I think these three defaults should be "utf-16le" to stay consistent with the rest of the PR., just veify this once?

subrata-ms: addressed in latest commit. Thanks for bringing this to my attention.

bewithgaurav (dismissed):
approving with test addition suggestions below.

logic lgtm. fix for #468 reproduces clean against sql2022, new TestIssue531Utf8CollationVarchar covers the SQL_C_WCHAR-default path on linux/macOS, all 185 encoding tests pass.

one ask before merge: the test coverage can be improved.

I went through each missing line. four python tests would close most of the gap:

1. fetchone with NULL + empty VARCHAR. covers null/empty arms in the new CHAR-via-WCHAR block.
```python
cursor.execute("SELECT CAST(NULL AS VARCHAR(10)), CAST('' AS VARCHAR(10))")
row = cursor.fetchone()
assert row[0] is None and row[1] == ''
```

2. fetchone with VARCHAR(MAX) and NVARCHAR(MAX) carrying 50k+ non-ASCII. exercises wide-char LOB streaming in ProcessChar and SQLGetData_wrap. large LOBs also surface SQL_NO_TOTAL on the first chunk, picks up that branch incidentally.
```python
payload = "caf \u00ad " * 6250  # ~50k
cursor.execute("CREATE TABLE #t (v VARCHAR(MAX) COLLATE Latin1_General_100_CI_AS_SC_UTF8, n NVARCHAR(MAX))")
cursor.execute("INSERT INTO #t VALUES (?, ?)", payload, payload)
cursor.execute("SELECT v, n FROM #t")
assert cursor.fetchone() == (payload, payload)
```

3. fetchone on a TEXT column. legacy LOB, deterministically returns SQL_NO_TOTAL on the first chunk. cheapest way to actually exercise that branch.
```python
cursor.execute("CREATE TABLE #t (v TEXT)")
cursor.execute("INSERT INTO #t VALUES (?)", "A" * 100_000)
cursor.execute("SELECT v FROM #t")
assert cursor.fetchone()[0] == "A" * 100_000
```

4. fetchall on VARCHAR(N) with UTF-8 collation at size boundary. catches off-by-one in the wide-char path with non-BMP chars.
```python
cursor.execute("CREATE TABLE #t (v VARCHAR(4) COLLATE Latin1_General_100_CI_AS_SC_UTF8)")
cursor.execute("INSERT INTO #t VALUES (N'\U0001F600')")
assert cursor.fetchall()[0][0] == '\U0001F600'
```

the decode-failure-to-bytes branch and the throw paths can't easily be hit from python. fine as a follow-up, or at minimum mark them as defensive in a code comment so a future refactor doesn't quietly strip them.

subrata-ms:
> approving with test addition suggestions below.
> 
> logic lgtm. fix for #468 reproduces clean against sql2022, new TestIssue531Utf8CollationVarchar covers the SQL_C_WCHAR-default path on linux/macOS, all 185 encoding tests pass.
> 
> one ask before merge: the test coverage can be improved.
> 
> I went through each missing line. four python tests would close most of the gap:
> 
> 1. fetchone with NULL + empty VARCHAR. covers null/empty arms in the new CHAR-via-WCHAR block.
> 
> ```python
> cursor.execute("SELECT CAST(NULL AS VARCHAR(10)), CAST('' AS VARCHAR(10))")
> row = cursor.fetchone()
> assert row[0] is None and row[1] == ''
> ```
> 
> 2. fetchone with VARCHAR(MAX) and NVARCHAR(MAX) carrying 50k+ non-ASCII. exercises wide-char LOB streaming in ProcessChar and SQLGetData_wrap. large LOBs also surface SQL_NO_TOTAL on the first chunk, picks up that branch incidentally.
> 
> ```python
> payload = "caf \u00ad " * 6250  # ~50k
> cursor.execute("CREATE TABLE #t (v VARCHAR(MAX) COLLATE Latin1_General_100_CI_AS_SC_UTF8, n NVARCHAR(MAX))")
> cursor.execute("INSERT INTO #t VALUES (?, ?)", payload, payload)
> cursor.execute("SELECT v, n FROM #t")
> assert cursor.fetchone() == (payload, payload)
> ```
> 
> 3. fetchone on a TEXT column. legacy LOB, deterministically returns SQL_NO_TOTAL on the first chunk. cheapest way to actually exercise that branch.
> 
> ```python
> cursor.execute("CREATE TABLE #t (v TEXT)")
> cursor.execute("INSERT INTO #t VALUES (?)", "A" * 100_000)
> cursor.execute("SELECT v FROM #t")
> assert cursor.fetchone()[0] == "A" * 100_000
> ```
> 
> 4. fetchall on VARCHAR(N) with UTF-8 collation at size boundary. catches off-by-one in the wide-char path with non-BMP chars.
> 
> ```python
> cursor.execute("CREATE TABLE #t (v VARCHAR(4) COLLATE Latin1_General_100_CI_AS_SC_UTF8)")
> cursor.execute("INSERT INTO #t VALUES (N'\U0001F600')")
> assert cursor.fetchall()[0][0] == '\U0001F600'
> ```
> 
> the decode-failure-to-bytes branch and the throw paths can't easily be hit from python. fine as a follow-up, or at minimum mark them as defensive in a code comment so a future refactor doesn't quietly strip them.

Hi @bewithgaurav , this suggested test does not improve code coverage ( ~1%). The lines that are flagged as not covered are hard to reach with current test framework. We need to add dummy/test code within the production code to create this scenario (one option). I tried that and it was looking pretty big change on the current coding framework.

yan-hic:
I hit the same error using `cursor.arrow()` and `.arrow_reader()`. 

Can you add tests for these cases too, as the `conn.setdecoding(_SQL_CHAR, "cp1252") ` workaround only affects DBAPI fetching.

bewithgaurav (approved):

gargsaumya (approved):

jahnvi480 (approved):

subrata-ms:
> I hit the same error using `cursor.arrow()` and `.arrow_reader()`.
> 
> Can you add tests for these cases too, as the `conn.setdecoding(_SQL_CHAR, "cp1252") ` workaround only affects DBAPI fetching.

@yan-hic , Thanks for the comment. However, arrow related fix will be tracked under below bug - https://github.com/microsoft/mssql-python/issues/553

Issue #506, opened 2026-04-09
cursor.execute() with reset_cursor=False raises ProgrammingError: Invalid cursor state


dlevy-msft-sql:
### Describe the bug

Calling `cursor.execute()` with `reset_cursor=False` raises `ProgrammingError: Invalid cursor state` even after fully consuming the previous result set with `fetchall()`.

The wiki documents `reset_cursor` as a parameter that, when `False`, skips freeing and reinitializing the statement handle so the prepared plan can be reused. In practice, any call with `reset_cursor=False` fails.

```
Exception message: Driver Error: Invalid cursor state; DDBC Error: [Microsoft]Invalid cursor state
```

### To reproduce

```python
import mssql_python

conn = mssql_python.connect(connection_string)
cursor = conn.cursor()

# First execute works fine
cursor.execute(
    "SELECT TOP 1 ProductID FROM Production.Product WHERE ProductSubcategoryID = ?",
    (1,),
)
row = cursor.fetchone()
print(row.ProductID)  # Works

# Consume remaining results
_ = cursor.fetchall()

# Second execute with reset_cursor=False fails
cursor.execute(
    "SELECT TOP 1 ProductID FROM Production.Product WHERE ProductSubcategoryID = ?",
    (2,),
    reset_cursor=False,
)
# Raises ProgrammingError: Invalid cursor state
```

Note: The same sequence works fine with the default `reset_cursor=True`.

### Expected behavior

`cursor.execute()` with `reset_cursor=False` should reuse the prepared statement handle and execute successfully, as documented in the wiki.

### Further technical details
Python version: 3.14.0
mssql-python version: 1.4.0
SQL Server version: SQL Server 2022
Operating system: Windows 11 (ARM64, running x64 emulation)

**Additional context**
Tested against AdventureWorks2022. The default behavior (`reset_cursor=True`) works correctly. Only `reset_cursor=False` triggers the error.

jahnvi480:
Closing this issue, we have fixed this and is now available in our new release v1.6.0.

PR #521, opened 2026-04-14
FIX: cursor.execute() with reset_cursor=False raises Invalid cursor state 

jahnvi480:

> [AB#44009](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44009)

> GitHub Issue: #506 

### Summary   
This pull request improves the handling of cursor state when executing prepared statements with `reset_cursor=False`, ensuring that the prepared plan is correctly reused and the cursor state is properly managed between executions. It introduces a new `close_cursor` method at the ODBC statement handle level, updates the Python bindings, and adds comprehensive tests to verify the new behavior.

**Enhancements to cursor state management:**

- Added logic in `cursor.py` to call `hstmt.close_cursor()` (which issues an ODBC `SQLFreeStmt(SQL_CLOSE)`) when `reset_cursor=False`, allowing prepared statements to be reused safely without resetting the entire cursor state. This resolves issues with invalid cursor state on re-execution.

**ODBC handle and Python bindings updates:**

- Implemented the `close_cursor` method in the `SqlHandle` C++ class, which closes only the cursor on the statement handle without freeing the prepared statement.
- Exposed the new `close_cursor` method to Python via the `ddbc_bindings` Pybind11 module, allowing it to be called from Python code.

**Testing improvements:**

- Added multiple test cases in `test_004_cursor.py` to verify that `reset_cursor=False` correctly reuses prepared plans, returns new results, works across multiple iterations, handles queries without parameters, and functions correctly when the previous result set was not fully consumed.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -1362,6 +1362,12 @@ void SqlHandle::free() {
     }
 }
 
+void SqlHandle::close_cursor() {
+    if (_handle && SQLFreeStmt_ptr && _type == SQL_HANDLE_STMT) {
+        SQLFreeStmt_ptr(_handle, SQL_CLOSE);
+    }
+}
+
 SQLRETURN SQLGetTypeInfo_Wrapper(SqlHandlePtr StatementHandle, SQLSMALLINT DataType) {
     if (!SQLGetTypeInfo_ptr) {
         ThrowStdException("SQLGetTypeInfo function not loaded");
@@ -5740,7 +5746,8 @@ PYBIND11_MODULE(ddbc_bindings, m) {
         .def_readwrite("ddbcErrorMsg", &ErrorInfo::ddbcErrorMsg);
 
     py::class_<SqlHandle, SqlHandlePtr>(m, "SqlHandle")
-        .def("free", &SqlHandle::free, "Free the handle");
+        .def("free", &SqlHandle::free, "Free the handle")
+        .def("close_cursor", &SqlHandle::close_cursor, "Close the cursor on the statement handle without freeing the prepared statement");
 
     py::class_<ConnectionHandle>(m, "Connection")
         .def(py::init<const std::string&, bool, const py::dict&>(), py::arg("conn_str"),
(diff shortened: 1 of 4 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1362,6 +1362,12 @@ void SqlHandle::free() {
     }
 }
 
+void SqlHandle::close_cursor() {
+    if (_handle && SQLFreeStmt_ptr && _type == SQL_HANDLE_STMT) {
```
bewithgaurav, line 1366: `_implicitly_freed` safety needs to be added here - potential use-after-free can happen if connection dies before cursor is cleaned up
one-line fix - just need to add `&& !_implicitly_freed` to this guard

bewithgaurav (changes requested):
we're proposing to revamp the full reset cursor logic in our upcoming perf PRs
the complete hstmt deletion and creation should eventually be discarded & we will move to cursor resetting by default - considering the scope of this PR, this looks good.
requesting a change to add a safety guard for implicit freeing post conn deletion.
else lgtm

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1362,6 +1362,12 @@ void SqlHandle::free() {
     }
 }
 
+void SqlHandle::close_cursor() {
```
bewithgaurav, line 1365: `close_cursor()` is an internal function to be called by cursor obj, I think we dont need to expose this to end users (its public as of now)

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1362,6 +1362,12 @@ void SqlHandle::free() {
...
+    if (_handle && SQLFreeStmt_ptr && _type == SQL_HANDLE_STMT) {
+        SQLFreeStmt_ptr(_handle, SQL_CLOSE);
+    }
```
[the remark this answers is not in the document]
sumitmsft: Agreed with this suggestion - check the return code and throw on failure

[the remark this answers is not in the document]
jahnvi480: Resolved this comment

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: I have created _closed_cursor() function on python side so that it is not exposed to end user

sumitmsft (approved):

bewithgaurav (approved):

PR #522, opened 2026-04-14
REFACTOR: Migrate Connection string sanitization from regex to parser-based

bewithgaurav:

> [AB#43979](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/43979)

### Summary   
  
Replaces the regex-based `sanitize_connection_string()` with a parser-based implementation 
that uses `_ConnectionStringParser` to correctly handle all ODBC connection string value 
formats including braced values per ODBC spec.

## Changes

- Moved `sanitize_connection_string()` to `connection_string_parser.py` where it naturally 
  belongs alongside the parser it depends on — eliminates circular import between helpers 
  and parser modules
- `helpers.py` retains a thin delegate for backward compatibility
- `connection.py` imports directly from `connection_string_parser`
- Added 5 new test cases covering braced values, escaped braces, and edge cases

## Testing

- All existing `TestPasswordSanitization` tests pass
- All connection string parser and allowlist tests pass (no regressions)


mssql_python/connection_string_parser.py
```diff
@@ -375,3 +377,48 @@ def _parse_braced_value(self, connection_str: str, start_pos: int) -> Tuple[str,
...
+        params = parser._parse(conn_str)
+
+        sanitized_params = {}
+        for key, value in params.items():
+            canonical = _ConnectionStringParser.normalize_key(key)
+            display_key = canonical if canonical else key
+            if key in _SENSITIVE_KEYS:
+                sanitized_params[display_key] = "***"
+            else:
+                sanitized_params[display_key] = value
```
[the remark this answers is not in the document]
bewithgaurav: changed

jahnvi480 (approved):
PR LGTM! 👍

mssql_python/connection_string_parser.py
```diff
@@ -21,6 +21,8 @@
 from mssql_python.helpers import sanitize_user_input
 from mssql_python.logging import logger
 
+_SENSITIVE_KEYS = frozenset({"pwd"})
```
sumitmsft, line 24: Can you validate if a connection string uses `Password=secret123`, the password will be masked and will not appear in logs in plaintext?

tests/test_007_logging.py
```diff
@@ -314,20 +314,86 @@ def test_pwd_sanitization(self, cleanup_logger):
...
+    def test_malformed_string_fully_redacted(self, cleanup_logger):
+        """Malformed connection string should be fully redacted, not partially leaked."""
+        from mssql_python.helpers import sanitize_connection_string
+
+        conn_str = "PWD={unclosed"
+        sanitized = sanitize_connection_string(conn_str)
+
+        assert "unclosed" not in sanitized
+        assert "PWD" not in sanitized
+        assert "redacted" in sanitized.lower()
```
sumitmsft, line 426: Can you validate if "redacted" in "".lower() evaluates to False? If yes, then this test may fail at runtime.

bewithgaurav: `sanitized` will never be empty. 
The test is test_malformed_string_fully_redacted with input "PWD={unclosed", which hits the except path and returns "<redacted – unparseable connection string>". 
The assertion "redacted" in sanitized.lower() correctly passes.

If sanitized were somehow empty, the assertion should fail - that would mean the redaction didn't work. This targeted test expects 
in case user enters a blank pwd (`PWD=;`) it does not pass from parser level itself.
The parser rejects it with `Empty value for keyword 'pwd'`. Empty values are not allowed for any parameter.

mssql_python/connection_string_parser.py
bewithgaurav: we dont have `password` as an allowed connstring param, so this will get rejected by the parser itself (before this code gets invoked)
had just kept "pwd" as the sensitive param since we're not accepting any other sensitive maskable params
below is the error that pops in

```
Validation error: Connection string parsing failed:
  Unknown keyword 'password' is not recognized
Normalized: {'Server': 'localhost'}
```

gargsaumya (approved):

Issue #514, opened 2026-04-11
Performance regression in Linux for NVARCHAR strings when using Arrow 


gargsaumya:
### Describe the bug
Seeing performance regression for NVARCHAR strings when using Arrow. 
mssql-python version:1.5.0

### Further technical details
Operating system: (e.g. Windows Server 2022, Ubuntu 24.04, macOS 14.7.1, Docker container): Linux

**Additional context**

PR #526, opened 2026-04-15
PERF: Handle utf16 strings using simdutf and std::u16string

ffelixg:

> GitHub Issue: #514 

### Summary   
  
- Use external dependency simdutf to decode utf16 when decoding in c++
- Use pybind11 support for std::u16string to decode utf16 when interacting with python
- Remove SQLWCHARToWString, WideToUTF8, Utf8ToWString and WStringToSQLWCHAR; refactor call sites


ffelixg:
I have been evaluating the external library simdutf as a high performance replacement for utf16 -> utf8 conversions, i.e. the functions SQLWCHARToWString, WideToUTF8 and Utf8ToWString. Rather than only using it for the arrow fetch path, I have been trying to make the switch for every location where one of these three functions is used, as the applications follow similar patterns. I didn't use simdutf in every case, another way to eliminate these function calls was to use std::u16string instead of std::wstring when passing strings to/from python. I think this avoids the whole issue where wchars are defined as 32 bit on some OSes but SQLWCHARs are always 16 bit.

This brings arrow performance on linux for nvarchars in line with what it should be.

If the std::u16string type works as well as I hope it does (will have to see what CI says about mac), there are some more spots where it could be used, for example to replace `WStringToSQLWCHAR`

ffelixg:
Only issue seems to be that some Linux CI containers don't have git. I'm trying to fetch simdutf via url instead.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -167,9 +102,36 @@
...
+        simdutf::convert_utf16le_to_utf8_with_replacement(utf16.data(), utf16.size(), utf8.data()));
+    return utf8;
+}
+
+inline std::u16string dupeSqlWCharAsUtf16Le(const SQLWCHAR* value, size_t length) {
+    std::u16string utf16(length, u'\0');
+    static_assert(sizeof(SQLWCHAR) == sizeof(char16_t), "SQLWCHAR must be 16-bit");
+
+    if (length > 0) {
+        std::memcpy(utf16.data(), value, length * sizeof(SQLWCHAR));
```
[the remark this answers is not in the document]
ffelixg: Should be safe and necessary

ffelixg:
@gargsaumya The second CI error was due to the ubuntu image using an older version of cmake. Should be fine now I hope.

I have also replaced the remaining std::wstring occurences with std::u16string and eliminated helper functions as well as platform specific code accordingly. Let me know if you are happy with this holistic update to utf16 string handling or if you would prefer to keep it contained to the arrow fetch path.

gargsaumya:
Thanks for the update @ffelixg. I actually prefer this and it LGTM overall. ButI’ll still take a closer look at the changes and get back to you.
I’ve approved the workflow for now so we can run the pipeline and get a passing test run.

ffelixg:
I think that this latest CI failure was due to a flaky test. The same test / platform passed in a past CI run and I don't think my changes affected that test. Could you rerun that one @gargsaumya?

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -36,140 +38,40 @@ using py::literals::operator""_a;
...
-        // SQLWCHAR is UTF-32, so just copy with validation
-        for (size_t i = 0; i < length; ++i) {
-            uint32_t cp = static_cast<uint32_t>(sqlwStr[i]);
-            if (IsValidUnicodeScalar(cp)) {
-                result.push_back(static_cast<wchar_t>(cp));
-            } else {
-                result.push_back(static_cast<wchar_t>(UNICODE_REPLACEMENT_CHAR));
-            }
-        }
+inline std::string utf16LeToUtf8Alloc(std::u16string utf16) {
```
gargsaumya: The parameter is taken by value, causing a heap allocation and copy on every call. Since simdutf only reads the data, taking it as const std::u16string&  will avoid the unnecessary copy.

mssql_python/pybind/CMakeLists.txt
```diff
@@ -215,6 +215,24 @@ endif()
 
 message(STATUS "Final Python library directory: ${PYTHON_LIB_DIR}")
 
+find_package(simdutf CONFIG QUIET)
+
+if(NOT simdutf_FOUND)
+    include(FetchContent)
+    message(STATUS "simdutf not found via find_package; downloading v8.2.0 source archive with FetchContent")
+    set(simdutf_fetchcontent_args
+        URL https://github.com/simdutf/simdutf/archive/refs/tags/v8.2.0.tar.gz
```
gargsaumya, line 224: Can we add URL_HASH SHA256=<hash> after the URL to verify the downloaded archive. Without it, a tampered tarball at the same URL would be consumed silently.
We can get the hash with: `curl -sL https://github.com/simdutf/simdutf/archive/refs/tags/v8.2.0.tar.gz | sha256sum`

gargsaumya (commented):
Really appreciate the effort here, this is a very meaningful contribution @ffelixg. I've reviewed the PR and left 2 inline comments, please take a look!

ffelixg:
Glad you like it @gargsaumya! I totally agree with both your comments and pushed changes accordingly.

ffelixg:
Seems like the same flaky test failed again

gargsaumya (dismissed):

sumitmsft:
I have a few generic comments before I delve into core review:
@gargsaumya 

- Add a short note to CONTRIBUTING.md or a pybind/README.md: "Use std::u16string for all SQLWCHAR data; use simdutf for any UTF transcoding. Do not introduce std::wstring in the C++ bindings." 
- The three inline helpers (`utf16LeToUtf8Alloc`, `dupeSqlWCharAsUtf16Le`, `reinterpretU16stringAsSqlWChar`) are in `ddbc_bindings.h`. Would recommend to them to a dedicated `utf_utils.h` so they can be unit-tested in isolation.
- We need to subscribe to simdutf releases. They have aggressive release timelines. We would want to evaluate before bumping its versions.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -36,140 +38,40 @@ using py::literals::operator""_a;
...
+    return utf16;
 }
-#endif
 
-#if defined(__APPLE__) || defined(__linux__)
-#include "unix_utils.h"  // Unix-specific fixes
-#endif
+inline SQLWCHAR* reinterpretU16stringAsSqlWChar(const std::u16string& utf16) {
+    static_assert(sizeof(std::u16string::value_type) == sizeof(SQLWCHAR),
+        "SQLWCHAR must same as u16string");
```
sumitmsft: Minor issue: "SQLWCHAR must have the same size as std::u16string::value_type"

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -36,140 +38,40 @@ using py::literals::operator""_a;
...
+    static_assert(alignof(std::u16string::value_type) == alignof(SQLWCHAR),
+        "SQLWCHAR must same as u16string");
+    return const_cast<SQLWCHAR*>(reinterpret_cast<const SQLWCHAR*>(utf16.c_str()));
```
sumitmsft: Minor issue: "SQLWCHAR must have the same size as std::u16string::value_type"

mssql_python/pybind/ddbc_bindings.h
sumitmsft: This statement is no longer true after this PR. Please remove or update.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -36,140 +38,40 @@ using py::literals::operator""_a;
...
-        // SQLWCHAR is UTF-32, so just copy with validation
-        for (size_t i = 0; i < length; ++i) {
-            uint32_t cp = static_cast<uint32_t>(sqlwStr[i]);
-            if (IsValidUnicodeScalar(cp)) {
-                result.push_back(static_cast<wchar_t>(cp));
-            } else {
-                result.push_back(static_cast<wchar_t>(UNICODE_REPLACEMENT_CHAR));
-            }
-        }
+inline std::string utf16LeToUtf8Alloc(const std::u16string& utf16) {
```
sumitmsft: Would recommend to move `utf16LeToUtf8Alloc`, `dupeSqlWCharAsUtf16Le`, and `reinterpretU16stringAsSqlWChar` from `ddbc_bindings.h` into a small `utf_utils.h` so they can be unit tested and the changes don't force recompilation of the whole module. 

I do not see dedicated unit tests for utf16LeToUtf8Alloc, dupeSqlWCharAsUtf16Le, or reinterpretU16stringAsSqlWChar. Edge cases (empty strings, surrogate pairs, lone surrogates) are only tested indirectly through integration tests. Once moved to utf_utils.h I would suggest to add targeted unit tests.

mssql_python/pybind/ddbc_bindings.h
ffelixg: Whoops, not sure what happened to my grammar there, fixed.

mssql_python/pybind/ddbc_bindings.h
ffelixg: Good catch, I removed it. I think it actually hasn't been truly windows specific for a long time, since many core structures live in that file. There is still some windows specific code, albeit less.

mssql_python/pybind/ddbc_bindings.h
ffelixg: By targeted unit tests, am I correct in assuming that you mean pure c++ unit tests? Testing them via python wrappers and pytest feels wrong. It would also involve changes to the build process as well as CI pipelines, since there is currently no infrastructure for c++ tests. The predecessors of these functions were also only tested via integration or even fully inlined as was the case for some `reinterpretU16stringAsSqlWChar` applications.

sumitmsft: Good point! I agree that setting up a full C++ unit test framework and CI pipeline changes is out of scope for this PR. Let's not block this PR on it, but I'd like to open a tracking issue for it in our internal backlog for future enhancements.  

I see you've already added `utf_utils.h`, thank you! Let's keep moving with this PR. We are expecting one more review on this before we merge it.

sumitmsft: @gargsaumya please track this: "but I'd like to open a tracking issue for it in our internal backlog for future enhancements" in ADO.

ffelixg:
Thanks for the review @sumitmsft.

Regarding the first point, There's actually one more method that's currently being used to convert unicode aside from simdutf and `std::u16string`, which is the native python c api functions in ProcessWChar (fetchall/fetchmany). I changed your suggestion accordingly and put it into the pybind README. Not sure if you meant Saumya to do the documentation, since you tagged her, feel free to change it in any way @gargsaumya.

I hope the added dependency is not too much of an inconvenience. I looked into native iconv as well, but didn't find it appealing in comparison. The simdutf releases look to me like they are mostly about adding new features, which we likely won't care about. Since unicode conversion itself is a solved problem, I was hoping that we would technically be fine using the same version forever, assuming no crazy bugs.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1641,16 +1563,11 @@ ErrorInfo SQLCheckError_Wrap(SQLSMALLINT handleType, SqlHandlePtr handle, SQLRET
...
-            errorInfo.sqlState = std::wstring(sqlState);
-            errorInfo.ddbcErrorMsg = std::wstring(message);
-#else
-            // On macOS/Linux, need to convert SQLWCHAR (usually unsigned short)
-            // to wchar_t
-            errorInfo.sqlState = SQLWCHARToWString(sqlState);
-            errorInfo.ddbcErrorMsg = SQLWCHARToWString(message, messageLen);
-#endif
+            std::u16string sqlStateUtf16 = dupeSqlWCharAsUtf16Le(sqlState, 5);
+            std::u16string messageUtf16 = dupeSqlWCharAsUtf16Le(message, static_cast<size_t>(messageLen));
```
bewithgaurav: heads up, this is a pre-existing bug on the unix path that surfaced while reviewing this PR. flagging it here since the PR is touching the same code and the fix is essentially free now.

any time SQL Server returns an error message longer than 512 chars, this branch leaks driver stack memory into the user-visible exception text. trigger is common: a stored proc `RAISERROR` with row context interpolated in, a syntax error where the server echoes back a long ORM-generated query, a linked-server failure with prefixed context. T-SQL's own `RAISERROR` ceiling is 2047 chars so the surface area is real.

minimal repro, just one stored proc call wrapped in `try/except` + `logging.exception`:

```python
proc = """
DECLARE @msg NVARCHAR(2048) = REPLICATE(N'Order validation failed for line items in batch #4218: ', 25);
RAISERROR(@msg, 16, 1);
"""
cur.execute(proc)
```

what lands in the log:

```
mssql_python.exceptions.ProgrammingError: Driver Error: ...; DDBC Error: 
[Microsoft][SQL Server]Order validation failed for line items in batch #4218: 
Order validation failed for line items in batch #4218: ... Order validation
#  42000 
    ...
   
```

first ~511 chars are the real error. then `42000` is the SQLSTATE leaking through, then ~1500 chars of stack memory rendered as text. the recurring high-codepoint chars (``, ``, ``) are aarch64 heap pointer high bits being read 16 bits at a time. user impact: log noise, anything pattern-matching on error strings breaks on long errors, and wherever exceptions get shipped (sentry / datadog / support tickets) now ingests our process memory.

the why: `SQLGetDiagRec`'s contract is that on a too-small buffer it truncates + NULs the buffer but sets the `messageLen` out-param to the *full untruncated length* so callers can re-call with a bigger buffer. `dupeSqlWCharAsUtf16Le(message, messageLen)` trusts `messageLen` and `memcpy`s that many chars out of the (smaller) buffer, walking off the stack. cap `messageLen` at the buffer size in the helper and both this site and `SQLGetAllDiagRecords` at L1605 are covered.

checked main: the unix path on main had this exact bug (`SQLWCHARToWString(message, messageLen)` reads the same way), windows path was safe (`std::wstring(message)`, NTS). this PR collapses the platform split into one path that always uses `messageLen`, so windows regresses, macos/linux carry it forward. since the helper is new in this PR and both sites already touch it, makes sense to clamp here rather than spin a separate fix later.

same shape (different buffer + length pair) exists in `SQLDescribeCol_wrap` at L2783 with `ColumnName[256]` + `NameLength`. SQL Server's `sysname=128` keeps it from firing today, but a one-line clamp inside the helper covers it for free as well.

mssql_python/pybind/README.md
```diff
@@ -1,6 +1,6 @@
 # Build Instructions for Developers
 
-This README provides instructions to build the DDBC Bindings for your system and documents the platform-specific dependencies.
+This Chapter provides instructions to build the DDBC Bindings for your system and documents the platform-specific dependencies.
```
bewithgaurav: nit: looks like an accidental find/replace, "README" got changed to "Chapter" in the intro line.

mssql_python/pybind/utf_utils.h
```diff
@@ -0,0 +1,42 @@
...
+    if (utf16.empty()) {
+        return {};
+    }
+
+    simdutf::result utf8Length =
+        simdutf::utf8_length_from_utf16le_with_replacement(utf16.data(), utf16.size());
+    std::string utf8(utf8Length.count, '\0');
+    utf8.resize(
+        simdutf::convert_utf16le_to_utf8_with_replacement(utf16.data(), utf16.size(), utf8.data()));
+    return utf8;
```
bewithgaurav, line 22: utf8 worst-case ratio is bounded at 3x the input code units (BMP code points are 3 utf-8 bytes each, supplementary pairs are 4 utf-8 bytes for 2 utf-16 units = 2 bytes per unit). so a single-pass version is doable:

```cpp
std::string utf8(utf16.size() * 3, '\0');
size_t n = simdutf::convert_utf16le_to_utf8_with_replacement(
    utf16.data(), utf16.size(), utf8.data());
utf8.resize(n);
```

tradeoff: exact-alloc (current) keeps memory tight, single-pass saves the second walk at the cost of transient over-allocation (trimmed back on `resize`). happy to leave as-is if the exact-alloc behavior was intentional.

bewithgaurav (changes requested):
one real bug (diag message overflow, stack memory leaking into exception text on long error messages, repro inline), one accidental README typo, one perf nit on the new helper. only the diag bug actually blocks, the other two can ride along or land separately.

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: Oh yeah, that's not great at all! I agree that this PR is a decent place to sneak in a fix for it.

Apparently the PRINT function is the way to produce the longest possible message with 8000 characters and THROW is somewhere between RAISERROR and PRINT. Ontop of that we get the driver name in front of the message, which does not count to the SQL Server limit of 8000 characters.

Since the `SQL_MAX_MESSAGE_LENGTH` constant defined in `sql.h` appears to be wrong, I instead defined a custom `SQL_MAX_MESSAGE_LENGTH_SQLSERVER` which I set at 10000 characters. That means we allocate 20kb on the stack, but we're not targeting embedded or doing any recursion, so I think it's fine. An alternative would be to call SQLGetDiagRec twice to get the length first and heap allocate accordingly, but that doesn't seem necessary.

With that, there should be enough space to retrieve arbitrary errors/messages. I also added `test_long_raiserror` and `test_long_print_message` to cover the extremes.

Clamping the lengths for SQLDescribeCol / SQLGetDiagRec of course makes sense as well just in case.

mssql_python/pybind/README.md
ffelixg: My thinking was that with the added note about std::u16string, the README is no longer strictly about the build process. But technically, saying that the README provides build instructions, doesn't necessarily mean that it doesn't also provide something else and "chapter" also sounds a bit awkward, so I changed it back.

mssql_python/pybind/utf_utils.h
ffelixg: I did a bit of research and apparently there's no clear answer on which is better and both approaches are used in various projects. `simdutf` themselves also benchmark both approaches and don't provide clear guidance on which one to use, so either seems fine. I implemented the suggestion though.

bewithgaurav:
@ffelixg - thanks for your responses - requesting you to please resolve conflicts and we'll do a last set of review

ffelixg:
@bewithgaurav Done, I also noticed that a couple `utf16LeToUtf8Alloc` calls could be replaced by std::u16string + pybind11 while I was doing the conflict resolution.

bewithgaurav (approved):
lgtm, great stuff - thanks @ffelixg !

gargsaumya (approved):

Issue #500, opened 2026-04-07
cursor.execute() with many bound parameters (~2000) is ~14x slower than pyodbc


dxdc:
### Describe the bug

When executing a multi-row INSERT with ~2000 bound parameters (the kind SQLAlchemy's `insertmanyvalues` generates), `cursor.execute()` is about 14x slower than the equivalent call through pyodbc.

`cursor.executemany()` is much closer between drivers (~1.6x), so the issue is specific to single `cursor.execute()` calls with large parameter counts.

### To reproduce

```python
import time
import mssql_python
import pyodbc

N = 100_000
COLS = 2
# ~1049 rows per INSERT, ~2098 params per call, ~95 calls total
ROWS_PER_BATCH = 2099 // COLS
SQL = "INSERT INTO bench (id, name) VALUES " + ",".join(
    [f"(?, ?)"] * ROWS_PER_BATCH
)
PARAMS = [None] * (ROWS_PER_BATCH * COLS)

# --- mssql-python ---
conn = mssql_python.connect(
    "SERVER=localhost;DATABASE=master;UID=sa;PWD=YourPassword;"
    "Encrypt=yes;TrustServerCertificate=yes;"
)
cursor = conn.cursor()
cursor.execute(
    "IF OBJECT_ID('bench') IS NOT NULL DROP TABLE bench;"
    "CREATE TABLE bench (id INT, name VARCHAR(50))"
)
conn.commit()

t0 = time.perf_counter()
for _ in range(N // ROWS_PER_BATCH):
    cursor.execute(SQL, PARAMS)
conn.commit()
ms_time = time.perf_counter() - t0
cursor.close()
conn.close()

# --- pyodbc ---
conn2 = pyodbc.connect(
    "DRIVER={ODBC Driver 18 for SQL Server};"
    "SERVER=localhost;DATABASE=master;UID=sa;PWD=YourPassword;"
    "Encrypt=yes;TrustServerCertificate=yes;"
)
cursor2 = conn2.cursor()
cursor2.execute("TRUNCATE TABLE bench")
conn2.commit()

t0 = time.perf_counter()
for _ in range(N // ROWS_PER_BATCH):
    cursor2.execute(SQL, PARAMS)
conn2.commit()
py_time = time.perf_counter() - t0
cursor2.close()
conn2.close()

print(f"mssql-python: {ms_time:.2f}s ({N / ms_time:,.0f} rows/s)")
[...]

PR #528, opened 2026-04-17
PERF: Optimize execute() hot path: soft reset, prepare caching, and guarded diagnostics 

bewithgaurav:

> [AB#44166](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44166)

> Related GitHub Issue: #500 

### Summary   
  
This pull request introduces performance benchmarking infrastructure improvements and optimizations to the `mssql_python` driver, as well as documentation and workflow updates to support a new `PERF:` pull request prefix. The most significant changes are grouped below.

**Driver Performance Optimizations:**

* Implemented a lightweight `_soft_reset_cursor` in `cursor.py` to reuse prepared statement handles, avoiding unnecessary SQLPrepare calls when executing the same SQL repeatedly. This improves performance for repeated statement execution. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280R752-R763) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L1352-R1368) [[3]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L1429-R1461) [[4]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L1457-R1482)
* Optimized parameter conversion logic to skip redundant conversions when re-executing the same SQL with the same parameter style, further reducing overhead.
* Added a new `DDBCSQLResetStmt` binding in the C++ layer to support the lightweight reset operation, and exposed it to Python. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R1365-R1383) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R5805)


mssql_python/cursor.py
```diff
@@ -1363,8 +1377,10 @@ def execute(  # pylint: disable=too-many-locals,too-many-branches,too-many-state
 
         self._check_closed()  # Check if the cursor is closed
         if reset_cursor:
```
saurabh500, line 1384: Should we soft-reset only when use_prepare is true? Is there a reason to only check for `reset_cursor` flag before resetting?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1378,6 +1378,25 @@ void SqlHandle::close_cursor() {
...
+        DriverLoader::getInstance().loadDriver();
+    }
+    SQLHANDLE hStmt = statementHandle->get();
+    if (!hStmt) {
+        return SQL_INVALID_HANDLE;
+    }
+
+    SQLRETURN rc = SQLFreeStmt_ptr(hStmt, SQL_CLOSE);
+    if (SQL_SUCCEEDED(rc)) {
+        rc = SQLFreeStmt_ptr(hStmt, SQL_RESET_PARAMS);
```
saurabh500: Before any ODBC call, it is good to release the GIL
Release GIL in the scope of the `if` statement's {} 
Once the block exits GIL will be reacquired. 
This is safe in the current context and will allow better parallelism 

```suggestion
        py::gil_scoped_release release;
        rc = SQLFreeStmt_ptr(hStmt, SQL_RESET_PARAMS);
```

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1378,6 +1378,25 @@ void SqlHandle::close_cursor() {
...
+    }
+    if (SQL_SUCCEEDED(rc) && SQLSetStmtAttr_ptr) {
+        SQLSetStmtAttr_ptr(hStmt, SQL_ATTR_PARAMSET_SIZE, (SQLPOINTER)1, 0);
```
saurabh500: Release GIL before making ODBC calls. 

```suggestion
        py::gil_scoped_release release;
        SQLSetStmtAttr_ptr(hStmt, SQL_ATTR_PARAMSET_SIZE, (SQLPOINTER)1, 0);
```

mssql_python/cursor.py
```diff
@@ -1491,8 +1512,8 @@ def execute(  # pylint: disable=too-many-locals,too-many-branches,too-many-state
             self._reset_cursor()
             raise
 
-        # Capture any diagnostic messages (SQL_SUCCESS_WITH_INFO, etc.)
-        if self.hstmt:
+        # Capture diagnostic messages only when the driver signalled info.
+        if ret == ddbc_sql_const.SQL_SUCCESS_WITH_INFO.value and self.hstmt:
```
saurabh500, line 1523: Why was the `if` condition enhanced? Wasn't the code already doing the right thing by capturing all the Diag Records ?

mssql_python/cursor.py
```diff
@@ -747,6 +749,18 @@ def _reset_cursor(self) -> None:
 
         # Reinitialize the statement handle
         self._initialize_cursor()
+        self.is_stmt_prepared = [False]
```
saurabh500, line 751: Why an array  ? For mutable reference to the bool ? If yes, might be good to specify intention in the comment.

saurabh500: Ignore this. I looked at the comment for the declaration. The comments specify the intention

benchmarks/perf-benchmarking.py
```diff
@@ -49,6 +49,10 @@ def _init_conn_strings():
         CONN_STR_PYODBC = f"Driver={{ODBC Driver 18 for SQL Server}};{CONN_STR}"
     else:
         CONN_STR_PYODBC = CONN_STR
+        # mssql-python manages its own driver — strip Driver= from its connection string
+        import re
+        CONN_STR = re.sub(r"Driver=[^;]*;?", "", CONN_STR, flags=re.IGNORECASE).strip(";")
```
saurabh500: Out of curiosity, why do we try to strip the Driver= keyword from the ODBC connection string? 
Can we not take the mssq-python connection string by default and Prefix `Driver= ` in the connection string? That would be simpler and we won;t have to deal with Regex.

mssql_python/cursor.py
```diff
@@ -1450,35 +1473,36 @@ def execute(  # pylint: disable=too-many-locals,too-many-branches,too-many-state
...
-        #       (Ex: huge number of parallel queries with huge query string sizes)
-        if operation != self.last_executed_stmt:
-            # Executing a new statement. Reset is_stmt_prepared to false
+        # Prepare caching: skip SQLPrepare when re-executing the same SQL
+        # with parameters. The HSTMT is reused via _soft_reset_cursor, so the
+        # server-side plan from the previous SQLPrepare is still valid.
+        same_sql = parameters and operation == self.last_executed_stmt and self.is_stmt_prepared[0]
+        if not same_sql:
             self.is_stmt_prepared = [False]
+        effective_use_prepare = use_prepare and not same_sql
```
saurabh500, line 1482: Curiosity question: What happens if we dispose the cursor, which would presumably free the underlying statement handle? In that case, if the same SQL statement is used across two cursors, wouldn’t the code assume it can re‑prepare the statement, even though the original statement may already have been freed? Am I missing any safeguards here?

benchmarks/perf-benchmarking.py
bewithgaurav: DB_CONNECTION_STRING env var is shared across CI for both pyodbc and mssql-python.
pyodbc needs Driver in the string & mssql-python rejects Driver= because it manages its own bundled driver.
The script adds Driver= for pyodbc when absent (the else branch you see on line 49), so the natural flow is: env var without Driver= → add for pyodbc, pass as-is for mssql-python.
This strip was for fallback - only for the case when someone sets the env var with Driver= locally or something.
However your suggestion is correct as this is supposed to be a breaking situation - we can avoid usage of re here, not at all needed. I'll probably make this path work without regex

mssql_python/cursor.py
bewithgaurav: `_soft_reset_cursor` benefits all re-executions regardless of `use_prepare`. 
Actually the savings come from avoiding the `SQLFreeHandle` + `SQLAllocHandle` cycle (which took about ~1.5ms per call), not from prepare caching. 
The soft reset is to `SQL_CLOSE` + `SQL_RESET_PARAMS`, since that is always cheaper than destroying and recreating the `HSTMT`. 
The `use_prepare` flag only controls whether we call `SQLPrepare`.

mssql_python/cursor.py
bewithgaurav: Each Cursor instance has its own `hstmt`, `last_executed_stmt`, and `is_stmt_prepared`. 
These are instance fields, not class-level state. 
e.g. Disposing cursor A calls `_reset_cursor()` which sets `hstmt = None` and `is_stmt_prepared = False`. 
Cursor B's state is completely independent. The `same_sql` check requires `self.hstmt` to exist AND `self.is_stmt_prepared[0]` to be `True` - both are per-cursor. 
No cross-contamination possible.

mssql_python/cursor.py
bewithgaurav: profiling in depth using our profiler showed `DDBCSQLGetAllDiagRecords` costs ~10ms per call even on `SQL_SUCCESS` (the ODBC driver still has to scan internal state to confirm there are no records). 
On a 100-call insertmany loop, that's ~1s of extra time. 
ideally we should only call diagnostic collection on `SQL_SUCCESS_WITH_INFO` or `SQL_ERROR` . On plain `SQL_SUCCESS` there are never diagnostic records to retrieve, so the call is pure overhead.
for `SQL_ERROR`, check_error()` already extracts diagnostics and raises exception hence kept `SQL_SUCCESS_WITH_INFO` in the if condition here

mssql_python/cursor.py
saurabh500: Thanks !

mssql_python/cursor.py
saurabh500: Nice ! Can you please enhance the comment with this intention

mssql_python/cursor.py
saurabh500: Sounds good.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1378,6 +1378,32 @@ void SqlHandle::close_cursor() {
...
+    if (!SQLFreeStmt_ptr) {
+        DriverLoader::getInstance().loadDriver();
+    }
+    SQLHANDLE hStmt = statementHandle->get();
+
+    SQLRETURN rc;
+    {
+        py::gil_scoped_release release;
+        rc = SQLFreeStmt_ptr(hStmt, SQL_CLOSE);
+        if (SQL_SUCCEEDED(rc)) {
```
gargsaumya, line 1397: Some ODBC drivers return `SQL_NO_DATA` from `SQLFreeStmt(SQL_CLOSE`) when no cursor is open. `SQL_SUCCEEDED` is false for `SQL_NO_DATA`, so parameter unbinding is silently skipped, leaving stale bindings on the handle. We may want to change the condition to `if (SQL_SUCCEEDED(rc) || rc == SQL_NO_DATA)`, just verify.

saurabh500: Very cool. Didn't think of this

bewithgaurav: Tested this locally for our driver(odbc 18) - `SQLFreeStmt(SQL_CLOSE)` returns `SQL_SUCCESS (0)` in all three scenarios:

- After INSERT (no cursor open) → `SQL_SUCCESS`
- After SELECT (cursor open) → `SQL_SUCCESS`
- Double reset (nothing to close, nothing to unbind) → `SQL_SUCCESS`

`SQL_NO_DATA` never occurs. The ODBC spec for `SQLFreeStmt` doesn't list `SQL_NO_DATA` as a valid return - `SQL_CLOSE` is a no-op when no cursor is open and returns `SQL_SUCCESS` either way. Since we bundle no other drivers, this edge case doesn't apply to us.

But out of curiosity, just wanted to understand more about other ODBC "drivers" - are these different versions of ODBC?

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: this is done

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: this is done

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: No, these are separate implementations, not versions. The `SQL_NO_DATA` guard is a pattern from pyodbc-style defensive coding for unknown drivers, hence the suggestion. Since we exclusively bundle ODBC Driver 18 and you've tested this thoroughly, we can close it.

gargsaumya (approved):

saurabh500 (approved):

PR #549, opened 2026-04-29
PERF: Native C++ parameter detection and execute pipeline

bewithgaurav:

> [AB#44979](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44979)
> AB#49196

> GitHub Issue: #500 

### Summary   

Moves parameter type detection and binding from Python into a native C++ pipeline using raw CPython API calls. The new `DDBCSQLExecute` handles type detection → parameter binding → SQLExecute in a single FFI crossing, eliminating per-parameter Python overhead entirely.

**What changed:**

* `DetectParamTypes` — C++ type detection using raw CPython API (`PyLong_Check`, `PyDateTime_Check`, `PyObject_RichCompareBool`, etc.) replacing the Python-side `_create_parameter_types_list` loop for the standard execute path.
* `DDBCSQLExecute` (formerly `DDBCSQLExecuteFast`) — single C++ pipeline: detect → bind → execute. ParamInfo never crosses the pybind11 boundary.  
* `DDBCSQLExecuteLegacy` (formerly `DDBCSQLExecute`) — retained for `setinputsizes` users only, and annotated in-code as slated for removal once those overrides are handled natively.
* MONEY/SMALLMONEY bounds cached once at init using exact Decimal comparison (not lossy double).
* `PyTypeCache` (`py_type_cache.hpp`) stores all type objects as raw `PyObject*` (not `py::object`), eliminating pybind11 wrapper overhead on every cache hit.
* NUMERIC mantissa built from four fixed-width `uint32` limbs instead of walking every Decimal digit through `PyNumber_Multiply`/`PyNumber_Add`. SQL Server caps NUMERIC at 38 digits, so the mantissa always fits 128 bits. Decimal detection measured 2.3–2.9x faster on its own.
* Reference ownership is RAII throughout. `ParamInfo::dataPtr` holds a `py::object` rather than a raw `PyObject*` with a hand-written rule of five, and `py_ref.hpp` provides `steal`/`borrow` so every adoption of a CPython reference states whether it is taking a new reference or borrowing one.
* Parameter detection lives in `param_detect.hpp` rather than inline in the 6600-line `ddbc_bindings.cpp`. Header-only because the build is `-O3` with no LTO, so a `.cpp` boundary would also be an inlining boundary.

**Routing (cursor.py):**

```python
# Standard path (99% of calls): no setinputsizes -> all in C++
# Legacy path: setinputsizes active -> Python type detection + DDBCSQLExecuteLegacy
```

### Performance Results 🚀

The Python-side type detection cost was **~2.0–2.3µs per parameter** — an `isinstance` check, `ParamInfo` object construction, and a pybind11 FFI boundary crossing _per parameter, per execute call_. The C++ path replaces this with **~35ns/param** (raw `PyLong_Check` + struct field write) — a **~60x faster per-parameter detection**.

#### macOS arm64 (Apple Silicon M-series), Python 3.13

| Scenario | After (C++) | Before | **Improvement** | Python overhead eliminated |
|----------|------------|--------|-----------------|---------------------------|
| 3 params | 435µs | 444µs | 2% faster | 10µs saved |
| 10 params | 321µs | 344µs | 7% faster | 22µs saved |
| 50 params | 394µs | 502µs | **21% faster** | 107µs saved |
| 100 params | 558µs | 770µs | **28% faster** | 212µs saved |
| 200 params | 644µs | 1.08ms | **40% faster** | 432µs saved |
| 500 params | 708µs | 1.82ms | **61% faster** | 1.12ms saved |
| 1024 params | 1.07ms | 3.32ms | **68% faster** | 2.25ms saved |

#### Linux aarch64 (Docker container), Python 3.13

| Scenario | After (C++) | Before | **Improvement** | Python overhead eliminated |
|----------|------------|--------|-----------------|---------------------------|
| 3 params | 158µs | 164µs | 4% faster | 6µs saved |
| 10 params | 159µs | 178µs | 11% faster | 19µs saved |
| 50 params | 170µs | 271µs | **37% faster** | 101µs saved |
| 100 params | 209µs | 411µs | **49% faster** | 202µs saved |
| 200 params | 280µs | 651µs | **57% faster** | 371µs saved |
| 500 params | 396µs | 1.35ms | **71% faster** | 956µs saved |
| 1024 params | 746µs | 2.75ms | **73% faster** | 2.0ms saved |

#### vs pyodbc (post-PR, macOS)

| Params | mssql-python | pyodbc | Gap |
|--------|-------------|--------|-----|
| 3 | 399µs | 388µs | **3% gap** (near parity) |
| 50 | 458µs | 452µs | **1% gap** (parity) |
| 200 | 604µs | 550µs | **10% gap** (down from ~14x pre-PR) |
| 1024 | 1.31ms | 979µs | 33% gap (binding overhead, addressable separately) |

#### Customer scenarios (end to end, macOS arm64)

The numbers above isolate driver overhead. These are whole insert workloads, so they also carry the network round trip and SQL Server actually writing the rows, which this PR does not change and which dilutes the percentage. Measured against the merge base (`d94debd`) with the two builds interleaved across 3 rounds, 7 iterations each, first 2 discarded.

| Scenario | Before | After | Speedup |
|----------|--------|-------|---------|
| Orders insert (int, varchar, decimal, datetime2) | 880.6ms | 561.0ms | **1.57x** |
| Event log insert (uuid, datetime2, varchar, int) | 827.2ms | 533.5ms | **1.55x** |
| Document insert (nvarchar(max) ~10KB, DAE) | 1622.0ms | 1021.3ms | **1.59x** |
| Wide row insert (50 mixed columns) | 1598.6ms | 1048.9ms | **1.52x** |
| Single-row execute x5000 (4 params) | 2103.5ms | 2095.9ms | 1.00x |

Between them these cover every parameter type whose detection moved: int, varchar, decimal, datetime2, uuid, and the nvarchar(max) DAE streaming path.

Single-row execute does not move, and that is the expected result rather than a disappointment. At roughly 420µs per call the cost is the network round trip; detection for 4 parameters was only ever about 9µs of it. The gain scales with parameters per execute, so batched and wide-row work benefits and one-row-at-a-time work stays where it was.

#### Bottom line

| Metric | Value |
|--------|-------|
| **Avg improvement (50+ params)** | **~50% faster execute()** |
| **Worst-case improvement (1024 params)** | **73% faster** (2ms saved per call) |
| **Per-param overhead reduction** | **~60x** (2.3µs → 35ns) |
| **pyodbc gap closed** | From **14x slower** (GH-500) to **<10% gap** at 200 params |
| **Real-world impact** | Bulk inserts see **1.5–1.6x throughput** end to end (customer scenarios above) |

### Checklist

- [x] Tested locally (macOS arm64 + Linux aarch64)
- [x] Verified perf gain with micro-benchmarks on both platforms, and with end-to-end customer scenarios against the merge base
- [x] CI passing (CodeQL, DevSkim, Black, C++ lint)
- [x] No breaking changes to public API
- [x] setinputsizes users unaffected (routed to legacy path)


mssql_python/pybind/param_detect.hpp
```diff
@@ -0,0 +1,598 @@
...
+
+            py::object digits_obj = steal(PyObject_GetAttrString(as_tuple_ptr.ptr(), "digits"));
+            if (!digits_obj) throw py::error_already_set();
+
+            if (!PyTuple_Check(digits_obj.ptr())) {
+                throw py::type_error("Decimal.as_tuple().digits must be a tuple");
+            }
+
+            Py_ssize_t num_digits = PyTuple_GET_SIZE(digits_obj.ptr());
+            int exponent = static_cast<int>(PyLong_AsLong(exponent_obj.ptr()));
```
jahnvi480: **Blocking:** `static_cast<int>(PyLong_AsLong(...))` narrows the exponent to 32 bits *before* we validate it. On LP64 platforms (Linux/macOS, where `long` is 64-bit), `Decimal("1E+4294967297")` has exponent `4294967297`, which truncates to `1`, sails past the `precision <= 38` gate, and silently binds `10` — while the legacy Python path (arbitrary-precision int) computes precision > 38 and raises. (On 64-bit Windows `long` is 32-bit, so `PyLong_AsLong` overflows and line 416 catches it instead — but the value still can't bind, and it's a platform-dependent divergence either way.) `INT_MIN`-style truncations also invite UB on the later negation, and huge positive results blow up the packing loop.

Please reject `num_digits > 38` and exponents outside `[-38, 38]` (treating a `PyLong_AsLong` overflow — returns `-1` with `PyErr_Occurred()` — as precision overflow) before narrowing/arithmetic. Worth tests around `INT_MIN`, `INT_MAX`, and `2**32`.

mssql_python/pybind/param_detect.hpp
```diff
@@ -0,0 +1,598 @@
...
+            } else {
+                info.columnSize = is_unicode ? utf16_len : length;
+                info.paramSQLType = is_unicode ? SQL_WVARCHAR : SQL_VARCHAR;
+                info.paramCType = is_unicode ? SQL_C_WCHAR : PARAM_C_TYPE_TEXT;
+            }
+            info.decimalDigits = 0;
+
+            // Check geometry prefixes (only for non-DAE strings; long geometry
+            // values stay on the DAE path with their already-set types).
+            if (!info.isDAE && length >= 5 && kind == PyUnicode_1BYTE_KIND) {
```
jahnvi480: The geometry override is gated behind `!info.isDAE`, so a WKT string over 4000 UTF-16 units never gets the geometry treatment — it stays on the generic long-string DAE path (`SQL_VARCHAR`, columnSize 0, DAE=true). The legacy `_map_sql_type` checks the geometry prefix *first* and returns `SQL_WVARCHAR` / `SQL_C_WCHAR` / `len(param)` / non-DAE regardless of length. Large polygons are realistic, so this changes both the SQL type and the execution path.

Please move geometry detection ahead of the length/DAE decision (and across all Unicode storage kinds, not just 1-byte) so both paths agree, then add a >4000-char POLYGON parity test.

mssql_python/pybind/param_detect.hpp
```diff
@@ -0,0 +1,598 @@
...
+                PARAM_C_TYPE_TEXT;  // matches slow path (its SQL_C_CHAR is -8 = SQL_C_WCHAR)
+            info.columnSize = 16;
+            info.decimalDigits = 6;
+            int hour = PyDateTime_TIME_GET_HOUR(obj);
+            int minute = PyDateTime_TIME_GET_MINUTE(obj);
+            int second = PyDateTime_TIME_GET_SECOND(obj);
+            int microsecond = PyDateTime_TIME_GET_MICROSECOND(obj);
+            // Always include microseconds (matches Python's isoformat(timespec="microseconds")).
+            char buf[32];
+            snprintf(buf, sizeof(buf), "%02d:%02d:%02d.%06d", hour, minute, second, microsecond);
```
jahnvi480: This hand-formats `HH:MM:SS.ffffff` from the raw time fields, which drops any `tzinfo` and bypasses subclass `isoformat` overrides. The legacy path calls `value.isoformat(timespec="microseconds")` (see `_normalize_time_param` in cursor.py), so an aware `datetime.time` produces e.g. `01:02:03.000004+05:30` there but `01:02:03.000004` here.

Simplest fix: call `isoformat("microseconds")` from C++ and bind the returned string/length instead of building it manually.

mssql_python/pybind/param_detect.hpp
```diff
@@ -0,0 +1,598 @@
...
+
+// Platform-specific text C type: unixODBC requires all text as wide chars on
+// Linux/macOS. On Windows the ODBC driver accepts SQL_C_CHAR for ASCII text.
+// This matches the Python slow path's behavior (its SQL_C_CHAR constant is
+// numerically -8, which is ODBC's SQL_C_WCHAR — a long-standing alias used
+// throughout the Python layer).
+#if defined(__APPLE__) || defined(__linux__)
+inline constexpr SQLSMALLINT PARAM_C_TYPE_TEXT = SQL_C_WCHAR;
+#else
+inline constexpr SQLSMALLINT PARAM_C_TYPE_TEXT = SQL_C_CHAR;
```
jahnvi480: On Windows `PARAM_C_TYPE_TEXT` resolves to the real ODBC `SQL_C_CHAR (1)`, but Python's `SQL_C_CHAR` constant is `-8` (identical to `SQL_C_WCHAR`). So the legacy path always binds text as *wide*, while the native path binds ASCII strings / money-text / time as *narrow* on Windows — a genuine C-type and encoding-path divergence. CI is green, but for exact parity I'd bind `SQL_C_WCHAR` on every platform.

If narrow Windows binding is intentional, let's state it as the parity contract and test it explicitly (incl. non-ASCII collation).

tests/test_023_execute_path_parity.py
```diff
@@ -0,0 +1,304 @@
...
+    """Standard path: no setinputsizes, so C++ detects the types."""
+    cursor.execute("SELECT ?", [value])
+    return cursor.fetchone()[0]
+
+
+def _slow_path_roundtrip(cursor, value, sql_type, column_size):
+    """Force the slow path by setting an explicit inputsizes entry. The fast
+    path is gated on `not (self._inputsizes and any(s is not None ...))`, so a
+    non-None tuple here flips us to the legacy Python type-detection path."""
+    cursor.setinputsizes([(sql_type, column_size, 0)])
```
jahnvi480, line 48: This "slow path" helper calls `setinputsizes(...)`, which forces the supplied SQL type and *bypasses `_map_sql_type` entirely*. So the suite is really comparing native detection against hand-specified types — it can't catch the geometry / aware-time / Windows C-type / column-size / DAE divergences that actually matter (and the file comment claiming it exercises `_map_sql_type` is inaccurate).

Suggest exposing a test-only native detector and asserting all five fields (SQL type, C type, column size, decimal digits, DAE) directly against `_map_sql_type`, using fresh parameter lists and the threshold / aware-time / Decimal-exponent edge cases.

mssql_python/pybind/param_detect.hpp
```diff
@@ -0,0 +1,598 @@
...
+            py::object as_tuple_ptr = steal(PyObject_CallMethod(obj, "as_tuple", NULL));
+            if (!as_tuple_ptr) throw py::error_already_set();
+
+            py::object exponent_obj = steal(PyObject_GetAttrString(as_tuple_ptr.ptr(), "exponent"));
+            if (!exponent_obj) throw py::error_already_set();
+
+            // NaN / Infinity / sNaN: refuse rather than silently writing 0.
+            if (PyUnicode_Check(exponent_obj.ptr())) {
+                throw py::value_error(
+                    "Cannot bind non-finite Decimal (NaN/Infinity) as SQL NUMERIC");
```
jahnvi480, line 449: Rejecting non-finite Decimals as `ValueError` here is reasonable, but the legacy path raises *different* types — NaN raises `decimal.InvalidOperation` (from the money `<=` comparison) and Infinity reaches `_get_numeric_data` and raises `TypeError`. If we want exception parity, add the same explicit non-finite check on the Python side so both raise `ValueError`, and assert the exact type.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2143,14 +1988,188 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle, const std::u16stri
...
+            if (!SQL_SUCCEEDED(rc)) return rc;
+            statementHandle->clearDescribeCache();
+            is_stmt_prepared[0] = py::bool_(true);
+        } else {
+            ThrowStdException("Cannot execute unprepared statement");
+        }
+    }
+
+    std::vector<std::shared_ptr<void>> paramBuffers;
+    rc = BindParameters(*statementHandle, hStmt, params, paramInfos, paramBuffers, charEncoding);
```
jahnvi480, line 2108: If `BindParameters` binds some parameters and then throws, we unwind without `SQL_RESET_PARAMS`, so the HSTMT keeps pointers into `paramBuffers` that are about to be destroyed, until the next reset. Exposure is limited (the next execute resets first), but an exception-only scope guard that resets params once binding has begun would make the invariant safe. Keep the existing ODBC-return error path untouched so diagnostics stay readable.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2143,14 +1988,188 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle, const std::u16stri
...
+
+            const ParamInfo* matchedInfo = nullptr;
+            for (auto& info : paramInfos) {
+                if (reinterpret_cast<SQLPOINTER>(const_cast<ParamInfo*>(&info)) == paramToken) {
+                    matchedInfo = &info;
+                    break;
+                }
+            }
+            if (!matchedInfo) {
+                ThrowStdException("SQLExecuteFast: unrecognized paramToken from SQLParamData");
```
jahnvi480: Leftover old name — `SQLExecuteFast:` here (and again at ~2134 and ~2158, plus a comment near line 1999). The function was renamed away from "fast"; these strings should follow so logs/messages aren't misleading.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2143,14 +1988,188 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle, const std::u16stri
...
+                }
+            } else if (PyBytes_Check(pyObj) || PyByteArray_Check(pyObj)) {
+                // Handle bytes and bytearray separately — pybind11's bytes
+                // caster does not safely convert bytearray.
+                const char* dataPtr = nullptr;
+                size_t totalBytes = 0;
+                std::string bytesStorage;  // lifetime must span the loop
+
+                if (PyBytes_Check(pyObj)) {
+                    bytesStorage = borrow<py::bytes>(pyObj);
```
jahnvi480: For immutable `bytes` this copies the whole payload into `bytesStorage` before streaming. `matchedInfo->dataPtr` holds a strong reference for the duration, so you can stream straight from `PyBytes_AS_STRING(pyObj)` / `PyBytes_GET_SIZE(pyObj)` and skip the copy. (The `bytearray` branch still needs its copy since it's mutable.)

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2143,14 +1988,188 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle, const std::u16stri
...
+        };
+        while (true) {
+            {
+                py::gil_scoped_release release;
+                rc = SQLParamData_ptr(hStmt, &paramToken);
+            }
+            if (rc != SQL_NEED_DATA) break;
+
+            const ParamInfo* matchedInfo = nullptr;
+            for (auto& info : paramInfos) {
```
jahnvi480: This linearly scans every param for each DAE token. Since the token *is* a stable `ParamInfo*` (never reallocated — the vector is sized up front), you can `reinterpret_cast` it back directly (or pass an index) instead of the O(N) search per chunk.

jahnvi480 (changes requested):

mssql_python/pybind/param_detect.hpp
bewithgaurav: thanks - fixed. the exponent is read at full width and range-checked before any narrowing, so an out-of-range or overflowing exponent raises `ValueError` on every platform instead of truncating and binding. tests around `2**32`, the `INT_MIN`/`INT_MAX` neighbours, and the in-range edges.

mssql_python/pybind/param_detect.hpp
bewithgaurav: fixed, with one deliberate difference from the suggestion

geometry runs before the length/DAE decision now and matches across all string kinds. 
I did not reuse the legacy tuple though: legacy is itself unbindable over 4000 chars (`Invalid precision value`), so geometry folds into the wide-type path and streams via DAE when large, which actually binds. added a >4000 polygon round-trip plus small and unicode-tagged checks.

mssql_python/pybind/param_detect.hpp
bewithgaurav: fixed. native calls `isoformat("microseconds")` now instead of hand-formatting, so aware times and subclass overrides match legacy. an aware time still fails to bind on both paths since `TIME` has no offset.

mssql_python/pybind/param_detect.hpp
bewithgaurav: fixed - text binds as `SQL_C_WCHAR` on every platform now, so windows no longer diverges
added round-trips including non-ascii and the 4000/4001 boundary

tests/test_023_execute_path_parity.py
bewithgaurav: this is now fixed, you are right that `setinputsizes` skipped `_map_sql_type` entirely

rather than a test-only detector, I've asserted `_map_sql_type` directly for all five fields, and keep native covered through `execute()` with `sql_variant` BaseType where the bound type is observable

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: no live exposure here. after a throw the HSTMT holds stale pointers into `paramBuffers`, but the next `execute()` rebinds every param before `SQLExecute`, so nothing dereferences them. no path today touches a bound-but-not-rebound statement, leaving as-is unless we add such a path - let me know if this sounds okay

mssql_python/pybind/param_detect.hpp
bewithgaurav: fixed. both `_map_sql_type` and `_get_numeric_data` raise `ValueError` for non-finite now, so NaN and Infinity stop raising different types across the two paths - added tests

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: good catch - fixed, the stale `SQLExecuteFast:` strings and the nearby comment are gone

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: fixed - immutable `bytes` streams straight from its buffer now with no copy, since the strong ref keeps it alive across the loop
`bytearray` keeps its copy because it is mutable across the GIL release

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: fixed - the token casts back to its `ParamInfo` directly now instead of scanning per chunk, with a range and alignment check so a bad token still throws

sumitmsft:
Nice work! I tried to trace the refcounting (steal/borrow RAII), GIL release around ODBC calls, DAE streaming, and legacy parity; it all holds up.

One optional, non-blocking follow-up:
"2**63" classifies as BIGINT and fails later at "param.cast<int64_t>()" with a generic message. This matches legacy "_map_sql_type" exactly, so it's faithful parity, not a regression. For a clearer "exceeds BIGINT range" error, raise at detect time on both paths. (Minor: the `int64_t` range guard in the bind case is dead code.) // Ref: param_detect.hpp

sumitmsft (dismissed):

jahnvi480 (dismissed):

gargsaumya (dismissed):
LGTM. Earlier reviews covered the substantive items, approved.

jahnvi480 (approved):

sumitmsft (approved):

PR #552, opened 2026-04-30
FEAT: Profiler

bewithgaurav:

> ADO Work Item: Fixed [AB#44819](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44819)

### Summary

Adds a performance profiler for diagnosing where time goes inside a database call, split across the two layers the driver is built from: the Python layer (`cursor.py`) and the native C++ layer (`ddbc_bindings`). A normal Python profiler sees the entire native layer as one opaque block; this instruments both layers with named timers so a slow call can be attributed to, for example, native parameter binding rather than Python.

This is an internal development tool. The C++ instrumentation is **compiled out of released wheels** (see below), so the native driver carries no profiler code. The Python-side phase markers are lightweight and remain in the shipped `cursor.py`; when profiling is disabled they are a no-op branch whose measured end-to-end cost is within run-to-run noise versus a no-profiler build.

**How it is built in**

- The C++ instrumentation is gated at compile time by an `ENABLE_PROFILING` CMake option, **off by default**. Normal builds expand every `PERF_TIMER` to a no-op and do not register the native `ddbc_bindings.profiling` submodule, so there is no native profiler code, no unwind tables, and no runtime branch in the released binary. Internal/dev builds opt in with `ENABLE_PROFILING=1 bash build.sh` (or `set ENABLE_PROFILING=1` + `build.bat` on Windows).
- The Python instrumentation (`perf_timer.py` and the `perf_phase(...)` markers in `cursor.py`) cannot be compiled out the same way, so it ships. When profiling is disabled, `perf_phase` returns a shared no-op context manager (measured at ~110 ns per marker in isolation). Some markers do sit on per-row paths (e.g. three in the `fetchone` loop), but that per-marker cost is orders of magnitude below per-row DB latency, so measured end-to-end overhead on execute/fetch workloads stays within run-to-run noise versus a no-profiler build.

**Components**

- C++ layer: `performance_counter.hpp` provides a `PERF_TIMER("name")` RAII macro and instruments `ddbc_bindings.cpp` and the connection sources, exposed to Python as the `ddbc_bindings.profiling` submodule.
- Python layer: `perf_timer.py` provides `perf_phase("name")` context managers around the execute/fetch phases in `cursor.py`. The `ddbc::` and `py::` prefixes identify which layer a timer belongs to.
- Runner: the `profiler/` CLI (`python -m profiler`) with built-in scenarios plus aggregate and timeline reports, and a `--script` flag for profiling an arbitrary workload. `profiler/` is dev-only and excluded from the shipped wheel.

**Controlled use and reentrancy**

Use controlled diagnostic workloads with one controller owning the process-wide measurement windows: enable, run the workload, wait for worker threads, then collect. Recording supports worker threads; independent concurrent profiling sessions and always-on production profiling are not supported.

Reports use detached snapshots. Native counters release their mutex before allocating Python report objects. Python bookkeeping uses immutable records and omits only same-thread recursively triggered samples; application cleanup, ordinary nested timers and other threads still run.

Enabled instrumentation adds bookkeeping overhead. Compare runs using the same profiling configuration; the disabled Python fast path and native compile-out remain unchanged.

**Tests**

Existing runtime-instrumentation tests are retained. Native cases require a profiling-enabled build; tests importing the dev-only `profiler` package skip when it is absent from an installed wheel. Broader profiler testing and profiling-enabled CI remain follow-up work.

### Follow-up work (intentionally out of scope for this PR)

This PR ships the internal profiler tooling only. The following are deliberately deferred to follow-up PRs and are **not** oversights:

- **CI does not build the profiling configuration.** No pipeline currently sets `ENABLE_PROFILING`, so the profiling build and its native-layer tests are not exercised in CI. A follow-up will add an `ENABLE_PROFILING` build/test leg (Linux + Windows). This is tied to the planned work to drive CI benchmarks from profiler data, where these builds and the scenario set become the fixed workloads a regression gate runs.
- **End-user profiling experience.** A minimal enable/dump API and a separate profiling wheel (so a customer can install one artifact, reproduce, and send a dump) are left for a follow-up designed with the team.
- **Cross-layer timeline ordering is approximate.** Python and native epochs are initialized separately, so cross-layer ordering and nesting remain approximate. This clock-origin mismatch does not affect aggregate durations. A shared origin or measured cross-layer offset remains follow-up work.


profiler/reporter.py
```diff
@@ -0,0 +1,89 @@
...
+    for name, s in (cpp_stats or {}).items():
+        merged[name] = s
+    for name, s in (py_stats or {}).items():
+        merged[name] = s
+    return merged
+
+
+def merge_timeline(cpp_timeline: list | None, py_timeline: list | None) -> list[dict]:
+    events = list(cpp_timeline or []) + list(py_timeline or [])
+    events.sort(key=lambda e: e["start_us"])
```
gargsaumya, line 19: Suggestion: Should we normalize the two timelines before sorting? The native and Python events use independently initialized epochs: _cpp.enable_timeline() runs before _py.enable_timeline(). Therefore, their start_us values are not relative to the same origin, but this sort treats them as directly comparable. The resulting offset can place a native child outside its Python wrapper and produce incorrect ordering and indentation. Should we establish a shared origin or calculate a cross-layer offset before merging the events?

mssql_python/pybind/performance_counter.hpp
```diff
@@ -0,0 +1,213 @@
...
+        timeline_.clear();
+        epoch_ = std::chrono::steady_clock::now();
+        timeline_enabled_ = true;
+    }
+    void disable_timeline() { timeline_enabled_ = false; }
+    bool is_timeline_enabled() const { return timeline_enabled_; }
+
+    void record(const std::string& name, int64_t duration_ns,
+                std::chrono::time_point<std::chrono::steady_clock> start) {
+        if (!enabled_) return;
```
gargsaumya, line 113: Suggestion: Can we keep samples tied to the window in which they started? ScopedTimer remembers only that profiling was enabled when it started, while this check observes the current global state. A timer can start in window A, remain active while another thread collects and resets A, and finish after window B is enabled. Its duration is then recorded in B; in timeline mode its start can also be negative relative to B’s new epoch. Can we capture a profiling-window generation in the timer and reject the sample when that generation is no longer current.

gargsaumya (commented):
The separation between opt-in native instrumentation and lightweight Python markers looks reasonable. I’ve added two comments about potential correctness issues. Let me know what you think.

profiler/reporter.py
bewithgaurav: agreed - the native and python layers stamp epochs from separate clock reads a few microseconds apart, so merging by raw start is approximate ordering across layers only

since this only affects the timeline view, the aggregate per-timer stats are exact. I'd rather take it as a follow-up with the other timeline and CI items since this would be a bigger scope 

added a note to the follow-up section in the PR description

mssql_python/pybind/performance_counter.hpp
bewithgaurav: fixed this one

profiler/core.py
```diff
@@ -0,0 +1,288 @@
...
+            "conn": self._conn,
+            "cursor": cursor,
+            "__name__": "__main__",
+            "__file__": str(path),
+        }
+
+        try:
+            # compile() is inside the guard so a SyntaxError in the user script
+            # still closes the cursor via the finally below.
+            code = compile(path.read_text(), str(path), "exec")
```
jahnvi480: `--script` retains the profiler’s `sys.path` and `sys.argv`. Adjacent-module imports fail, and scripts using `argparse` reject the profiler’s own arguments. Temporarily establish—and restore—the target script’s import path and argument vector.

profiler/core.py
```diff
@@ -0,0 +1,288 @@
...
+        print(f"\n{'#' * 100}")
+        print(f"# CUSTOM: {path.name}")
+        print(f"{'#' * 100}")
+
+        ns = {
+            "conn": self._conn,
+            "cursor": cursor,
+            "__name__": "__main__",
+            "__file__": str(path),
+        }
```
jahnvi480: `compile()` inherits `core.py`’s future-annotations flag, changing scripts that never requested it. A valid dataclass using `ClassVar` consequently fails with `TypeError`. Pass `dont_inherit=True`.

profiler/scenarios.py
```diff
@@ -0,0 +1,373 @@
...
+        t0 = time.perf_counter()
+        for _ in range(num_batches):
+            cursor.execute(sql, params)
+        conn.commit()
+        wall_ms = (time.perf_counter() - t0) * 1000
+        cpp, py = ctx.collect()
+        actual = num_batches * rows_per_batch
+        rps = actual / (wall_ms / 1000) if wall_ms > 0 else 0
+    finally:
+        cursor.close()
```
sumitmsft: **Medium - Roll back failed insertmanyvalues batches before reusing the connection**

Please roll back the transaction when this workload fails, rather than only closing the cursor. If the second INSERT fails, the first batch remains pending on the connection retained by `Profiler`. Catching that error and running a table-backed scenario on the same profiler reaches `setup_test_data()` -> `conn.commit()`, committing the failed workload's earlier rows. I reproduced 1,000 pending rows after the failure and confirmed they survived rollback after the next scenario. Add failure-path rollback with guaranteed cursor closure and a regression test for this reuse path.

The affected rows are in the temporary benchmark table; this is a programmatic-reuse and transaction-cleanup issue, not production-table corruption.

mssql_python/perf_timer.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __init__(self, name: str):
+        self._name = name
+
+    def __enter__(self):
+        self._t0 = time.perf_counter_ns()
+        return None
+
+    def __exit__(self, *exc):
+        _record(self._name, time.perf_counter_ns() - self._t0, self._t0)
```
sumitmsft: **Low - Reject Python timing samples from closed measurement windows**

Please associate Python samples with their starting measurement window, matching the native `ScopedTimer`. A live query spanning `collect()`/reset/re-enable records its old `py::execute::cpp_call` duration in the new window, while the native sample is correctly dropped. This exit path also records after `disable()`. A generation check for both phase and manual timers would keep the two layers' aggregate snapshots consistent. This is a non-blocking robustness improvement for callers changing windows while work remains in flight.

sumitmsft (changes requested):
The opt-in native instrumentation and development-package separation look sound. Two findings remain after the independent review and reconciliation:

- **Medium:** Roll back failed `insertmanyvalues` batches before reusing the profiler connection. A later scenario can otherwise commit pending inserts from the failed workload. This concerns temporary benchmark data and supported programmatic reuse, not production-table corruption.
- **Low:** Reject Python samples from closed measurement windows. In-flight Python timers can enter the next window's aggregate statistics even when the native counterpart is correctly dropped. This is non-blocking for the intended synchronous, single-threaded runner.

**Recommendation:** Request Changes for the transaction-cleanup issue.
**Overall risk:** Medium. **Design compliance:** Not Applicable; no separately approved profiler design was accessible.

mssql_python/perf_timer.py
```diff
@@ -0,0 +1,201 @@
...
+                "name": ev["name"],
+                "start_us": ev["start_ns"] // 1000,
+                "duration_us": ev["duration_ns"] // 1000,
+            }
+            for ev in _timeline
+        ]
+
+
+def get_stats() -> dict:
+    with _lock:
```
sumitmsft: **[High] Avoid a same-thread deadlock during statistics collection**

`get_stats()` holds the new non-reentrant `threading.Lock` while allocating report dictionaries. Those allocations can trigger cyclic garbage collection and run a user object's `__del__`. If that finalizer executes SQL for cleanup, the cursor’s profiling phase calls `perf_stop()` → `_record()`, which tries to acquire the same lock already held by this thread. Neither operation can finish.

I reproduced this with a finalizer dropping a temporary table: the current implementation hangs, while the previous timer implementation completes the same workload. This requires enabled profiling but does **not** require multiple threads; normal profiling-disabled use is unaffected.

Please make snapshot construction and recording reentrancy-safe and add a bounded subprocess regression. Simply replacing `Lock` with `RLock` is insufficient without ensuring a stable snapshot, because reentrant recording can mutate the dictionary being iterated.

bewithgaurav: fixed in both layers
reports use detached snapshots, and recursive python bookkeeping no longer blocks application cleanup.
the bounded GC/finalizer cases now complete, including native snapshots on python 3.11. ordinary nested timers and worker-thread counts are preserved
recursive samples caused by python profiler bookkeeping are intentionally omitted and documented

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -3964,6 +3995,7 @@ SQLRETURN SQLFetchScroll_wrap(SqlHandlePtr StatementHandle, SQLSMALLINT FetchOri
 // TODO: Move to anonymous namespace, since it is not used outside this file
 SQLRETURN SQLBindColums(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& columnNames,
                         SQLUSMALLINT numCols, int fetchSize, int charCtype = SQL_C_WCHAR) {
+    PERF_TIMER("SQLBindColums");
     SQLRETURN ret = SQL_SUCCESS;
```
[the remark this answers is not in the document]
bewithgaurav: discarding this one

profiler/scenarios.py
```diff
@@ -0,0 +1,376 @@
...
+    cursor = conn.cursor()
+    try:
+        ctx.enable()
+        t0 = time.perf_counter()
+        cursor.execute(f"SELECT * FROM {table} WHERE id <= 100")
+        rows = cursor.fetchall()
+        wall_ms = (time.perf_counter() - t0) * 1000
+        cpp, py = ctx.collect()
+    finally:
+        cursor.close()
```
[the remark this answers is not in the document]
bewithgaurav: leaving this as-is. `Profiler.run()` owns the profiling lifecycle and already disables profiling in its `finally` when a scenario fails.

these functions are internal helpers, not standalone entry points. the documented CLI/runner path already handles cleanup, so I'm keeping that responsibility in the runner rather than duplicating it across helpers

mssql_python/perf_timer.py
sumitmsft: The runtime deadlock is fixed; the remaining request is the regression test mentioned in my original comment.

Please commit a bounded subprocess test for GC/finalizer reentry. `tests/test_025_profiler.py` is unchanged by this fix, and its existing tests also passed the previously deadlocking implementation.

The test should trigger a finalizer during statistics collection and assert that cleanup actually runs, collection completes within a timeout, and the returned snapshot remains valid. It should fail against the old implementation and pass against this fix, so a future regression cannot silently bring the hang back.

A focused Python-only case can cover this without SQL Server or a profiling-enabled native build; the finalizer can invoke the Python timing API directly to exercise the same lock reentry. This does not require the broader profiling-enabled CI work deferred by the PR.

bewithgaurav: added this
the focused python-only subprocess regression passes with the fix, times out on the old implementation, and rejects an `RLock`-only change through its snapshot assertion.
it checks that finalizer cleanup completes and enforces a 10-second timeout. no sql server or profiling-enabled native build is required

mssql_python/perf_timer.py
```diff
@@ -0,0 +1,286 @@
...
+    if getattr(_local, "depth", 0):
+        return
+    with _bookkeeping():
+        # Prebind the no-argument release: RLock.__exit__ and method lookup can
+        # allocate before unlocking, which is unsafe around GC finalizers.
+        release = _lock.release
+        while True:
+            _lock.acquire()
+            try:
+                if not _enabled or (start_ns and start_ns < _window_start_ns):
```
gargsaumya, line 248: Could we use monotonic generation counters here instead of treating `perf_counter_ns()` as a unique window ID? The clock is monotonic, but consecutive reads may return the same value at the platform clock resolution. If a phase starts at tick T and `reset()` or `enable_timeline()` also records T, the `<` and `>=` checks accept that old phase into the new aggregate or timeline window. I reproduced the aggregate case with controlled ticks: enable=100, phase start=200, reset=200, phase end=300; `get_stats()` then contained `old-window`. Separate aggregate-window and timeline generations, captured when the timer starts, would make boundary identity independent of clock resolution.

gargsaumya (commented):
The latest head addresses the earlier window-boundary, finalizer, cleanup, and test-packaging findings. I found one remaining clock-resolution edge case in the Python window identity logic.

gargsaumya (approved):
I’ve added a suggestion, but it is non-blocking, so you can consider addressing it as part of future scope.
Approving this PR.

gargsaumya (approved):
I’ve added a suggestion, but it is non-blocking, so you can consider addressing it as part of future scope.
Approving this PR.

sumitmsft (approved):

**Recommendation:** Approve

The previously reported runtime issues and the requested bounded GC/finalizer regression test are addressed. An independent rubber-duck pass found no additional actionable issues. No separate approved profiler design was available; the documented internal diagnostic scope was used for this review.

At submission, 27 checks have succeeded; the macOS SQL2022 job and parent validation status remain pending. This approval does not bypass CI or authorize merging.

jahnvi480 (approved):
Reviewed the PR! LGTM!

mssql_python/perf_timer.py
bewithgaurav: will incorporate this in the next CI PR, thanks for the suggestion

PR #556, opened 2026-05-05
CHORE: Exclude diagnostic LOG statements from coverage metrics

gargsaumya:

> [AB#44903](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44903)

### Summary   
  

This pull request introduces a new workflow for improving code coverage reporting by automatically joining multi-line `LOG()` macro calls into single lines before coverage analysis. This makes it easier to exclude logging statements from LCOV coverage results without modifying the original source code. The changes include a new helper script, updates to the build script to use this helper during coverage builds, and adjustments to the coverage merge process to exclude `LOG` statements.

**Coverage build improvements:**

* Added a new script `eng/scripts/join_logs_for_coverage.py` that joins multi-line `LOG()` macro calls into single lines in `.cpp` and `.hpp` files, simplifying LCOV filtering for logging statements. This script operates on temporary copies of the source code and does not modify the originals.
* Updated `mssql_python/pybind/build.sh` to invoke the new helper script during coverage builds. The script backs up original source files, processes them to join `LOG` statements, and restores the originals after the build, ensuring the source code remains unchanged.

**Coverage reporting adjustments:**

* Modified `generate_codecov.sh` to use LCOV's `--omit-lines` option with a regex that matches all `LOG` macro calls (e.g., `LOG`, `LOG_ERROR`, `LOG_WARNING`) during the coverage merge step, ensuring these statements are excluded from the final coverage report.


bewithgaurav (changes requested):
## A revamped take on the LOG-coverage gap

I called this exact gap out in the logging framework PR a few months ago: https://github.com/microsoft/mssql-python/pull/312#issuecomment-3525784436. at the time I rejected the LCOV markers approach for the same reasons that show up here: 150+ manual changes, clutters the codebase, hides executable code paths. coming back to it now I think there's a third option we didn't consider then that gets the same metric lift without any of those costs.

the 600 LCOV_EXCL_LINE markers and the python filter in `generate_codecov.sh` can be replaced with one tiny pre-build step plus one extra lcov flag. same coverage number, cleaner source tree, no helper scripts.

ran it end-to-end on macos arm64. full pytest suite including the logging integration tests:

```
PR's machinery:    3916/5079 = 77.10%
proposed flow:     3912/5052 = 77.43%
```

`1733/1733` tests pass. logging still works at runtime, LoggerBridge gets normal coverage, nothing about behavior changes.

## the proposal

1. add a small script (`scripts/join_logs_for_coverage.py`, ~50 lines) that joins multi-line LOG calls onto single lines. only used in the codecov build, never committed source changes.
2. in `mssql_python/pybind/build.sh` codecov mode, snapshot the source, run the join, build, restore on EXIT via `trap` (~12 lines).
3. in `generate_codecov.sh`, swap the embedded python filter for `--omit-lines '\bLOG[A-Z_]*\('` on the existing `lcov -a` line.

joining preserves runtime behavior because adjacent string literals concatenate at compile time. the .so built from joined source is bit-for-bit equivalent to one built from the original.

## what gets deleted

- all 600 `// LCOV_EXCL_LINE` markers from the cpp files
- `add_lcov_exclusions.py` and `fix_multiline_log_exclusions.py` at repo root
- the embedded ~90-line `PYTHON_FILTER` block in `generate_codecov.sh`

`tests/test_021_coverage_edge_cases.py` and `.gitignore` additions stay.

## small bonus

the helper scripts use `'LOG(' in line` which misses `LOG_ERROR(` and `LOG_WARNING(`. 11 sites currently unmarked (9 LOG_ERROR + 1 LOG_WARNING + their continuations). the proposed regex catches them. that's the +0.33pp.

generate_codecov.sh
```diff
@@ -74,36 +74,61 @@ fi
...
 
-# Merge LCOV reports (ignore inconsistencies in Python LCOV export)
-echo "[ACTION] Merging Python and C++ coverage"
-lcov -a python-coverage.info -a cpp-coverage.info -o total.info \
+# Merge LCOV reports and filter LOG statements using lcov's built-in exclusion
+# The --rc option sets lcov_excl_line to match any line containing LOG macros
+# Since we joined multi-line LOGs during build, they're now on single lines
+echo "[ACTION] Merging Python and C++ coverage with LOG exclusion"
+lcov -a python-coverage.info -a cpp-coverage.info -o total-unfiltered.info \
+  --rc lcov_excl_line='\bLOG[A-Z_]*\s*\(' \
```
bewithgaurav: `--rc lcov_excl_line=...` on `lcov -a` doesn't actually exclude anything unless `--filter region` is also passed. that flag controls the regex lcov uses to find exclusion-comment markers in source, but lcov only scans source when the region filter is on. so right now the regex is set but never applied.

measured against pr head:

```
as-is:                                  3949/5236 = 75.42%
with --filter region added:             3820/4947 = 77.22%
with --omit-lines '\bLOG[A-Z_]*\s*\(':  3820/4947 = 77.22%
```

fix either by adding `--filter region` to the `lcov -a` call, or swapping `--rc lcov_excl_line=...` for `--omit-lines '\bLOG[A-Z_]*\s*\('`. 
same output

eng/scripts/join_logs_for_coverage.py
```diff
@@ -0,0 +1,197 @@
...
+    else:
+        # Default to current directory
+        base_dir = Path.cwd()
+    
+    if not base_dir.exists():
+        print(f"[ERROR] Directory not found: {base_dir}", file=sys.stderr)
+        sys.exit(1)
+    
+    # Find all C++ source files
+    cpp_files = list(base_dir.rglob('*.cpp')) + list(base_dir.rglob('*.hpp'))
```
bewithgaurav: `.h` files aren't included. `ddbc_bindings.h:468` has a single-line `LOG_ERROR` so the regex catches it either way, no impact today. but a future multi-line LOG in any header would silently skip the join and its continuation lines would slip into the denominator. one-character fix: `'*.h*'` instead of `'*.hpp'` so both `.h` and `.hpp` get picked up.

eng/scripts/join_logs_for_coverage.py
```diff
@@ -0,0 +1,197 @@
...
+    
+    if not cpp_files:
+        print(f"[WARNING] No .cpp or .hpp files found in {base_dir}")
+        return
+    
+    print(f"[INFO] Processing {len(cpp_files)} C++ files in {base_dir}")
+    for filepath in cpp_files:
```
bewithgaurav, line 191: `rglob` walks `build/` too, so the script ends up processing `build/CMakeFiles/<ver>/CompilerIdCXX/CMakeCXXCompilerId.cpp` on every run - that isn't part of the codebase, it's a cmake-generated probe to detect the C++ compiler. 
can consider adding a cheap guard: 
`if 'build' in filepath.parts: continue` 
right after the loop opens, or filter it out when building `cpp_files`.

bewithgaurav (changes requested):
need one more refinement w.r.t code coverage and minor fixes on the join logs script, rest is good

bewithgaurav (dismissed):
lgtm, thanks!

mssql_python/pybind/build.sh
```diff
@@ -31,6 +31,45 @@ COVERAGE_MODE=false
...
+    echo "[ACTION] Preparing source for coverage build (joining LOG statements)"
+    
+    # Save current directory
+    ORIGINAL_DIR=$(pwd)
+    
+    # Create backup using tar to preserve directory structure
+    BACKUP_FILE="${ORIGINAL_DIR}/.source_backup_coverage.tar.gz"
+    echo "[INFO] Creating backup of source files"
+    tar -czf "$BACKUP_FILE" --exclude='build' --exclude='.source_backup*' \
+        $(find . -maxdepth 2 -type f \( -name "*.cpp" -o -name "*.hpp" \) -o -type d -name connection) 2>/dev/null || true
```
bewithgaurav: minor: this tar pattern is now out of sync with the join script's glob. the joiner picks up `*.cpp` + `*.h*` (so `ddbc_bindings.h` and `utf_utils.h` get processed), but this find still catches `*.cpp` + `*.hpp` only. root-level `.h` files end up modified by the joiner but not in the backup, so the restore in `generate_codecov.sh` can't bring them back.

zero impact today (both root `.h` files have only single-line LOGs, joiner doesn't touch them). but if someone writes a multi-line `LOG_ERROR(...)` in `ddbc_bindings.h` later, every successful codecov run will leave that file in the joined state and the contributor has to `git checkout` to recover.

fix is the same shape as the script glob change: swap `"*.hpp"` for `"*.h*"` here.

bewithgaurav (commented):
just saw a corner case and left a minor comment
can choose to fix here or later, not p0

subrata-ms (dismissed):

mssql_python/pybind/build.sh
gargsaumya: Fixed here

bewithgaurav (approved):

subrata-ms (approved):

PR #558, opened 2026-05-07
PERF: Optimize Row construction and repeated cursor bookkeeping

jahnvi480:

> [AB#44921](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44921)

### Summary
Reduce Python-side overhead in `fetchone()`, `fetchmany()`, and `fetchall()` while preserving decoding, converter, diagnostic, and Row-access behavior. The branch incorporates main through `e279a4f6` and includes the review fixes in `56c3d3b5`.

#### Fetch optimizations
- Construct eligible `fetchmany()`/`fetchall()` rows in native `construct_rows`, bypassing the Python per-row initialization loop. Use `Row._fast_create()` for the corresponding `fetchone()` path.
- Add `Row.__slots__` and retain zero-copy storage of fetched values when no conversion is required. Initialize the shared lowercase column map on both Python and native construction paths.
- Cache CHAR/WCHAR encoding strings and the CHAR C type. Refresh them only when connection decoding settings change, including for existing cursors.
- Reuse the SQL-to-C type mapping and use `str.isascii()` for Unicode detection.
- Keep the no-conversion fast path available when converters are registered but none applies to the current result set. An explicit empty mapping avoids unnecessary row copies or per-row fallback lookups.

#### Correctness and compatibility
- Track converter-generation changes so adding, replacing, removing, or clearing converters after `execute()` updates subsequent fetches without rebuilding mappings on unchanged cache hits.
- Preserve current-main dispatch precedence: raw ODBC SQL type, Python type, then the string/bytes-only WVARCHAR fallback. Unrelated integer and UUID columns are not passed to the string fallback.
- Preserve direct `Row` construction's connection-converter fallback when no precomputed mapping is supplied, including UUID stringification.
- Retain main's lowercase/string-key column access, profiling scopes, and CHAR decoding C-type forwarding. Native WCHAR decoding semantics are unchanged.
- Validate `row_class` with `PyType_Check` before casting in the Python-callable native helper. Invalid arguments raise `TypeError` rather than crashing the interpreter.
- Restore unconditional diagnostic retrieval after non-error fetch returns. The native wrappers can overwrite intermediate statuses, so checking only the final `SQL_SUCCESS_WITH_INFO` is unsafe. The earlier diagnostic-skip optimization has been removed; this restores prior behavior rather than claiming to repair pre-existing native diagnostic-record overwrites.
- Check errors in `fetchone()` and `fetchmany()` before updating row positions or constructing rows, consistent with `fetchall()`.

### Performance: current main versus this PR

| Workload (rows x columns) | API | Main median | PR median | Time reduction |
|---|---|---:|---:|---:|
| Narrow: 10,000 x 3 | `fetchmany(1000)` | 9.36 ms | 6.54 ms | **30.18%** |
| Narrow: 10,000 x 3 | `fetchall()` | 8.67 ms | 6.99 ms | **19.30%** |
| Wide: 10,000 x 24 | `fetchmany(1000)` | 27.57 ms | 24.72 ms | **10.34%** |
| Wide: 10,000 x 24 | `fetchall()` | 30.59 ms | 26.07 ms | **14.79%** |

Positive reduction means less time, calculated from unrounded medians. Narrow/wide batch cases improved in 10/10 and 9/10 pairs respectively. `fetchmany(1)` medians were only 1.15-1.80% lower, with narrow results inconclusive. Narrow `fetchone()` was 5.02% lower; wide `fetchone()`.

These are workload-specific results, not a universal or cross-platform speedup. Full results for all 11 fetch cases, control, dispersion, paired intervals, build identities, and raw repetitions are retained locally; older pre-merge numbers are not mixed in.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -160,9 +160,9 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         self._has_result_set = False  # Track if we have an active result set
         # Cache decoding encoding strings — these don't change between fetches,
         # so we avoid 2 method calls + 2 dict.get() per fetch call.
-        self._cached_char_encoding = self._get_decoding_settings(
-            ddbc_sql_const.SQL_CHAR.value
-        ).get("encoding", "utf-8")
+        self._cached_char_encoding = self._get_decoding_settings(ddbc_sql_const.SQL_CHAR.value).get(
+            "encoding", "utf-8"
+        )
         self._cached_wchar_encoding = self._get_decoding_settings(
             ddbc_sql_const.SQL_WCHAR.value
         ).get("encoding", "utf-16le")
@@ -2569,9 +2569,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
             uuid_idx = self._uuid_str_indices
             # Fast path: build Row objects in C++ — avoids Python loop overhead
             if not converter_map and not uuid_idx:
-                return ddbc_bindings.construct_rows(
-                    rows_data, Row, column_map, self
-                )
+                return ddbc_bindings.construct_rows(rows_data, Row, column_map, self)
             return [
                 Row(
                     row_data,
@@ -2635,9 +2633,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
             uuid_idx = self._uuid_str_indices
             # Fast path: build Row objects in C++ — avoids Python loop overhead
             if not converter_map and not uuid_idx:
-                return ddbc_bindings.construct_rows(
-                    rows_data, Row, column_map, self
-                )
+                return ddbc_bindings.construct_rows(rows_data, Row, column_map, self)
             return [
                 Row(
                     row_data,
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -158,6 +158,14 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         self._conn_native_uuid = getattr(self.connection, "_native_uuid", None)
         self._next_row_index = 0  # internal: index of the next row the driver will return (0-based)
         self._has_result_set = False  # Track if we have an active result set
+        # Cache decoding encoding strings — these don't change between fetches,
+        # so we avoid 2 method calls + 2 dict.get() per fetch call.
+        self._cached_char_encoding = self._get_decoding_settings(
+            ddbc_sql_const.SQL_CHAR.value
+        ).get("encoding", "utf-8")
+        self._cached_wchar_encoding = self._get_decoding_settings(
+            ddbc_sql_const.SQL_WCHAR.value
+        ).get("encoding", "utf-16le")
         self._skip_increment_for_next_fetch = (
             False  # Track if we need to skip incrementing the row index
         )
@@ -173,11 +181,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         Returns:
             True if the string contains non-ASCII characters, False otherwise.
         """
-        try:
-            param.encode("ascii")
-            return False  # Can be encoded to ASCII, so not Unicode
-        except UnicodeEncodeError:
-            return True  # Contains non-ASCII characters, so treat as Unicode
+        return not param.isascii()
 
     def _parse_date(self, param: str) -> Optional[datetime.date]:
         """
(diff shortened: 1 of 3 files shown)


mssql_python/cursor.py
```diff
@@ -1451,7 +1472,9 @@ def _get_column_and_converter_maps(self):
         # Fallback to legacy column name map if no cached map
         column_map = column_map or getattr(self, "_column_name_map", None)
 
-        # Get cached converter map
+        # Refresh once per settings change, not once per row.
+        if self._cached_converters_generation != self._connection._converters_generation:
+            self._cached_converter_map = self._build_converter_map()
```
sumitmsft, line 1489: **PR #558: Preserve value-type gating when converters are registered after execution.**

The late-registration case of the earlier SQL_VARIANT converter concern remains here. If no converters are registered at `execute()` and an `SQL_WVARCHAR` converter is added afterward, this refresh switches variant columns from the legacy value-gated fallback to the metadata-only converter map. `_map_data_type` defaults `SQL_SS_VARIANT` to `str`, even though native fetching returns the underlying integer, date, UUID, or other Python value.

An integer variant containing `42` can therefore become the string converter's replacement value. The base preserves `42` without invoking that converter. This affects `fetchone()`, `fetchmany()`, and `fetchall()`.

Please keep fallback string conversion gated on the actual `str`/`bytes` value for variant or unknown columns, while preserving explicitly registered converter precedence. Add late-registration regression coverage across all three fetch APIs.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -6008,6 +6045,65 @@ void DDBCSetDecimalSeparator(const std::string& separator) {
...
+    py::str attr_column_map("_column_map");
+    py::str attr_cursor("_cursor");
+    py::str attr_column_map_lower("_column_map_lower");
+    py::str attr_column_names("_column_names");
+
+    py::list result(n);
+
+    for (Py_ssize_t i = 0; i < n; ++i) {
+        // Allocate Row without calling __init__
+        PyObject* row = row_type->tp_alloc(row_type, 0);
```
sumitmsft: **PR #558: Reject incompatible types before direct Row allocation.**

The `PyType_Check` guard addresses non-type inputs, but it does not make `tp_alloc` safe for arbitrary Python types. On the pinned Windows CPython 3.13 build, this bounded call terminates the interpreter with access violation `0xC0000005` instead of raising a Python exception:

```python
ddbc_bindings.construct_rows([[1]], types.FunctionType, {"v": 0}, None)
```

Direct allocation bypasses initialization required by that type. Ordinary cursor calls pass `Row`; this is an invalid-input crash in the Python-callable helper, not a demonstrated server-data trigger.

Please restrict `row_class` to supported Row types before allocation, or use a path that safely rejects incompatible types. Add an isolated subprocess regression for unrelated type objects as well as non-type inputs.

mssql_python/cursor.py
jahnvi480: Fixed in `e9be1880` by gating string fallback conversion on the actual value type while preserving explicit converter precedence, with regression coverage across all three fetch APIs.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Fixed in `e9be1880` by restricting allocation to `Row` and its subclasses, with isolated subprocess tests confirming unrelated types raise `TypeError` instead of crashing.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -6008,6 +6045,70 @@ void DDBCSetDecimalSeparator(const std::string& separator) {
...
+//
+// By doing tp_alloc + slot assignment in a tight C loop, this avoids:
+//   - Python bytecode dispatch (FOR_ITER, LOAD_FAST, CALL_FUNCTION)
+//   - Keyword argument processing overhead per Row
+//   - Python function call frame setup per iteration
+//
+// Accepts Row and its subclasses, whose internal fields can be assigned directly.
+// Semantically identical to _fast_create — no converter or UUID processing.
+// ---------------------------------------------------------------------------
+py::list construct_rows(const py::list& rows_data,
```
bewithgaurav: suggestion: keep the new Row-construction logic in a focused header, with only its registration here.
we already separated detection, type caching and reference ownership this way.
this keeps another independent responsibility out of `ddbc_bindings.cpp`.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -6008,6 +6045,70 @@ void DDBCSetDecimalSeparator(const std::string& separator) {
...
+    py::str attr_column_map("_column_map");
+    py::str attr_cursor("_cursor");
+    py::str attr_column_map_lower("_column_map_lower");
+    py::str attr_column_names("_column_names");
+
+    py::list result(n);
+
+    for (Py_ssize_t i = 0; i < n; ++i) {
+        // Allocate Row without calling __init__
+        PyObject* row = row_type->tp_alloc(row_type, 0);
```
bewithgaurav: suggestion: use our existing `steal()` owner for the newly allocated row, then release it into the result list.
the cleanup works today, but this brings back manual refcount handling where we already have an agreed pattern.
the borrowed `row_data` pointer can stay as-is.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Resolved

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Resolved

bewithgaurav (approved):

sumitmsft (approved):
PR #558: Both previously reported findings are addressed. Late string-converter registration preserves non-string variant values and explicit converter precedence. Native Row construction rejects incompatible types before allocation and handles reference ownership and failure cleanup safely.

No remaining actionable findings in the reviewed changes.

Issue #557, opened 2026-05-06
cursor.executemany throws RuntimeError when decimals change signs


taehoon-song:
### Describe the bug
I'm loading a polars DataFrame and one of the columns has negative and positive decimals. If there is a sign change, the driver throws a RuntimeError. I've tried to narrow it down, but the behavior is a little inconsistent.

```
Exception message: RuntimeError: Input string exceeds allowed column size at parameter index 0
Stack trace:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    df_to_sql(table_name="pytest", df=df, credential=cred)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\tsong\repo\pam\src\pam\sql\write_db.py", line 37, in df_to_sql
    cursor.executemany(stmt, df.rows())
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "C:\Users\tsong\repo\pam\.venv\Lib\site-packages\mssql_python\cursor.py", line 2338, in executemany
    ret = ddbc_bindings.SQLExecuteMany(
        self.hstmt, operation, columnwise_params, parameters_type, row_count, encoding_settings
    )
```

At the moment, I work around this by casting all Decimal columns to Strings.
```python
df = df.with_columns(pl.col(pl.Decimal).cast(pl.String).name.keep())
```

### To reproduce
Include a complete code listing (or project/solution) that we can run to reproduce the issue.

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to reproduce the issue.

```python
# I have a table named [pytest]
# CREATE TABLE [dbo].[pytest] ([col_1] [decimal](28, 14))

import polars as pl
import mssql_python as mssql
from decimal import Decimal

df = pl.DataFrame(
        {
            "col_1": [
                Decimal("-0.1"),
                Decimal("1.0"),
            ]
        }
    )

stmt = "INSERT INTO [pytest] ([col_1]) VALUES (?)"

connection_url = "Server=<SERVER_ADDRESS>;Database=<DATABASE>;Encrypt=yes;Authentication=ActiveDirectoryDefault;"
conn = mssql.connect(connection_url)
cursor = conn.cursor()
cursor.executemany(stmt, df.rows())
cursor.commit()
conn.close()
```

### Expected behavior
Insert should be successful.

### Further technical details
[...]

PR #560, opened 2026-05-08
FIX: executemany RuntimeError when decimals change signs (GH-557)

jahnvi480:

> [AB#44997](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44997)

> GitHub Issue: #557 

### Summary   
This pull request fixes an issue with inserting decimal values of varying sign (positive/negative) using `executemany`, where the allocated column size could be too small if a negative value's string representation is longer than the sample used for sizing. It ensures the correct column size is used for all rows, preventing runtime errors. A new test is added to verify the fix.

Bug fix for decimal handling in `executemany`:

* In `mssql_python/cursor.py`, when inserting decimals as `SQL_VARCHAR`, the code now scans all rows to determine the maximum formatted string length, ensuring the column size is large enough even when negative values are present.

Testing improvements:

* Added `test_executemany_decimal_sign_change` in `tests/test_004_cursor.py` to cover cases where decimals change sign, verifying that no runtime error occurs and the data is correctly inserted and retrieved.


mssql_python/cursor.py
```diff
@@ -2225,7 +2233,7 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
                 if hasattr(seq_of_parameters, "__getitem__")
                 else []
             )
-            sample_value, min_val, max_val = self._compute_column_type(column)
+            sample_value, min_val, max_val, _ = self._compute_column_type(column)
```
subrata-ms, line 2236: "_'  is that (4th param) correct coding standard or do we need add a variable name?

mssql_python/cursor.py
subrata-ms, line 2281: Why above code (line 2277 to 2281) is not able to find the length of the negative number properly?

mssql_python/cursor.py
```diff
@@ -2322,6 +2330,17 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
...
                     paraminfo.columnSize = 1
 
+                # Correct column size for Decimal columns sent as SQL_VARCHAR (GH-557).
+                # The sample value's formatted string may be shorter than another
+                # row's (e.g. positive sample "1.0" = 3 chars vs negative "-0.1" = 4).
+                # max_decimal_len was already computed during _compute_column_type
+                # so no extra iteration is needed.
+                if (
+                    paraminfo.paramSQLType == ddbc_sql_const.SQL_VARCHAR.value
+                    and max_decimal_len > paraminfo.columnSize
```
subrata-ms, line 2340: paraminfo.columnSize is nothing but max ( max_binary_size ,1) which is expected to work correctly. If we determined that it is not working properly then we should do cleanup in this block of code and remove the unnecessary code and keep the current/latest modified code.

mssql_python/cursor.py
jahnvi480: This is correct based on the coding standards, reason we didn't use any variable name here is because it won't be used.

mssql_python/cursor.py
jahnvi480: That `max_binary_size` loop is inside an if `sql_type` in (`SQL_BINARY, SQL_VARBINARY, SQL_LONGVARBINARY`) block - it only runs for binary (bytes/bytearray) columns, not for Decimal columns. These are completely separate type paths:

Binary data -> `SQL_BINARY/SQL_VARBINARY` -> has its own max-size scan
Decimal data (MONEY range) -> `SQL_VARCHAR` -> had no max-size scan (the bug)
Decimal data (outside MONEY range) -> `SQL_NUMERIC` -> uses precision, not string length 
The binary loop never sees Decimal values because Decimal values never enter the `SQL_BINARY` type check. They take a different path through `_map_sql_type` -> `SMALLMONEY/MONEY` range -> `SQL_VARCHAR`

mssql_python/cursor.py
jahnvi480: The `max_binary_size` loop is working correctly for its purpose - binary data sizing is fine. The Decimal bug is a separate issue in a separate type path. There's no broken code to clean up

bewithgaurav (dismissed):
ran the #557 reproducer matrix on macos arm64 against sql2022, all seven cases pass on this branch (`[0.1, -0.1]`, `[-0.1, 1.0]`, `[1.0, -12345678.9]` all fail on origin/main with `Input string UTF-16 length exceeds allowed column size`, fixed here). hoisting `max_decimal_formatted_len` into `_compute_column_type` keeps it to a single pass, and the bump matches what gets written at the per-row `format(val, "f")` step, so sizing stays consistent end-to-end. diff coverage 100%, all existing reviewer threads addressed. one tiny ride-along on the test docstring inline. lgtm.

subrata-ms (approved):

gargsaumya (approved):

Issue #532, opened 2026-04-21
Login failures are raised as RuntimeError instead of a mssql_python exception


martinstuder:
Login failures raised during `connect` are raised as `RuntimeError` instead of a mssql_python exception (see https://github.com/microsoft/mssql-python/wiki/Exception):

```
Traceback (most recent call last):
  File "mssql_connect.py", line 5, in <module>
    connection = connect(
                 ^^^^^^^^^^^^^^
  File ".\site-packages\mssql_python\db_connection.py", line 46, in connect
    conn = Connection(
           ^^^^^^^^^^^
  File ".\site-packages\mssql_python\connection.py", line 319, in __init__
    self._conn = ddbc_bindings.Connection(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Login failed for user ''.
```

I would expect an appropriate mssql_python error to be raised.

The specific use case we have is that we use access tokens to connect to SQL Server. These tokens expire after a certain period of time after which a retry mechanism retries the operation. A mssql_python specific error such as `OperationalError` would help to better isolate the exceptions that have to be retried.

gargsaumya:
Hi @martinstuder, thanks for the details and the clear use case!
You're right, connection-time failures (login, network, driver errors) are currently being raised as RuntimeError instead of the appropriate DB-API 2.0 exceptions. We've root-caused the issue and are working on a fix. We’ll share an update once the fix is in.

gargsaumya:
Thanks for reporting this issue. The fix is now available in our latest release 1.7.1. Please continue trying out mssql-python, and feel free to keep reporting any issues.

PR #562, opened 2026-05-11
FIX: GH-Issue (#532) - Login failures are raised as RuntimeError instead of a mssql_python exception

gargsaumya:

> [AB#44632](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44632)

> GitHub Issue: #532 
### Summary   
  
This pull request improves error handling in the `mssql_python` connection layer by ensuring that low-level runtime errors from the C++ pybind layer are consistently mapped to the correct DB-API 2.0 exceptions using SQLSTATE codes. This makes error reporting more robust and predictable, especially during connection, commit, and rollback operations. The changes also add comprehensive tests to verify this behavior.

**Error handling improvements:**

* Introduced the `_raise_connection_error` function in `connection.py` to parse `RuntimeError` messages from the C++ layer, extract SQLSTATE codes, and map them to appropriate DB-API exceptions using `sqlstate_to_exception`. If no SQLSTATE is found, a fallback `OperationalError` is raised. (`[mssql_python/connection.pyR61-R89](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R61-R89)`)
* Updated the connection initialization, `setautocommit`, `commit`, and `rollback` methods in `connection.py` to wrap calls to the native layer in try/except blocks that utilize `_raise_connection_error`, ensuring consistent exception mapping throughout the connection lifecycle. (`[[1]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R362-R367)`, `[[2]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R528-R531)`, `[[3]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R1516-R1519)`, `[[4]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R1542-R1545)`)
* Modified the C++ `Connection::checkError` method to format error messages as `"SQLSTATE:XXXXX:<odbc_error_message>"`, enabling the Python layer to parse and handle errors correctly. (`[mssql_python/pybind/connection/connection.cppR182-R186](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR182-R186)`)

**Testing enhancements:**

* Added a regression test (`test_connect_runtime_error_mapped_to_correct_dbapi_exception`) to verify that connection, commit, and rollback errors from the C++ layer are mapped to the correct DB-API exceptions, including handling of unknown or missing SQLSTATE codes. (`[tests/test_006_exceptions.pyR193-R257](diffhunk://#diff-16f96f3facfc774c12b49a09116108763139a8a92ceef526fbb4f4f01600c6fdR193-R257)`)

**Dependency update:**

* Imported `sqlstate_to_exception` in `connection.py` to support the new error mapping logic. (`[mssql_python/connection.pyR41](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R41)`)

### DIFF
--- a/mssql_python/connection.py
+++ b/mssql_python/connection.py
@@ -38,6 +38,7 @@ from mssql_python.exceptions import (
     InternalError,
     ProgrammingError,
     NotSupportedError,
+    sqlstate_to_exception,
 )
 from mssql_python.auth import extract_auth_type, process_connection_string
 from mssql_python.constants import ConstantsDDBC, GetInfoConstants
@@ -56,6 +57,35 @@ INFO_TYPE_STRING_THRESHOLD: int = 10000
 # UTF-16 encoding variants that should use SQL_WCHAR by default
 # Note: "utf-16" with BOM is NOT included as it's problematic for SQL_WCHAR
 UTF16_ENCODINGS: frozenset[str] = frozenset(["utf-16le", "utf-16be"])
+
+_SQLSTATE_RE = re.compile(r"^SQLSTATE:([A-Z0-9]{5}):(.*)", re.DOTALL)
+
+
+def _raise_connection_error(e: RuntimeError) -> None:
+    """Map a RuntimeError from the C++ pybind layer to the correct DB-API 2.0 exception.
+
+    Connection::checkError() throws "SQLSTATE:XXXXX:<odbc_message>" so the SQLSTATE
+    can be mapped via sqlstate_to_exception(), consistent with cursor-level error handling.
+    """
+    error_msg = str(e)
+    match = _SQLSTATE_RE.match(error_msg)
+    if match:
+        sqlstate, ddbc_error = match.group(1), match.group(2)
+        exc = sqlstate_to_exception(sqlstate, ddbc_error)
+        if exc is None:
+            logger.error("Unknown SQLSTATE %s, raising DatabaseError", sqlstate)
+            raise DatabaseError(
+                driver_error=f"An error occurred with SQLSTATE code: {sqlstate}",
+                ddbc_error=ddbc_error,
+            ) from None
+        logger.error("Connection error (SQLSTATE %s): %s", sqlstate, ddbc_error)
+        raise exc from None
+    # Fallback: no SQLSTATE prefix — e.g. "Connection handle not allocated"
+    logger.error("Connection error: %s", error_msg)
+    raise OperationalError(
+        driver_error="Connection operation failed",
+        ddbc_error=error_msg,
+    ) from None
 
 
 def _validate_utf16_wchar_compatibility(
@@ -333,9 +363,12 @@ class Connection:
         if not PoolingManager.is_initialized():
             PoolingManager.enable()
         self._pooling = PoolingManager.is_enabled()
-        self._conn = ddbc_bindings.Connection(
-            self.connection_str, self._pooling, self._attrs_before
-        )
+        try:
+            self._conn = ddbc_bindings.Connection(
+                self.connection_str, self._pooling, self._attrs_before
+            )
+        except RuntimeError as e:
+            _raise_connection_error(e)
         self.setautocommit(autocommit)
 
         # Register this connection for cleanup before Python shutdown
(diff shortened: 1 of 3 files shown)


mssql_python/pybind/connection/connection.cpp
```diff
@@ -179,9 +179,11 @@ void Connection::disconnect() {
 // DB spec compliant
 void Connection::checkError(SQLRETURN ret) const {
     if (!SQL_SUCCEEDED(ret)) {
+        // Format: "SQLSTATE:XXXXX:<odbc_error_message>" — parsed by _raise_connection_error()
         ErrorInfo err = SQLCheckError_Wrap(SQL_HANDLE_DBC, _dbcHandle, ret);
+        std::string sqlState = WideToUTF8(err.sqlState);
         std::string errorMsg = WideToUTF8(err.ddbcErrorMsg);
-        ThrowStdException(errorMsg);
+        ThrowStdException("SQLSTATE:" + sqlState + ":" + errorMsg);
```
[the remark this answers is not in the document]
gargsaumya: Valid. Done

mssql_python/connection.py
```diff
@@ -57,6 +58,35 @@
...
+            ) from None
+        logger.error("Connection error (SQLSTATE %s): %s", sqlstate, ddbc_error)
+        raise exc from None
+    # Fallback: no SQLSTATE prefix — e.g. "Connection handle not allocated"
+    logger.error("Connection error: %s", error_msg)
+    raise OperationalError(
+        driver_error="Connection operation failed",
+        ddbc_error=error_msg,
+    ) from None
+
```
bewithgaurav, line 96: C++ `checkError` now prefixes every error with `SQLSTATE:XXXXX:`, but only `__init__`, `setautocommit`, `commit`, `rollback` are wrapped. the other connection methods (`get_autocommit`, `setencoding`/`setdecoding`, `getinfo`, `cursor()`) still propagate the raw RuntimeError, now with the prefix visible to users.

easiest fix is to wrap those call sites too, or move the parse-and-map into the pybind layer so python doesn't need to know about the prefix.

bewithgaurav (commented):
repro'd #532 on this branch and main (macos): bad-host connect now raises `OperationalError` with the real ODBC text, main still raises raw `RuntimeError`. new test passes.

one inline note about the `SQLSTATE:` prefix leaking through the connection methods that aren't wrapped.

mssql_python/connection.py
bewithgaurav: follow-up on the same theme: the regex parsing here is a workaround for the C++ side serializing structured data (sqlstate, message) into a string just so it can ride pybind's default `RuntimeError` translation. then python reparses it. there's already a TODO above `checkError` ("Add an exception class in C++ for error handling, DB spec compliant") that points at the right fix, register a custom pybind exception type (or `py::register_exception`) carrying sqlstate and message as fields, throw and catch it directly. drops both the regex and the prefix-leak failure mode.

if it's beyond the scope of #532, fine to defer, please file as a follow-up task so it doesn't get lost.

bewithgaurav (changes requested):
requesting changes per the inline follow-up on `_raise_connection_error`. fine to defer to a separate task if out of scope for #532.

gargsaumya:
> requesting changes per the inline follow-up on `_raise_connection_error`. fine to defer to a separate task if out of scope for #532.

Thanks for the review! I've extended exception mapping to the relevant connection methods. [getinfo()](vscode-file://vscode-app/c:/Users/gargsaumya/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [set_attr()](vscode-file://vscode-app/c:/Users/gargsaumya/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) don't need wrapping because These methods already handle [RuntimeError](vscode-file://vscode-app/c:/Users/gargsaumya/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in their existing exception handlers.

Re: C++ exception class approach: I agree with implementing a custom exception hierarchy in C++, I've added a task to the backlog to track this.

bewithgaurav (dismissed):
lgtm

bewithgaurav (approved):

subrata-ms (approved):

Issue #565, opened 2026-05-11
Reopen Issue #491: SSH tunneling fails using paramiko+sshtunnel


jschuba:
### Describe the bug
`mssql_python.connect()` hangs when connecting using an SSH Tunnel created with `paramiko `+ `sshtunnel`.  This same setup works fine with `pymssql `or `pyodbc`.  

In [issue 491](https://github.com/microsoft/mssql-python/issues/491), using `mssql-python 1.4.0`,  the connection would time out after 15 seconds.   Using `mssql-python 1.6.0`, the connection simply hangs indefinitely.   The fix reported in [PR 541](https://github.com/microsoft/mssql-python/pull/541) does not appear to have been tested for this scenario. 

### To reproduce

See also [issue 491](https://github.com/microsoft/mssql-python/issues/491)

```python
import sshtunnel

import mssql_python
import pymssql
import pyodbc


# Fill in with your credentials
ssh_host = {"ssh_address_or_host": "host",
            "ssh_username": "",
            "ssh_password": "",
            "ssh_port": 22,
            "local_bind_address": ('127.0.0.1', 0),
            }

sql_server = {"host": "DB_HOST_IP",
              "port": 1433,
              "database": "my_database",
              "user": "",
              "password": "",
              }

tunnel = sshtunnel.open_tunnel(
    (ssh_host['ssh_address_or_host'], ssh_host["ssh_port"]),
    ssh_username=ssh_host["ssh_username"],
    ssh_password=ssh_host["ssh_password"],
    local_bind_address=ssh_host["local_bind_address"],
    remote_bind_address=(sql_server['host'], sql_server['port']))

try:
    tunnel.start()

    ### Connect with pyodbc
    pyodbc_conn_str = (
        f"Driver={{ODBC Driver 17 for SQL Server}};"
        f"Server={tunnel.local_bind_host},{tunnel.local_bind_port};"
        f"DATABASE={sql_server['database']};"
        f"UID={sql_server['user']};"
        f"PWD={sql_server['password']};"
        "Encrypt=no;TrustServerCertificate=yes")
    conn = pyodbc.connect(pyodbc_conn_str)
    curs = conn.cursor()
    curs.execute("SELECT 'hello from pyodbc'")
    print(curs.fetchall())  ### This works

    ### Connect with mssql_python
    CONNECTION_STRING = (
        f"Server={tunnel.local_bind_host},{tunnel.local_bind_port};"
        f"DATABASE={sql_server['database']};"
[...]

Issue #491, opened 2026-03-26
SSH tunneling fails using paramiko+sshtunnel


joefrancia:
### Describe the bug

`mssql_python.connect()` fails with a TCP timeout error when connecting through an SSH tunnel established via `paramiko` + `sshtunnel`. The identical connection works correctly when the SSH tunnel is established manually via OpenSSH in the terminal (`ssh -L`), and the identical `paramiko`+`sshtunnel` setup works correctly when using `pymssql` as the driver.

The SSH session dies approximately 15 seconds after authentication with `SSH session not active` before the channel can be opened. Logging suggests `mssql_python.connect()` is delaying ~15 seconds before attempting to open the tunnel channel, long enough for the server-side SSH session idle timeout to kill the session. `pymssql` does not exhibit this delay and connects successfully through the same tunnel code.

```
Exception message: RuntimeError: [Microsoft][ODBC Driver 18 for SQL Server]TCP Provider: Timeout error [258].
Stack trace:
  File "test_mssql.py", line 30, in <module>
    conn = mssql_python.connect(CONNECTION_STRING)
RuntimeError: [Microsoft][ODBC Driver 18 for SQL Server]TCP Provider: Timeout error [258].
```

Paramiko/sshtunnel logs at point of failure:
```
INFO  | Authentication (publickey) successful!
DEBUG | Received global request "hostkeys-00@openssh.com"
DEBUG | Rejecting "hostkeys-00@openssh.com" global request from server.
DEBUG | Debug msg: b'/home/ubuntu/.ssh/authorized_keys:1: key options: ...'
# ~15 seconds of silence here
ERROR | Could not establish connection from local ('127.0.0.1', 52316) to remote ('DB_HOST_INTERNAL_IP', 1433) side of the tunnel: open new channel ssh error: SSH session not active
```

For comparison, `pymssql` with identical tunnel code opens the channel immediately after auth with no delay:
```
INFO  | Authentication (publickey) successful!
DEBUG | [chan 0] Max packet in: 32768 bytes
DEBUG | [chan 0] Max packet out: 32768 bytes
DEBUG | Secsh channel 0 opened.
# connects successfully
```

### To reproduce

```python
import mssql_python
import paramiko
from sshtunnel import SSHTunnelForwarder

pkey = paramiko.Ed25519Key.from_private_key_file('/path/to/key.pem')

with SSHTunnelForwarder(
    ssh_address_or_host=("JUMPBOX_IP", 22),
[...]

PR #568, opened 2026-05-11
FIX: Release GIL around blocking SQLSetConnectAttr calls (#565)

saurabh500:

> GitHub Issue: #565

### Summary

Fixes the indefinite hang of `mssql_python.connect()` when the target server is reached through an in-process SSH tunnel created with `paramiko` + `sshtunnel` (issue #565, originally reported as #491).

**Root cause**

PR #541 released the GIL around `SQLDriverConnect`, `SQLDisconnect`, and `SQLEndTran`, but a handful of `SQLSetConnectAttr` call sites in `mssql_python/pybind/connection/connection.cpp` were left holding the GIL:

- `Connection::setAutocommit` — `SQL_ATTR_AUTOCOMMIT`. This is the call that hangs in the issue's repro: it runs by default immediately after every `connect()`.
- `Connection::setAttribute` (int / wide-string / bytes paths) — user-facing `set_attr()`.
- `Connection::reset` — `SQL_ATTR_RESET_CONNECTION` and `SQL_ATTR_TXN_ISOLATION`, called by the connection pool on acquire.

When the GIL is held during a call that ends up doing blocking work, paramiko's transport thread cannot run and so cannot forward bytes through the SSH channel — the driver waits forever for a response that never comes. `pyodbc` does not exhibit this because it releases the GIL around its blocking ODBC calls.

**Fix**

Wrap each of these call sites in `py::gil_scoped_release`, matching the pattern PR #541 already established for `SQLDriverConnect` / `SQLEndTran`. Local-only attribute reads (`SQLGetConnectAttr` for `SQL_ATTR_CONNECTION_DEAD` / `SQL_ATTR_AUTOCOMMIT`, `SQLGetInfo`, `SQLAllocHandle`) are intentionally left alone.

**Additional fix: Connection pool mutex/GIL deadlock**

The initial GIL-release change in `Connection::reset()` introduced a secondary deadlock in `ConnectionPool::acquire()`. The pool's `acquire()` method called `conn->isAlive()` and `conn->reset()` **while holding the pool `_mutex`**. With `reset()` now releasing the GIL internally, this created a classic ABBA lock-ordering inversion:

| Thread A | Thread B |
|----------|----------|
| Holds `_mutex`, calls `reset()` | Holds GIL, waiting on `_mutex` |
| `reset()` releases GIL, ODBC call | Gets nothing (blocked on mutex) |
| ODBC returns, tries to reacquire GIL — **blocked** | Still blocked on `_mutex` |
| **Deadlock:** holds mutex, needs GIL | **Deadlock:** holds GIL, needs mutex |

The fix restructures `ConnectionPool::acquire()` to pop one candidate connection from the pool under the mutex, then validate it (`isAlive()` + `reset()`) **outside** the mutex. This follows the same pattern already established in the pool for `connect()` (Phase 2.5) and `disconnect()` (Phase 3/4), where blocking ODBC calls that release the GIL are performed without holding the pool lock.

**Why this is safe:**

1. **No data races on `_pool`:** The candidate is popped from the deque atomically under the mutex before any validation. No other thread can access the same connection object.
2. **`_current_size` stays accurate:** Stale connections are decremented during pruning (Phase 1). Failed candidates are decremented individually after validation. New connections increment the count under the mutex before `connect()` is called.
3. **Pool size limit is respected:** The "reserve a new slot" check (`_current_size < _max_size`) still happens under the mutex, after confirming no viable candidates remain.
4. **Exception safety:** If `isAlive()` or `reset()` throws, the connection is caught and treated as dead — it goes to the disconnect list and `_current_size` is decremented.
5. **Single-threaded behavior unchanged:** With one thread, the loop simply pops, validates, and returns — identical to the previous while-loop but without holding the mutex during ODBC calls.

**Verification**

- Reproduced #565 locally with an in-process Python TCP forwarder (simulating paramiko/sshtunnel). Before the GIL-release fix, `mssql_python.connect()` hung indefinitely; after the fix it completes in ~0.3s.
- The pool deadlock was reproduced by returning connections to the pool and having 2+ threads call `connect()` concurrently — this hung indefinitely before the pool fix and completes instantly after.
- Full test suite: **1707 passed, 64 skipped, 0 failures** (~2 minutes).

**Diff:** 2 files changed — `connection.cpp` (+42/-12 GIL release) and `connection_pool.cpp` (+39/-26 mutex restructure).

### DIFF
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -315,9 +315,13 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
 
             SQLPOINTER ptr;
             SQLINTEGER length;
-            
-            ptr = reinterpretU16stringAsSqlWChar(this->wstrStringBuffer);
-            length = static_cast<SQLINTEGER>(this->wstrStringBuffer.length() * sizeof(SQLWCHAR));
+            // Copy to a stack-local buffer so that releasing the GIL below
+            // doesn't expose a race where another thread overwrites the
+            // member wstrStringBuffer (and reallocates) while the ODBC
+            // driver is still reading from ptr.
+            std::u16string localStrBuffer = this->wstrStringBuffer;
+            ptr = reinterpretU16stringAsSqlWChar(localStrBuffer);
+            length = static_cast<SQLINTEGER>(localStrBuffer.length() * sizeof(SQLWCHAR));
 
             SQLRETURN ret;
             {
@@ -336,11 +340,12 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
         }
     } else if (py::isinstance<py::bytes>(value) || py::isinstance<py::bytearray>(value)) {
         try {
-            std::string binary_data = value.cast<std::string>();
-            this->strBytesBuffer.clear();
-            this->strBytesBuffer = std::move(binary_data);
-            SQLPOINTER ptr = const_cast<char*>(this->strBytesBuffer.c_str());
-            SQLINTEGER length = static_cast<SQLINTEGER>(this->strBytesBuffer.size());
+            // Copy to a stack-local buffer so that releasing the GIL
+            // doesn't expose a race where another thread overwrites the
+            // member strBytesBuffer while the driver reads from ptr.
+            std::string localBytesBuffer = value.cast<std::string>();
+            SQLPOINTER ptr = const_cast<char*>(localBytesBuffer.c_str());
+            SQLINTEGER length = static_cast<SQLINTEGER>(localBytesBuffer.size());
 
             SQLRETURN ret;
             {
--- a/mssql_python/pybind/connection/connection.cpp
+++ b/mssql_python/pybind/connection/connection.cpp
@@ -211,9 +211,16 @@ void Connection::setAutocommit(bool enable) {
     }
     SQLINTEGER value = enable ? SQL_AUTOCOMMIT_ON : SQL_AUTOCOMMIT_OFF;
     LOG("Setting autocommit=%d", enable);
-    SQLRETURN ret =
-        SQLSetConnectAttr_ptr(_dbcHandle->get(), SQL_ATTR_AUTOCOMMIT,
-                              reinterpret_cast<SQLPOINTER>(static_cast<SQLULEN>(value)), 0);
+    SQLRETURN ret;
+    {
+        // Release the GIL during the blocking ODBC call. Holding the GIL
+        // here can deadlock when the network path goes through another
+        // Python thread (e.g. an in-process SSH tunnel via paramiko +
+        // sshtunnel), since that thread also needs the GIL to run.
+        py::gil_scoped_release release;
+        ret = SQLSetConnectAttr_ptr(_dbcHandle->get(), SQL_ATTR_AUTOCOMMIT,
+                                    reinterpret_cast<SQLPOINTER>(static_cast<SQLULEN>(value)), 0);
+    }
     checkError(ret);
     if (value == SQL_AUTOCOMMIT_ON) {
         LOG("Autocommit enabled");
@@ -286,9 +293,15 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
         // Get the integer value
         int64_t longValue = value.cast<int64_t>();
 
-        SQLRETURN ret = SQLSetConnectAttr_ptr(
-            _dbcHandle->get(), attribute,
-            reinterpret_cast<SQLPOINTER>(static_cast<SQLULEN>(longValue)), SQL_IS_INTEGER);
+        SQLRETURN ret;
+        {
+            // Release the GIL around the ODBC call for consistency with the
+            // other connection-attribute paths; some attributes can block.
+            py::gil_scoped_release release;
+            ret = SQLSetConnectAttr_ptr(
+                _dbcHandle->get(), attribute,
+                reinterpret_cast<SQLPOINTER>(static_cast<SQLULEN>(longValue)), SQL_IS_INTEGER);
+        }
 
         if (!SQL_SUCCEEDED(ret)) {
             LOG("Failed to set integer attribute=%d, ret=%d", attribute, ret);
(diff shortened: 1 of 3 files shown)


gargsaumya (dismissed):
LGTM.

tests/test_023_ssh_tunnel_gil_release.py
```diff
@@ -0,0 +1,254 @@
...
+    env = os.environ.copy()
+    env["DB_CONNECTION_STRING"] = base_conn_str
+    env["PYTHONUNBUFFERED"] = "1"
+    # Propagate sys.path so the subprocess can find mssql_python even when
+    # the package is only available via pytest's path manipulation or an
+    # editable install rooted in the workspace.
+    env["PYTHONPATH"] = os.pathsep.join(sys.path)
+
+    proc = subprocess.Popen(
+        [sys.executable, __file__],
```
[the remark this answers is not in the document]
saurabh500: Fixed

mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -39,39 +40,51 @@ std::shared_ptr<Connection> ConnectionPool::acquire(const std::wstring& connStr,
 
         size_t pruned = before - _pool.size();
         _current_size = (_current_size >= pruned) ? (_current_size - pruned) : 0;
+    }
```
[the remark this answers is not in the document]
saurabh500: Pre-existing issue. Out of scope of this PR

bewithgaurav (commented):
spotted a buffer race on the windows wide-string path in `setAttribute`. ptr aims into the shared `wstrStringBuffer` member field and the new GIL release lets a second thread sharing the connection re-enter `setAttribute` and reallocate the buffer mid-call. the field was always shared, but the GIL was serializing access by accident. couldn't repro from macOS since that path uses a stack-local. 
inline comment below has a sketch.

rest looks good - tested locally, ssh repro from #565 passes here, pool restructure holds up under concurrent acquire stress.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -342,7 +355,11 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
             length = static_cast<SQLINTEGER>(this->wstrStringBuffer.length() * sizeof(SQLWCHAR));
 #endif
 
-            SQLRETURN ret = SQLSetConnectAttr_ptr(_dbcHandle->get(), attribute, ptr, length);
+            SQLRETURN ret;
+            {
+                py::gil_scoped_release release;
+                ret = SQLSetConnectAttr_ptr(_dbcHandle->get(), attribute, ptr, length);
```
bewithgaurav, line 329: ptr aims into `wstrStringBuffer` (a member field) and we drop the GIL across `SQLSetConnectAttr`. if the connection is shared across threads, a second thread can re-enter `setAttribute` mid-call:

```
T1: wstrStringBuffer = "us_english"; ptr = &buf[0]; release GIL; SQLSetConnectAttr(ptr) ...
T2:                                                              wstrStringBuffer = "british_english_with_a_long_name"
                                                                 // std::wstring reallocates, T1's ptr now dangles
T1: ...driver reads from ptr -> freed memory
```

the shared buffer was already there, but the GIL was quietly serializing access. dropping it here makes the race reachable.

macOS/Linux is safe by accident because `sqlwcharBuffer` is a stack-local vector. windows path and bytes path on 384 aim straight into the member field.

fix is one stack copy before the release block, same shape as the unix branch already does.

tests/test_023_ssh_tunnel_gil_release.py
```diff
@@ -0,0 +1,254 @@
...
+        f"Unexpected subprocess output.\n"
+        f"stdout:\n{out.decode(errors='replace')}\n"
+        f"stderr:\n{err.decode(errors='replace')}"
+    )
+
+
+# Subprocess entry point: when this file is run as a script (by the test
+# above), execute the forwarded-connect body. Pytest collection ignores
+# this block.
+if __name__ == "__main__":
```
bewithgaurav, line 253: worth a test for the buffer race
two threads sharing a connection, hammering `set_attr(SQL_ATTR_CURRENT_CATALOG, ...)` with short and long values to force the buffer to move. on a buggy build the driver eventually returns errors with a catalog name that's neither input.

didn't repro it myself - it might be non-deterministic & would need tuning, is windows-only since the unix path is already safe.
can skip and defer to stress tests if non-reproable

saurabh500: Whats the other inline in "worth a test for the buffer race in the other inline" mean?

saurabh500: OK you mean the inline comment above. I think I understand now :D

bewithgaurav: just edited, I meant the other inline comment.

mssql_python/pybind/connection/connection.cpp
saurabh500: This is a good one. 
I didn't realize that Windows and Unix have different paths. 
Looking into it.

saurabh500: Great catch — fixed in `05c6ca2c`. Both the Windows wstring path and the bytes path now copy to stack-local buffers before releasing the GIL, matching the pattern the Unix branch already uses.

Re: the stress test — there's no existing test that exercises concurrent `set_attr` on a shared connection. The fix is structural (`ptr` can no longer dangle since it points into a stack-local that outlives the ODBC call), so a race test would always pass on the fixed code. That said, a defensive concurrent `set_attr` stress test is good due diligence and would catch regressions on both platforms.

Opened #578 to track adding a cross-platform concurrent `set_attr` stress test.

tests/test_023_ssh_tunnel_gil_release.py
saurabh500: Agreed on the value of a stress test here. Since there are no existing concurrent `set_attr` stress tests in the repo, adding one feels out of scope for this PR — it would be new test infrastructure rather than a regression test for the fix itself. Opened #578 to track this as a follow-up so it gets proper attention without holding up the deadlock fix.

gargsaumya (dismissed):

bewithgaurav (dismissed):

bewithgaurav (approved):

gargsaumya (approved):

Issue #534, opened 2026-04-22
Support all auth options using bulk copy


MeinAccount:
### Describe the bug
Using `Authentication=ActiveDirectoryServicePrincipal` causes `bulkcopy` to carsh with `RuntimeError: Failed to connect to SQL Server: Connection Error: Authentication method 'ActiveDirectoryServicePrincipal' is not supported. No token provider was registered for this method.`

[The docs to not mention supported auth options.](https://github.com/microsoft/mssql-python/wiki/Bulk-Copy-(BCP)-API-Reference)

### Further technical details
Python version: Python 3.12.9
SQL Server version: Azure SQL
Library Version: 1.5.0

MeinAccount:
Using the trick from #127  also doesn't work: `RuntimeError: Failed to connect to SQL Server: Sql Error: 18456: Class 14: State 1: Login failed for user ''. on Unknown in Unknown at line 0`

saurabh500:
@bewithgaurav what are you thinking? Take the token and pass it to mssql-py-core with the Access token API ?

bewithgaurav:
@MeinAccount - thanks for raising this issue - this is a bug on token collection logic, and the PR is under development, we'll post more updates as we go.
 
@saurabh500 - yes same lines, this is happening due to missed auth cases in token acquistion path.

nspcc-tv:
I have the same issue. Could you update with possible fix & ETA please?

MeinAccount:
Bump @bewithgaurav . Can you give us an estimate for our planning?

bewithgaurav:
@MeinAccount @nspcc-tv - Thanks for your patience, we're in process of developing the PRs for this since the changes are extending to the TDS auth internals as well.
As of now the bulkcopy path supports `ActiveDirectoryDefault`, `ActiveDirectoryDeviceCode` & `ActiveDirectoryInteractive` for EntraID auth. 
We're planning to add `ActiveDirectoryMSI` by v1.8.0 (release date 28/05)
We will add an estimation update on the other auth options - esp. ServicePrincipal by the end of this week
Thanks again for your continued patience, we will update the wikis & docs shortly.

bewithgaurav:
The support for ActiveDirectoryMSI is live now with our newly released v1.8.0, please uograde and start using the same.
The support for ServicePrincipal is under development and will go live with the next release (v1.9.0)
Thanks for your continued support and patience!

MeinAccount:
[...]

PR #573, opened 2026-05-12
FEAT: Add ActiveDirectoryMSI support for bulk copy

bewithgaurav:

> [AB#45137](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45137)

> GitHub Issue: #534

### Summary

Adds `Authentication=ActiveDirectoryMSI` support to bulk copy. Partial fix for #534.

- Zero-arg `ManagedIdentityCredential()` for system-assigned MSI.
- `ManagedIdentityCredential(client_id=UID)` for user-assigned MSI. Matches ODBC convention where `UID` carries the identity's `client_id` under MSI.
- Threads optional `credential_kwargs` through `get_auth_token` / `get_raw_token` / `_acquire_token` so future auth methods that need constructor args plug in via the same channel.
- Credential cache key stays a plain string for zero-arg auth types and becomes a tuple `(auth_type, sorted_kwargs)` when kwargs are present, so different `client_id`s get separate cached credentials.

`mssql-py-core` (the Rust path used by bulk copy) doesn't itself acquire Entra tokens. Python pre-acquires a JWT and passes it as `access_token`. Today this works for `Default`, `DeviceCode`, and `Interactive`. MSI was missing, the most common Azure-hosted-service auth method.

ServicePrincipal and Password are explicitly **out of scope** for this PR. They need different design work and ship separately.

#### How user-assigned MSI survives the bulkcopy fresh-token path

`Connection.__init__` calls `process_connection_string`, which sanitizes `self.connection_str` (strips `UID=`, `Authentication=`, `PWD=`). Bulkcopy needs a fresh token at copy time. Re-parsing `self.connection_str` at that point would lose UID and silently degrade user-assigned MSI to system-assigned (wrong-identity bug).

Fix shape: `process_connection_string` returns a 4-tuple including the captured `credential_kwargs`. `Connection.__init__` persists `_credential_kwargs` alongside `_auth_type`. `cursor.bulkcopy()` reads `self.connection._credential_kwargs` directly. No re-parse, no chance for the sanitized string to drop the `client_id`.

#### Connection string examples

```python
# System-assigned MSI
"Server=tcp:foo.database.windows.net;Authentication=ActiveDirectoryMSI;Database=mydb;"

# User-assigned MSI
"Server=tcp:foo.database.windows.net;Authentication=ActiveDirectoryMSI;UID=<client_id_guid>;Database=mydb;"

### CODE REVIEW

File: mssql_python/auth.py [hunk 1]
```diff
@@ -130,6 +150,20 @@ def _acquire_token(auth_type: str) -> Tuple[bytes, str]:
...
 
 
+def _extract_msi_client_id(parameters: List[str]) -> Optional[str]:
+    """Pull UID out of connection parameters for user-assigned MSI.
+
+    For ActiveDirectoryMSI, UID (when present) carries the user-assigned
+    identity's client_id. Returns None for system-assigned MSI.
+    """
+    for param in parameters:
+        key, _, value = param.strip().partition("=")
```
saurabh500: Is it possible to get this from the conn str parser map? Assuming we need the conn str UID

saurabh500: I am not sure what will happen if someone provides a UID={hello=world}

bewithgaurav: thanks - missed this, parser should be used as a standard across 
that handles `{hello=world}` correctly

mssql_python/auth.py
```diff
@@ -130,6 +151,27 @@ def _acquire_token(auth_type: str) -> Tuple[bytes, str]:
...
+    ``hello=world`` (no surrounding braces, no false split on the inner
+    ``=``), and a semicolon inside a legitimate braced value (e.g.
+    ``Database={foo;uid=victim;bar}``) cannot spoof a top-level ``UID=``.
+    """
+    try:
+        parsed = _ConnectionStringParser(validate_keywords=False)._parse(connection_string)
+    except Exception:  # noqa: BLE001 — parser raises ConnectionStringParseError on malformed input;
+        # absence of UID is the safe answer for credential extraction.
+        return None
+    uid = (parsed.get("uid") or "").strip()
```
saurabh500, line 175: I dont know if this is the right place, but we are kind of downgrading to SAMI if UAMI uid is not found.

Since we are falling back, do you think it makes sense to add a logging statement to help debug if needed?

mssql_python/auth.py
```diff
@@ -89,20 +108,22 @@ def _acquire_token(auth_type: str) -> Tuple[bytes, str]:
             credential_class.__name__,
         )
 
+        kwargs = credential_kwargs or {}
+        cache_key = _credential_cache_key(auth_type, kwargs)
         try:
             with _credential_cache_lock:
-                if auth_type not in _credential_cache:
+                if cache_key not in _credential_cache:
```
saurabh500, line 118: There is no pruning of credential cache. I don't think its needed right now, but I would encourage adding a comment about it. 

This case is not usually possible but in some crazy scenario where the same app is serving requests for Multi-tenants (I don't think ODBC supports multi tenant Managed Identity yet).

mssql_python/auth.py
bewithgaurav: yes you're correct - the fallback to SAMI is expected and as per the ODBC behaviour
added logger statements for better debugging help, thanks for the suggestion!

mssql_python/auth.py
bewithgaurav: added a comment to reference this just above the _credential_cache declaration

mssql_python/auth.py
```diff
@@ -130,6 +154,28 @@ def _acquire_token(auth_type: str) -> Tuple[bytes, str]:
...
+    ``hello=world`` (no surrounding braces, no false split on the inner
+    ``=``), and a semicolon inside a legitimate braced value (e.g.
+    ``Database={foo;uid=victim;bar}``) cannot spoof a top-level ``UID=``.
+    """
+    # Connection.__init__ already parsed the same string through
+    # _ConnectionStringParser via _construct_connection_string, so by the
+    # time we get here the input is guaranteed parseable. No defensive
+    # try/except: a parse failure now means a real bug upstream and should
+    # propagate, not silently degrade user-assigned MSI to system-assigned.
+    parsed = _ConnectionStringParser(validate_keywords=False)._parse(connection_string)
```
saurabh500, line 174: Functionally correct but a tiny perf hit due to double parsing of connections string. 
Perfect solution would require parsing done earlier in the connect api and passing around the map to this function. That's a bigger change. I recommend a follow up PR and tracking with a GH issue. 

An adhoc fix is to maintain a hashmap of connection string and uid. But that's prone to other problems esp concurrency.

saurabh500 (dismissed):

mssql_python/connection.py
```diff
@@ -298,6 +304,7 @@ def __init__(
             # On Windows Interactive, process_connection_string returns None
             # (DDBC handles auth natively), so fall back to the connection string.
             self._auth_type = connection_result[2] or extract_auth_type(self.connection_str)
+            self._credential_kwargs = connection_result[3]
```
jahnvi480, line 344: `process_connection_string` changed from 3-tuple to 4-tuple return. This is internal so it's fine, but worth noting in the docstring that callers must unpack all 4. If a third-party or test is doing `a, b, c = process_connection_string(...)`, they'll get a `ValueError` with no obvious cause. The existing tests were all updated - just flagging for awareness.

jahnvi480 (changes requested):

mssql_python/connection.py
bewithgaurav: good call. the 4-tuple is interim. I have filed #580 to refactor process_connection_string to take the parsed connection-string map instead of the raw string (also avoids the 3x parse on the connect path). once that lands, the 4-tuple goes away and callers move to dict access.

for now, added a short docstring note about the 4-tuple to cover anyone reaching in before #580 lands.

saurabh500 (approved):

jahnvi480 (approved):

Issue #553, opened 2026-05-03
VARCHAR non-ascii character parsing


benmatwil:
### Describe the bug
Found mssql-python recently and started testing to see about usage in our team to replace pyodbc. Generally worked well and enjoyed the arrow functionality to quickly get data to pandas.

However most of our DB tables use non-unicode/VARCHAR string column types with similar collations to that in the example. This means the data coming back from the DB was coming back with incorrect unicode python strings e.g. from `.fetchall()` and the `.arrow().to_pandas()` method would fail as the strings were invalid unicode - see MWE.

Using `.setdecoding` would fix these string issues. Although no sure if that then plays other havoc elsewhere. Converting to nvarchar in the query also fixes the strings.

No conversion seems to be necessary from pyodbc so wondered if this is an issue or just a different workflow in this library.

### To reproduce
```python
import mssql_python

SERVER = ...
DATABASE = ...


connection_string = ";".join(
    [
        f"Server={SERVER}",
        f"Database={DATABASE}",
        "Authentication=ActiveDirectoryIntegrated",
        "TrustServerCertificate=yes",
    ]
)


conn = mssql_python.connect(connection_string)

cursor = conn.cursor()

# Create a temp table on a database with Latin1_General_CI_AS collation.
# The £ sign is 0xA3 in Windows-1252 — valid in Latin1 collations,
# but 0xA3 is not a valid single-byte UTF-8 sequence.
cursor.execute("""
    CREATE TABLE #test_encoding (
        id   INT,
        name VARCHAR(100) COLLATE Latin1_General_CI_AS
    )
""")
cursor.execute("INSERT INTO #test_encoding VALUES (1, 'Price: £100')")
cursor.execute("INSERT INTO #test_encoding VALUES (2, 'Résumé')")
# ASCII only, works fine
cursor.execute("INSERT INTO #test_encoding VALUES (3, 'Hello')")
conn.commit()

# --- fetchall() ---
cursor.execute("SELECT id, name FROM #test_encoding")
rows = cursor.fetchall()
print("fetchall() result:", rows)
# Expected: [(1, 'Price: £100'), (2, 'Résumé'), (3, 'Hello')]

conn.setdecoding(mssql_python.SQL_CHAR, encoding="windows-1252")

# --- fetchall() works correctly after setting decoding ---
cursor.execute("SELECT id, name FROM #test_encoding")
rows = cursor.fetchall()
print("fetchall() result:", rows)

[...]

PR #575, opened 2026-05-13
FIX: Request SQL_CHAR as SQL_C_WCHAR in arrow fetch path

ffelixg:

> [AB#44922](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/44922)

> GitHub Issue: #553

### Summary   
  
 
Due to #495, we can now request SQL_CHAR data as SQL_C_WCHAR, i.e. utf16le strings. Doing this for the arrow path ensures that arrow methods always return correct data no matter the encoding settings / locale / operating system. There does not seem to be any significant negative performance impact.


tests/test_004_cursor_arrow.py
```diff
@@ -313,6 +313,32 @@ def test_arrow_long_string(cursor: mssql_python.Cursor):
...
+        "",
+        None,
+    ]
+
+    cursor.execute(f"create table {table} (id int primary key, v varchar(64) collate {collation})")
+
+    try:
+        for index, value in enumerate(expected, start=1):
+            cursor.execute(f"insert into {table} (id, v) values (?, ?)", index, value)
+        tbl = cursor.execute(f"select v from {table} order by id").arrow()
```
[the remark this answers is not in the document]
ffelixg: Fixed, added an additional mandatory cp1252 test too

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4841,16 +4840,7 @@ SQLRETURN FetchArrowBatch_wrap(SqlHandlePtr StatementHandle, py::list& capsules,
...
-                            ret = GetDataVar(hStmt, idxCol + 1, SQL_C_CHAR,
-                                             buffers.charBuffers[idxCol],
-                                             buffers.indicators[idxCol].data());
-                            if (!SQL_SUCCEEDED(ret)) {
-                                LOG("Error fetching CHAR LOB for column %d", idxCol + 1);
-                                return ret;
-                            }
-                            break;
-                        }
+                        case SQL_LONGVARCHAR:
```
gargsaumya: Can we add a comment explaining these now fall through to WCHAR handling because SQL_CHAR is bound as SQL_C_WCHAR at line 4784.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -5093,24 +5083,7 @@ SQLRETURN FetchArrowBatch_wrap(SqlHandlePtr StatementHandle, py::list& capsules,
...
-                            target_vec->resize(target_vec->size() * 2);
-                        }
-
-                        std::memcpy(&(*target_vec)[start],
-                                    &buffers.charBuffers[idxCol][idxRowSql * fetchBufferSize],
-                                    dataLen);
-                        arrowColumnProducer->varVal[idxRowArrow + 1] = start + dataLen;
-                        break;
-                    }
+                    case SQL_LONGVARCHAR:
```
gargsaumya: similarly here.

tests/test_004_cursor_arrow.py
```diff
@@ -313,6 +313,63 @@ def test_arrow_long_string(cursor: mssql_python.Cursor):
...
+        assert tbl.column(0).type.equals(pa.large_string())
+        assert tbl.column(0).to_pylist() == expected
+    finally:
+        cursor.execute(f"drop table if exists {table}")
+
```
gargsaumya, line 391: The fix applies to SQL_CHAR, SQL_VARCHAR, and SQL_LONGVARCHAR, but only VARCHAR is tested. Can we add a test for CHAR (fixed-length) type.

tests/test_004_cursor_arrow.py
```diff
@@ -313,6 +313,63 @@ def test_arrow_long_string(cursor: mssql_python.Cursor):
...
+
+def test_arrow_varchar_utf8_collation_cp1252(cursor: mssql_python.Cursor):
```
gargsaumya: The test uses SQL_Latin1_General_CP1_CI_AS (CP1252), NOT a UTF-8 collation. The current name is slightly misleading, would it be better to name it as test_arrow_varchar_cp1252_collation_unicode?

ffelixg: Oh yeah that makes much more sense

tests/test_004_cursor_arrow.py
ffelixg: Yes, I included "text" type with cp1252 as well

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: Sure, I added the comment

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4780,9 +4780,8 @@ SQLRETURN FetchArrowBatch_wrap(SqlHandlePtr StatementHandle, py::list& capsules,
     ColumnBuffers buffers(numCols, fetchSize);
 
     if (!hasLobColumns && fetchSize > 0) {
-        // Bind columns — Arrow always uses SQL_C_CHAR for VARCHAR because
-        // it processes raw byte buffers directly, not via Python codecs.
-        ret = SQLBindColums(hStmt, buffers, columnNames, numCols, fetchSize, SQL_C_CHAR);
+        // Always request WCHARs so we don't have to deal with CHAR encodings
+        ret = SQLBindColums(hStmt, buffers, columnNames, numCols, fetchSize, SQL_C_WCHAR);
```
subrata-ms: @ffelixg , I think we should avoid hardcoding SQL_C_WCHAR here. With the recent design update introduced in PR #495(https://github.com/microsoft/mssql-python/pull/495) for CP1252 character set handling, we’ve moved toward a more flexible approach. It would be good to align with that design for Arrow support as well to ensure consistency and maintainability.

subrata-ms (changes requested):

mssql_python/pybind/ddbc_bindings.cpp
ffelixg: @subrata-ms  I've pushed a commit that enables fetching narrow chars on Linux/MacOS if configured by the user. That brings Linux/MacOS behavior in line with the other fetch paths, which also substitute any narrow encoding with utf-8.
Note that the current implementation even for the regular python fetch path isn't perfect, as the following example silently corrupts data on Linux even though everything is configured as cp1252:

```py
def test_locale_varchar_decode_iso885915():
    import locale
    assert locale.getlocale() == ('en_US', 'UTF-8')

    # change locale BEFORE connecting
    locale.setlocale(locale.LC_ALL, 'en_US.iso885915')
    assert locale.getlocale()[1] == 'ISO8859-15'
    connection = connect()
    connection.setdecoding(mssql_python.SQL_CHAR, 'ISO8859-15')
    cursor = connection.cursor()

    (val,) = cursor.execute("SELECT '€'").fetchone()
    # without changing locale, we would get val == '€'
    assert val == b'\xa4', val
    assert val.decode('ISO8859-15') == '€', val

    cursor.close()
    connection.close()
```

The fact that configuring utf-8 + SQL_C_CHAR on Windows is interpreted as utf-16 + SQL_C_WCHAR is also a bit strange, albeit understandable. To fetch utf-8 correctly on Windows the reinterpretation as utf-16 should not exist, instead the user would have to set the locale to utf-8.

But to be honest, as a user I don't enjoy thinking about those kinds of configurations anyway, the driver should just figure out how it can fetch data without corruption.

Anyway, decoding behavior should now mirror python fetching for everything except when narrow data is configured to some non-Unicode encoding on Windows. Ontop of what I mentioned above, another reason to just fetch wchars in that case is that simdutf doesn't support much aside from Unicode. Therefore we may not even see performance benefits from decoding cp1252 over utf-16 for example. Combine that with the fact that most users won't want to touch this kind of configuration, it seems hard to justify the added maintenance burden.

Let me know if you thought of some alternative way to address this internally, I can't actually read the Azure DevOps ticket.

subrata-ms: Hi @ffelixg, The code looks good now. I understand your concern, and I’d like to share that we have a future plan to leverage streaming data types (e.g., string) to handle CHAR, WCHAR, and similar cases. For now, our priority is to maintain a consistent design and code flow for CP1252 encoding and decoding logic across all supported OS platforms. This consistency will help improve maintainability and enable us to more effectively address any related bugs going forward.

subrata-ms (approved):

gargsaumya (approved):

PR #576, opened 2026-05-13
FEAT: Add ActiveDirectoryServicePrincipal support for bulk copy

bewithgaurav:

> [AB#45137](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45137)

> GitHub Issue: #534

### Summary

Adds `Authentication=ActiveDirectoryServicePrincipal` support for bulk copy. Partial fix for #534. Addresses the exact error in the issue report (`"Authentication method 'ActiveDirectoryServicePrincipal' is not supported. No token provider was registered for this method."`).

### Why a callback rather than pre-acquire

The other Entra ID modes already supported in bulk copy (Default, DeviceCode, Interactive) pre-acquire a JWT in Python and pass it as `access_token`. **ServicePrincipal can't work that way:**

- `ClientSecretCredential(tenant_id, client_id, client_secret)` requires `tenant_id` as a constructor argument.
- The connection string doesn't carry tenant_id; azure-identity doesn't discover it from anywhere.
- The only place tenant is available client-side is the STS URL the server hands back during the FedAuth pre-login (this is what ODBC does internally; that's why ODBC has never needed `tenant_id` on the connection string).

So bulk copy must **register a callback** that's invoked mid-handshake. mssql-py-core hands the callback `(spn, sts_url, auth_method)` and we parse the tenant from the STS URL on every call.

### Flow

```
mssql-python                              mssql-py-core (Rust)              mssql-tds (Rust)

cursor.bulkcopy()
  parses conn string (UID/PWD)
  ServicePrincipalAuth
    .make_token_factory(uid, pwd)
  returns _factory closure

  pycore_context = {
      "authentication": "AD-SP",
      "user_name":  client_id,
      "password":   client_secret,
      "entra_id_token_factory": _factory,
      ...
  }
                                          PyCoreConnection(ctx)
                                            validate_auth / transform_auth
                                            register factory in
                                              auth_method_map[SP]
                                                                                TDS pre-login, open TCP/TLS
                                                                                server: FedAuthInfo
                                                                                  (spn, sts_url)
                                            context.entra_id_
                                              token_factory()
                                            factory.create_token invoked
  _factory(spn, sts_url, method)        <-- (spn, sts_url, m)
    _parse_tenant_id(sts_url)
    ClientSecretCredential(
        tenant_id, client_id,
        client_secret)
    .get_token(scope=spn + "/.default")
    return jwt.encode("utf-16-le")
                                                                                FedAuth token packet
                                                                                SQL Server validates
```

### Test

```
$ python -m pytest tests/test_008_auth.py -q
65 passed in 0.77s
```

27 new tests cover:

- `TestAuthType::test_auth_type_constants` extended for `SERVICE_PRINCIPAL`
- `TestProcessAuthParameters::test_service_principal_auth_leaves_odbc_path_alone`, `test_service_principal_auth_case_insensitive`
- `TestExtractAuthType::test_serviceprincipal`
- `TestParseTenantId`: GUID, GUID without trailing slash, domain, query string, extra path segments, empty string, no path
- `TestServicePrincipalAuth`: factory returns callable, requires non-empty `client_id` and `client_secret`, returns UTF-16LE bytes, forwards credentials to `ClientSecretCredential`, builds scope from SPN (and keeps `/.default` suffix when already present), errors on unparseable STS URL, propagates `ClientAuthenticationError` as `RuntimeError`

### Connection string example

```python
import mssql_python

conn = mssql_python.connect(
    "Server=tcp:myserver.database.windows.net;"
    "Database=mydb;"
    "Authentication=ActiveDirectoryServicePrincipal;"
    "UID=<client-id-guid>;"
    "PWD=<client-secret>;"
    "Encrypt=yes;"
)
cur = conn.cursor()
cur.bulkcopy("mytable", [(1, "a"), (2, "b")])  # uses the callback under the hood
```

Same connection string keeps working for regular queries (ODBC handles ServicePrincipal natively in the non-bulk-copy path).

### Out of scope

- `ActiveDirectoryPassword` needs the same callback shape but a different azure-identity credential class. Separate PR; py-core still rejects it today with a clear error.
- `ActiveDirectoryIntegrated` needs SSPI/Kerberos wiring at the Rust layer, not a Python callback. Separate work.

### Dependencies

- Requires `mssql-py-core` 0.1.5+. This Python PR will fail at import or runtime against earlier py-core versions because the `entra_id_token_factory` dict key is silently ignored there. Will update the dependency pin once a new mssql-py-core wheel is published.


mssql_python/auth.py
```diff
@@ -171,6 +173,143 @@ def _acquire_token(
...
+    """
+
+    @staticmethod
+    def make_token_factory(client_id: str, client_secret: str):
+        """Return a callable suitable for ``entra_id_token_factory``.
+
+        Signature: ``(spn: str, sts_url: str, auth_method: str) -> bytes``.
+        Returns the JWT encoded as UTF-16LE bytes (the TDS FedAuth wire format).
+
+        ``ClientSecretCredential`` instances are reused across calls via the
```
jahnvi480, line 235: @bewithgaurav , quick question on the SP auth scope.

The PR currently handles ClientSecretCredential (tenant_id + client_id + client_secret). But ActiveDirectoryServicePrincipal in the broader Azure ecosystem also supports certificate-based auth (CertificateCredential in azure-identity), which takes a certificate path/PEM instead of a client secret.

Two questions:
- Is certificate-based SP auth planned as a follow-up, or out of scope? Some customers use certificates over secrets since they're harder to leak, if they set Authentication=ActiveDirectoryServicePrincipal without PWD (because they expect cert-based flow), the current error says "requires UID (client_id) and PWD (client_secret)" which doesn't tell them certificates aren't supported yet.
- If it's out of scope for now, would it make sense to update the error message to explicitly say certificate-based SP is not yet supported? Something like: "ActiveDirectoryServicePrincipal currently supports client secret only. Certificate-based authentication is not yet supported. Provide UID (client_id) and PWD (client_secret)." -that way users who rely on certificates get a clear signal instead of thinking they misconfigured something.

Not a blocker for the PR, just want to make sure we're not leaving a confusing gap for cert-based SP users.

bewithgaurav: thanks @jahnvi480

cert-based SP is out of scope for this PR. 

- The connection-string surface today (UID+PWD) mirrors msodbcsql, and the natural place for native cert support is mssql-rs (`TdsAuthenticationMethod::ActiveDirectoryServicePrincipal`).

- Alternative: env vars + `Authentication=ActiveDirectoryDefault` works for bulkcopy too via the existing codepath. I will update the secret-missing error in cursor.py to point cert users there so they don't end up debugging a missing PWD.

mssql_python/auth.py
```diff
@@ -171,6 +174,181 @@ def _acquire_token(
...
+        works across handshake retries, reconnects, and separate bulkcopy
+        invocations using the same identity.
+        """
+        if not client_id:
+            raise ValueError("ServicePrincipal auth requires a non-empty client_id (UID)")
+        if not client_secret:
+            raise ValueError("ServicePrincipal auth requires a non-empty client_secret (PWD)")
+
+        def _factory(spn: str, sts_url: str, auth_method: str) -> bytes:
+            # pylint: disable=import-outside-toplevel,unused-argument
```
jahnvi480: `auth_method` is silenced via `unused-argument`. Are we expecting it to be used in furture?

mssql_python/auth.py
```diff
@@ -171,6 +174,181 @@ def _acquire_token(
...
+                        # Tracked as a follow-up; the fix is to derive
+                        # ``authority`` from ``urlparse(sts_url).netloc``.
+                        # Out of scope for the initial #534 work.
+                        credential = ClientSecretCredential(
+                            tenant_id=tenant_id,
+                            client_id=client_id,
+                            client_secret=client_secret,
+                            transport=transport,
+                        )
+                        _credential_cache[cache_key] = credential
```
jahnvi480, line 334: `_credential_cache` is never evicted. After a client_secret rotation, a fresh entry is added under the new `secret_hash`, but the old `ClientSecretCredential` stays in the dict for the life of the process - holding the old secret in memory plus its internal token cache. For long-lived processes that rotate secrets periodically this is a slow leak.

Consider evicting any prior entries matching `(serviceprincipal, tenant_id, client_id, *)` when inserting a new `secret_hash`, or documenting the long-running-process caveat. Not blocking.

mssql_python/auth.py
```diff
@@ -171,6 +174,181 @@ def _acquire_token(
...
+                # surfaced message is intentionally generic so that any
+                # secret-bearing provider text never reaches the user-facing
+                # exception chain.
+                logger.error(
+                    "ServicePrincipal authentication failed: tenant=%s, error=%s",
+                    tenant_id,
+                    str(e),
+                )
+                raise RuntimeError(
+                    "ServicePrincipal authentication failed; " "see debug logs for provider details"
```
jahnvi480: message says "see debug logs for provider details" but the provider error is logged at `logger.error(...)` a few lines up, not `debug`. Either drop the log to `debug` or update the message to "see error logs" so the user knows where to actually look.

mssql_python/auth.py
```diff
@@ -171,6 +174,181 @@ def _acquire_token(
...
+                # as the other AD methods.
+                #
+                # The cache key includes a hash of client_secret so a rotated
+                # secret produces a different cache entry. Without this, an
+                # external secret rotation would not invalidate the cached
+                # ClientSecretCredential: azure-identity's internal token cache
+                # would keep returning the previously-issued token (good for
+                # up to ~1 hour) until expiry, masking the rotation. Hashing
+                # avoids storing the raw secret in the dict key.
+                secret_hash = hashlib.sha256(client_secret.encode("utf-8")).hexdigest()
```
jahnvi480: `secret_hash` is recomputed on every factory invocation, but `client_secret` is fixed for the lifetime of the closure returned by `make_token_factory`. Compute the hash once in the outer scope and capture it:

```python
secret_hash = hashlib.sha256(client_secret.encode("utf-8")).hexdigest()
def _factory(spn, sts_url, auth_method):
    ...
    cache_key = _credential_cache_key(_AuthInternal.SERVICE_PRINCIPAL,
        {"tenant_id": tenant_id, "client_id": client_id, "secret_hash": secret_hash})
```

Save the SHA-256 per call (tens of μs each) and remove a redundant import of `hashlib` from the hot path.

mssql_python/auth.py
```diff
@@ -171,6 +174,181 @@ def _acquire_token(
...
+                        )
+                        _credential_cache[cache_key] = credential
+                # mssql-tds passes the resource SPN; azure-identity wants a scope.
+                scope = spn if spn.endswith("/.default") else spn.rstrip("/") + "/.default"
+                token = credential.get_token(scope).token
+                logger.info(
+                    "ServicePrincipal token factory: token acquired, length=%d chars",
+                    len(token),
+                )
+                return token.encode("utf-16-le")
```
jahnvi480, line 345: `AADAuth.get_token_struct` (used for the `access_token` path) returns a `<I` length-prefixed UTF-16LE struct. This factory returns *bare* `token.encode("utf-16-le")` with no length prefix.

one line in the docstring like *"py-core handles the FedAuth length-prefix wrapping; do not pre-wrap"* so a future reader doesn't try to "fix" this to match `get_token_struct`.

jahnvi480 (dismissed):
PR lgtm!
I have just posted few questions and wanted clarification and understanding on others.
Comments are non-blocking, hence approving the PR.

mssql_python/auth.py
bewithgaurav: yeah I had set three-arg contract in the Rust EntraIdTokenFactory trait in mssql-tds since I wanted to keep it generic across all auth methods
the pylint workaround is not nice, so I've done an underscore prefix so that pylint stops complaining

mssql_python/auth.py
bewithgaurav: makes sense, real concern for long-lived processes
thanks for catching - I've fixed this in my latest commit

mssql_python/auth.py
bewithgaurav: corrected to "error logs" thing

mssql_python/auth.py
bewithgaurav: hoisted to make_token_factory outer scope - thanks for catching this

mssql_python/auth.py
bewithgaurav: added a comment above the return clarifying that py-core handles the FedAuth length-prefix wrapping, so we return bare JWT bytes here - this would be helpful for future readers.

mssql_python/auth.py
```diff
@@ -171,6 +174,197 @@ def _acquire_token(
...
+                raise RuntimeError(
+                    "ServicePrincipal token factory: empty SPN from server "
+                    "(cannot construct token scope)"
+                )
+            tenant_id = _parse_tenant_id(sts_url)
+            if not tenant_id:
+                raise RuntimeError(f"Could not extract tenant_id from STS URL: {sts_url!r}")
+
+            logger.info(
+                "ServicePrincipal token factory: acquiring token for tenant=%s, spn=%s",
```
sumitmsft: Is it intentional to log tenant id and spn in the logs? Can we redact this?

mssql_python/auth.py
```diff
@@ -171,6 +174,197 @@ def _acquire_token(
...
+                        # mssql-py-core blocking-pool worker (tokio
+                        # spawn_blocking), so a stuck callback ties that
+                        # worker up for the duration. SP is non-interactive
+                        # and token issuance is typically <1s; 10s/15s is
+                        # generous and still bounded.
+                        transport = RequestsTransport(
+                            connection_timeout=10,
+                            read_timeout=15,
+                        )
+                        # KNOWN LIMITATION: ``authority=`` is not passed,
```
sumitmsft, line 318: Do we plan to allow SPN to be used in Sovereign clouds? Any timeline? 
Also, please plan to document this limitation in our public docs.

sumitmsft (dismissed):
approved with some minor comments.

mssql_python/auth.py
bewithgaurav: yes we can, had intentionally kept it since I thought neither technically is a secret or confidential
either ways, it doesn't add much value actually - im fine with eliminating this one, committed the change

mssql_python/auth.py
bewithgaurav: yes, as of now - sovereign cloud support is blocked on native token acquisition in mssql-rs
today all auth methods except MSI and Default are `todo!()` in the mssql-rs layer, so we shim through Python (callback for SP, pre-acquire for the rest)
the shim defaults to public cloud because `azure-identity` credential constructors need an explicit `authority=` that we don't pass.

the longer term solution is implementing the auth methods natively in mssql-rs (same pattern msodbcsql uses internally: receive the STS URL from the server during FedAuth, post the OAuth grant directly to that URL). once we do that, sovereign clouds work inherently for all methods because we'd just hit whatever endpoint the server tells us, exactly like ODBC does today

will be adding this in backlog & document the limitation in public docs for bcp in the meantime

sumitmsft (approved):

jahnvi480 (approved):

Issue #582, opened 2026-05-14
Row does not support string-key indexing


dlevy-msft-sql:
## Summary

In mssql-python 1.6.0, `Row` no longer supports string-key indexing (`row["ColumnName"]`). This worked in 1.5.0 and is not called out as a breaking change in the [v1.6.0 release notes](https://github.com/microsoft/mssql-python/releases/tag/v1.6.0).

## Repro

```python
import mssql_python

conn = mssql_python.connect("<your conn str>")
cur = conn.cursor()
cur.execute("SELECT 1 AS ProductID, 'foo' AS Name")
row = cur.fetchone()

print(row[0])           # works
print(row.ProductID)    # works (attribute access)
print(row["ProductID"]) # TypeError on 1.6.0; worked on 1.5.0
```

### Output on 1.6.0

```
1
1
Traceback (most recent call last):
  File "probe.py", line N, in <module>
    print(row["ProductID"])
          ~~~^^^^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
```

## Versions

- mssql-python 1.6.0 (verified via `pip show mssql-python`)
- Python 3.14 on Windows
- Server: Fabric SQL endpoint (not relevant; reproes with any `SELECT ... AS Name`)

The same script returns `1` on 1.5.0.

## Why this matters

1. **Silent regression.** The exception message (`list indices must be integers or slices`) reads like a Python typo, not a deliberate API change. Users will spend time debugging their query and column aliases before they suspect the driver.
2. **Not in release notes.** The v1.6.0 notes list 1 enhancement and 6 bug fixes, none of which mention Row access semantics or list any "Breaking Changes" section. v0.14.0 set the precedent that breaking changes get their own section and migration notes — v1.6.0 should follow that convention.
3. **Public documentation uses this pattern.** The official mssql-python quickstart on Microsoft Learn currently shows `print(f"Inserted Product ID : {result['ProductID']}")` ([source](https://github.com/MicrosoftDocs/sql-docs-pr/blob/main/docs/connect/python/mssql-python/python-sql-driver-mssql-python-quickstart.md)). Anyone following the quickstart on 1.6.0 hits this on the last code block. (A docs fix to use attribute access is already staged in our v1.6.0 doc PR, but the broader user base relying on `row["col"]` in their own code is the real concern.)
[...]

PR #589, opened 2026-05-19
FEAT: Row string-key indexing 

jahnvi480:

> [AB#45119](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45119)

> GitHub Issue: #582 

### Summary   
This pull request enhances the usability of the `Row` class by allowing access to row values using column names as string keys (e.g., `row["col"]`), in addition to existing integer index and attribute access. It also introduces case-insensitive string-key access when the cursor's `lowercase` attribute is enabled. Comprehensive tests have been added to ensure correct behavior for these new access patterns.

### Enhancements to Row Access Patterns

* Updated the `Row.__getitem__` method in `mssql_python/row.py` to support accessing values by column name as a string key, including case-insensitive lookup when the cursor's `lowercase` attribute is set.

### Testing Additions

* Added `test_row_string_key_indexing` and `test_row_string_key_case_insensitive_with_lowercase` to `tests/test_001_globals.py` to verify string-key and case-insensitive access, as well as backward compatibility with integer and slice access.
* Added `test_row_string_key_indexing` to `tests/test_004_cursor.py` to test string-key access on real query results, ensuring consistency with index and attribute access and correct error handling for missing keys.


tests/test_004_cursor.py
```diff
@@ -2876,6 +2876,41 @@ def test_row_attribute_access(cursor, db_connection):
...
+
+        # Non-existent key raises KeyError
+        with pytest.raises(KeyError):
+            row["nonexistent"]
+
+    except Exception as e:
+        pytest.fail(f"Row string-key indexing test failed: {e}")
+    finally:
+        cursor.execute("DROP TABLE #pytest_row_strkey")
+        db_connection.commit()
```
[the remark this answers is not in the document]
jahnvi480: Fixed this

mssql_python/row.py
```diff
@@ -156,8 +156,18 @@ def _apply_output_converters_optimized(self, values, converter_map):
...
+        if isinstance(index, str):
+            if index in self._column_map:
+                return self._values[self._column_map[index]]
+            # Case-insensitive lookup when lowercase is enabled
+            if hasattr(self._cursor, "lowercase") and self._cursor.lowercase:
+                index_lower = index.lower()
+                for col_name in self._column_map:
+                    if col_name.lower() == index_lower:
+                        return self._values[self._column_map[col_name]]
+            raise KeyError(f"Row has no column '{index}'")
```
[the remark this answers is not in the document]
jahnvi480: Fixed this

tests/test_001_globals.py
```diff
@@ -996,3 +996,53 @@ def test_stringify_uuids_with_tuple_values():
...
+        [1, "bar"],
+        {"productid": 0, "name": 1},
+        cursor=mock_cursor,
+    )
+
+    # Exact match
+    assert row["productid"] == 1
+    # Case-insensitive match
+    assert row["ProductID"] == 1
+    assert row["NAME"] == "bar"
```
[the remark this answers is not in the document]
jahnvi480: Fixed this.

mssql_python/row.py
```diff
@@ -156,8 +156,18 @@ def _apply_output_converters_optimized(self, values, converter_map):
...
-        """Allow accessing by numeric index: row[0]"""
+    def __getitem__(self, index) -> Any:
+        """Allow accessing by numeric index (row[0]) or column name (row["col"])."""
+        if isinstance(index, str):
+            if index in self._column_map:
+                return self._values[self._column_map[index]]
+            # Case-insensitive lookup when lowercase is enabled
+            if get_settings().lowercase:
+                index_lower = index.lower()
+                for col_name in self._column_map:
```
sumitmsft: The current implementation does a linear scan with .lower() on every column name each time a case-mismatched string key is used. In latency-sensitive applications iterating over large result sets (e.g., 1M rows with wide tables), this creates repeated O(columns) scans with string allocations on every row access. For a 200-column table over 1M rows, that's ~200M .lower() calls and comparisons which potentially adding seconds of overhead.

Same issue exists in __getattr__ () as well. 

You can build a lowercase original-key lookup map once, so both __getitem__ and __getattr__ get O(1) case-insensitive access.

mssql_python/row.py
```diff
@@ -156,8 +156,18 @@ def _apply_output_converters_optimized(self, values, converter_map):
...
+                    if col_name.lower() == index_lower:
+                        return self._values[self._column_map[col_name]]
+            raise KeyError(f"Row has no column '{index}'")
```
sumitmsft, line 182: Currently, any index type that isn't str falls through to self._values[index]. If a user accidentally passes an unsupported type (e.g., row[3.5], row[None]), they get a confusing error from Python's list internals, rather than a helpful message from the Row API itself.

Add an explicit type guard so Row clearly communicates its supported index types:

Something like this:
```python
raise TypeError(
        f"Row indices must be integers, slices, or strings, not {type(index).__name__}"
```

sumitmsft (changes requested):

mssql_python/row.py
jahnvi480: Makes sense, I have made the required changes,

mssql_python/row.py
jahnvi480: I have added explicit check and have added testcase for same.

sumitmsft (dismissed):

bewithgaurav (approved):

mssql_python/row.py
```diff
@@ -60,6 +60,9 @@ def __init__(self, values, column_map, cursor=None, converter_map=None, uuid_str
 
         self._column_map = column_map
         self._cursor = cursor
+        # Pre-built lowercase-key map case-insensitive lookups.
+        # Built once per Row, shared cost amortised across all accesses.
+        self._column_map_lower = {k.lower(): v for k, v in column_map.items()} if column_map else {}
```
bewithgaurav: this builds a new dict per Row unconditionally, even when `lowercase` is off (the default). `column_map` itself is cached once per result set in the cursor and shared across all rows, so this is the only per-row allocation the PR adds.

quick bench results on the cost:

- 5 cols: ~460 ns/row → 4.6 ms for 10k rows
- 20 cols: ~900 ns/row → 9 ms for 10k rows
- 50 cols: ~2.3 µs/row → 23 ms for 10k rows, 235 ms for 100k rows

for typical result sets that's noise, but it scales linearly with both row count and column count, and main branch pays zero here.

suggestion: build it once in the cursor alongside `_cached_column_map`, only when lowercase is on:

```python
# in cursor.py, right after _cached_column_map is built
self._cached_column_map_lower = (
    {k.lower(): v for k, v in self._cached_column_map.items()}
    if get_settings().lowercase else None
)
```

then pass it to `Row.__init__` and store the reference. same O(1) lookup at access time, but:
- **zero per-row cost** (built once per result set, shared across all rows, same as `column_map`)
- **zero cost when lowercase is off** (stays `None`, no dict allocated at all)

bewithgaurav (changes requested):
on a 2nd look, found a perf problem with column map (lowercase), can benefit from caching
requesting changes for the same

mssql_python/row.py
jahnvi480: Made the required changes

tests/test_001_globals.py
```diff
@@ -996,3 +996,66 @@ def test_stringify_uuids_with_tuple_values():
     assert row[2] == "hello"
     # Internal storage should now be a list (converted from tuple)
     assert isinstance(row._values, list)
+
+
+def test_row_string_key_indexing():
```
bewithgaurav, line 1001: nit: suggesting a test for the real cursor-built lowercase path end-to-end. the unit test covers Row with an injected `column_map_lower`, and this integration test covers string-key access with `lowercase=False`. what's missing is mixed-case DB columns + `lowercase=True` through an actual cursor. `test_lowercase_attribute` already has that exact table setup but only checks `description` column

rough shape for a test which can be e2e:
- create table with mixed-case columns (`ID INT, UserName VARCHAR(50)`)
- insert a row
- set `mssql_python.lowercase = True`, open a fresh cursor, execute SELECT
- verify string-key access: `row["username"]` and `row["USERNAME"]` both return the value
- verify attribute access: `row.username` and `row.USERNAME` both work
- verify exact-case still works: `row["id"]` (lowercase, since that's what the cursor emits when lowercase is on)
- verify missing key still raises `KeyError`
- restore `mssql_python.lowercase` in `finally`

bewithgaurav (approved):
just added a test suggestion for e2e scenario - non-blocker
diff code coverage covers unit tests and is already at 100% so can skip
PR lgtm

sumitmsft (approved):

gargsaumya (approved):

Issue #580, opened 2026-05-14
Avoid parsing the connection string multiple times in the auth path


bewithgaurav:
## What

The connection string gets parsed by `_ConnectionStringParser` more than once on every connect when Entra ID auth is in play.

For one Entra ID connect today:

1. `Connection.__init__` calls `_construct_connection_string`, which parses once.
2. `process_connection_string` (in `auth.py`) splits the string again to pick out auth params.
3. For MSI with UID, `_extract_msi_client_id` parses it a third time via `_ConnectionStringParser`.

Same string, three parses. Small per-connect cost, but it's wasted work and it's also why `auth.py` keeps reimplementing parsing instead of trusting the canonical parser.

## Why fix it

- Removes redundant work on the connect path.
- Lets `auth.py` stop hand-rolling its own parsing (the naive `split(";")` in `process_connection_string` and `extract_auth_type` only exists because the parsed map isn't available at that point).
- Makes the bracing-value bug ([separate issue](TBD link)) trivial to fix as a side effect.

## Suggested approach

1. `Connection.__init__` already produces a parsed map via `_construct_connection_string`. Hold on to it (or re-run the parser once and pass the dict around).
2. Change `process_connection_string` (and `extract_auth_type`, and `_extract_msi_client_id`) to take the parsed dict instead of the raw string. Read keys from the dict directly.
3. Build the sanitized output via `_ConnectionStringBuilder` from the same dict (minus the sensitive keys).

Same call shape from the outside, one parse instead of three, and `auth.py` stops carrying its own mini-parser.

## Workaround mentioned in the review

A connection-string → uid hashmap was suggested as an ad-hoc fix. Skipping it because it has its own concurrency and lifecycle problems and doesn't address the deeper "auth.py shouldn't reparse" issue.

## Background

Surfaced during PR #573 review.

PR #590, opened 2026-05-19
FIX: Avoid parsing connection string multiple times in auth path (#580)

jahnvi480:

> [AB#45135](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45135)

> GitHub Issue: #580 

### Summary   
This pull request significantly refactors the authentication handling and connection string construction in the `mssql_python` package. The main improvements include migrating from string-based to dictionary-based parameter handling, consolidating authentication logic, and simplifying sensitive parameter removal. These changes modernize the codebase, reduce redundant parsing, and make it easier to extend or maintain authentication features.

**Authentication and connection string processing improvements:**

* Refactored authentication parameter extraction and processing to operate directly on parsed parameter dictionaries instead of raw connection string fragments, reducing multiple parsing passes and improving clarity (`mssql_python/auth.py`, `mssql_python/connection.py`). 
* Consolidated and simplified the removal of sensitive parameters (such as `UID`, `PWD`, `Trusted_Connection`, and `Authentication`) by using a canonical set and dictionary operations, ensuring robust sanitization before ODBC handoff (`mssql_python/auth.py`, `mssql_python/connection.py`).

**API and codebase modernization:**

* Changed `Connection._construct_connection_string` to return both the constructed string and the normalized parameter dictionary, enabling downstream logic to avoid reparsing and to use normalized parameters directly (`mssql_python/connection.py`).
* Updated tests to accommodate the new return type of `_construct_connection_string`, ensuring continued correctness and coverage (`tests/test_003_connection.py`).

**Cleanup and removal of legacy code:**

* Removed old, string-based authentication helpers and replaced them with dictionary-based logic, reducing code duplication and improving maintainability (`mssql_python/auth.py`, `mssql_python/connection.py`, `tests/test_008_auth.py`). 

Overall, these changes streamline authentication handling, reduce parsing complexity, and lay a stronger foundation for future enhancements.


sumitmsft:
PR looks good to me, didnt find any critical issue.

sumitmsft (dismissed):

mssql_python/connection.py
```diff
@@ -328,20 +335,32 @@ def __init__(
...
-        # If authentication is specified, it will be processed to handle
-        # different authentication types like interactive, device code, etc.
-        if re.search(r"authentication", self.connection_str, re.IGNORECASE):
-            connection_result = process_connection_string(self.connection_str)
-            self.connection_str = connection_result[0]
-            if connection_result[1]:
-                self._attrs_before.update(connection_result[1])
+        # Handle Entra ID authentication if specified.
+        # The parsed dict is used directly — no re-parsing of the connection string.
+        if "Authentication" in self._parsed_params:
```
bewithgaurav: the dict lookup works today because `_normalize_params` maps `"authentication"` to `"Authentication"`. but the canonical string is hardcoded as a raw `"Authentication"` in multiple places here instead of referencing the source of truth.

three spots that break silently if someone renames the canonical key in `_ALLOWED_CONNECTION_STRING_PARAMS`:

- line 340: `if "Authentication" in self._parsed_params`
- line 348: `self._parsed_params.get("UID")`
- line 363: `extract_auth_type(self._parsed_params)` (which internally does `.get("Authentication")`)

the whole point of this refactor is to stop hand-rolling string matching against connection params. using raw strings here reintroduces the same fragility, just at the dict-key level instead of the split-on-semicolon level.

ask: define short constants (or reuse existing ones from `constants.py`) for the canonical key names and reference those. something like:

```python
# in constants.py or at top of connection.py
_KEY_AUTHENTICATION = "Authentication"
_KEY_UID = "UID"
```

then:

```python
if _KEY_AUTHENTICATION in self._parsed_params:
    ...
    uid = (self._parsed_params.get(_KEY_UID) or "").strip()
```

this way the parser's normalization output and the consumer's lookup key are always in sync from one definition.

mssql_python/connection.py
```diff
@@ -328,20 +335,32 @@ def __init__(
...
+                self._credential_kwargs = credential_kwargs
+
             # Store auth type so bulkcopy() can acquire a fresh token later.
-            # On Windows Interactive, process_connection_string returns None
-            # (DDBC handles auth natively), so fall back to the connection string.
-            self._auth_type = connection_result[2] or extract_auth_type(self.connection_str)
-            self._credential_kwargs = connection_result[3]
+            # On Windows Interactive, process_auth_parameters returns None
+            # (DDBC handles auth natively), so fall back to extract_auth_type.
+            self._auth_type = auth_type or extract_auth_type(self._parsed_params)
```
bewithgaurav: `self._parsed_params` persists UID and PWD in cleartext on the Connection object for its entire lifetime. this is new. on `main` today there is no dict retaining credentials after the auth block sanitizes the connection string.

what changed:

- **`main` (before this PR):** `_construct_connection_string` returns a plain string. after the auth block rebuilds it via `process_connection_string`, UID/PWD are gone from `self.connection_str`. no other attribute holds them.
- **this PR:** `_construct_connection_string` returns `(str, dict)`. the dict is stored as `self._parsed_params` and survives the entire connection lifetime with UID/PWD intact, even though the string was sanitized.

repro on the PR branch:

```python
conn = connect(
    "Server=test;Database=testdb;"
    "Authentication=ActiveDirectoryDefault;"
    "UID=myuser;PWD=secret123"
)
conn.connection_str         # no UID, no PWD (good, sanitized)
conn._parsed_params["PWD"]  # 'secret123' -- still there
conn._parsed_params["UID"]  # 'myuser'   -- still there
```

on `main`, `hasattr(conn, "_parsed_params")` is `False`.

the dict is only read during `__init__` for auth processing. after that it is dead weight holding credentials.

fix options (pick whichever fits):

- `self._parsed_params = None` right after the auth block (around line 364)
- `del self._parsed_params` at end of `__init__`
- don't store it on `self` at all. keep it as a local in `__init__` and pass to the auth helpers directly

bewithgaurav (changes requested):
two things to address before merge, added both inline

mssql_python/connection.py
jahnvi480: This makes sense, I have addressed this comment

mssql_python/connection.py
jahnvi480: This makes sense, I have addressed this comment

gargsaumya (approved):

mssql_python/constants.py
```diff
@@ -519,6 +519,14 @@ def get_attribute_set_timing(attribute):
     "packetsize": "PacketSize",
 }
 
+# Canonical normalized key names produced by _ConnectionStringParser._normalize_params.
+# Consumer code should reference these instead of hard-coding raw strings so that
+# a rename in _ALLOWED_CONNECTION_STRING_PARAMS is caught at import time.
+_KEY_AUTHENTICATION = "Authentication"
```
bewithgaurav, line 525: nit: not blocking this PR: the new `_KEY_*` constants duplicate the canonical values already in `_ALLOWED_CONNECTION_STRING_PARAMS`:

```python
# _ALLOWED_CONNECTION_STRING_PARAMS (source of truth today)
"authentication": "Authentication",
"uid": "UID",

# new constants (separate definition, same strings)
_KEY_AUTHENTICATION = "Authentication"
_KEY_UID = "UID"
```

these can drift. the cleaner shape is one set of constants that both the dict and consumer code reference:

```python
_CONNECTION_STRING_AUTH_KEY = "Authentication"
_CONNECTION_STRING_UID_KEY = "UID"

_ALLOWED_CONNECTION_STRING_PARAMS = {
    "authentication": _CONNECTION_STRING_AUTH_KEY,
    "uid": _CONNECTION_STRING_UID_KEY,
    ...
}
```

follows the existing pattern (`_CONNECTION_STRING_DRIVER_KEY`, `_CONNECTION_STRING_APP_KEY` already work this way). could you open a small cleanup task for this?

bewithgaurav (approved):
discovered one minor tangential task, approving this PR

sumitmsft (approved):

PR #596, opened 2026-05-21
FIX: Store deferred connect-attribute values in member buffers (#594)

### LINKED ISSUE #594: SIGBUS in ddbc_bindings.Connection on 1.7.1 (macOS arm64) when authenticating with SQL_COPT_SS_ACCESS_TOKEN against Fabric DW
### Describe the bug

On mssql-python **1.7.1** (macOS 26.3.1 arm64, Python 3.13.7), calling `mssql_python.connect()` with an Azure access token supplied via `attrs_before` / `SQL_COPT_SS_ACCESS_TOKEN` (1256) against a real Microsoft Fabric Data Warehouse **crashes the host Python interpreter with `SIGBUS` (`Bus error: 10`)**. No Python exception is raised, no traceback is emitted — the process is terminated by the kernel after the TCP/TLS handshake completes and the access-token authentication exchange begins.

The same script works correctly on 1.4.0, 1.5.0, and 1.6.0 from the same machine against the same warehouse.

`faulthandler` pins the fault to the native `ddbc_bindings.Connection(...)` call:

```
Fatal Python error: Bus error

Current thread 0x00000001ef8eb100 (most recent call first):
  File ".../mssql_python/connection.py", line 374 in __init__
  File ".../mssql_python/db_connection.py", line 55 in connect
  File ".../repro_mssql_python_bus_error.py", line 65 in main

Extension modules: _cffi_backend (total: 1)
```

`connection.py:374` is the `self._conn = ddbc_bindings.Connection(self.connection_str, self._pooling, self._attrs_before)` call.

### Scope (bisect)

Tested against the same Fabric DW endpoint, same Python, same machine:

| Variant | 1.4.0 | 1.5.0 | 1.6.0 | 1.7.1 |
|---|---|---|---|---|
| `connect()` + `attrs_before` (SQL_COPT_SS_ACCESS_TOKEN) + real Fabric DW | ✅ Connected, `SELECT 1` round-trips | ✅ | ✅ | ❌ **SIGBUS** |
| `connect()` without `attrs_before` + fake server | n/a | n/a | ✅ raises `RuntimeError` | ✅ raises `OperationalError` |
| `connect()` with `attrs_before` (bogus token bytes) + fake server | n/a | n/a | ✅ raises `RuntimeError` | ✅ raises `OperationalError` |

So the crash requires **all** of: 1.7.1, a real server that completes the TLS handshake, and `attrs_before` carrying `SQL_COPT_SS_ACCESS_TOKEN`. It is not triggered by argument-parsing of `attrs_before` alone, and it is not triggered without `attrs_before` even when the server is real. Smells like the bug lives in the token-auth handshake path in the native ddbc layer.

> 1.7.0 was not testable on macOS arm64 / Python 3.13.7 because no wheel was published for that platform (see #588). I cannot confirm whether 1.7.0 has the same fault.

### To reproduce

Prereqs: a Microsoft Fabric workspace with a Data Warehouse, and `az login` against the right tenant.

```python
"""Reproducer for SIGBUS in mssql-python 1.7.1 on macOS arm64.

Usage:
    pip install mssql-python==1.7.1 azure-identity
    az login
    export FABRIC_HOST='<workspace>-<id>.datawarehouse.fabric.microsoft.com'
    export FABRIC_DB='<warehouse name>'
    python repro_mssql_python_bus_error.py
"""
from __future__ import annotations

import faulthandler
import os
import struct
import sys
from itertools import chain, repeat

faulthandler.enable()

import mssql_python
from azure.identity import AzureCliCredential

SQL_COPT_SS_ACCESS_TOKEN = 1256
SQL_SCOPE = "https://database.windows.net/.default"


def main() -> int:
    host = os.environ["FABRIC_HOST"]
    database = os.environ["FABRIC_DB"]

    print(f"mssql-python version: {mssql_python.__version__}", flush=True)
    print(f"Python: {sys.version}", flush=True)

    token = AzureCliCredential().get_token(SQL_SCOPE).token
    token_bytes = bytes(token, "utf-8")
    encoded = bytes(chain.from_iterable(zip(token_bytes, repeat(0))))
    attrs_before = {SQL_COPT_SS_ACCESS_TOKEN: struct.pack("<i", len(encoded)) + encoded}

    conn_str = ";".join([
        f"Server={host}",
        f"Database={database}",
        "Encrypt=Yes",
        "TrustServerCertificate=No",
    ])
    print("Calling mssql_python.connect() ...", flush=True)

    handle = mssql_python.connect(
        conn_str, attrs_before=attrs_before, autocommit=True, timeout=60,
    )
    print("Connected.", flush=True)

    cur = handle.cursor()
    cur.execute("SELECT 1")
    print(f"Round-trip select returned: {cur.fetchone()}", flush=True)

    cur.close()
    handle.close()
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

Expected on 1.7.1: stops after `Calling mssql_python.connect() ...`, kernel kills the process with `Bus error: 10` and a multiprocessing semaphore leak warning. No Python traceback.

Expected on 1.6.0: prints `Connected.` and `Round-trip select returned: (1,)`.

### Expected behavior

`mssql_python.connect()` with `attrs_before={SQL_COPT_SS_ACCESS_TOKEN: ...}` against Fabric DW should either succeed (as on 1.6.0) or raise a Python exception (`OperationalError`, `DatabaseError`, etc.). It must not terminate the host process.

### Further technical details

- mssql-python version: **1.7.1** (uploaded 2026-05-20T15:23, same day as the bug surfaced downstream)
- Python version: 3.13.7
- SQL Server version: Microsoft Fabric Data Warehouse (SQL endpoint at `*.datawarehouse.fabric.microsoft.com`)
- Operating system: macOS 26.3.1 arm64 (Apple Silicon)
- Auth: Azure CLI access token via `attrs_before` / `SQL_COPT_SS_ACCESS_TOKEN` (1256)
- Installer: `uv pip install` into a fresh Python 3.13.7 venv

saurabh500:
Fixes #594.

### Summary

PR #568 changed `Connection::setAttribute` to copy bytes/string attribute values into **stack-local buffers** before calling `SQLSetConnectAttr`, on the theory that the released GIL could otherwise expose a race with another thread mutating the member buffer.

That premise is unsound for *deferred* attributes and introduced a **use-after-free**.

`SQL_COPT_SS_ACCESS_TOKEN` (1256) is a deferred ODBC connect attribute: the MS driver stores the caller's pointer at `SQLSetConnectAttr` time and dereferences it later, during `SQLDriverConnect`, to build the FedAuth login packet. With the stack-local copy, by the time the driver reads the pointer the buffer has already been freed.

## Observed symptoms (issue #594)

| Platform | Symptom |
|---|---|
| macOS arm64 | `Fatal Python error: Bus error` (SIGBUS) |
| Windows x64 | `ProgrammingError: Authentication token is missing in the federated authentication message` |
| Azure SQL DB | `OperationalError: Communication link failure; TCP Provider Error code 0x2746` (server reset) |

All three are the same root cause — the driver dereferences a pointer to a freed stack frame; the symptom depends on what the OS leaves on that page.

## Fix

Store the value in the Connection-owned member buffers (`strBytesBuffer` for bytes/bytearray, `wstrStringBuffer` for str) so the memory remains valid for the lifetime of the `Connection`, which covers the deferred dereference during `SQLDriverConnect`.

PR #568's `gil_scoped_release` around `SQLSetConnectAttr` is **preserved**.

## Why the 'race with another thread' rationale from PR #568 doesn't apply

1. `attrs_before` is applied once, sequentially, during `connect()`. No other thread is mutating attributes at that moment.
2. Concurrent `setAttribute()` on the same `Connection` from different threads is not a supported pattern — ODBC connection handles aren't designed for concurrent attribute mutation.
3. 1.6.0 used the persistent member buffer with no reported issue.
4. Even granting the race premise, a stack local that is freed *before* the driver finishes using the pointer is strictly worse than a member that at least lives as long as the connection.

Comments in the patched code document both the deferred-read requirement and the non-thread-safe contract.

## Verification

Built the patched `.so` from this branch on macOS arm64, overlaid into the installed 1.7.1 wheel, and ran the issue's repro script against an Azure SQL DB with an `AzureCliCredential` access token:

| Build | Result |
|---|---|
| Stock 1.7.1 | `Fatal Python error: Bus error` (SIGBUS) / `OperationalError: ... TCP Provider Error code 0x2746` (run-to-run variance) |
| This fix | Clean `OperationalError: Login failed for user '<token-identified principal>'` surfaced from the server (token transmitted intact; server-side rejection because that test target has no AAD admin configured) |

## Files touched

- `mssql_python/pybind/connection/connection.cpp` — both the `py::bytes`/`py::bytearray` branch and the `py::str` branch of `Connection::setAttribute` reverted to use the persistent member buffers.


mssql_python/pybind/connection/connection.cpp
```diff
@@ -320,17 +320,23 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
...
-            SQLPOINTER ptr;
-            SQLINTEGER length;
-            // Copy to a stack-local buffer so that releasing the GIL below
-            // doesn't expose a race where another thread overwrites the
-            // member wstrStringBuffer (and reallocates) while the ODBC
-            // driver is still reading from ptr.
-            std::u16string localStrBuffer = this->wstrStringBuffer;
-            ptr = reinterpretU16stringAsSqlWChar(localStrBuffer);
-            length = static_cast<SQLINTEGER>(localStrBuffer.length() * sizeof(SQLWCHAR));
+            SQLPOINTER ptr = reinterpretU16stringAsSqlWChar(this->wstrStringBuffer);
```
saurabh500: I am not sure if this change should be scoped only to 1256 or be made to all the attributes.

saurabh500:
Pushed `2f8db0b3`: replaced the two shared `Connection` member buffers (`wstrStringBuffer`, `strBytesBuffer`) with `std::unordered_map<SQLINTEGER, ...>` keyed by attribute id, so a second deferred string/bytes attribute can't invalidate the pointer the driver stashed for an earlier one. Only `SQL_COPT_SS_ACCESS_TOKEN` (1256) is deferred in practice today, so the previous single-buffer form was sufficient to fix #594 as observed; this hardens the fix against any future deferred string/bytes attribute. Re-verified end-to-end against an Azure SQL DB with an `AzureCliCredential` token — same clean server-side `OperationalError`, no SIGBUS / TCP reset.

saurabh500:
Pushed `031c7175` to expand the comments around the new per-attribute buffers, explaining the lifetime contract more fully.

**Why `_attrStringBuffers` / `_attrBytesBuffers` are NOT cleared on `Connection::reset()`** (was a reviewer question):

`SQL_ATTR_RESET_CONNECTION` is the pool-checkin reset. It asks the MS ODBC driver to wipe per-session state (temp tables, open cursors, `SET` options, isolation level, etc.) on the next use. It does **not**:

1. tear down the underlying TCP/TLS socket,
2. discard the driver-side authentication context, or
3. release the deferred connect attributes the driver has stashed (notably the `SQL_COPT_SS_ACCESS_TOKEN` pointer used to build the FedAuth Login7 packet).

The driver may still dereference those stashed pointers after the reset — most importantly during **Idle Connection Resiliency**: if the underlying TCP connection drops while the connection sits idle in the pool, the driver transparently re-establishes it on the next use and re-runs the Login7 / FedAuth handshake, reading the same access-token pointer it stashed at original-connect time.

If we cleared the per-attribute buffers on `reset()` we'd reintroduce issue #594 in a new and harder-to-debug form: a UAF during silent reconnect rather than initial connect. The buffers are therefore intentionally retained for the lifetime of the `Connection` object and freed implicitly via the destructors of the unordered_maps in `~Connection()`.

Memory cost is bounded: one entry per unique attribute id ever set on the connection (in practice = 1, for token 1256). Token rotation overwrites the same map slot in place, so it's O(1) memory per connection regardless of how often the token is refreshed.

mssql_python/pybind/connection/connection.h
```diff
@@ -68,8 +69,13 @@ class Connection {
...
     std::chrono::steady_clock::time_point _lastUsed;
-    std::u16string wstrStringBuffer;  // UTF-16 buffer for wide ODBC attributes
-    std::string strBytesBuffer;     // string buffer for byte attributes setting
+    // Per-attribute owned buffers for connect attributes whose pointer the
+    // driver may dereference *after* SQLSetConnectAttr returns (deferred
+    // attributes, e.g. SQL_COPT_SS_ACCESS_TOKEN). Keyed by attribute ID so
+    // that setting a second deferred attribute does not invalidate the
+    // pointer the driver stashed for the first. See issue #594.
+    std::unordered_map<SQLINTEGER, std::u16string> _attrStringBuffers;
+    std::unordered_map<SQLINTEGER, std::string>    _attrBytesBuffers;
```
subrata-ms, line 78: This is not a bug but good to have - 
**step1:** Driver stores pointer → P1
**step2:** Same attribute set again
**step3:** std::string reallocates → P1 invalid
**step4**: If driver still dereferences old pointer → ❌ potential UAF

The fix will work perfectly fine because of following reasons -
    1. Attribute is set once
    2. No repeated mutation
    3. No concurrent update

If we want to make it bulletproof, then probably we can define the containers as below - 
    std::unordered_map<SQLINTEGER, std::unique_ptr<std::string>> _attrStringBuffers;
    std::unordered_map<SQLINTEGER, std::unique_ptr<std::string>>    _attrBytesBuffers;

mssql_python/pybind/connection/connection.cpp
```diff
@@ -320,17 +320,38 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
...
+            // they are released only when the Connection object itself is
+            // destroyed.
+            //
+            // Note: attrs_before is applied once, sequentially, during
+            // connect(); the Connection's attribute setters are not designed
+            // for concurrent mutation from multiple threads.
+            auto& buf = this->_attrStringBuffers[attribute];
+            buf = value.cast<std::u16string>();
+
+            SQLPOINTER ptr = reinterpretU16stringAsSqlWChar(buf);
```
subrata-ms, line 352: This casting may cause data corruption or misaligned input on non-Windows environments.
unixODBC / Linux / macOS (problematic) as char16_t is 2 bytes and encoding is UTF16 while SQLWCHAR is 4 bytes and encoding UTF-32 ( via wchar_t).

reinterpret_cast works only if the memory layout matches exactly.Since SQLWCHAR is platform-dependent, this makes the code non-portable and prone to undefined behavior.

subrata-ms (approved):

mssql_python/pybind/connection/connection.cpp
saurabh500: Microsoft's ODBC Driver also explicitly uses UTF-16 SQLWCHAR on Linux/macOS to match Windows

Also Not new code: reinterpretU16stringAsSqlWChar is already used ~30+ times in ddbc_bindings.cpp for SQLExecDirect, SQLTables, SQLColumns, SQLForeignKeys, etc. If the cast were unsafe on Linux/macOS, the driver wouldn't function on those platforms at all — this PR just reuses the existing pattern.

mssql_python/pybind/connection/connection.h
saurabh500: I will keep this out of scope of the current PR, since the API usage and its impact on the current data structures is not well understood. 

In ODBC when the connection is established there is no supported access token overwrite. 
Hence repeated set_attr behavior should be considered undefined and we shouldn't make an effort to make anything work till it really is needed for a valid mssql-python scenario. 

The right approach could be to error out on repeated set_attr in mssql-python for attributes where the repeated settings don't make sense (which could be all the attributes).

mssql_python/pybind/connection/connection.cpp
```diff
@@ -349,12 +370,42 @@ SQLRETURN Connection::setAttribute(SQLINTEGER attribute, py::object value) {
...
+            //     driver-side authentication context and the stashed
+            //     deferred-attribute pointer are intentionally retained.
+            // For these reasons the per-attribute buffers are NOT cleared
+            // on reset()/checkin; they are released only when the
+            // Connection object itself is destroyed.
+            //
+            // Note: attrs_before is applied once, sequentially, during
+            // connect(); concurrent setAttribute() on the same Connection
+            // from different threads is not a supported pattern.
+            auto& buf = this->_attrBytesBuffers[attribute];
```
sumitmsft, line 405: The diff coverage report shows 60% most likely because the `py::bytes` / `py::bytearray` branch in `setAttribute()` is never covered by the test suite. A test on this code path would be of high value. 

Although this is not a blocker for merge since the fix is solid and validated end-to-end, but would be great to have for CI regression detection.

saurabh500: Can I do a follow up PR on this so that the critical path is unblocked and the regression fix makes its way. I will keep the bug opened and have a follow up PR to get the test coverage in.

saurabh500: @sumitmsft

sumitmsft (approved):

Issue #577, opened 2026-05-13
Azure Entra Auth Options


benmatwil:
Are there any plans to enable the user to customise the authentication via Azure?
I understand that the ODBC connection string schema options might restrict what's possible for customisation.
Under the current version, I know you could always generate a token manually send that to auth so there's always that alternative. 

We have a couple of use cases where logging into Azure needs a more custom use of the Azure Identity Credential classes. Examples include
- `DefaultAzureCredential` but excluding the managed identity check
- Directly using the `AzureCliCredential` or 

Perhaps being able to pass a Credential instance from the `azure.identity` library to `.connect` would be useful?

jahnvi480:
@benmatwil Thanks for raising this, it's a valid use case. We're currently discussing this internally with the team to evaluate the best approach. I'll update this issue once we have a conclusion. Appreciate your patience!

stuartleeks:
+1 to this.

I've had cases where I am using a remote dev machine but want to use the `AzreuCliCredential` (instead of `ManagedIdentityCredential` which picks up the VM identity).

I've also encountered a scenario where I want to use `mssparkutils` to get an access token for SQL DB when running in a pyspark notebook in Fabric. In this scenario, being able to pass a custom credential class to expose the credential would be really useful.

jahnvi480:
@benmatwil @stuartleeks The PR has been merged and it will be available to use in our next release. Thank you for helping us make our driver better.

PR #603, opened 2026-05-27
FEAT: Add "token_provider= " parameter for custom Azure Identity credential support

jahnvi480:

> [AB#45121](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45121)

> GitHub Issue: #577

### Summary   
This pull request introduces a new `token_provider` parameter for Microsoft Entra ID (Azure AD) authentication, enabling the use of any credential object with a `.get_token(scope)` method (such as those from `azure-identity`). It also adds robust support for custom credential objects, improves token acquisition error handling, and refines identity key computation for connection pooling. Several internal utilities and documentation comments have been added to clarify and harden the authentication process.

**Authentication and Token Provider Support:**

* Added a `token_provider` parameter to `connect()` and the `Connection` class, accepting any object with a `get_token(scope)` method for Microsoft Entra ID authentication (e.g., `DefaultAzureCredential`, `AzureCliCredential`, `ManagedIdentityCredential`). This is mutually exclusive with `Authentication=` in the connection string and with pre-acquired tokens in `attrs_before`. Bulk copy operations now re-acquire a fresh token from the provider for each operation. 
* Introduced the `TokenProvider` protocol to define the expected interface for custom credential objects, ensuring compatibility with both minimal implementations and all `azure-identity` credentials.

**Token Acquisition and Error Handling:**

* Added `_get_token_from_credential`, `acquire_token_from_credential`, and `acquire_raw_token_from_credential` utility functions to centralize token acquisition, handle errors, and log token details, including expiry. These functions provide clear error messages for common mistakes (e.g., async credentials, missing `.token` attribute) and warn if an already-expired token is returned.
* Improved documentation and logging for the Azure SQL Database OAuth scope, clarifying that only the Azure commercial cloud is supported via the `token_provider` path.

**Connection Pooling and Identity Key Computation:**

* Refined the logic for computing connection pool identity keys: managed identities now use distinct prefixes (`msi:client:<client_id>` vs. `msi:system`), and token-based identities use a SHA-256 hash of the token. The computation is now expiry-aware and avoids collisions between system- and user-assigned identities.
* Added `compute_token_identity` to encapsulate the hashing logic for token-based identity keys.

**Type Checking and Coverage:**

* Updated `.coveragerc` to exclude type-checking-only imports from coverage, improving test accuracy.

These changes significantly improve the flexibility, reliability, and clarity of Microsoft Entra ID authentication support in the codebase.


mssql_python/connection.py
```diff
@@ -334,10 +335,33 @@ def __init__(
...
+            if not callable(getattr(credential, "get_token", None)):
+                raise TypeError(
+                    f"credential must have a .get_token() method. "
+                    f"Got {type(credential).__name__}."
+                )
+            from mssql_python.auth import acquire_token_from_credential
+
+            token = acquire_token_from_credential(credential)
+            self._attrs_before[ConstantsDDBC.SQL_COPT_SS_ACCESS_TOKEN.value] = token
+            self._custom_credential = credential
```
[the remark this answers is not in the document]
jahnvi480: Good catch, made the required changes,

mssql_python/connection.py
```diff
@@ -401,6 +464,26 @@ def __init__(
         if not PoolingManager.is_initialized():
             PoolingManager.enable()
         self._pooling = PoolingManager.is_enabled()
+
+        # Access-token connections must NOT be pooled. The native pool is keyed
```
sdebruyn: Users should have an option to enable it as most cases will only have a single user. Pooling might be crucial for some workloads. Especially Fabric Data Warehouse fails fast when opening lots of connections.

sdebruyn (dismissed):

mssql_python/connection.py
jahnvi480: This is resolved now.

sumitmsft:
Looks good.. have some minor comments but not very critical to solve and can be ignored.

sumitmsft (dismissed):

mssql_python/db_connection.py
```diff
@@ -35,6 +39,34 @@ def connect(
             This per-connection override is useful for migration from pyodbc:
             connections that need string UUIDs can pass native_uuid=False, while the default (True)
             returns native uuid.UUID objects.
+        token_provider (object, optional): A token provider for Microsoft Entra ID
+            authentication. This must be any object with a ``.get_token(scope)`` method that
```
bewithgaurav, line 40: nit: inconsistencies in docstring explanation and can cause type check issues for end users:

the docstring says any object with a `.get_token(scope)` method works. the type hint in line 21 says otherwise, and the two disagree.

the hint is `Optional[TokenCredential]`, and that protocol needs `get_token(self, *scopes, **kwargs)`. so an object written exactly like the docstring says gets rejected by both type checkers:

```py
class MyToken:
    def get_token(self, scope):
        return AccessToken(my_jwt, exp)

mypy:    incompatible type "MyToken"; expected "TokenCredential | None"
pyright: "MyToken" cannot be assigned to parameter "token_provider"
```

it runs fine - just fails type checking, and we ship `py.typed`, so the user sees red in their editor and their CI, this can come from the fabric case from #577, where the token comes from `mssparkutils`, not from an azure-identity credential.

the docstring is the promise we want to keep, so we can possibly widen the hint

mssql_python/auth.py
```diff
@@ -660,3 +670,168 @@ def extract_auth_type(parsed_params: Dict[str, str]) -> Optional[str]:
...
+        pre-acquired token via ``attrs_before[SQL_COPT_SS_ACCESS_TOKEN]``.
+
+    Raises:
+        InterfaceError: If the provider returns no valid ``.token`` string.
+        OperationalError: If the underlying ``get_token()`` call fails.
+    """
+    start_time = time.perf_counter()
+    try:
+        token_result = credential.get_token(_SQL_SCOPE)
+    except TypeError as e:
```
bewithgaurav, line 764: this `except TypeError` blames the scope argument for every TypeError, even ones thrown inside the user's own `get_token`.

if their `get_token` body has a bug, say it does `"a" + 1`, they get:

```
InterfaceError: token_provider.get_token() must accept a scope positional argument
ddbc_error:     can only concatenate str (not "int") to str
```

so the message tells them to fix their scope argument, when the real bug is inside their function
the tests only cover the missing-scope case, so this wrong message never shows up in coverage

suggestion: only show the scope message when the call itself couldn't bind, and let a TypeError from inside `get_token` keep its own message and fall through to `OperationalError`.

tests/test_020_bulkcopy_auth_cleanup.py
```diff
@@ -112,6 +117,171 @@ def capture_context(ctx, **kwargs):
...
+
+        with patch.dict("sys.modules", {"mssql_py_core": mock_pycore_module}):
+            with pytest.raises(OperationalError):
+                cursor.bulkcopy("dbo.t", [(1, "a")], timeout=10)
+            # Cursor still works after the transient failure.
+            cursor.bulkcopy("dbo.t", [(2, "b")], timeout=10)
+
+        assert credential.get_token.call_count == 2
+
+        mock_logger.is_debug_enabled = False
```
bewithgaurav, line 272: this looks a bit weird - I guess two tests got mashed into one here

a decorator and a def line went missing in an edit

`test_transient_failure_then_recovery` starts at 226 and ends at 260. from line 262 down, this is a second, separate test: it resets the logger, makes a new credential with an empty token, and asserts its own thing. but it has no name and no decorator of its own - need to fix this one

mssql_python/connection.py
```diff
@@ -576,6 +640,15 @@ def _make_token_factory():
         if not PoolingManager.is_initialized():
             PoolingManager.enable()
         self._pooling = PoolingManager.is_enabled()
+
+        # Access-token connections are safe to pool: the pool key computed above
```
bewithgaurav: nit: this comment explains why a pooling guard was removed, but that guard was both added and removed inside this same PR - it never existed on main, and the pool-key comment at 604 already says what actually holds
can drop this one

mssql_python/cursor.py
```diff
@@ -3013,7 +3013,27 @@ def bulkcopy(
...
+                raw_token, _ = acquire_raw_token_from_credential(self.connection._token_provider)
+            except (OperationalError, InterfaceError) as e:
+                raise OperationalError(
+                    driver_error=(
+                        "Bulk copy failed: unable to acquire token from custom credential"
+                    ),
+                    ddbc_error=str(e),
+                ) from e
+            pycore_context["access_token"] = raw_token
+            for key in ("authentication", "user_name", "password"):
```
bewithgaurav: this loop removes nothing, since the parsed connection string is already stripped of auth params
`authentication`, `user_name`, `password` are never present here anyways - can remove this pop loop

mssql_python/cursor.py
```diff
@@ -3013,7 +3013,27 @@ def bulkcopy(
...
         # Token acquisition — only thing cursor must handle (needs azure-identity SDK)
-        if self.connection._auth_type:
+        if self.connection._token_provider is not None:
+            # User-supplied credential — use it directly for a fresh token.
+            from mssql_python.auth import acquire_raw_token_from_credential
+
+            try:
+                raw_token, _ = acquire_raw_token_from_credential(self.connection._token_provider)
+            except (OperationalError, InterfaceError) as e:
+                raise OperationalError(
```
bewithgaurav: minor: same failure, different exception

- this wraps an `InterfaceError` into an `OperationalError`, so an empty token from the provider surfaces as `InterfaceError` from `connect()` but `OperationalError` from `bulkcopy()`.
- DB-API treats those as different categories, so someone catching `InterfaceError` around both would miss it at bulk copy 

suggestion: keep the same failure catchable the same way from both calls, so an empty token raises `InterfaceError` whether it comes from `connect()` or `bulkcopy()`

tests/test_008_auth.py
```diff
@@ -5,9 +5,13 @@
 """
 
 import pytest
+import collections
+import inspect
```
bewithgaurav: this is unused (in tests)

tests/test_008_auth.py
```diff
@@ -1259,6 +1268,998 @@ def test_token_output_correct_on_cache_miss_and_hit(self):
...
+        conn = connect("Server=test;Database=testdb", token_provider=mock_cred)
+        assert conn._token_provider is mock_cred
+        assert conn._token_expires_on == 1893456000
+        assert ConstantsDDBC.SQL_COPT_SS_ACCESS_TOKEN.value in conn._attrs_before
+        # Existing auth_type should be None (no Authentication= in conn str)
+        assert conn._auth_type is None
+        conn.close()
+
+    @patch("mssql_python.connection.ddbc_bindings.Connection")
+    def test_token_provider_plus_authentication_raises_valueerror(self, mock_ddbc_conn):
```
bewithgaurav: a batch of test names here say one exception and the body raises another. reading these later will mislead:

```
1423  ..._plus_authentication_raises_valueerror          raises InterfaceError
1439  ..._via_kwargs_raises_valueerror                    raises InterfaceError
1456  ..._attrs_before_access_token_raises_valueerror     raises InterfaceError
1473  ..._without_get_token_raises_typeerror              raises InterfaceError
1511  ..._get_token_failure_raises_runtime_error          raises OperationalError
1522  ..._non_callable_get_token_raises_typeerror         raises InterfaceError
1736  test_suspicious_signature_warns_but_does_not_block  nothing warns
```

also `test_scope_is_commercial_cloud` is defined twice (1308 and 1570), and 1894 checks the same scope string a third time

mssql_python/mssql_python.pyi
bewithgaurav, line 1: not from this change, but since this stub is being edited: it does not reach anyone & is irrelevant

this pyi needs an mssql_python.py for it to describe and validate - and also, `*.pyi` is not listed in `package_data` in `setup.py`, so it is not packaged in the wheel

type checking of `connect(token_provider=...)` comes from the real hint in `db_connection.py` instead. passing an `int` is already rejected with this file removed.

so the additions here are harmless but does nothing for users
worth deciding separately whether to ship this stub or delete it - could you add a backlog task for this?

bewithgaurav (changes requested):
added comments inline, the PR description needs revision - requesting changes
can skip some minor ones if needed

mssql_python/db_connection.py
jahnvi480: Fixed — swapped TokenCredential for a local TokenProvider Protocol defined as get_token(self, scope), so the hint now matches the docstring. A Fabric/mssparkutils-style get_token(scope) object type-checks clean under both mypy and pyright.

mssql_python/auth.py
jahnvi480: Fixed — the scope-binding message is now only emitted when the call itself fails to bind (traceback stops at this frame). A TypeError raised inside the user's get_token keeps its own message and falls through to OperationalError.

tests/test_020_bulkcopy_auth_cleanup.py
jahnvi480: Fixed — split into test_transient_failure_then_recovery and test_empty_token_from_provider_raises_interface_error, each with its own name and decorator.

mssql_python/connection.py
jahnvi480: Dropped it — the guard it described was added and removed within this PR, and the pool-key comment already states the real invariant.

mssql_python/cursor.py
jahnvi480: Removed the pop loop from the token_provider path (the parsed connstr is already stripped there). The remaining pop stays in the built-in-auth branch, where those fields are actually present.

mssql_python/cursor.py
jahnvi480: Fixed — bulkcopy now re-raises via type(e)(...), so an empty/invalid token stays an InterfaceError from both connect() and bulkcopy() and callers catch the same DB-API category.

tests/test_008_auth.py
jahnvi480: Removed

tests/test_008_auth.py
jahnvi480: Renamed each test to match the exception it actually raises (InterfaceError/OperationalError) and the no-warn case; de-duplicated test_scope_is_commercial_cloud.

mssql_python/mssql_python.pyi
jahnvi480: It's not in package_data and has no .py to describe, so it's inert; real type-checking comes from db_connection.py. I'll file a backlog task to decide ship-vs-delete and revert the stub additions from this PR.

sumitmsft (dismissed):

bewithgaurav (dismissed):

bewithgaurav (dismissed):

sumitmsft (dismissed):

gargsaumya (approved):

sumitmsft (approved):

bewithgaurav (approved):

Issue #609, opened 2026-05-30
mssql-python 1.7.1 and 1.8.0 both fail on executemany


sravikrishna:
python 3.13.13 on Windows
mssql-python 1.7.1/1.8.0

I have a python script to copy a table from one env to another.  The script makes two connections, one to the source database and one to the target database.  Reads rows on the source as fetchmany(5000) and then writes on the target side as one executemany with fast_executemany set to True.

Works great most of the time.  However I am seeing this issue with one table.

After copying about 923,000 rows, it errors out

   ret = ddbc_bindings.SQLExecuteMany(
        self.hstmt, operation, columnwise_params, parameters_type, row_count, encoding_settings
    )

RuntimeError: Parameter's object type does not match parameter's C type. paramIndex - 10, C type - SQL_C_NUMERIC

Seem to happen at the same row.

I understand this information may not be enuf.
Pls advise what information should I provide for the support team to find a resolution.

dlevy-msft-sql:
@sravikrishna - This may unblock you while we get to the bottom of your issue. 

[Quickstart: Bulk copy with the mssql-python driver for Python](https://learn.microsoft.com/sql/connect/python/mssql-python/python-sql-driver-mssql-python-bulk-copy-quickstart)

sravikrishna:
Thanks David.

Yes I am aware of bulk_copy and I am using it for file copy purposes.
But this is direct table to table copy.

Update:  I have an old version of the script which uses pyodbc.  I was very happy with pyodbc fast_executemany feature.  But pyodbc can not take datetime offset and few other variables and throws out -155 error.  mssql-python is an improvement there.

Anyhow this table does not contain any data type which can be -155 issue.  So I copied the same using the pyodbc version.  Two things:

1. It is noticeably faster than mssql-python.  There is no difference in the code except the connection driver.
2. I didn't get any error and the copy completed successfully.

This rules out any issue with the table.

here is the relevant code, which is same for pyodbc and mssql-python.

t_cursor.fast_executemany = True
s_cursor.execute(sql_stmt)
while rows := s_cursor.fetchmany(5000) :
      t_cursor.executemany(ins_sql,rows)
      t_cxn.commit()
      rows_copied += len(rows)
      write_log(str(rows_copied) + " rows committed")

dlevy-msft-sql:
> Thanks David.
> 
> Yes I am aware of bulk_copy and I am using it for file copy purposes. But this is direct table to table copy.
> 
[...]

PR #611, opened 2026-06-01
FIX: executemany SQL_C_NUMERIC mismatch

jahnvi480:

> [AB#45380](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45380)

> GitHub Issue: #609 

### Summary   
This pull request addresses a critical bug in the handling of `executemany` with large `Decimal` values in the `mssql_python` driver, specifically when values exceed the SQL Server `MONEY` range. The main fix ensures that parameter type detection and conversion are consistent, preventing runtime errors when binding large decimal values. Extensive unit and integration tests are added to verify the fix and cover edge cases involving `Decimal` values, including scenarios with `NULL`s and multi-column inserts.

**Bug Fix: Executemany Decimal Handling**

* In `cursor.py`, the `executemany` method now explicitly overrides the C type for parameters with SQL type `DECIMAL` or `NUMERIC` to use `SQL_C_CHAR` (string binding) when the data is converted to strings. This prevents mismatches that previously caused runtime errors when inserting large decimal values. The column size is also adjusted to fit the longest string representation.

**Testing: Unit and Integration Tests for Decimal Handling**

* Added comprehensive unit tests in `test_001_globals.py` to verify type detection, mapping, and the override logic for `executemany` with large `Decimal` values, both within and outside the `MONEY` range. These tests confirm that the C type override is necessary and correctly applied.
* Added integration tests in `test_004_cursor.py` to exercise the fixed behavior in real database scenarios, including:
  - Inserting batches with decimals inside and outside the `MONEY` range.
  - Handling `NULL` values alongside large decimals.
  - Multi-column inserts where one column contains large decimals.


bewithgaurav (dismissed):
reproduced #609 against sql2022, fix is clear and well-scoped. lgtm.

tests/test_001_globals.py
```diff
@@ -1059,3 +1059,164 @@ def test_row_string_key_case_insensitive_with_lowercase():
...
+# GH-609: Unit tests for executemany SQL_C_NUMERIC mismatch
+# ---------------------------------------------------------
+import decimal as _decimal
+from mssql_python.cursor import Cursor, MONEY_MAX
+from mssql_python.constants import ConstantsDDBC as _C
+
+
+def _make_bare_cursor():
+    """Create a Cursor instance without a connection for unit testing."""
+    cur = Cursor.__new__(Cursor)
```
bewithgaurav: [non-blocking] suggestion to not have cursor tests in global test file

jahnvi480: Moved the tests to cursor

subrata-ms (approved):

bewithgaurav (approved):

Issue #606, opened 2026-05-29
Add .to_dict to Row objects


amachanic:
### Is your feature request related to a problem? Please describe.
I've just finished testing the newly released row-string indexing feature. It's great and extremely handy for cases where a mapping interface is needed. However, I think it can go a bit further: Sometimes we need to pass a true dict into some legacy thing, and here mssql-python could help, by providing a built-in `to_dict()` function.

### Describe the solution you'd like
I would like a function on the `Row` object that will return a Python `dict` with all of the row's values.

### Describe alternatives you've considered
Interrogating the cursor's `description` property and producing my own dict is easy enough. But it would certainly be a lot easier, and probably faster, to let mssql-python do the work instead.

Thank you!

amachanic:
Looking a bit more heavily through my use cases, I'd also like to suggest the addition of `items()` and `values()` functions, that will behave just like their dict counterparts, in addition to the `to_dict()` function suggested above.

jahnvi480:
Hi @amachanic, thanks for the feature request! We agree this is a great addition - having a built-in to_dict() is much cleaner than manually zipping cursor.description.

We've raised a PR that adds the following methods to Row:
- to_dict() - returns a plain dict mapping column names to values
- keys() - column names
- values() - column values
- items() - (name, value) pairs
- "col" in row - membership testing (also respects case-insensitive mode)

These are all lazy, zero cost at fetch time, only run when you call them. We'll update this issue once the PR is merged. Thanks again for the suggestion!

amachanic:
Thank you @jahnvi480!

jahnvi480:
Hello @amachanic, the PR has merged to main, so this ships in our next release.

One clarification on the final API - we landed on a single row._mapping view (a read-only collections.abc.Mapping) rather than separate to_dict() / keys() / values() / items() methods on Row directly. Everything you asked for is available through it:
```py
row = cursor.fetchone()

dict(row._mapping)            # {'id': 1, 'name': 'Alice'}
row._mapping.keys()           # column names
row._mapping.values()         # values
row._mapping.items()          # (name, value) pairs
row._mapping["name"]          # value by column name
"name" in row._mapping        # membership by column name
```

Positional/tuple access is unchanged: row[0], row.name, and row["name"] all still work (and row["name"] / row.name honor case-insensitive lowercase mode).
Thanks again for the suggestion!

amachanic:
[...]

PR #613, opened 2026-06-01
FEAT: Add read-only dict-style Row access via row._mapping

jahnvi480:

> [AB#45369](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45369)

> GitHub Issue: #606 

### Summary   
This pull request adds dictionary-style access to the `Row` class through a single, read-only `row._mapping` view instead of adding standalone dict methods to `Row` itself. This keeps the `Row` tuple/attribute surface clean while making rows fully compatible with Python's mapping protocol.

Enhancements to the `Row` class (`mssql_python/row.py`):

* Added a read-only `row._mapping` property that returns a `RowMapping` — a `collections.abc.Mapping` view of column name → value. `RowMapping` is exported from the package.
* Through the standard `Mapping` protocol, `row._mapping` supports `dict(row._mapping)`, `row._mapping["col"]`, `.keys()`, `.values()`, `.items()`, the `in` operator, iteration over column names, and equality comparison with plain `dict`s.
* The view is read-only — assignment and deletion raise `TypeError` — so a row cannot be mutated through it.
* Column names are order-preserving and de-duplicated (the last column wins for a repeated name), consistent with subscript (`row["col"]`) and attribute (`row.col`) access.
* The cursor snapshots the canonical column names once per result set and shares them with every row, so `row._mapping` stays correct even after the cursor is reused for another query (previously `cursor.description` could be stale).

Note: the earlier standalone `keys()`, `values()`, `items()`, and `to_dict()` methods on `Row` are intentionally not included; all dict-like access now goes through `row._mapping` (e.g. `dict(row._mapping)` replaces `row.to_dict()`).

Testing improvements:

* Added comprehensive DB-backed tests for `row._mapping` covering: dict conversion; `keys`/`values`/`items`; duplicate-key de-duplication; `__getitem__`/`get`/`in` including missing-key `KeyError` and non-string-key `KeyError`; read-only enforcement (`TypeError` on assign/delete); `repr` and equality; duplicate column names (last-wins); rows produced by `fetchall`/`fetchmany`; reflection of converted values (DECIMAL, UNIQUEIDENTIFIER, DATETIME2, NVARCHAR, NULL); custom output converters; the global `lowercase` setting; and stability across cursor reuse.
* Preserved the existing no-DB unit tests for `Row` string-key indexing (`row["col"]`), including case-insensitive lookup and `KeyError`/`TypeError` edge cases.


mssql_python/row.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         converter_map=None,
         uuid_str_indices=None,
         column_map_lower=None,
+        column_names=None,
```
bewithgaurav: flagging a repro'd perf regression on this which ultimately deviates from the the "lazy, zero cost at fetch time, only run when you call them." claim:

`column_names` is computed on the cursor and passed to every `Row.__init__()` call, even when no one calls `keys()` or `to_dict()`. I benchmarked it:

- main branch: 0.08 us/row
- this PR: 0.16 us/row
- at 1M rows that's +76ms of overhead for a feature the caller didn't use

the fix is lazy-compute. drop the `column_names` kwarg entirely, add `self._column_names = None` in `__init__`, and resolve on first access:

```python
def _get_column_names(self):
    if self._column_names is not None:
        return self._column_names

    if self._cursor and hasattr(self._cursor, "description") and self._cursor.description:
        column_names = tuple(desc[0] for desc in self._cursor.description)
    elif self._column_map:
        idx_to_name = {}
        for name, idx in self._column_map.items():
            if idx not in idx_to_name:
                idx_to_name[idx] = name
        column_names = tuple(idx_to_name[i] for i in sorted(idx_to_name))
    else:
        column_names = ()

    self._column_names = column_names
    return column_names
```

this also removes the cursor-side `_cached_column_names` plumbing, smaller diff overall.

mssql_python/row.py
```diff
@@ -209,6 +227,36 @@ def __getattr__(self, name: str) -> Any:
 
         raise AttributeError(f"Row has no attribute '{name}'")
 
+    def keys(self):
```
bewithgaurav: the existing methods on Row all have return type annotations (`__getitem__ -> Any`, `__len__ -> int`, etc.). the new ones don't. for consistency:

```python
def keys(self) -> tuple:
def values(self) -> list:
def items(self) -> list:
def to_dict(self) -> dict:
```
adding them would help with not having typing visibility & issues.

mssql_python/row.py
bewithgaurav: not a blocker, but worth calling out:

now that Row has `keys()`/`values()`/`items`, devs/LLMs can confuse it for a `dict`
it gives and impression of `dict` when reading the API overall, even though it still behaves like a tuple when you iterate it.

but `for x in row` still yields values (tuple behavior), not keys (dict behavior) - which is correct and matches `pyodbc` behaviour as well, but worth calling out in the class docstring for visibility. something like:

```python
class Row:
    """
    A row of data from a cursor fetch operation. Provides both tuple-like indexing
    and attribute access to column values.

    Row supports dict-like access via keys(), values(), items(), and to_dict().
    However, iteration (for x in row) yields values, not keys — consistent with
    pyodbc.Row and sqlite3.Row. Use row.keys() to iterate column names.

    Column attribute access behavior depends on the global 'lowercase' setting:
    - When enabled: Case-insensitive attribute access
    - When disabled (default): Case-sensitive attribute access matching original column names

    Example:
        row = cursor.fetchone()
        print(row[0])              # Access by index
        print(row.column_name)     # Access by column name
        print(row.to_dict())       # Convert to dict
        for value in row:          # Iterates values, not keys
            print(value)
    """
```

bewithgaurav (changes requested):
perf regression on `Row.__init__` from the `column_names` kwarg  - fix suggestion + details inline
two other suggestions on type annotations, docstring)

mssql_python/row.py
jahnvi480: Thanks for calling this out, I'll keep this in mind while raising other PRs.

mssql_python/row.py
jahnvi480: Made required changes

mssql_python/row.py
jahnvi480: Added required doc string.

mssql_python/row.py
```diff
@@ -209,6 +217,55 @@ def __getattr__(self, name: str) -> Any:
...
+                if idx not in idx_to_name:
+                    idx_to_name[idx] = name
+            column_names = tuple(idx_to_name[i] for i in sorted(idx_to_name))
+        else:
+            column_names = ()
+
+        self._column_names = column_names
+        return column_names
+
+    def keys(self) -> tuple:
```
sumitmsft, line 239: keys() returns tuple, but values() returns list. Can we discuss and pick one and apply consistently. Both could return tuples; if list semantics are desired, both could return lists. 

possible? 

@bewithgaurav @jahnvi480

jahnvi480: values() returns self._values which is already a list internally, returning it directly avoids a copy and is consistent with how __iter__ already works (iterates the same list). Converting to tuple would add an O(n) copy on every call.

keys() returns a tuple because _column_names is cached as a tuple (immutable, computed once, shared across calls). Converting to list would create a new list on every call.

Both are sequences, both support len(), indexing, iteration, and in. The difference is mutability, if we are planning for consistency then I will suggest to go with tuple since column names shouldn't be modified.

mssql_python/row.py
```diff
@@ -209,6 +217,55 @@ def __getattr__(self, name: str) -> Any:
...
+        return list(zip(self._get_column_names(), self._values))
+
+    def to_dict(self) -> dict:
+        """Return the row as a plain dict mapping column names to values."""
+        return dict(zip(self._get_column_names(), self._values))
+
+    def __contains__(self, key) -> bool:
+        """Support 'col_name in row' membership testing.
+
+        Uses the full _column_map (which includes lowercase aliases),
```
sumitmsft: The docstring claims:

_Uses the full _column_map (which includes lowercase aliases), so 'ProductID' in row and 'productid' in row both return True._

But this is only true for rows produced by metadata methods (cursor.columns(), cursor.tables(), etc.) which inject lowercase aliases in _prepare_metadata_result_set

For a normal SELECT with default lowercase=False.

_cached_column_map contains only the original-cased name (e.g. {"ProductID": 0}).
_cached_column_map_lower is None.

So "productid" in row returns False, not True as the docstring promises. The PR's own test_row_contains test doesn't catch this because it only asserts column names from cursor.description (which preserve original casing).

Can you check this?

tests/test_004_cursor.py
```diff
@@ -3081,6 +3081,271 @@ def test_row_string_key_indexing(cursor, db_connection):
...
+    finally:
+        try:
+            cursor.execute("DROP TABLE IF EXISTS #pytest_row_kvi")
+            db_connection.commit()
+        except Exception:
+            pass
+
+
+def test_row_contains(cursor, db_connection):
+    """Test 'column_name in row' membership testing from a real cursor row."""
```
sumitmsft: `column_name` in row now tests for column-name membership (previously tested value membership via iteration). Seems to be a silent breaking change. Can you please add a changelog/release-note entry calling it out.

mssql_python/row.py
```diff
@@ -209,6 +217,55 @@ def __getattr__(self, name: str) -> Any:
...
+            column_names = ()
+
+        self._column_names = column_names
+        return column_names
+
+    def keys(self) -> tuple:
+        """Return column names, like dict.keys()."""
+        return self._get_column_names()
+
+    def values(self) -> list:
```
sumitmsft: One more thing to add:

`values()` returns internal list by reference 

def values(self) -> list:
    return self._values  # _>>>> caller can mutate row state_

A caller can do `row.values().append("x") `and corrupt the row. Can we return a copy for safety and to stay consistent with `keys()` (which returns an immutable tuple)?

This also resolves the keys() -> tuple vs values() -> list inconsistency I raised earlier? Though I'm fine either way as long as the choice is documented and applied consistently.

mssql_python/row.py
jahnvi480: I'm removing __contains__ from this PR. Users who need column-name membership can use "col" in row.keys() or "col" in row.to_dict().

We can revisit __contains__ in a follow-up with proper design discussion on the intended semantics and a documented breaking change note if we decide to ship it.

tests/test_004_cursor.py
jahnvi480: Removing __contains__ from this PR. Users who need column-name membership can use "col" in row.keys() or "col" in row.to_dict().

We can revisit __contains__ in a follow-up with proper design discussion on the intended semantics and a documented breaking change note if we decide to ship it.

mssql_python/row.py
jahnvi480: Will send it as tuple as mentioned in earlier comment to avoid manipulation or corruption of data.

mssql_python/row.py
bewithgaurav: both being tuples is the right call, I'm fine with this change done

tests/test_004_cursor.py
sumitmsft: Does this need doc change? @jahnvi480 ?

jahnvi480: No, but we need to add that we are now supporting these new APIs as well and mention how they work

mssql_python/row.py
```diff
@@ -209,6 +217,41 @@ def __getattr__(self, name: str) -> Any:
 
         raise AttributeError(f"Row has no attribute '{name}'")
 
+    def _get_column_names(self) -> tuple:
```
bewithgaurav, line 220: regression caused from previous suggestion:

`_get_column_names()` reads `self._cursor.description` lazily on first call. if the cursor re-executes before `keys()`/`to_dict()` is called on an old row, the old row picks up the new query's column names. repro:

```python
cursor.execute("SELECT 1 AS OrderID, 'Widget' AS Product, 9.99 AS Price")
order = cursor.fetchone()

cursor.execute("SELECT 100 AS CustomerID, 'Alice' AS Name")
customer = cursor.fetchone()

order.to_dict()
# expected: {'OrderID': 1, 'Product': 'Widget', 'Price': 9.99}
# actual:   {'CustomerID': 1, 'Name': 'Widget'}
# Price silently dropped (zip truncates 3 values to 2 columns)
```

fix: snapshot `cursor.description` at construction time instead of reading it live. one pointer assignment, no computation:

```python
def __init__(self, ...):
    self._column_names = None
    self._description = cursor.description if cursor else None

def _get_column_names(self) -> tuple:
    if self._column_names is not None:
        return self._column_names
    if self._description:
        column_names = tuple(desc[0] for desc in self._description)
    elif self._column_map:
        ...
```

still lazy, still zero-cost if `keys()`/`to_dict()` is never called, but safe on cursor reuse.

mssql_python/row.py
```diff
@@ -209,6 +217,41 @@ def __getattr__(self, name: str) -> Any:
...
+                if idx not in idx_to_name:
+                    idx_to_name[idx] = name
+            column_names = tuple(idx_to_name[i] for i in sorted(idx_to_name))
+        else:
+            column_names = ()
+
+        self._column_names = column_names
+        return column_names
+
+    def keys(self) -> tuple:
```
bewithgaurav, line 239: just checked during testing, adding `keys()`, `values()`, `items()`, `to_dict()` shadows column attribute access for any column with those names. repro:

```python
cursor.execute("SELECT 'timeout' AS keys, '30' AS [values]")
row = cursor.fetchone()

row.keys    # before PR: 'timeout'  (via __getattr__)
row.keys    # after PR:  <bound method Row.keys>
row["keys"] # still works: 'timeout'
```

class methods win over `__getattr__` in Python's MRO. this is a silent breaking change for existing code using `row.keys` or `row.values` as column attribute access.

mssql_python/row.py
```diff
@@ -209,6 +217,41 @@ def __getattr__(self, name: str) -> Any:
...
+
+    def values(self) -> tuple:
+        """Return column values as a tuple, like dict.values()."""
+        return tuple(self._values)
+
+    def items(self) -> list:
+        """Return (column_name, value) pairs, like dict.items()."""
+        return list(zip(self._get_column_names(), self._values))
+
+    def to_dict(self) -> dict:
```
bewithgaurav, line 251: `to_dict()` silently loses data when the query has duplicate column names. common in joins:

```python
cursor.execute("""
    SELECT e.id, e.name, m.id, m.name
    FROM employees e LEFT JOIN employees m ON e.manager_id = m.id
""")
row = cursor.fetchone()

list(row)       # [1, 'Alice', 2, 'Bob']  — 4 values
row.to_dict()   # {'id': 2, 'name': 'Bob'} — 2 entries, Alice gone
```

lets discuss approaching this

bewithgaurav (dismissed):

mssql_python/row.py
amachanic: @bewithgaurav The Row object's name-based indexing already has this problem:

```
cur.execute('select 123 as xx, 234 as xx;')
f = cur.fetchone()
print(f['xx'])
```
--> 234

I'd argue that as long as the behavior is consistent, to_dict() can silently dump the value. But instead, ideally, I'd like to see this handled a bit higher up the stack. If there are duplicates, maybe add `_1`, `_2`, etc, based on ordinality. I guess that could be a breaking change, but I don't think the current behavior is great. Better to prevent accidents than support them IMO.

sumitmsft (dismissed):
The `row._mapping` implementation preserves tuple-style Row behavior and keeps saved mappings stable across cursor reuse. The native build and 19 focused SQL Server tests passed.

One narrow mapping-contract issue remains at `8e9a31a3`. I acknowledge that [the earlier discussion](https://github.com/microsoft/mssql-python/pull/613#discussion_r4027007456) is marked resolved; however, I reproduced the reported behavior on the current head:

```python
mapping = Row([], None)._mapping

mapping["missing"]                 # TypeError; expected KeyError
mapping.get("missing", "fallback")  # TypeError; expected "fallback"
"missing" in mapping               # TypeError; expected False
```

This construction is explicitly covered by the new `test_row_mapping_none_column_map`, but that test only checks iteration, conversion, and length. Normal cursor-produced rows are not implicated by this reproduction.

Please handle the absent column map as a missing key, without broadly catching unrelated `TypeError`s, and extend that test to cover these three operations.

**Recommendation: Request Changes** for this limited error-handling defect. No other actionable issues identified.

sumitmsft (dismissed):
Re-reviewed at `63b199ea`. The previously reported empty-mapping issue is fixed: missing-key indexing raises `KeyError`, `.get()` returns the expected default, and membership returns `False`. The regression test now covers these operations.

Reviewed the PR for correctness, security, reliability, performance, test coverage, repository conventions, and applicable architecture and design specifications. No actionable issues were identified. The implementation is consistent with repository standards and the stated API contract. No separate approved design was accessible for verification.

Validation: rebuilt the native extension; all 19 focused tests passed against local SQL Server. Additional live checks passed for duplicate labels, `nextset()`, saved-row stability, and metadata mappings. Black passed across 93 files. Cross-platform ADO, coverage, and performance checks were still pending at the end of the re-review; this approval does not replace those checks.

**Recommendation:** Approve

mssql_python/row.py
```diff
@@ -209,6 +234,33 @@ def __getattr__(self, name: str) -> Any:
...
+        row was built without that snapshot, reconstructs names from ``_column_map``
+        (one name per column index); returns ``()`` when neither is available.
+        """
+        if self._column_names is not None:
+            return self._column_names
+        if self._column_map:
+            idx_to_name: dict = {}
+            for name, idx in self._column_map.items():
+                idx_to_name.setdefault(idx, name)
+            return tuple(idx_to_name[i] for i in sorted(idx_to_name))
```
bewithgaurav, line 283: nonblocking: directly constructed rows can expose a different key order from cursor-produced rows with the same labels.

```python
Row([1, 2, 3], {"a": 2, "b": 1})._mapping
# keys: ["b", "a"]

Row([1, 2, 3], {"a": 2, "b": 1}, column_names=("a", "b", "a"))._mapping
# keys: ["a", "b"]
```

sorting by the surviving indices moves repeated names to their last position. can we align the fallback ordering with the first-seen-key contract, or explicitly document its limitation? normal cursor fetches already supply the snapshot.

mssql_python/row.py
```diff
@@ -209,6 +234,33 @@ def __getattr__(self, name: str) -> Any:
...
 
+    @property
+    def _mapping(self) -> "RowMapping":
+        """Read-only dict-like view (column name -> value) over this row.
+
+        Returns a ``collections.abc.Mapping``; use ``dict(row._mapping)`` for a plain
+        dict, ``row._mapping.items()`` for name/value pairs, and ``iter(row._mapping)``
+        for column names. Names are order-preserving and de-duplicated (last column
+        wins for a repeated name, matching subscript and attribute access).
+        """
```
bewithgaurav, line 262: suggestion: add a user-facing example for `dict(row._mapping)` and `row._mapping.items()`, plus the duplicate-label and reserved-name caveats. the docstring explains the feature, but users also need to know that a column named `_mapping` must be accessed through `row["_mapping"]`.

bewithgaurav (dismissed):
this addresses dictionary conversion and named-field iteration through the agreed mapping view, without adding common method names to the row itself. two nonblocking suggestions on fallback ordering and user-facing documentation. approving.

mssql_python/row.py
jahnvi480: Resolved

bewithgaurav (approved):

sumitmsft (approved):

Issue #610, opened 2026-05-30
Lots of calls to sp_describe_undeclared_parameters


amachanic:
Any time a parameterized query sends a `None`, I'm seeing an internal call made to `sp_describe_undeclared_parameters`. Since my workload is entirely parameterized and sends a lot of Nones in the course of its work, I'm getting thousands of these calls a minute. I'm sure that's adding up to a significant performance hit in aggregate. 

To make matters worse, the vast majority of these calls are internally throwing an exception, so I'm not sure how much value they're providing in general to the driver. For example, if a stored procedure is called with named parameters, the call will fail with a message like: "@first_parameter is not a parameter for procedure my_stored_procedure."

So, I guess I'd like to ask a few things:
- Are these calls really all that necessary? I assume the goal is to try to find the correct data type for the NULL, but since the calls are raising an exception anyway most of the time and the queries are still working just fine, presumably that information isn't really useful...
- If the calls are really necessary, can the driver use a small internal cache so that repeated calls for the same query shape don't have to incur the cost every single time? Maybe cache a hash of the most recent 512 query shapes or something like that..? That should use a small amount of memory and will cover the situation nicely.
- Is there any way to avoid the situation in the meantime? Can I add some type info somewhere that will work around whatever internal path is trying to get this information?

jahnvi480:
Thanks for reporting this, @amachanic. A few questions to help us scope the fix:
1) Is your workload using cursor.execute(), cursor.executemany(), or a mix of both? (executemany() already has an optimization that skips the describe call for all-NULL columns, so we want to confirm which path you're hitting.)
2) Are the queries primarily stored procedure calls (e.g., EXEC myproc @p1=?, @p2=?), ad-hoc parameterized SQL (e.g., SELECT ... WHERE col = ?), or a mix? This helps us understand why the describe calls are failing - stored procs with named parameters are a known failure case for sp_describe_undeclared_parameters.
3) When the describe call fails and the driver falls back to SQL_VARCHAR for the NULL, are your queries still returning correct results? Or are you seeing any type-related errors or wrong data? We're trying to determine if we can safely skip the describe round-trip for NULLs entirely, rather than just caching the results.

amachanic:
@jahnvi480 

1. I'm using only `.execute` at the moment
[...]

PR #614, opened 2026-06-02
FIX:  Cache SQLDescribeParam results for NULL parameters to reduce server round-trips

jahnvi480:

> [AB#45399](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45399)

> GitHub Issue: #610
### Summary   
This pull request implements a robust caching mechanism for SQL parameter type resolution when binding `None` (NULL) parameters in the MSSQL Python driver, addressing issues with repeated round-trips to `SQLDescribeParam` and previous incorrect type fallbacks. The changes affect both the Python and C++ layers, improving performance and correctness, especially for all-NULL columns and VARBINARY types. Comprehensive tests are added to ensure correctness and coverage of the new cache logic.

**Enhancements to NULL parameter type resolution and caching:**

* Introduced a thread-safe cache in `ddbc_bindings.cpp` to store `SQLDescribeParam` results per statement handle and parameter index, reducing redundant round-trips and ensuring correct type inference for NULL parameters. This replaces the previous fallback to `SQL_VARCHAR`, which caused issues with certain column types. 
* Updated both single and array parameter binding (`BindParameters` and `BindParameterArray`) to use the new cache for resolving types of NULL parameters, ensuring consistency and correctness for all-NULL columns.

**Cache invalidation and integration:**

* Added cache invalidation logic when a new SQL statement is prepared, ensuring that parameter type caches are not reused across different statements. This is handled in both `SQLExecute_wrap` and `SQLExecuteMany_wrap`. 

**Python interface and type mapping:**

* Modified `_map_sql_type` in `cursor.py` to always return `SQL_UNKNOWN_TYPE` for `None` parameters, delegating type resolution to the C++ cache. Removed the previous Python-side fallback to `SQL_VARCHAR` for all-NULL columns in `executemany`.

**Testing improvements:**

* Added extensive tests in `test_004_cursor.py` to cover cache hit/miss scenarios, cache invalidation, all-NULL columns, and bypassing the cache with explicit type hints, ensuring the new logic is robust and correct.

**Codebase maintenance:**

* Included necessary C++ headers for thread-safety and updated function pointer loading logs for clarity. 

* Minor loop index and logging improvements for clarity and correctness in parameter binding.

These changes collectively resolve previous issues with NULL parameter handling, improve driver efficiency, and provide comprehensive test coverage for the new behavior.


subrata-ms (dismissed):

mssql_python/cursor.py
```diff
@@ -412,8 +412,15 @@ def _map_sql_type(  # pylint: disable=too-many-arguments,too-many-positional-arg
...
+            # Use SQL_VARCHAR directly instead of SQL_UNKNOWN_TYPE to avoid
+            # a costly SQLDescribeParam / sp_describe_undeclared_parameters
+            # round-trip to the server for every NULL parameter.  The describe
+            # call fails most of the time (especially for stored-procedure
+            # calls) and the C++ layer already falls back to SQL_VARCHAR, so
+            # this is safe.  Matches the executemany() all-NULL optimisation.
+            # See GH-610.
             return (
-                ddbc_sql_const.SQL_UNKNOWN_TYPE.value,
+                ddbc_sql_const.SQL_VARCHAR.value,
```
sumitmsft: @jahnvi480 I thought through this change and have following observation:
This change looks risky because mapping every `None` to `SQL_VARCHAR` is not equivalent to preserving an unknown/typed NULL. It may work for common text-compatible cases, but it can break strongly typed nullable parameters such as `VARBINARY`, `UNIQUEIDENTIFIER`, `XML`, `DECIMAL`, and stored procedure params. 

Previously the driver attempted metadata resolution via SQLDescribeParam before falling back to SQL_VARCHAR; this PR removes that behavior globally without adding integration coverage for typed NULL cases. This could be a high-severity compatibility risk and would want end-to-end tests before merging.


Add integration tests for None against:
VARBINARY
UNIQUEIDENTIFIER
XML
DECIMAL
stored procedure nullable parameters
If those fail, narrow the optimization instead of applying it globally.

jahnvi480: Made changes based on the discussion we had

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -399,6 +404,76 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+// executions of the same prepared statement with NULL parameters.
+struct DescribedParamInfo {
+    SQLSMALLINT sqlType;
+    SQLULEN columnSize;
+    SQLSMALLINT decimalDigits;
+    bool succeeded;
+};
+
+static std::shared_mutex g_describeCacheMutex;
+static std::unordered_map<SQLHANDLE, std::unordered_map<int, DescribedParamInfo>> g_describeCache;
```
sumitmsft: The cache is only ever evicted via the two InvalidateDescribeCache calls after SQLPrepare. Correctness today relies on the invariant that any code path reaching ResolveNullParamType has either just run SQLPrepare on the handle, or is re-executing the same SQL on the same HSTMT. Under that invariant, stale-cache-on-recycled-pointer reads cannot happen.

But the one issue remain:

Memory leak. When an HSTMT is freed without a subsequent SQLPrepare on the recycled pointer, its entries are orphaned forever. Three concrete sources (please confirm this yourself):
Connection::disconnect() marks child STMTs as _implicitly_freed, their cache entries are never touched.
Pooled connections (SQL_ATTR_RESET_CONNECTION path) keep the DBC alive across checkin, so freed STMT pointers rarely get recycled back to the same DBC and cache entries accumulate over the pool's lifetime.
Cursors that never call SQLPrepare (no-param execute >> SQLExecDirect, metadata methods, failed executes) leak on close even with pointer recycling.

Suggested fix — make cache lifetime ≤ handle lifetime by construction:

void SqlHandle::free() {
    if (_handle && SQLFreeHandle_ptr) {
        // ... existing shutdown / implicitly_freed checks ...
        if (_type == SQL_HANDLE_STMT) {
            InvalidateDescribeCache(_handle);   // GH-610
        }
        SQLFreeHandle_ptr(_type, _handle);
        _handle = nullptr;
    }
}

Plus the same call in SQLFreeHandle_wrap and inside the _implicitly_freed short-circuit branch so disconnect-time STMTs are evicted too. The existing SQLPrepare invalidations can stay as harmless safety nets.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -399,6 +404,76 @@ SQLTablesFunc SQLTables_ptr = nullptr;
...
+
+static DescribedParamInfo ResolveNullParamType(SQLHANDLE hStmt, int paramIndex) {
+    // 1. Check cache (shared/read lock — concurrent readers allowed)
+    {
+        std::shared_lock<std::shared_mutex> lock(g_describeCacheMutex);
```
sumitmsft: Process-global mutex `g_describeCacheMutex` becomes a serialization point for all cursors across all connections that bind NULL parameters. While the read path uses shared_lock, every cache miss takes the exclusive lock for the writeback. Under multi-threaded workloads with many distinct prepared statements, this is a contention hotspot.

One way to avoid this contention is to move the cache onto `SqlHandle` (one mutex per statement handle, no contention across statements).

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Made the required changes.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Made the required changes.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -232,12 +233,23 @@ class DriverLoader {
...
 // RAII wrapper around ODBC handles (ENV, DBC, STMT).
 // Use `std::shared_ptr<SqlHandle>` (alias: SqlHandlePtr) for shared ownership.
 //-------------------------------------------------------------------------------------------------
+
+// GH-610: Cached result of SQLDescribeParam for a single parameter.
+struct DescribedParamInfo {
+    SQLSMALLINT sqlType;
+    SQLULEN columnSize;
+    SQLSMALLINT decimalDigits;
+    bool succeeded;  // false = SQLDescribeParam failed, used SQL_VARCHAR fallback
```
bewithgaurav: `succeeded` is set but never read on cache hit, is it actually needed?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2562,7 +2591,14 @@ SQLRETURN BindParameterArray(SQLHANDLE hStmt, const py::list& columnwise_params,
 
                     dataPtr = nullBuffer;
                     bufferLength = 1;
-                    LOG("BindParameterArray: SQL_C_DEFAULT bound - param_index=%d", paramIndex);
+
+                    // Override info fields so SQLBindParameter below uses resolved type
+                    const_cast<ParamInfo&>(info).paramSQLType = resolvedSqlType;
```
bewithgaurav: `paramInfos` has a pre-existing `const` on its signature (`SQLExecuteMany_wrap` / `BindParameterArray`), but this cache resolve path needs to write back into it 

the `const_cast` workaround works, but since the function does mutate `paramInfos` now, dropping the `const` from the signature would be cleaner.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2673,7 +2711,7 @@ SQLRETURN SQLExecuteMany_wrap(const SqlHandlePtr statementHandle, const std::u16
             py::list rowParams = columnwise_params[rowIndex];
 
             std::vector<std::shared_ptr<void>> paramBuffers;
-            rc = BindParameters(hStmt, rowParams, const_cast<std::vector<ParamInfo>&>(paramInfos),
+            rc = BindParameters(*statementHandle, hStmt, rowParams, const_cast<std::vector<ParamInfo>&>(paramInfos),
```
bewithgaurav: same thing here as stated in prev comment

tests/test_004_cursor.py
```diff
@@ -2184,6 +2184,127 @@ def test_executemany_MIX_NONE_parameter_list(cursor, db_connection):
         db_connection.commit()
 
 
+def test_map_sql_type_none_returns_sql_unknown_type():
```
bewithgaurav, line 2187: coverage bot shows 69% diff coverage on `ddbc_bindings.cpp`. 
I understand coverage on failure cases might be tricky - but could you recheck if any of the uncovered lines are reachable from existing tests?
If not thats fine, I checked but most of uncovered lines look difficult to repro normally, hence this is more of an FYI comment.

bewithgaurav (changes requested):
moving the cache to per-handle lgtm
however I've left a few comments below, nothing very major but worth a look before we merge

mssql_python/pybind/ddbc_bindings.h
jahnvi480: Made the required changes

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Removed the unnecessary constants

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Made the fix

tests/test_004_cursor.py
jahnvi480: I tried, but uncovered lines are SQLDescribeParam failure paths + cache invalidation logging, which require specific ODBC error conditions difficult to reproduce in tests

mssql_python/cursor.py
sumitmsft: @jahnvi480 were these integration tests added?

jahnvi480: Yes

bewithgaurav (dismissed):
lgtm

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -262,6 +273,13 @@ class SqlHandle {
     // before freeing the DBC handle.
     void markImplicitlyFreed();
 
+    // GH-610: Per-handle SQLDescribeParam result cache.
+    // Keyed by 0-based parameter index. Populated on first NULL param
+    // execution, cleared on SQLPrepare (new SQL = new param types) and
+    // on handle free. No mutex needed — each handle is used by one thread.
```
gargsaumya: The comment assumes "each handle is used by one thread," but this assumption is not enforced. What happens if a user shares a cursor across threads (even though it's not recommended)?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1787,6 +1808,8 @@ SQLRETURN SQLExecute_wrap(const SqlHandlePtr statementHandle,
                     rc, (void*)hStmt);
                 return rc;
             }
+            // GH-610: Clear per-handle describe cache (new prepare = new param types)
+            statementHandle->clearDescribeCache();
```
gargsaumya, line 1812: The cache is cleared when a new statement is prepared, but what if statement is re-used with same SQL but different parameter semantics?
Test executing INSERT INTO t1 VALUES (?) then INSERT INTO t2 VALUES (?, ?) on the same cursor to ensure cache doesn't mix parameter types. And can we add a test for this scenario.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -2621,6 +2657,8 @@ SQLRETURN SQLExecuteMany_wrap(const SqlHandlePtr statementHandle, const std::u16
         LOG("SQLExecuteMany: SQLPrepare failed - rc=%d", rc);
         return rc;
     }
+    // GH-610: Clear per-handle describe cache (new prepare = new param types)
+    statementHandle->clearDescribeCache();
```
gargsaumya, line 2661: same comment as line 1812

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -627,33 +669,15 @@ SQLRETURN BindParameters(SQLHANDLE hStmt, const py::list& params,
...
-                        decimalDigits = 0;
-                    } else {
-                        sqlType = describedType;
-                        columnSize = describedSize;
-                        decimalDigits = describedDigits;
-                    }
+                    auto resolved = ResolveNullParamType(handle, hStmt, paramIndex);
+                    sqlType = resolved.sqlType;
+                    columnSize = resolved.columnSize;
+                    decimalDigits = resolved.decimalDigits;
```
gargsaumya, line 680: If ResolveNullParamType fails and returns a fallback type, this fallback is cached. On the next execution, if the actual parameter is not NULL, the code might use the cached (wrong) fallback type.

Example scenario:
First call: All params are NULL, SQLDescribeParam fails, cache stores `SQL_VARCHAR`
Second call: Param 0 is now an integer 42
Code might still use cached `SQL_VARCHAR` from step 1

Actually looking at the code, the cache is only consulted when `sqlType == SQL_UNKNOWN_TYPE`, which only happens when the param is NULL. So this might be OK, but it's subtle and deserves a test. So can we add a test: Execute with NULL, then execute with non-NULL value on same statement.

sumitmsft (dismissed):

mssql_python/pybind/ddbc_bindings.h
jahnvi480: Updated the comment to reference the ODBC spec

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: SQLPrepare always calls clearDescribeCache() when SQL changes. Added test_gh610_cache_no_mix_across_different_sql which executes INSERT INTO t1 VALUES (?) then INSERT INTO t2 VALUES (?, ?) on the same cursor and verifies both tables get correct data with no type mixing.

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: the cache is only consulted inside the SQL_C_DEFAULT switch case, which is only entered when paramCType == SQL_C_DEFAULT. This only happens when _map_sql_type(None) returns SQL_C_DEFAULT. For non-NULL params (e.g., 42), _map_sql_type returns SQL_C_TINYINT, which enters the SQL_C_STINYINT case — cache is never touched. Added test_gh610_cache_null_then_nonnull_same_statement which executes (None) → (42) → (None) on the same prepared statement and verifies all values are correct.

sumitmsft (approved):

gargsaumya (approved):

Issue #482, opened 2026-03-23
Bulkcopy: Cannot cast Row object as tuple


ebj-jyskebank-dk:
### List of Rows from cursor.fetchall() cannot be used directly in bulkcopy

Data retrieved using `cursor.fetchall()` is of type `List[Row]`. Intuitively I would expect to be able to use these objects directly in `cursor.bulkcopy()`, but I get an exception.

```
  File "/home/vscode/.local/lib/python3.14/site-packages/mssql_python/cursor.py", line 2706, in bulkcopy
    raise type(e)(str(e)) from None
ValueError: Expected tuple, got: 'Row' object cannot be cast as 'tuple'
```

### To reproduce

```python
import mssql_python

conn = mssql_python.connect("...")
cursor = conn.cursor()

cursor.execute(f"SELECT TOP 10 * FROM table_A")
data = cursor.fetchall()
copy_result = cursor.bulkcopy("table_B", data)

conn.close()
```

(This is a simplified example - in reality I stumpled on this while trying to migrate data between two databases on different servers).

### Expected behavior
I expected the `bulkcopy` method to accept a data argument of type `List[Row]`.

### Further technical details
Python version: 3.14
SQL Server version: Microsoft SQL Azure 12.0.2000.8
Operating system: Linux 39a7e4bda3b8 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux

bewithgaurav:
@ebj-jyskebank-dk - thanks a lot for raising this issue & sharing the examples.
we'll get back to you on this as we triage this further this week

jahnvi480:
@ebj-jyskebank-dk Thank you for raising this issue and making our driver better. We have merged the PR for supporting Row object, it will be available in our upcoming release.

PR #615, opened 2026-06-02
FEAT: Accept Row objects in bulkcopy without manual tuple conversion

jahnvi480:

> [AB#43796](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/43796)

> GitHub Issue: #482

### Summary   
This pull request improves the robustness of the `bulkcopy` function in `mssql_python/cursor.py` by ensuring that all data rows passed to the Rust layer are properly formatted as tuples. This prevents type errors when using different iterable types, such as lists or custom `Row` objects, as input.

Data validation and conversion:

* Added a helper function `_ensure_tuples` to convert each item in the provided data to a tuple if it is a list or `Row` object, and raise a `TypeError` for unsupported types. This ensures compatibility with the strict type requirements of the Rust backend.
* Updated the call to `bulkcopy` to use `_ensure_tuples(data)` instead of simply iterating over `data`, enforcing consistent tuple formatting for each row.


bewithgaurav (commented):
EDIT: adding a new comment as per discussions from the review call.

mssql_python/cursor.py
```diff
@@ -2997,11 +2997,26 @@ def bulkcopy(
             )
             pycore_cursor = pycore_connection.cursor()
 
+            # Auto-convert Row/list objects to tuples for the Rust layer.
+            # mssql_py_core expects native tuples; Row objects (from fetchmany)
+            # are iterable but fail the strict type check in Rust.
+            def _ensure_tuples(iterable):
```
bewithgaurav: this helper is pure Python and testable without a SQL connection. a small unit test would lock down the Row/list/tuple/TypeError behavior and prevent silent regressions.

mssql_python/cursor.py
```diff
@@ -2997,11 +2997,26 @@ def bulkcopy(
...
+                for item in iterable:
+                    if isinstance(item, tuple):
+                        yield item
+                    elif isinstance(item, (list, Row)):
+                        yield tuple(item)
+                    else:
+                        raise TypeError(
```
bewithgaurav: just flagging a pre-existing behavior (saw this while reviewing):
- the Rust layer (`mssql-py-core/src/cursor.rs`) wraps the Python iterator in a `filter_map` that **silently drops** rows on error instead of propagating
- there is a `TODO` comment in the Rust code acknowledging this
- so if this `TypeError` fires mid-stream, the row gets skipped rather than raising
- I'll open a follow-up task to check this on the `mssql-rs` side

mssql_python/cursor.py
bewithgaurav: this still says `Iterable[Union[Tuple, List]]`. since the PR now explicitly handles `Row`, `list`, and `tuple`, worth updating to `Iterable[Union[Tuple, List, "Row"]]` (or `Iterable[Sequence]` if you want to stay generic and future-proof). same for the docstring a few lines below.

jahnvi480, line 2861: Create a doc for this 1 pager

jahnvi480: Check if we can checkfirst object and then convert directly to tuple

jahnvi480: check how other drivers are doing this, also make sure if this can be solved using Row.py

mssql_python/cursor.py
```diff
@@ -2997,11 +2997,26 @@ def bulkcopy(
             )
             pycore_cursor = pycore_connection.cursor()
 
+            # Auto-convert Row/list objects to tuples for the Rust layer.
+            # mssql_py_core expects native tuples; Row objects (from fetchmany)
+            # are iterable but fail the strict type check in Rust.
+            def _ensure_tuples(iterable):
+                for item in iterable:
+                    if isinstance(item, tuple):
+                        yield item
```
bewithgaurav: Tuple instance checking will have perf overhead, if the iterator already yields tuple (which is the normal scenario) 
This section will add a type checking perf overhead (will be significant in bcp scale)

mssql_python/cursor.py
```diff
@@ -2997,11 +2997,26 @@ def bulkcopy(
...
+                    elif isinstance(item, (list, Row)):
+                        yield tuple(item)
```
bewithgaurav: As discussed, this seems like a very targeted fix on the flow of `cursor.execute` -> `fetchall()` -> `bulkcopy()`

Getting `Row` objects from iterator is very specific to using `cursor.fetchall()` as a source, instead of doing this we might also thing of returning tuples from `fetchall()` (can be param enabled e.g. `fetchall(tuples=True)`) if the fix is JUST to tackle this scenario

bewithgaurav (changes requested):
Adding a few more comments post the PR review discussion just for reference, will wait for the doc

mssql_python/cursor.py
```diff
@@ -3034,11 +3036,33 @@ def bulkcopy(
...
+            # Row objects from fetch methods are converted using direct _values
+            # access (4x faster than __iter__). All other types raise TypeError.
+            def _ensure_tuples(iterable):
+                it = iter(iterable)
+                first = next(it, None)
+                if first is None:
+                    return
+                if isinstance(first, tuple):
+                    yield first
+                    yield from it
```
bewithgaurav: one perf note. the generator (`yield from`) adds measurable overhead per row on the tuple passthrough path. switching to `itertools.chain` eliminates this since chain is implemented in C and avoids per-item Python generator frame cost.

since `next(it)` already consumed the first row, `chain((first,), it)` replays that first row before the rest of the iterator. no copying involved.

the changes:
- add `from itertools import chain` inside the function (cached after first call)
- `return iter(())` for empty case instead of bare `return`
- `return chain((first,), it)` instead of `yield first` / `yield from it`
- `return (tuple(item._values) for item in chain((first,), it))` instead of the explicit yield loop

benchmarked at 1M rows (10 cols), the chain approach brings tuple passthrough overhead down from ~45% to ~6% compared to no wrapper at all.

mssql_python/cursor.py
```diff
@@ -3034,11 +3036,33 @@ def bulkcopy(
             )
             pycore_cursor = pycore_connection.cursor()
 
+            # Enforce the bulkcopy type contract: only tuple and Row accepted.
+            # Rust (mssql_py_core) requires native PyTuple via cast::<PyTuple>().
+            # Row objects from fetch methods are converted using direct _values
```
bewithgaurav: could you add a regression test for the mentioned bulkcopy Row path? 
something that passes `Row` objects from `fetchmany`/`fetchall` into `bulkcopy` and verifies the data lands correctly.

bewithgaurav (changes requested):
added a note on perf improvement and requesting adding regression tests

mssql_python/cursor.py
jahnvi480: Made required changes

mssql_python/cursor.py
jahnvi480: Added tests.

bewithgaurav (approved):
looks good.

one known constraint of the check-first pattern in this PR (we can't fix this since we're optimizing for perf):

- only the first item's type is validated. mixed-type iterables (e.g. tuples + Row in the same batch) would pass the first-item check, then fail with a Rust cast error.

I'd say this not a real concern since bulkcopy data is always homogeneous in practice, but adding this here just for our reference later on

subrata-ms (approved):
looks good to me as we are iterating only the first record to determine the type and that is agreed upon during design discussion.

Issue #612, opened 2026-06-01
mysql-python loses anything past the first message


sam-leonard-ct:
### Describe the bug

mssql-python doesn't return anything after the first message

### To reproduce

This is the minimum test case I could get to reproduce the bug, this script passes when cursor is from pyodbc, but not when cursor is from mysql-python.

```python
cursor.execute("""
PRINT('hi');
PRINT('ih');
""")

assert cursor.messages == [('[01000] (0)', '[Microsoft][ODBC Driver 18 for SQL Server][SQL Server]hi')]

assert cursor.nextset()
# on mysql-python this second message is missing
assert cursor.messages == [('[01000] (0)', '[Microsoft][ODBC Driver 18 for SQL Server][SQL Server]ih')]

assert not cursor.nextset()
```

### Expected behavior
All messages from the operation should be returned via cursor.messages at some point before cursor.nextset() returns False, preferably in the order detailed in the above script.

### Further technical details
Python version: 3.12.11
SQL Server version: SQL SERVER 2017
Operating system: macOS 26.5

jahnvi480:
Thanks for the detailed report and repro, @sam-leonard-ct!

We've triaged this and confirmed it's a bug. The root cause is that cursor.nextset() calls SQLMoreResults under the hood but doesn't check for SQL_SUCCESS_WITH_INFO to capture diagnostic messages (like PRINT output). The first PRINT is captured correctly by execute(), but any subsequent messages delivered via SQLMoreResults were silently dropped.

We'll have a PR up shortly with the fix.

sam-leonard-ct:
> Thanks for the detailed report and repro, [@sam-leonard-ct](https://github.com/sam-leonard-ct)!
> 
> We've triaged this and confirmed it's a bug. The root cause is that cursor.nextset() calls SQLMoreResults under the hood but doesn't check for SQL_SUCCESS_WITH_INFO to capture diagnostic messages (like PRINT output). The first PRINT is captured correctly by execute(), but any subsequent messages delivered via SQLMoreResults were silently dropped.
> 
> We'll have a PR up shortly with the fix.

Awesome, thanks for the quick debugging!

jahnvi480:
@sam-leonard-ct We have merged the PR and it this bug will be resolved in our next release. Thank you for detecting bug and helping us improve our driver. Keep trying out our driver!

PR #618, opened 2026-06-02
FIX: Capture PRINT messages in nextset() via SQL_SUCCESS_WITH_INFO

jahnvi480:

> [AB#45395](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45395)

> GitHub Issue: #612

### Summary   
This pull request addresses an issue where diagnostic messages (such as SQL Server PRINT output) from subsequent result sets were not properly captured when using the `nextset()` method. The fix ensures that all messages are collected, even across multiple result sets, and adds comprehensive tests to verify this behavior.

**Bug Fix: Diagnostic Message Handling**
- Updated the `nextset()` method in `cursor.py` to capture diagnostic messages (e.g., PRINT output) when SQL returns `SQL_SUCCESS_WITH_INFO`, ensuring messages from all result sets are preserved.

**Testing Improvements**
- Added several new tests in `test_004_cursor.py` to verify that PRINT messages are correctly captured across multiple result sets and that messages from previous result sets are cleared appropriately:
  - Test for multiple PRINT statements across result sets via `nextset()`
  - Test for PRINT messages interleaved with SELECT statements
  - Test for three consecutive PRINT messages across `nextset()` calls
  - Test to ensure `nextset()` clears messages from the previous result set

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -2759,12 +2759,12 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
 
         return pyarrow.RecordBatchReader.from_batches(schema, batch_generator())
 
-    def nextset(self) -> Union[bool, None]:
+    def nextset(self) -> bool:
         """
         Skip to the next available result set.
 
         Returns:
-            True if there is another result set, None otherwise.
+            True if there is another result set, False otherwise.
 
         Raises:
             Error: If the previous call to execute did not produce any result set.
@@ -2784,6 +2784,11 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         # Skip to the next result set
         ret = ddbc_bindings.DDBCSQLMoreResults(self.hstmt)
         check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
+
+        # Capture diagnostic messages (e.g. PRINT output) on SQL_SUCCESS_WITH_INFO.
+        # Without this, messages from subsequent result sets are silently lost (GH-612).
+        if ret == ddbc_sql_const.SQL_SUCCESS_WITH_INFO.value and self.hstmt:
+            self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
 
         if ret == ddbc_sql_const.SQL_NO_DATA.value:
             logger.debug("nextset: No more result sets available")
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -810,6 +810,25 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
                 ddbc_error="",
             )
 
+    def _capture_diagnostics(self, ret: int) -> None:
+        """Append diagnostic messages to self.messages when the return code
+        indicates records may be present.
+
+        Captures on SQL_SUCCESS_WITH_INFO (info/warning messages) and
+        SQL_NO_DATA (trailing diagnostics attached by SQLMoreResults,
+        e.g. a PRINT after the final result set).
+
+        Skips SQL_SUCCESS to avoid the ~10 ms overhead of scanning the
+        driver's internal state when no records exist.  SQL_ERROR is
+        handled separately by check_error() which extracts diagnostics
+        and raises.
+        """
+        if self.hstmt and ret in (
+            ddbc_sql_const.SQL_SUCCESS_WITH_INFO.value,
+            ddbc_sql_const.SQL_NO_DATA.value,
+        ):
+            self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
+
     def _ensure_pyarrow(self) -> Any:
         """
         Import and return pyarrow or raise ImportError accordingly.
@@ -1518,13 +1537,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
             self._reset_cursor()
             raise
 
-        # Capture diagnostic messages only on SQL_SUCCESS_WITH_INFO.
-        # SQL_SUCCESS has no records — calling DDBCSQLGetAllDiagRecords on it
-        # costs ~10ms/call (driver scans internal state to find nothing).
-        # SQL_ERROR is already handled by check_error() above which extracts
-        # diagnostics and raises.
-        if ret == ddbc_sql_const.SQL_SUCCESS_WITH_INFO.value and self.hstmt:
-            self.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(self.hstmt))
+        self._capture_diagnostics(ret)
 
         self.last_executed_stmt = operation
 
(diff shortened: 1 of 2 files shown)


KellyKr:
I just ran into this and see the fix will be coming. Nice!

To ask for a bit more: the type of cursor.messages is list[Unknown], it would be nice to have correct typing on it.

mssql_python/cursor.py
```diff
@@ -2799,6 +2799,11 @@ def nextset(self) -> Union[bool, None]:
         ret = ddbc_bindings.DDBCSQLMoreResults(self.hstmt)
         check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
 
+        # Capture diagnostic messages (e.g. PRINT output) on SQL_SUCCESS_WITH_INFO.
+        # Without this, messages from subsequent result sets are silently lost (GH-612).
+        if ret == ddbc_sql_const.SQL_SUCCESS_WITH_INFO.value and self.hstmt:
```
sumitmsft: Can you validate if this capture pattern is duplicated in other places? If so, I think making a helper method would help to avoid duplicating this pattern in the code base.

sumitmsft: `SQLMoreResults` can return `SQL_NO_DATA` with diagnostic records still attached. The trailing `PRINT` after a final `SELECT` lands on exactly this return code:

``` python
PRINT('before'); 
SELECT 1; 
PRINT('after');  -- 'after' attaches to SQL_NO_DATA
```
The current fix only captures on `SQL_SUCCESS_WITH_INFO`, so PRINT('after') will still be silently dropped.

Can you tak care of this case as well?

tests/test_004_cursor.py
```diff
@@ -6452,6 +6452,106 @@ def test_cursor_messages_with_error(cursor):
...
+
+    while cursor.nextset():
+        all_messages.extend(cursor.messages)
+        if cursor.description:
+            all_rows.extend(cursor.fetchall())
+            # fetchall may also produce diagnostics
+            all_messages.extend(cursor.messages)
+
+    # Verify both PRINT messages were captured somewhere across all result sets
+    combined_text = " ".join(m[1] for m in all_messages)
```
sumitmsft: `all_messages` includes messages captured by `execute(), nextset(), and fetchall()`. So even if `nextset()` silently drops the "after select" PRINT, the test still passes because `fetchall()` independently captures the same diagnostics from the statement handle.

So, this test passes today, but it would also pass if someone reverted the entire fix in this PR. It can't tell the difference between "nextset() works" and "nextset() is broken but fetchall() saved us."

I'd suggest only collect messages from the `nextset()` call itself because the test fails if and only if `nextset()` actually drops the message which is exactly what we want.

tests/test_004_cursor.py
```diff
@@ -6452,6 +6452,106 @@ def test_cursor_messages_with_error(cursor):
     assert "After error" in cursor.messages[0][1], "Message should be from after the error"
 
 
+def test_cursor_messages_nextset_multiple_prints(cursor):
```
sumitmsft, line 6455: Add a test that covers the case where `nextset()` returns `SQL_NO_DATA` while diagnostic info is still attached (the most common cause of bug reports). The trailing-`PRINT-after-final-SELECT` case is the main customer pain point and is not unambiguously tested.

tests/test_004_cursor.py
```diff
@@ -6452,6 +6452,106 @@ def test_cursor_messages_with_error(cursor):
...
+    """Test that PRINT messages from subsequent result sets are captured via nextset().
+
+    Regression test for GH-612: PRINT messages after the first one were lost
+    because nextset() did not capture SQL_SUCCESS_WITH_INFO diagnostics.
+    """
+    cursor.execute("""
```
sumitmsft, line 6461: Tests use `PRINT('hi')` syntax (with parentheses). T-SQL `PRINT ` accepts both with and without parens; consistency with the rest of the test file (use PRINT 'Hello world!') is preferable.

mssql_python/cursor.py
```diff
@@ -2773,12 +2773,12 @@ def batch_generator():
 
         return pyarrow.RecordBatchReader.from_batches(schema, batch_generator())
 
-    def nextset(self) -> Union[bool, None]:
+    def nextset(self) -> bool:
```
sumitmsft: Per [PEP 249](https://peps.python.org/pep-0249/#nextset), `nextset() `should return `True `or `None ` never `False`. `mssql-python` returns True/False, which is a pre-existing deviation. By tightening the type to `bool`, we're now formally committing to it. Future PEP-249 alignment would become a breaking change.

Can we either 
(a) revert to Optional[bool] to keep the door open - to migrate to PEP-249 semantics (True/None) later without a breaking signature change.
or (b) keep bool but call out the deviation in the CHANGELOG so it's a deliberate decision? This will let users know not to expect pyodbc-compatible `None`.

sumitmsft:
> I just ran into this and see the fix will be coming. Nice!
> 
> To ask for a bit more: the type of cursor.messages is list[Unknown], it would be nice to have correct typing on it.

@jahnvi480 While we're touching the messages handling, can we also fix the typing? Today `self.messages = []` is inferred as `list[Unknown]` by IDEs and type-checkers.

The C++ binding (`SQLGetAllDiagRecords`) always returns `Tuple[str, str]` per entry. `(sql_state_with_error_code, message_text)` and the tests already assert this shape. So. do you think it's a safe fix?

mssql_python/cursor.py
jahnvi480: Yes while working on this issue I did see that we have missed capturing this in other APIs as well, I have created a helper method which will be later reused.

jahnvi480: Have added capturing for SQL_NO_DATA as well

tests/test_004_cursor.py
jahnvi480: Made required changes

tests/test_004_cursor.py
jahnvi480: Added test to check this as well

mssql_python/cursor.py
jahnvi480: Reverted to Optional[bool] with docstring noting the PEP-249 deviation

tests/test_004_cursor.py
jahnvi480: The CI run disproved this assumption, the ODBC Driver 18 delivers a trailing PRINT as its own result set (SQL_SUCCESS_WITH_INFO), not as diagnostics attached to SQL_NO_DATA. The actual flow for SELECT 1; PRINT 'trailing' is:

SQLMoreResults → SQL_SUCCESS_WITH_INFO (carries the PRINT message) → nextset() returns True
SQLMoreResults → SQL_NO_DATA (no diagnostics) → nextset() returns False
I've updated the test to match the real driver behavior — it now asserts nextset() returns True and captures the trailing PRINT via SQL_SUCCESS_WITH_INFO.

That said, _capture_diagnostics() still defensively captures on SQL_NO_DATA too, in case other ODBC drivers behave differently. It's zero-cost when no records exist.

jahnvi480:
Thanks @KellyKr and @sumitmsft for detailed review, I have verified and updated the typing for messages.

sumitmsft (approved):

subrata-ms (approved):

Issue #621, opened 2026-06-03
FEAT: Add session audit context setter/getter API (set_audit_context / get_audit_context) for compliance tracing


kapilsamant:
### Is your feature request related to a problem? Please describe.
There is no built-in way to attach session-level auditing or tracing metadata to a Connection in mssql-python. Applications that need to propagate end-user identity, module names, or action context to SQL Server for auditing, or Extended Events tracing must manually execute raw sp_set_session_context calls, handle input validation themselves, and track which keys have been set. This is error-prone, verbose, and inconsistent across codebases.

### Describe the solution you'd like
A setter/getter pair on the connection class
set_audit_context()
Accepts well-known keys (application, module, action, user_id) as named parameters and arbitrary extra keys via **kwargs.
Validates key/value lengths against SQL Server limits (128-char keys, 8000-char values) before executing.
Uses parameterized queries to call sp_set_session_context, preventing SQL injection.
Supports a read_only flag to lock keys for the remainder of the session.
Raises appropriate ProgrammingError / InterfaceError exceptions on invalid input or closed connections.
get_audit_context()
Returns the current session audit context as a dictionary. 
The getter reads from the local cache without making a server round-trip

### Describe alternatives you've considered
Raw cursor execution — Users call cursor.execute("EXEC sp_set_session_context @key=?, @value=?", ...) directly. This works but requires boilerplate, duplicates validation logic, and lacks discoverability.
Connection string properties — Encoding audit metadata into the connection string (e.g., Application Name). This only covers a single field, is immutable after connect, and doesn't flow to SESSION_CONTEXT().

### Additional context
Applications that need to comply with regulations (GDPR, PCI-DSS etc) must propagate end-user identity, module context, and action tracking to SQL Server for audit trail generation and compliance reporting. Without a dedicated setter API, developers are forced to manually execute raw sp_set_session_context stored procedure calls, implement their own input validation, and maintain their own getter/cache logic

jahnvi480:
Thanks @kapilsamant for the PR! We'll review this based on our current priority list and get back to you. Appreciate the contribution!

saurabh500:
@kapilsamant can you open a discussion item with an API proposal. I see that the design and intention is scattered across the issue and the PR, allowing little chance to ask questions about the API.

saurabh500:
An example discussion where we built a non-standard API with community is https://github.com/microsoft/mssql-python/discussions/414

kapilsamant:
Done #624

PR #622, opened 2026-06-03
FEAT: Add session audit context API (set_audit_context / get_audit_context)

kapilsamant:

> GitHub Issue: #621

### Summary   
Add session-level auditing/tracing API to the Connection class via two new methods:

set_audit_context() -> Sets session metadata (application, module, action, user_id, and arbitrary extra keys) on the current connection using sp_set_session_context. Supports read_only mode, input validation with length limits, parameterized query execution, and local caching of set values.
get_audit_context()-> Returns a copy of the locally cached session context dictionary.
Values set through this API are visible to SESSION_CONTEXT() in T-SQL, Extended Events, sys.dm_exec_sessions, and audit specifications.


kapilsamant:
> ```
> @microsoft-github-policy-service agree
> ```

Issue #629, opened 2026-06-08
1.8.0 fails in writing to varchar(max)


sravikrishna:
The code is same as was mentioned in another bug I recently created.

https://github.com/microsoft/mssql-python/issues/609

mssql-python : 1.8.0
Windows python 3.13.13

my program reads few thousand rows from fetchmany and passes it to another connection where it is written via executemany.

Traceback (most recent call last):
  File "C:\mytools\dbatools\pybin\sqlcopy.py", line 183, in <module>
    copy_data_from_sql_file(driver,sql_file,t_table,s_cxn,t_cxn)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\mytools\dbatools\pybin\sqlcopy.py", line 125, in copy_data_from_sql_file
    t_cursor.executemany(ins_sql,rows)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "C:\winpython\Lib\site-packages\mssql_python\cursor.py", line 2389, in executemany
    self.execute(operation, row)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "C:\winpython\Lib\site-packages\mssql_python\cursor.py", line 1476, in execute
    paraminfo = self._create_parameter_types_list(param, param_info, parameters, i)
  File "C:\winpython\Lib\site-packages\mssql_python\cursor.py", line 1003, in _create_parameter_types_list
    sql_type, c_type, column_size, decimal_digits, is_dae = self._map_sql_type(
                                                            ~~~~~~~~~~~~~~~~~~^
        parameter, parameters_list, i, min_val=min_val, max_val=max_val
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "C:\winpython\Lib\site-packages\mssql_python\cursor.py", line 701, in _map_sql_type
    raise TypeError(
        "Unsupported parameter type: The driver cannot safely convert it to a SQL type."
    )
TypeError: Unsupported parameter type: The driver cannot safely convert it to a SQL type.

As the error is coming from executemany, it means that it is the write which is failing.

The structure of the table being written to is

	[database_name] [varchar](50) NOT NULL,
	[login_name] [varchar](100) NULL,
	[host_name] [varchar](100) NULL,
[...]

PR #630, opened 2026-06-10
FIX: Handle Row objects in executemany DAE fallback path (GH-629)

gargsaumya:

> [AB#45679](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45679)

> GitHub Issue: #629 

### Summary   
  
This pull request introduces a compatibility improvement to the `executemany` method in `mssql_python/cursor.py`. It ensures that when iterating through rows of parameters, any Row objects are converted to tuples before execution. This conversion is necessary because the internal `_map_sql_type` logic only recognizes primitive types, and not Row objects, which prevents potential type errors during execution.

Parameter compatibility improvement:

* Updated the `executemany` method to convert Row objects to tuples before passing them to `execute`, ensuring compatibility with the internal type mapping logic.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -2413,6 +2413,11 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
                 "DAE parameters detected. Falling back to row-by-row execution with streaming.",
             )
             for row in seq_of_parameters:
+                # Convert Row objects to tuples for compatibility with execute()
+                # Row objects from fetch APIs must be converted since _map_sql_type
+                # only recognizes primitive types (str, int, etc.), not Row objects
+                if hasattr(row, "_values") and hasattr(row, "__iter__"):
+                    row = tuple(row)
                 self.execute(operation, row)
             return
 
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -1454,6 +1454,12 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
             if isinstance(parameters, tuple) and len(parameters) == 1:
                 if isinstance(parameters[0], (tuple, list, dict)):
                     actual_params = parameters[0]
+                elif isinstance(parameters[0], Row):
+                    # A Row (e.g. from fetchone()) is a sequence of column values.
+                    # Normalize it to a tuple so the downstream binding logic, which
+                    # only handles tuple/list/dict, can unwrap it into individual
+                    # parameters instead of treating the whole Row as one value.
+                    actual_params = tuple(parameters[0])
                 else:
                     actual_params = parameters
             else:
@@ -2413,11 +2419,6 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
                 "DAE parameters detected. Falling back to row-by-row execution with streaming.",
             )
             for row in seq_of_parameters:
-                # Convert Row objects to tuples so execute() can unwrap them as parameters.
-                # execute() only unwraps tuple/list/dict arguments, not Row objects,
-                # so passing a Row directly would treat it as a single scalar parameter.
-                if isinstance(row, Row):
-                    row = tuple(row)
                 self.execute(operation, row)
             return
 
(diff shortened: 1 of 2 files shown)


mssql_python/cursor.py
```diff
@@ -2413,6 +2413,11 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
                 "DAE parameters detected. Falling back to row-by-row execution with streaming.",
             )
             for row in seq_of_parameters:
+                # Convert Row objects to tuples so execute() can unwrap them as parameters.
+                # execute() only unwraps tuple/list/dict arguments, not Row objects,
+                # so passing a Row directly would treat it as a single scalar parameter.
+                if isinstance(row, Row):
```
bewithgaurav: uncovered a day 0 bug on `Row` ops - bug's root cause is in `execute()` itself and hence suggesting the fix there.

`cursor.execute(sql, row_obj)` hits the same TypeError on `main` and on this branch. quick repro:

```
cur.execute("SELECT * FROM t WHERE id = 1")
r = cur.fetchone()
cur.execute("INSERT INTO t VALUES (?, ?, ?)", r)
# TypeError: Unsupported parameter type
```

same root cause. 
execute() only unpacks (tuple, list, dict) - line 1455, so a `Row` falls through and `_map_sql_type` treats the whole row as one value.

if we unpack Row there itself well, this block isn't needed.
the per-row self.execute(operation, row) below gets solved & no pre-patch will be needed.

ends up smaller diff, broader fix, and execute(sql, row) also fixes.

tests/test_004_cursor.py
```diff
@@ -16594,3 +16594,54 @@ def test_executemany_multi_column_with_large_decimal(cursor, db_connection):
...
+        large_text = "X" * 5000
+        cursor.execute("INSERT INTO #pytest_gh629_source VALUES (?, ?)", (1, large_text))
+        cursor.execute("INSERT INTO #pytest_gh629_source VALUES (?, ?)", (2, large_text))
+        db_connection.commit()
+
+        # Fetch rows as Row objects
+        cursor.execute("SELECT * FROM #pytest_gh629_source")
+        rows = cursor.fetchmany(10)  # Returns Row objects
+        assert len(rows) == 2
+        assert type(rows[0]).__name__ == "Row"
```
bewithgaurav: `type(rows[0]).__name__ == "Row"` is fragile. Row is already imported in this module, better to just use `isinstance(rows[0], Row)`.

bewithgaurav (changes requested):
requesting changes, there's an underlying day 0 bug here, added details. 
happy to re-review once the unwrap moves into execute().

tests/test_004_cursor.py
```diff
@@ -16594,3 +16594,54 @@ def test_executemany_multi_column_with_large_decimal(cursor, db_connection):
     finally:
         cursor.execute("DROP TABLE IF EXISTS #pytest_gh609_multi")
         db_connection.commit()
+
+
+def test_executemany_row_objects_with_varchar_max_dae(cursor, db_connection):
```
bewithgaurav, line 16766: the fix now lives in execute() but the test still only covers the executemany DAE path 
a few lines in the same fixture would guard this fix surface:

```python
cursor.execute("SELECT * FROM #pytest_gh629_source WHERE id = 1")
row = cursor.fetchone()
cursor.execute("INSERT INTO #pytest_gh629_target VALUES (?, ?)", row)
```

otherwise a future refactor of execute()'s unwrap can re-break this

bewithgaurav (dismissed):
lgtm, thanks
approving with a regression test suggestion

tests/test_004_cursor.py
sumitmsft: @gargsaumya I do not see any test directly calls `execute(sql, row)`. Can you pls validate and add?

sumitmsft:
looks good to me.. Approved with a minor change to add one additional test

sumitmsft (dismissed):

tests/test_004_cursor.py
gargsaumya: Done

bewithgaurav (approved):

sumitmsft (approved):

Issue #620, opened 2026-06-03
cursor.fetchall, fetchmany and fetchone fail to type-check under `ty`


sam-leonard-ct:
### Describe the bug
`cursor.fetchall`, `cursor.fetchmany` and `cursor.fetchone` both fail to type-check under `ty`

### To reproduce

```python
# setup a connection
cursor.execute("SELECT 1")
cursor.fetchone()
# or any of the below:
cursor.fetchmany()
cursor.fetchall()
```

The above snippet when checked with `ty check repro.py` gives the following:
```
error[missing-argument]: No argument provided for required parameter `self` of function `fetchone`
   --> repro.py:103:1
    |
102 | cursor.execute("SELECT 1")
103 | cursor.fetchone()
    | ^^^^^^^^^^^^^^^^^
    |
info: Union variant `def fetchone(self) -> None | Row` is incompatible with this call site
info: Attempted to call union type `Unknown | (bound method Cursor.fetchone() -> None | Row) | (def fetchone(self) -> None | Row)`
info: rule `missing-argument` is enabled by default
```

It seems `ty` thinks this is a union of either a bound method or static method?

I believe this could also be an issue in ty as well but thought I'd ask here first.

### Expected behavior
The methods should type check correctly.

### Further technical details
Python version: 3.12.11
SQL Server version: SQL Server 2017
Operating system: MacOs 26.5
ty version: Observed on both 0.29 and 0.40

I have just been getting around this with `# ty: ignore[missing-argument]` but I'd rather not silence the lint.

sam-leonard-ct:
Curiously `cursor.fetchval` does type check correctly.

jahnvi480:
Hi @sam-leonard-ct, thanks for the detailed report!
We've been able to reproduce and confirm this is a bug on our side. We'll take this up based on our current priority list and get back to you. Will keep you posted. Thank you for trying out our driver!

gargsaumya:
The fix has now been merged and will be included in the next release on Friday. Please continue testing mssql-python, and feel free to reopen this issue if the problem persists.

PR #631, opened 2026-06-10
FIX: fetchone/fetchmany/fetchall type-checking under ty (#620)

gargsaumya:

> [AB#45498](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45498)

> GitHub Issue: #620 

### Summary   
  
This pull request refactors how column name mappings are handled for catalog/metadata result sets in the `mssql_python/cursor.py` module, specifically to avoid dynamically reassigning fetch methods (like `fetchone`, `fetchmany`, and `fetchall`) as instance attributes. This change improves compatibility with static type checkers and ensures cleaner, more predictable method behavior. A regression test is also added to prevent future regressions.

Refactoring of column mapping and fetch method handling:

* Refactored the `_prepare_metadata_result_set` method to build and cache the column name to index map (`_column_map`) directly, merging in specialized aliases when provided, and removed all logic that dynamically wraps or reassigns fetch methods on the instance. This ensures fetch methods remain class methods and improves static type checking compatibility. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L1626-R1641) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L1592-R1584)
* Updated the `__init__` method of the cursor to initialize the new `_column_map` attribute, and clarified the usage of cached column maps.
* Removed obsolete logic in `execute` that restored original fetch methods, since fetch methods are no longer replaced at runtime.

Testing and regression prevention:

* Added `test_fetch_methods_not_shadowed_on_instance` to verify that fetch methods are never shadowed as instance attributes, preventing regressions related to GH #620 and ensuring compatibility with static type checkers.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -1642,9 +1642,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         # specialized aliases) to keep case-insensitive access working for catalog
         # rows (e.g. row.TABLE_NAME when lowercase=True).
         self._cached_column_map_lower = (
-            {k.lower(): v for k, v in column_map.items()}
-            if get_settings().lowercase
-            else None
+            {k.lower(): v for k, v in column_map.items()} if get_settings().lowercase else None
         )
 
         # Initialize rownumber tracking so fetchone() and iteration work
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -141,16 +141,17 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
             False
         ]  # Indicates if last_executed_stmt was prepared by ddbc shim.
         # Is a list instead of a bool coz bools in Python are immutable.
-
-        # Initialize attributes that may be defined later to avoid pylint warnings
-        # Note: _original_fetch* methods are not initialized here as they need to be
-        # conditionally set based on hasattr() checks
         # Hence, we can't pass around bools by reference & modify them.
         # Therefore, it must be a list with exactly one bool element.
 
         self._rownumber = -1  # DB-API extension: last returned row index, -1 before first
 
+        # Column-name -> index map for the current result set. For catalog/metadata
+        # result sets this also carries lowercase and friendly aliases (see
+        # _prepare_metadata_result_set).
+        self._column_map = None
         self._cached_column_map = None
+        self._cached_column_map_lower = None
         self._cached_converter_map = None
         self._uuid_str_indices = None  # Pre-computed UUID column indices for str conversion
         # Cache the effective native_uuid setting for this cursor's connection.
@@ -1392,16 +1393,6 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         # Log the actual query being executed
         logger.debug("Executing query: %s", operation)
 
-        # Restore original fetch methods if they exist
-        if hasattr(self, "_original_fetchone"):
-            logger.debug("execute: Restoring original fetch methods")
-            self.fetchone = self._original_fetchone
-            self.fetchmany = self._original_fetchmany
-            self.fetchall = self._original_fetchall
-            del self._original_fetchone
-            del self._original_fetchmany
-            del self._original_fetchall
-
         self._check_closed()  # Check if the cursor is closed
         if reset_cursor:
             if self.hstmt:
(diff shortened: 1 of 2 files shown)


mssql_python/cursor.py
```diff
@@ -1626,94 +1616,34 @@ def _prepare_metadata_result_set(  # pylint: disable=too-many-statements
...
+        # per-call method reassignment (the previous approach broke static type
+        # checking - see GH #620).
+        self._cached_column_map = column_map
+        # Mirror execute(): when ``lowercase`` is enabled the description names are
+        # already lowercased, so build the lowercase lookup map (including any
+        # specialized aliases) to keep case-insensitive access working for catalog
+        # rows (e.g. row.TABLE_NAME when lowercase=True).
+        self._cached_column_map_lower = (
+            {k.lower(): v for k, v in column_map.items()} if get_settings().lowercase else None
+        )
```
sumitmsft, line 1655: This same conditional is now in five places in cursor.py:

_get_column_and_converter_maps()
execute()
_prepare_metadata_result_set() 
executemany() 
nextset()

Four of them existed before this PR, so this isn't a regression, but adding the fifth feels like a good moment to extract a small helper so any future change to lowercase behavior only has to be made in one place.

tests/test_004_cursor.py
```diff
@@ -10578,6 +10578,62 @@ def test_setinputsizes_sql_injection_protection(db_connection):
...
+
+    # A catalog helper historically reassigned the fetch methods. Make sure it
+    # no longer shadows them on the instance.
+    cursor.getTypeInfo().fetchall()
+    for name in fetch_methods:
+        assert name not in cursor.__dict__, f"{name} was shadowed after getTypeInfo()"
+
+    # A normal execute must also leave the class methods intact.
+    cursor.execute("SELECT 1 AS one")
+    assert cursor.fetchall() == [[1]]
```
sumitmsft: The check `cursor.fetchall() == [[1]]` only compares row values, not column names. So this test would still pass even if the column-name cache from the earlier `getTypeInfo()` call leaked into this `SELECT`

This PR makes `_prepare_metadata_result_set()` write into the same `_cached_column_map` that `execute()` uses. That cache must be rebuilt on every new query. If a future change accidentally breaks that rebuild, users would silently get wrong results. for example, `row.TYPE_NAME` on a `SELECT 1 AS` one result would return `1` instead of raising an error.

Today the rebuild works correctly, but the test doesn't actually verify it. A few extra lines would lock it in.

The two pytest.raises lines are the important ones, they're what would catch a future regression that today's test would silently miss.

sumitmsft (dismissed):
approved with minor changes

mssql_python/cursor.py
```diff
@@ -1626,94 +1616,34 @@ def _prepare_metadata_result_set(  # pylint: disable=too-many-statements
...
             self.description = fallback_description
 
-        # Define column names in ODBC standard order
-        self._column_map = {}  # pylint: disable=attribute-defined-outside-init
+        # Build the column-name -> index map for this metadata result set.
+        # Both the exact name and its lowercase alias are stored so that rows
+        # support case-insensitive lookup regardless of the global ``lowercase``
+        # setting (catalog column names are well-known and case-stable).
+        column_map = {}
         for i, (name, *_) in enumerate(self.description):
```
sumitmsft: Just happen to stumble upon it.. Good to fix it,

If `DDBCSQLDescribeCol `raises (the try/except above swallows it), column_metadata stays empty, `_initialize_description([])` sets `self.description = None`, and this loop then crashes with TypeError: 'NoneType' object is not iterable.

`getTypeInfo()` is the one exposed caller, it doesn't pass a `fallback_description`, so the fallback check at the top of the method can't save it. The other catalog helpers (tables(), columns(), primaryKeys(), etc.) all pass a fallback and are fine.

The old code had the same loop with no guard, so this isn't a regression,but since this PR is already restructuring this block, it's a good moment to fix it with a one-character change:

```python
for i, (name, *_) in enumerate(self.description or ()):
    column_map[name] = i
    column_map[name.lower()] = i
```
If description is None, the loop runs zero times, column_map stays empty, and the cursor returns no rows on the next fetch, which matches what the driver actually has. No behavior change for the happy path.

subrata-ms (dismissed):

subrata-ms (approved):

bewithgaurav (approved):

Issue #635, opened 2026-06-15
Connection Context Manager and Commits


benmatwil:
### Describe the bug
When using a context manager as described in the docs:
https://github.com/microsoft/mssql-python/wiki/Connection#context-manager
there does not seem to be any commit on exit. Are the docs incorrect?

Also do bulkcopies completely ignore the current connection object and create a new one?

### To reproduce
```python
import mssql_python

CONN_STR = ...

# Create the table inside a context manager block — no explicit commit
with mssql_python.connect(CONN_STR) as conn:
    conn.execute("CREATE TABLE dbo.no_commit_test (id INT)")
    print("Tried creating table inside context manager")

# Open a new connection and check whether the table exists
with mssql_python.connect(CONN_STR) as conn:
    with conn.cursor() as cursor:
        cursor.tables(table="no_commit_test", schema="dbo", tableType="TABLE")
        exists = cursor.fetchone() is not None

print(f"Table exists after context manager exit: {exists}")
# Expected output: False — the CREATE TABLE was rolled back
```

### Expected behavior
The process described by the docs

### Further technical details
Python version: 3.10
SQL Server version: SQL Server 2025
Operating system: Windows/Docker

bewithgaurav:
Hey @benmatwil, thanks for reporting this in detail. 
You're right, this is a bug.

The `__exit__` implementation today just calls `close()`, which always rolls back uncommitted changes when `autocommit=False`. 
There's no commit path in the context manager at all, so clean exits and exception exits both end up rolling back.

The [wiki](https://github.com/microsoft/mssql-python/wiki/Connection#context-manager) and the docstrings all describe commit-on-success / rollback-on-exception. , and that's what the implementation should be doing. 
We'll fix `__exit__` to match.

On bulkcopy and connections: Yes, `bulkcopy()` opens a separate internal `PyCoreConnection` under the hood. 
This is a known limitation of the design and has been discussed across & tracked in #492.

benmatwil:
Ah thanks for looking! 

Given the bulkcopy does not, does the arrow functionality share a connection with python or it's own? When the new underlying connection is created, is it with the same parameters? I'm guessing it does it's own commiting too for bulkcopy?
Is the plan to being these connections together at some point?

bewithgaurav:
@benmatwil - you're correct, its with same params (we're fixing some bugs on that area e.g. #626)
yes, we do have a plan and there's some development undergoing to facilitate that.

[...]

PR #639, opened 2026-06-22
FIX: Context manager commits on clean exit, rolls back on exception

bewithgaurav:
> GitHub Issue: #635

### Summary

Fixes the Connection context manager (`__exit__`) to implement commit-on-success / rollback-on-exception semantics, matching the documented behavior in the [wiki](https://github.com/microsoft/mssql-python/wiki/Connection#context-manager), the [DevBlogs post](https://devblogs.microsoft.com/python/simplifying-resource-management-in-mssql-python-through-context-manager/), and the docstrings.

Previously `__exit__` only called `close()`, which always rolled back uncommitted changes regardless of whether the block exited cleanly or via exception. Now:

- Clean exit + `autocommit=False`: commits the transaction
- Exception + `autocommit=False`: rolls back the transaction
- `autocommit=True`: no explicit commit/rollback (same as before)
- Connection always closed on exit

`close()` is unchanged. Its internal rollback becomes a harmless no-op on an already-resolved transaction.

24 subprocess-isolated tests cover: commit/rollback semantics, autocommit toggling, manual commit/rollback inside block, DDL+DML, pending cursor results, doomed transactions (XACT_ABORT), KeyboardInterrupt, generator abandonment, killed SPID, 10k-row transactions, double-exit idempotency, and GC safety. Resource-intensive tests (rapid connection cycling, 10k inserts) are marked `@pytest.mark.stress`.


mssql_python/connection.py
```diff
@@ -1698,16 +1698,38 @@ def __enter__(self) -> "Connection":
         logger.info("Entering connection context manager.")
         return self
 
-    def __exit__(self, *args: Any) -> None:
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
```
bewithgaurav, line 1706: **why `*args` → `exc_type, exc_val, exc_tb`:**

the old `__exit__` didn't inspect its arguments at all (just called `close()`), so `*args` was fine. now we branch on `exc_type` to decide commit vs rollback, so we use the canonical `__exit__(self, exc_type, exc_val, exc_tb)` signature from the [Python data model](https://docs.python.org/3/reference/datamodel.html#object.__exit__). 

this is what `sqlite3`, `psycopg2`, and every other DB-API driver uses. no behavioral change from the rename itself - just makes the branching logic readable.

mssql_python/connection.py
```diff
@@ -1703,16 +1703,43 @@ def __enter__(self) -> "Connection":
...
-        if not self._closed:
+        if self._closed:
+            return
+        try:
+            if not self.autocommit:
+                if exc_type is None:
+                    self.commit()
+                else:
+                    self.rollback()
+        except Exception:
```
sumitmsft, line 1730: The `except Exception` swallows the close error after a failed rollback (exception path) silently with no logging. We may have no visibility into why the connection failed to close. Worth adding a logging here.

mssql_python/connection.py
```diff
@@ -1703,16 +1703,43 @@ def __enter__(self) -> "Connection":
...
+            try:
+                self.close()
+            except Exception:
```
sumitmsft, line 1733: Similarly here as well... Would logging make sense?

sumitmsft (dismissed):
Approved with minor comments

mssql_python/connection.py
bewithgaurav: you're correct, have added logging here

mssql_python/connection.py
bewithgaurav: yes you're correct, added logging here

jahnvi480 (approved):

gargsaumya (approved):

Issue #643, opened 2026-06-24
Question: RecordBatchReader.close() behavior


amachanic:
When working with the `arrow_reader()` function on a cursor object, the returned type is a `RecordBatchReader`. I was hoping that I could call `close()` on this object partway through iteration and have it cancel any server-side work and clean up the underlying resources, but instead this appears to be a no-op. (Or at least, I don't see any effect on this end.)

So: What is `close()` _supposed_ to do today, if anything? And would it possible for it to do what I'm asking for?

subrata-ms:
Hi @amachanic , Thanks for raising the bug.

Effectively, [reader.close()] today is a client-side, pyarrow-internal no-op with respect to mssql_python and the database. The only paths that actually free anything are:

1. Iterating the reader to completion (the generator returns, [arrow_batch] returns 0 rows).
2. Calling [cursor.close()] or letting the [Cursor] be garbage-collected.
3. Closing the connection.

However, this looks like a valid scenario, and we are evaluating the design and potential fix.

amachanic:
@subrata-ms Thank you for the prompt reply! 

Just to let you know my overall goal, in case it helps with design and/or prioritization: I have a data access layer which is effectively a thin wrapper on top of mssql-python. I've added a function supporting `arrow_reader()` -- i.e. it allows tabular queries to stream back data in record batches. I've implemented that as a context manager, and I'd like to put some cleanup code in a `finally` block: Ideally if the client were iterating over several million of rows and some upstream (Python) exception occurred close to the beginning, I'd like it to be able to cancel, rather than forcing it to pull back all of the rows to get things back to a steady state. Closing the cursor is not an ideal option here, as there might be other things going on such as open transactions that I may or may not want to commit.

Thanks again, I really appreciate your attention here.

subrata-ms:
Hi @amachanic, This change was merged into the main branch today and will be available in the upcoming mssql-python driver release.
Closing this bug for now. Please reopen it if you encounter any issues during testing or usage.

PR #644, opened 2026-06-25
FIX: Implementing RecordBatchReader.Close() functionality for Arrow

subrata-ms:

> [AB#45978](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45978)

> GitHub Issue: #643 

### Summary   
  
This pull request introduces a new `_ArrowReader` wrapper to enhance the behavior of the `arrow_reader` method in the `Cursor` class. The new wrapper ensures that server-side resources are properly released when the reader is closed, supporting robust cleanup and allowing the parent cursor to remain usable. The tests are updated and extended to verify the improved semantics, including close behavior and context manager support.

Enhancements to Arrow batch reading and resource management:

* Added the `_ArrowReader` class to `cursor.py`, which wraps a `pyarrow.RecordBatchReader` and implements an 8-step close sequence to properly release server-side resources, reset cursor state, and support idempotent and context-manager-based cleanup. The parent `Cursor` remains usable after closing the reader.
* Modified the `Cursor.arrow_reader` method to return an instance of `_ArrowReader` instead of a raw `pyarrow.RecordBatchReader`, ensuring that closing the reader stops fetching, releases the server-side cursor, and resets cursor state. [[1]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L2708-R2909) [[2]](diffhunk://#diff-deceea46ae01082ce8400e14fa02f4b7585afb7b5ed9885338b66494f5f38280L2729-R2924)

Test improvements for Arrow reader behavior:

* Updated the `test_arrow_reader` test to check for duck-typed compatibility with `pyarrow.RecordBatchReader`, reflecting the new wrapper class.
* Added new tests: `test_arrow_reader_close_semantics` to verify that `.close()` stops fetching, marks the reader as closed, is idempotent, and leaves the parent cursor usable; and `test_arrow_reader_context_manager` to verify that the reader is closed on context manager exit and the cursor remains usable.


mssql_python/cursor.py
```diff
@@ -65,6 +65,150 @@ def _normalize_time_param(value, c_type):
...
+        """Synchronously stop fetching, release the server-side cursor, and
+        reset parent-cursor bookkeeping.  Idempotent.
+
+        Most of the actual cleanup work lives in the generator's ``finally``
+        clause (see ``Cursor.arrow_reader``); this method just unblocks any
+        in-flight fetch and closes the generator, which triggers that
+        ``finally`` block.
+        """
+        if self._closed:
+            return
```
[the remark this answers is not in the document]
subrata-ms: fixed

mssql_python/cursor.py
```diff
@@ -2722,11 +2879,60 @@ def arrow_reader(self, batch_size: int = 8192) -> "pyarrow.RecordBatchReader":
...
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    close_failed = True
+                    logger.debug("arrow_reader cleanup: _close_cursor failed: %s", e)
+
+                # 3) If SQL_CLOSE itself produced diags, pick them up too.
+                if close_failed:
+                    try:
+                        cur.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(cur.hstmt))
+                    except Exception:  # pylint: disable=broad-exception-caught
+                        pass
```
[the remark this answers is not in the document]
subrata-ms: fixed

tests/test_004_cursor_arrow.py
```diff
@@ -297,12 +297,138 @@ def test_arrow_table(cursor: mssql_python.Cursor):
...
+    # finally block must run, releasing the cursor so the next execute()
+    # succeeds without ProgrammingError("connection busy") etc.
+    del reader
+    gc.collect()
+
+    cursor.execute("select 5")
+    assert cursor.fetchone()[0] == 5
+
+
+def test_arrow_reader_cancel_from_other_thread(cursor: mssql_python.Cursor):
```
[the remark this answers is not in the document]
subrata-ms: fixed

tests/test_004_cursor_arrow.py
```diff
@@ -297,12 +297,138 @@ def test_arrow_table(cursor: mssql_python.Cursor):
...
+    try:
+        for batch in reader:
+            rows += batch.num_rows
+            if rows > 2_000_000:  # safety net — should never reach this
+                pytest.fail("reader was not cancelled by the other thread")
+    except pa.ArrowInvalid:
+        pass  # acceptable: reader was closed mid-iteration
+
+    closer_done.wait(timeout=5)
+    t.join(timeout=5)
```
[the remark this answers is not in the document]
subrata-ms: fixed

mssql_python/cursor.py
```diff
@@ -65,6 +65,150 @@ def _normalize_time_param(value, c_type):
...
+        if gen is not None:
+            try:
+                gen.close()
+            except Exception as e:  # pylint: disable=broad-exception-caught
+                logger.debug("arrow_reader.close: generator.close raised: %s", e)
+
+        # Drop strong refs so the wrapper does not extend the lifetime of
+        # the parent Cursor or the inner pyarrow reader.
+        self._cursor = None
+        self._inner = None
```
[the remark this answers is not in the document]
subrata-ms: fixed

ffelixg:
Huh, I didn't know that arrow readers supported a close method! I think that `_ArrowReader` should be subclassing `pyarrow.RecordBatchReader`. That way, we won't have to worry about duck typing. It also might save on some implementation overhead. @subrata-ms

subrata-ms:
> Huh, I didn't know that arrow readers supported a close method! I think that `_ArrowReader` should be subclassing `pyarrow.RecordBatchReader`. That way, we won't have to worry about duck typing. It also might save on some implementation overhead. @subrata-ms

Thanks for the suggestion — I looked into it and unfortunately subclassing doesn't work cleanly here because [pyarrow.RecordBatchReader](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is a Cython extension type, not a normal Python class. Specifically:

[RecordBatchReader.from_batches(...)](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is a Cython factory that hard-codes the return type to the base class; [Sub.from_batches(...)](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) returns a plain [RecordBatchReader](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html), so overridden [close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) / [read_next_batch()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) are never called.
__class__ reassignment is rejected on Cython types (TypeError: __class__ assignment only supported for mutable types or ModuleType subclasses), so downcasting a factory-produced reader isn't an option either.
A Cython instance also rejects arbitrary attributes, so self._cursor, self._generator, etc. can't live on the base instance directly.
The bare Sub.__new__(Sub) path leaves the underlying C stream unset — [super().close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) segfaults immediately (I reproduced it locally on pyarrow 24.0.0).
So there's no path where the object is simultaneously (a) an [isinstance](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) of [pa.RecordBatchReader](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html), (b) able to intercept [close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for our SQLCancel + generator teardown, and (c) able to carry our wrapper state.

The current composition/duck-typing shape is intentional for that reason. Happy to add __arrow_c_stream__ to the wrapper so consumers that follow the Arrow PyCapsule Protocol (polars, duckdb, [RecordBatchReader.from_stream](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html), etc.) can consume it without any [isinstance](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) check — that gives us the "no duck-typing worries" property without the subclass constraints.

mssql_python/cursor.py
```diff
@@ -2751,11 +2975,61 @@ def arrow_reader(self, batch_size: int = 8192) -> "pyarrow.RecordBatchReader":
...
+                #    result is no longer meaningful.
+                try:
+                    cur._clear_rownumber()  # pylint: disable=protected-access
+                    cur.rowcount = -1
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    logger.debug("arrow_reader cleanup: bookkeeping reset failed: %s", e)
+
+        gen = batch_generator()
+        inner = pyarrow.RecordBatchReader.from_batches(schema, gen)
+        return _ArrowReader(self, inner, gen, pyarrow.ArrowInvalid)
```
sumitmsft, line 3096: `_ArrowReader` has no internal lock, so two threads calling `close()` concurrently (or `close()` racing with `__del__`) can both pass the `self._generator is not `None` guard and both call `gen.close()/_cancel()`. The individual native calls are guarded/idempotent so this is unlikely to corrupt state, but do you think we need to consider a lightweight `threading.Lock` around the close body to make the documented cross-thread semantics fully robust?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1606,6 +1608,31 @@ void SqlHandle::close_cursor() {
     }
 }
 
+void SqlHandle::cancel() {
```
sumitmsft, line 1611: We are allowing `cancel()` to be called from a different thread than the one owning the statement (that's the cross-thread `close()` feature). But `SqlHandle` itself has no mutex. 

Because of that, `cancel()` checks `_handle `/ `_implicitly_freed` and then calls `SQLCancel_ptr(h)` with no lock. If another thread runs `free()` / `close_cursor()` in between, we can hit a use-after-free (calling `SQLCancel` on a handle that was just released), which may crash.

Since this PR is what makes concurrent access possible, could we protect the check + `SQLCancel` call with a lock, and have `free() `/ `close_cursor()` take the same lock? 

Alternatively, if we're confident `free()` can never race with `cancel()` in practice, a short comment explaining why it's safe would resolve the concern.

mssql_python/cursor.py
```diff
@@ -65,6 +65,217 @@ def _normalize_time_param(value, c_type):
...
+        # SQLCancel (cross-thread safe) — unblocks a fetch running on another
+        # thread so that the generator's finally clause can then run
+        # SQLFreeStmt(SQL_CLOSE) without risking the undefined-behaviour
+        # window of closing an HSTMT mid-fetch.  Safe no-op for an idle stmt.
+        cursor = self._cursor
+        if cursor is not None and not cursor.closed and cursor.hstmt is not None:
+            try:
+                cursor.hstmt._cancel()  # pylint: disable=protected-access
+            except Exception as e:  # pylint: disable=broad-exception-caught
+                logger.debug("arrow_reader.close: SQLCancel raised: %s", e)
```
sumitmsft, line 277: In the cleanup path, each step is wrapped in a broad except ... logger.debug(...). That's fine for the diagnostic-drain steps - if those fail we just lose some warning text.

But `_close_cursor()` is different: if it fails, the server-side cursor (and its locks) leak on SQL Server, and over time that can cause real production issues. Right now that failure is logged at DEBUG, which is typically disabled in production (INFO/WARN), so the leak would be completely invisible - there'd be no signal that anything went wrong.

I suggest we raise just the `_close_cursor()` failure log to `WARNING` (a failed cursor release is operationally significant and should show up in normal logs). The diagnostic-drain failures can stay at `DEBUG`.

mssql_python/cursor.py
sumitmsft, line 3096: Previously `arrow_reader()` returned a real `pyarrow.RecordBatchReader`; it now returns `_ArrowReader`, which (for good reasons documented in the code) does not subclass `pyarrow.RecordBatchReader`. This is a genuine behavioral change and can silently break existing users in two ways:

- `isinstance` checks fail: Any consumer doing `isinstance(reader, pyarrow.RecordBatchReader)` will now get `False`. The new `__arrow_c_stream__` protocol nicely covers polars/duckdb/from_stream, but a direct `isinstance` check has no way to benefit from it.

- The type hint is now inaccurate: it still advertises -> "pyarrow.RecordBatchReader" while returning `_ArrowReader`, which misleads users and type checkers. Pls check above.

So,here is that I was thinking as suggestions:
1. Update the return annotation to reflect reality
`def arrow_reader(self, batch_size: int = 8192) -> "_ArrowReader":`
3. Add a changelog entry and a line in the docstring explicitly calling out: "this is a pyarrow-compatible reader, not a pyarrow.RecordBatchReader; consume it via the Arrow C-stream protocol rather than an isinstance check."
Please confirm no internal callers rely on the exact `pyarrow.RecordBatchReader` type.
Keeping the wrapper is the right call - this is just about making the contract change visible so users aren't caught off guard.

sumitmsft (changes requested):

mssql_python/cursor.py
subrata-ms: The existing design already gets full robustness from three independent guarantees stacked together:

**1. Native cleanup** (SQLCancel, [SQLFreeStmt(SQL_CLOSE)](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) is idempotent and cross-thread-safe at the ODBC layer.
**2. The generator's finally** — which does all the parent-cursor mutation — runs exactly once, enforced by CPython.
**3. Close()** [close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is retry-safe: if a race causes [gen.close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to raise [ValueError](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) because another thread is still in the body, [_generator](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is preserved and the next [close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[__del__](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) completes the job ([cursor.py:269-273](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html)).

subrata-ms: My recommendation: no, and adding one would trade a benign cosmetic race for real risks:

**1. A lock held across** [gen.close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) while a third thread is inside the generator body doesn't prevent the [ValueError: generator already executing](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) case — the ValueError is a CPython invariant, not a locking issue — so the lock adds nothing there.
**2. Del -** [__del__](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) acquiring a lock is a well-known anti-pattern (GC-time waits, and it can deadlock during interpreter shutdown when the owning thread is being torn down). The current [__del__](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is already guarded against [sys.is_finalizing()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and swallows exceptions; layering a lock over it regresses that safety.
**3. The documented** cross-thread contract of this class is specifically "call SQLCancel from another thread to unblock a fetch" — which is precisely the pattern ODBC blesses without a lock. Introducing a Python-level lock around cancel semantics muddies that contract.

mssql_python/cursor.py
subrata-ms: agreed and fixed under follow-up commit.

mssql_python/cursor.py
```diff
@@ -65,6 +65,217 @@ def _normalize_time_param(value, c_type):
     return None
 
 
+class _ArrowReader:
```
bewithgaurav, line 68: went through the wrapper and understood we need to have a separate wrapper since its a cython extension - there is no other way to intercept close()
however - I found that the hand-rolled method surface inside it misses many cases and can be simplified.

the below user scenario works on 1.11.0 from pypi today and breaks on this branch:

```python
cur.execute("select * from orders where region = ?", "APAC")
table = cur.arrow_reader().read_all()
```

```
AttributeError: '_ArrowReader' object has no attribute 'read_all'
```

`cast()` and `read_pandas()` go the same way. `read_all()` is the usual way to drain a reader, so this is the first thing an existing user hits on upgrade, and an `AttributeError` reads like a driver bug rather than an intentional change.

as discussed in the comments - the reason those three are missing is that the class enumerates the pyarrow surface by hand, so anything not copied across is silently gone. `__arrow_c_stream__` landing a month after the class is the same gap showing up earlier.

**suggestion:**
delegating instead of enumerating closes the category rather than the three methods:

```python
def __getattr__(self, name):
    return getattr(self._inner, name)
```

I built this locally to check - this reduced the 210 line change to 75 lines with all existing tests passing, and against sql2022 `read_all`, `cast`, `pa.table(reader)`, `from_stream(reader)`, `__arrow_c_stream__`, `schema`, `read_next_batch` and iteration all work, while close mid-iteration, idempotency, the context manager and parent-cursor reuse all behave the same

this is required since every method added to `RecordBatchReader` in future goes missing the same way.

bewithgaurav (changes requested):
suggesting a better impl of the wrapper

mssql_python/pybind/ddbc_bindings.cpp
subrata-ms: Technically you are right that a race exists at the language-VM level

**1.** The mutex doesn't close the wider [free()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html)-vs-fetch window, so it's uneven defensive depth.
**2.** The narrow race requires the app to violate the DB API 2.0 "don't share cursors across threads" rule — which the codebase explicitly documents it does not support ([cursor.py:292-299](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html)).
**3.** The single new cross-thread contract added by this PR (SQLCancel from a cancel thread to unblock the fetch thread) is exactly the one ODBC blesses; that's why it works without a lock.

added a short comment on [SqlHandle::cancel()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [ddbc_bindings.cpp:1611](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/1b6a188127/resources/app/out/vs/code/electron-browser/workbench/workbench.html) documenting the two-thread invariant behavior.

mssql_python/cursor.py
subrata-ms: fixed. Added a change-log

mssql_python/cursor.py
subrata-ms: fixed in latest commit.

mssql_python/cursor.py
```diff
@@ -2759,15 +2994,28 @@ def arrow(self, batch_size: int = 8192) -> "pyarrow.Table":
 
     def arrow_reader(self, batch_size: int = 8192) -> "pyarrow.RecordBatchReader":
```
bewithgaurav: need to change the return type annotation here to `_ArrowReader`

mssql_python/cursor.py
bewithgaurav: the return type hint might've gotten missed, added a separate comment to highlight
else lgtm

subrata-ms: Thanks for catching this. fixed in follow-up commit.

subrata-ms: fixed

mssql_python/cursor.py
subrata-ms: fixed

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1606,6 +1608,53 @@
...
+    // observe the cancel and return.
+    //
+    // Cross-thread invariant (why no mutex is needed):
+    //   The only cross-thread pattern this driver blesses is exactly the one
+    //   ODBC blesses: cancel() may be called from a thread *other than* the
+    //   fetch thread to unblock an in-flight SQLFetch/SQLExecute on the same
+    //   HSTMT. Per the ODBC spec, SQLCancel (with the SQLGetDiagRec/Field
+    //   family) is the only entry point safe to call across threads on the
+    //   same statement handle. All other operations on a Cursor/SqlHandle
+    //   are single-owner: per DB API 2.0 and the Cursor thread-safety TODO
```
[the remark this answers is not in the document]

bewithgaurav (approved):

sumitmsft (approved):

Issue #627, opened 2026-06-05
SQLDescribeParam fails for VARBINARY/BINARY NULL when non-NULL parameter is bound first (ordinal remapping bug)


jahnvi480:
### Describe the bug
SQLDescribeParam returns SQLSTATE 07009 (Invalid descriptor index) when called after a non-NULL parameter has already been bound via SQLBindParameter on the same statement handle. This causes NULL parameters to silently fall back to SQL_VARCHAR, which breaks VARBINARY and BINARY columns:

```
Exception message:
mssql_python.exceptions.ProgrammingError: Driver Error: Syntax error or access violation; 
DDBC Error: [Microsoft][SQL Server]Implicit conversion from data type varchar to varbinary(max) 
is not allowed. Use the CONVERT function to run this query.

Stack trace:
  File "mssql_python/cursor.py", line 1518, in execute
    check_error(ddbc_sql_const.SQL_HANDLE_STMT.value, self.hstmt, ret)
  File "mssql_python/helpers.py", line 38, in check_error
    raise_exception(error_info.sqlState, error_info.ddbcErrorMsg)
  File "mssql_python/exceptions.py", line 563, in raise_exception
    raise exception_class
```

Root cause: Our BindParameters loop calls SQLDescribeParam and SQLBindParameter interleaved in a single pass. The ODBC driver internally passes already-bound params as "declared" to sp_describe_undeclared_parameters, which returns remaining params with ordinals restarting from 1. Our code still uses the original parameter index, causing a mismatch.

Proven by calling sp_describe_undeclared_parameters directly:

```
-- All undeclared → ordinals match
EXEC sp_describe_undeclared_parameters @tsql = N'INSERT INTO t VALUES (@P1, @P2)'
-- ordinal=1 → @P1(int), ordinal=2 → @P2(varbinary) ✅

-- @P1 declared → ordinals restart
EXEC sp_describe_undeclared_parameters @tsql = N'INSERT INTO t VALUES (@P1, @P2)',
    @params = N'@P1 int'
-- ordinal=1 → @P2(varbinary)  ← shifted from 2 to 1! Our code asks for index 2 → 07009 ❌
```

### To reproduce

```python
import mssql_python

conn = mssql_python.connect("Server=your_server;Database=your_db;Trusted_Connection=yes;TrustServerCertificate=yes;")
cur = conn.cursor()

[...]

PR #654, opened 2026-06-30
FIX: SQLDescribeParam ordinal remapping for VARBINARY/BINARY NULL

jahnvi480:

> [AB#45521](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/45521)

> GitHub Issue: #627 

### Summary   
This pull request significantly improves how the driver resolves SQL types for NULL parameters, especially when inserting into temp tables or table variables. It proactively resolves unknown NULL parameter types before binding, emits actionable Python warnings when automatic type resolution fails, and provides clear guidance on using `setinputsizes` to avoid SQL Server implicit conversion errors. Additionally, the code is refactored for clarity and efficiency, and minor formatting improvements are made.

**Improvements to NULL parameter type resolution:**
- Added a new `PreResolveUnknownNullTypes` function to proactively resolve unknown NULL SQL types for both single and batch execution paths, ensuring all types are determined before any parameters are bound. This prevents driver errors with binary NULLs and improves reliability. 
- When SQL type resolution falls back to `SQL_VARCHAR` (because `SQLDescribeParam` fails), the code now emits a Python warning with detailed guidance on how to use `cursor.setinputsizes()` to specify the correct type, preventing common SQL Server errors with BINARY/VARBINARY columns.

**Documentation and user guidance:**
- Expanded the `setinputsizes` docstring in `cursor.py` to document the temp table/NULL issue and provide a concrete usage example for binary columns.

**Code structure and clarity:**
- Refactored parameter binding code to avoid redundant type resolution during binding, since types are now resolved in advance. 
- Added a `bool isFallback` field to `DescribedParamInfo` to track when a fallback type is used.

**Minor improvements:**
- Made minor formatting and whitespace changes for improved readability.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -499,7 +499,7 @@ static DescribedParamInfo ResolveNullParamType(SqlHandle& handle, SQLHANDLE hStm
 
     DescribedParamInfo info;
     if (SQL_SUCCEEDED(rc)) {
-        info = {type, size, digits, false};
+        info = {type, size, digits};
         LOG("ResolveNullParamType: SQLDescribeParam succeeded for param[%d] "
             "-> sqlType=%d, columnSize=%lu, decimalDigits=%d",
             paramIndex, type, (unsigned long)size, digits);
@@ -515,7 +515,7 @@ static DescribedParamInfo ResolveNullParamType(SqlHandle& handle, SQLHANDLE hStm
         // Example:
         //   cursor.setinputsizes([None, (SQL_VARBINARY, 100, 0)])
         //   cursor.execute("INSERT INTO #t (id, data) VALUES (?, ?)", [1, None])
-        info = {SQL_VARCHAR, 1, 0, true};
+        info = {SQL_VARCHAR, 1, 0};
         LOG_WARNING("ResolveNullParamType: SQLDescribeParam failed for "
                     "param[%d] (rc=%d), falling back to SQL_VARCHAR",
                     paramIndex, rc);
@@ -549,6 +549,9 @@ static DescribedParamInfo ResolveNullParamType(SqlHandle& handle, SQLHANDLE hStm
 static void PreResolveUnknownNullTypes(SqlHandle& handle, SQLHANDLE hStmt,
                                        std::vector<ParamInfo>& paramInfos,
                                        const py::list* params = nullptr) {
+    if (paramInfos.empty())
+        return;
+
     // Quick scan: skip entirely if no unknown NULL params exist.
     bool hasUnknownNull = false;
     for (size_t i = 0; i < paramInfos.size(); ++i) {
@@ -560,7 +563,6 @@ static void PreResolveUnknownNullTypes(SqlHandle& handle, SQLHANDLE hStmt,
     }
     if (!hasUnknownNull)
         return;
-
     for (size_t paramIndex = 0; paramIndex < paramInfos.size(); ++paramIndex) {
         ParamInfo& paramInfo = paramInfos[paramIndex];
         if (paramInfo.paramCType != SQL_C_DEFAULT || paramInfo.paramSQLType != SQL_UNKNOWN_TYPE) {
@@ -590,7 +592,6 @@ SQLRETURN BindParameters(SqlHandle& handle, SQLHANDLE hStmt, const py::list& par
 
     // GH-627: resolve unknown NULL param SQL types before binding any param.
     PreResolveUnknownNullTypes(handle, hStmt, paramInfos, &params);
-
     for (int paramIndex = 0; paramIndex < params.size(); paramIndex++) {
         const auto& param = params[paramIndex];
         ParamInfo& paramInfo = paramInfos[paramIndex];
@@ -740,8 +741,7 @@ SQLRETURN BindParameters(SqlHandle& handle, SQLHANDLE hStmt, const py::list& par
                 if (!py::isinstance<py::none>(param)) {
                     ThrowStdException(MakeParamMismatchErrorStr(paramInfo.paramCType, paramIndex));
                 }
-                // GH-627: SQL type already resolved by PreResolveUnknownNullTypes.
-                dataPtr = nullptr;
+                dataPtr = nullptr;  // GH-627: type resolved by PreResolveUnknownNullTypes.
                 strLenOrIndPtr = AllocateParamBuffer<SQLLEN>(paramBuffers);
                 *strLenOrIndPtr = SQL_NULL_DATA;
                 bufferLength = 0;
@@ -2051,7 +2051,6 @@ SQLRETURN BindParameterArray(SqlHandle& handle, SQLHANDLE hStmt, const py::list&
     try {
         // GH-627: resolve unknown NULL array param SQL types before binding any param.
         PreResolveUnknownNullTypes(handle, hStmt, paramInfos);
-
         for (int paramIndex = 0; paramIndex < columnwise_params.size(); ++paramIndex) {
             const py::list& columnValues = columnwise_params[paramIndex].cast<py::list>();
             ParamInfo& info = paramInfos[paramIndex];
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -473,7 +473,8 @@ std::string DescribeChar(unsigned char ch) {
 // GH-610: Resolve SQL type for a NULL parameter using per-handle cache.
 // On cache miss, calls SQLDescribeParam and stores the result.
 static DescribedParamInfo ResolveNullParamType(SqlHandle& handle, SQLHANDLE hStmt, int paramIndex) {
-    // Check per-handle cache (no mutex — one handle per thread)
+    // Check per-handle cache. ODBC mandates one handle per thread, so no
+    // mutex is needed. Violating this contract causes undefined behavior.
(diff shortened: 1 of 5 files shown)


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -507,17 +508,98 @@ static DescribedParamInfo ResolveNullParamType(
...
                     "param[%d] (rc=%d), falling back to SQL_VARCHAR",
                     paramIndex, rc);
     }
 
-    // Store in per-handle cache
+    // Cache both successful and fallback results BEFORE emitting warnings.
+    // For fallbacks, this avoids repeated SQLDescribeParam network calls on
+    // statement reuse (same_sql path). The cache is cleared on SQLPrepare,
+    // so if the user changes the SQL or recreates the table, the next
+    // execution will retry SQLDescribeParam.
```
gargsaumya: This caches a negative result: on describe failure we store [{SQL_VARCHAR,1,0}] (Line 533). The comment says "cache is cleared on SQLPrepare," but clearDescribeCache() only runs inside [if (usePrepare)] (Line 1914) - it's skipped on the same_sql reuse path (else at Line 1916). So a transient describe failure is pinned as [SQL_VARCHAR] for the life of the reused prepared statement and never retried, even if the metadata later becomes resolvable. The comment is therefore slightly misleading. Please confirm this tradeoff is intentional, or cache only successful describes.

gargsaumya (dismissed):

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: +1

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -507,17 +508,98 @@ static DescribedParamInfo ResolveNullParamType(
...
     handle.describeCache[paramIndex] = info;
+
+    // Emit a Python warning on fallback so users get actionable guidance.
+    // Wrapped in try-catch: if warning emission fails (e.g. interpreter
+    // shutdown), we still return the cached fallback rather than aborting.
+    if (!SQL_SUCCEEDED(rc)) {
+        try {
+            auto builtins = py::module_::import("builtins");
+            py::module_::import("warnings")
+                .attr("warn")(
```
bewithgaurav: this warning fires on every `SQLDescribeParam` miss, not just the binary case the text describes. any parameterized query with a `None` that isn't backed by a describable table column trips it, and they all succeed:
```python
cur.execute("SELECT ? + ?", [1, None])                     # -> (None,)   warns
cur.execute("SELECT COALESCE(?, ?)", ["abc", None])        # -> ('abc',)  warns
cur.execute("SELECT ISNULL(?, ?)", [None, 5])              # -> ('5',)    warns
cur.execute("SELECT CASE WHEN ? = 1 THEN ? ELSE ? END", [1, None, "x"])  # warns
```
each one prints a paragraph about "Implicit conversion ... to varbinary" on a query with no binary column anywhere, the fallback to varchar is harmless for these (`1 + NULL` is `NULL` regardless of type), so the warning is a false alarm for the whole expression-parameter surface

real tables stay quiet since describe succeeds there, and the temp-table binary case is the only spot it's genuinely useful

suggestion: don't warn eagerly on every describe miss, let the fallback bind happen and only when the actual varbinary conversion error fires, attach the `setinputsizes` guidance to that exception

that keeps the help exactly where it matters and stops warning users whose query works fine.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -507,17 +508,98 @@ static DescribedParamInfo ResolveNullParamType(
...
+                    "SQLDescribeParam failed for parameter index " + std::to_string(paramIndex) +
+                        " (0-based). Falling back to SQL_VARCHAR for NULL binding. "
+                        "This may cause 'Implicit conversion from data type varchar to varbinary "
+                        "is not allowed' errors for BINARY/VARBINARY columns. "
+                        "To fix: from mssql_python.constants import ConstantsDDBC; then call "
+                        "cursor.setinputsizes() with explicit type info for every parameter. "
+                        "Example (2 params): cursor.setinputsizes("
+                        "[(ConstantsDDBC.SQL_INTEGER.value, 10, 0), "
+                        "(ConstantsDDBC.SQL_VARBINARY.value, column_size, 0)])",
```
bewithgaurav: this fallback guidance doesn't run, and the size it asks for is a no-op for the case it's about.

the example in the warning string ends with a bare `column_size`:

```py
cur.setinputsizes([(ConstantsDDBC.SQL_INTEGER.value, 10, 0), (ConstantsDDBC.SQL_VARBINARY.value, column_size, 0)])
# NameError: name 'column_size' is not defined
```

and this comment's version uses a None slot, which setinputsizes rejects:

```py
cur.setinputsizes([None, (ConstantsDDBC.SQL_VARBINARY.value, 100, 0)])
# ValueError: Invalid SQL type: None
```

past the paste errors, `column_size` is misleading, this path only fires for a NULL param, and for a NULL the size arg is ignored
verified against a VARBINARY(50) temp table, every size binds the NULL fine:

```py
size=0 -> ok    size=1 -> ok    size=50 -> ok    size=8000 -> ok
```

size only matters when you bind real bytes (too small truncates). so pointing the user at their column size sends them after a number that has no effect on the NULL they're inserting.

suggestion: give a concrete, runnable example with a throwaway size:

```py
cur.setinputsizes([(ConstantsDDBC.SQL_INTEGER.value, 10, 0), (ConstantsDDBC.SQL_VARBINARY.value, 0, 0)])
```

one concrete (sql_type, size, digits) tuple per parameter, no None slots, no placeholder names. the cursor.py docstring already has the right shape.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -507,17 +508,98 @@ static DescribedParamInfo ResolveNullParamType(
...
+                    builtins.attr("UserWarning"));
+        } catch (const py::error_already_set& e) {
+            LOG_WARNING("ResolveNullParamType: Failed to emit Python warning: %s", e.what());
```
bewithgaurav: the try/catch swallows the warn when a user runs warnings as errors, so it ignores `warnings.simplefilter("error")` while every other warning obeys it:

```
warnings.simplefilter("error")
cur.execute("SELECT ? + ?", [1, None])   # returns (None,), warning eaten
warnings.warn("plain", UserWarning)       # raises, as expected
```

affects more when someone who wants the temp-table binary insert to fail fast here instead gets the vague ProgrammingError later on in the code
dropping the catch is safe since this runs before any bind. just let it propagate:

```
py::module_::import("warnings").attr("warn")(msg);
```

`builtins` import above can go too, UserWarning is the default category.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -507,17 +508,98 @@ static DescribedParamInfo ResolveNullParamType(
...
+// all values in that column are NULL, so no Python-level check is needed.
+static void PreResolveUnknownNullTypes(SqlHandle& handle, SQLHANDLE hStmt,
+                                       std::vector<ParamInfo>& paramInfos,
+                                       const py::list* params = nullptr) {
+    if (paramInfos.empty())
+        return;
+
+    // Quick scan: skip entirely if no unknown NULL params exist.
+    bool hasUnknownNull = false;
+    for (size_t i = 0; i < paramInfos.size(); ++i) {
```
bewithgaurav: this quick-scan loop just re-checks what the loop right below already handles (iterating the same params and `continue` on anything that isn't `SQL_C_DEFAULT + SQL_UNKNOWN_TYPE`) 

so the pre-scan walks the vector once only to decide whether to walk it again.

this is more of a code improvement suggestion
suggest dropping it and keeping just the `empty()` guard:

```cpp
static void PreResolveUnknownNullTypes(...) {
    if (paramInfos.empty()) return;
    for (size_t paramIndex = 0; paramIndex < paramInfos.size(); ++paramIndex) {
        ...
    }
}
```

bewithgaurav (changes requested):
core fix is right, and solves the issue
comments mostly are all on the warning introduced and impl around it which can cause future problems, requesting changes

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Rewritten to explicitly state negative caching is intentional and only cleared on usePrepare path

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Removed warning entirely — no more UserWarning on any path

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Fixed code comment to use concrete runnable example with 0 for size

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Eliminated (no warning = no try/catch)

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Removed — kept just empty() guard + single loop

bewithgaurav (approved):
lgtm

gargsaumya (approved):

Issue #659, opened 2026-07-02
Entra ID access token is acquired on every connect(), even on a pool hit


sdebruyn:
## Summary

With Entra ID authentication (e.g. `Authentication=ActiveDirectoryDefault`), `Connection.__init__` acquires an access token on **every** `connect()`, before the native connection pool is consulted. On a pool hit the freshly acquired token is never used: the pooled physical connection is already authenticated, and the pool keys only on the sanitized connection string, so the token in `attrs_before` is not reapplied. The token acquisition and struct packing are therefore wasted work on every reused connection.

This partially defeats the purpose of pooling for token-auth workloads: pooling is enabled to avoid per-connection cost, yet a token is still materialized for each connection.

## Where (v1.10.0)

`mssql_python/connection.py`, `Connection.__init__`:

```python
# token acquired unconditionally, before the pool is consulted
sanitized = remove_sensitive_params(parsed_params)
self.connection_str = _ConnectionStringBuilder(sanitized).build()
token = get_auth_token(auth_type, credential_kwargs)      # <-- always runs
if token:
    self._attrs_before[ConstantsDDBC.SQL_COPT_SS_ACCESS_TOKEN.value] = token

# ... later ...

# pool checkout happens here, in the C layer, keyed on connection_str
if not PoolingManager.is_initialized():
    PoolingManager.enable()
self._pooling = PoolingManager.is_enabled()
self._conn = ddbc_bindings.Connection(
    self.connection_str, self._pooling, self._attrs_before
)
```

`get_auth_token` -> `AADAuth._acquire_token` reuses a cached *credential instance*, but still calls `credential.get_token("https://database.windows.net/.default")` and `get_token_struct()` (UTF-16-LE encode + `struct.pack`) on every call. azure-identity serves the token from its own in-memory cache while it is still valid, so this is not a full network round-trip each time, but it is per-connection CPU work whose result is discarded on a pool hit.

## Evidence

A `dlt` pipeline loading many tables to a Fabric Warehouse with `Authentication=ActiveDirectoryDefault` and native pooling enabled:

- The native pool is active and reusing connections: `SQL_ATTR_RESET_CONNECTION` (pool checkout reset) is logged ~212 times, with a single real cold login (~2.9s) followed by a uniform ~0.28s per open.
- Yet `get_token: Azure AD token acquired successfully` is logged once per `connect()` (146 token acquisitions for 146 opens, exactly 1:1), i.e. a token is produced even for the reused connections.

## Impact

[...]

Issue #651, opened 2026-06-30
Connection pool should separate identities for access-token and integrated auth connections


jahnvi480:
## Summary

The current connection pool in mssql-python keys only on the sanitized connection string. It does not account for the **identity/security context** of the user. This means different principals connecting to the same Server/Database can collide in the same pool bucket — a silent privilege escalation risk.

As a stopgap, PR #603 (`token_provider=`) disables pooling for all access-token connections. This issue tracks the proper fix: **extending the pool key to include identity information** so pooling can remain enabled safely.

## Problem

The native pool (`ddbc_bindings`) receives `(connection_string, pooling_bool, attrs_before)` but only the connection string is used as the pool key. The access token lives in `attrs_before[SQL_COPT_SS_ACCESS_TOKEN]` and is invisible to the pool. Two users authenticating with different tokens to the same server/database get the same pool bucket.

## Auth paths affected

| Auth path | Identity visible to pool? | Risk |
|---|---|---|
| `token_provider=` (custom credential) | ❌ No | Different principals share pool bucket |
| `Authentication=ActiveDirectoryDefault` | ❌ No | Same — token injected via attrs_before |
| `Authentication=ActiveDirectoryInteractive` | ❌ No | Same |
| `Authentication=ActiveDirectoryDeviceCode` | ❌ No | Same |
| `Authentication=ActiveDirectoryServicePrincipal` | ❌ No | Same (bulk copy path uses raw token) |
| Windows Integrated Auth (NTLM/Kerberos) | ❌ No | SSPI token negotiated by driver, not exposed to pool key |
| SQL Auth (UID/PWD) | ✅ Yes | Credentials are in the connection string (pool key) |

## Proposed solution

Extend the native pool key to include an **identity discriminator** alongside the connection string. Options to explore:

1. **For `token_provider=`**: Use `id(token_provider)` (object identity) as a pool key component. Same credential instance → same pool bucket. New instance → new bucket. This aligns with the principle that a new security context = a new provider instance.

2. **For built-in `Authentication=ActiveDirectory*`**: Derive a discriminator from the credential type + principal (e.g., hash of auth type + UID if present).

3. **For Windows Integrated Auth (NTLM/Kerberos)**: The SSPI token is negotiated internally by the driver and not exposed to the Python layer. Needs investigation into whether the native layer can extract the principal identity for the pool key.

## Cross-driver reference

| Driver | Pool identity separation? | How |
|---|---|---|
[...]

PR #660, opened 2026-07-03
FEAT: token-expiry capture and identity-aware pool key

jahnvi480:

> [AB#46159](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46159)
> [AB#46160](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46160)

> GitHub Issue: #651 
> GitHub Issue: #659 

### The problem
The connection pool keyed only on the **connection string**. When two callers hit the same server with different Entra (Azure AD) identities, they collided onto one pool — so a connection authenticated as **user A could be handed to user B**. Two secondary issues: a token was acquired on **every** connect (even pool hits), and a pooled connection could be reused with a **near-expired token**.

### What this PR does
Keys the pool by **security context**, not just the connection string, upholding one invariant: **a token-backed connection is never keyed on the bare connection string.** It also defers token acquisition to pool-misses and refreshes pooled connections whose token is near expiry.

### Design decisions (aligned with the cross-driver discussion)
- **Key by security context.** Non-token auth (SQL, trusted, **Service Principal** — handled natively by msodbcsql, Windows-interactive) is already isolated by the connection string and stays unchanged. When a token is involved, the key becomes `connStr + "\x00" + <identity>`.
- **Discriminator** (`compute_identity_key`): MSI → `msi:<client_id>`/`msi:system` (derived from params, no token needed); Interactive/DeviceCode → `acct:<home_account_id>`; DefaultAzureCredential / raw token → `tok:<sha256(token)>`.
- **Token expiry** (per the agreed guidance): as the pool owner, only ask the provider for a new token when the pooled token is **near expiry** (threshold = 5 min). Compare against the pooled token — **same token → reuse; rotated → discard those connections and reopen** with the new token. No background refresh; the SDK's silent refresh covers caching.
- **Interactive/device-code:** silent refresh from the cached account first; prompt only when the SDK forces it. One signed-in account per process.
- **MSI object-id vs client-id alignment** is a customer-facing breaking change — **deferred** as a separate item, not folded into this PR.

### Implementation highlights
- **Identity-aware pool key** + invariant, with a fail-safe token hash. The embedded `\x00` separator means the composite key can never collide with a bare connStr key.
- **Lazy token acquisition:** a token factory is invoked only when a physical connection is actually opened (pool miss) — same-identity pool hits acquire no token.
- **Silent-first interactive auth:** credentials built with `disable_automatic_authentication=True`; interactive prompt only on `AuthenticationRequiredError`; `home_account_id` cached so reconnects stay silent.
- **Expiry-aware checkout (5-min threshold):** mint fresh → byte-compare vs the pooled token → same+safe = reuse, rotated = discard + reopen (no double factory call) and drain siblings holding the stale token. Applies to MSI/interactive pools.
- **Idle identity-pool eviction:** single-use identities don't leak — aged-out pools are reclaimed on a later acquire (throttled sweep, `use_count()==1` guard, closed outside the manager mutex to respect lock ordering).
- **Fail-closed everywhere:** a token-auth failure raises `InterfaceError` rather than connecting token-less; a non-`bytes` access-token attribute is rejected at **both** the Python and native boundaries so it can't slip through with a bare pool key.

### Behavior matrix
| Auth type | Pool key | Token acquired | Expiry-aware? |
|-----------|----------|----------------|---------------|
| SQL / trusted / Service Principal / Windows-interactive | bare connStr | native | n/a |
| Managed Identity | `msi:<client_id>` / `msi:system` | lazy (on miss) | ✅ |
| Interactive / Device-code | `acct:<home_account_id>` | lazy, silent-first | ✅ |
| DefaultAzureCredential / raw token | `tok:<sha256>` | eager (bound to token) | ❌ (see limitations) |

### Testing
- **Unit** (`tests/test_008_auth.py`): identity-key mapping, silent-first interactive auth, fail-closed acquisition, pool-key wiring, and adversarial non-binary-token cases.
- **Integration** (`tests/test_009_pooling.py`, require a live SQL Server): cross-identity isolation, lazy factory invoked-on-miss / skipped-on-reuse, near-expiry refresh & token-bytes extraction, idle-pool eviction, checked-out-pool-not-evicted, enable/disable re-arm, concurrent-disable, and slot recovery on factory failure.

### Documented limitations (README)
- Interactive/device-code share one signed-in account per process.
- Bring-your-own-token lifetime is the application's responsibility (not expiry-aware).
- `DefaultAzureCredential` is not recommended for multi-user pooling (one-time warning emitted; keyed on token hash, so not expiry-aware).

### Out of scope / follow-ups
- MSI object-id → client-id alignment (breaking change).
- DB-API exception-type consistency on the eager auth path.
- Optional LRU bound on the credential/account caches.


tests/test_008_auth.py
```diff
@@ -1380,3 +1600,304 @@ def test_factory_cache_key_does_not_contain_raw_secret(self):
...
+        k1 = compute_identity_key("default", token_struct=b"same")
+        k2 = compute_identity_key("default", token_struct=b"same")
+        assert k1 == k2
+
+    def test_none_when_token_required_but_absent(self):
+        # default/interactive/devicecode need the token to form the key;
+        # without it, signal the caller to acquire then recompute.
+        assert compute_identity_key("default") is None
+        assert compute_identity_key("interactive") is None
+        assert compute_identity_key("devicecode") is None
```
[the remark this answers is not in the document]
jahnvi480: The acct: branch is covered in TestComputeIdentityKey in test_008_auth.py:1763 by three tests, which go beyond the single assertion Copilot suggested:

test_interactive_uses_home_account_id — compute_identity_key("interactive", home_account_id="acct-xyz") == "acct:acct-xyz"
test_devicecode_uses_home_account_id — same for "devicecode"
test_account_id_preferred_over_token_hash — asserts the account id wins over the token hash when both are present

jahnvi480:
##  Reviewer's guide: identity-aware connection pooling

### The 2-minute background

**The bug:** the label was built **only from the connection string**. That's fine for username/password logins (the password is in the string). But with **Entra ID (Azure AD)**, you authenticate with a short-lived **access token**, not a password — so two different people hitting the same server produce the **same connection string**, get the **same label**, and land in the **same pool**. **User B could be handed a connection authenticated as User A.** That identity leak is what this PR fixes.

**The fix in one line:** when a token is involved, add the caller's **identity** to the pool key.

**Terms you'll see:**
- **Identity key** — a short string standing for "who this caller is" (account ID, or a hash of their token), glued onto the pool key.
- **Token factory** — a callback that mints a fresh token *only when we actually need to open a real connection* (lazy).
- **Silent-first auth** — for interactive logins, get a token from the cached session *without* prompting; only prompt if that fails.
- **Expiry-aware checkout** — before reusing a pooled connection, check its token isn't about to expire.

**The one rule ("invariant") to keep in mind while reviewing:**
> If a token is involved, the pool key is `connection-string + <identity>` — **never** the bare connection string.

---

### The big picture

Python figures out *who* is connecting and builds the pool key; the native C++ layer owns the actual pool. **Note for C++ reviewers:** the pool has two locks (manager, then pool), and any network call (connect / disconnect / validate) must run with **no lock held** — that's the main thing to watch for in Passes 4–6.

```mermaid
flowchart TB
    subgraph PY["Python layer"]
      C["connection.py<br/>builds pool key + factory"]
      A["auth.py<br/>picks identity, gets tokens"]
      P["pooling.py<br/>PoolingManager on/off"]
    end
    subgraph NATIVE["Native C++ — pool lives here"]
      M["Pool manager<br/>key to pool + switch"]
      Pool["One pool<br/>shelf of connections"]
      Conn["Connection<br/>ODBC handle + expiry"]
    end
    C --> A
    C --> M
    P --> M
    M --> Pool --> Conn
```

---

## Pass 1 — How the identity key is chosen
🔗 [`auth.py::compute_identity_key`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/auth.py#L600-L651) · test: [`TestComputeIdentityKey`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_008_auth.py#L1726)

**What it does:** given the auth type and its parameters, returns a short string for "who is connecting" — or `None` when the connection string already isolates callers.

**Why it matters:** this function is the source of truth for the whole fix. Wrong answer here = identities can collide.

**What to check:**
- Password-style logins (SQL / trusted / Service Principal / Windows) → `None` → keep the bare connection string, exactly as before. *(Service Principal is safe: msodbcsql isolates it via the string.)*
- Managed Identity → `msi:<client_id>` / `msi:system`, from parameters, **no token needed**.
- Interactive / Device-code → `acct:<home_account_id>`.
- DefaultAzureCredential / raw token → `tok:<sha256 of token>` — the **catch-all fallback**.

```mermaid
flowchart TD
    S["connect()"] --> AT{"auth type?"}
    AT -->|"password-style"| BARE["bare connStr key<br/>unchanged"]
    AT -->|"Managed Identity"| MSI["msi:client_id"]
    AT -->|"Interactive"| ACCT["acct:account_id"]
    AT -->|"DAC / raw token"| HASH["tok:sha256"]
    BARE --> CMP["pool key = connStr + identity"]
    MSI --> CMP
    ACCT --> CMP
    HASH --> CMP
```

---

## Pass 2 — Where the pool key gets attached
🔗 [token-info helper](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/connection.py#L403-L420) · [token-factory helper](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/connection.py#L422-L442) · [key-building block](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/connection.py#L444-L502) · [raw-token safety net + fail-closed guard](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/connection.py#L549-L572) · [same guard in native](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection.cpp#L300-L316) · test: [`TestConnectionPoolKey`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_008_auth.py#L1785)

**What it does:** when a `Connection` is created, the identity from Pass 1 becomes an actual pool key (and, for some auth types, a lazy token factory), then gets handed to native.

**Why it matters:** this is the enforcement point for the invariant — every token path must set a key; nothing token-related may fall through to the bare-string path.

**What to check:**
- Every token branch sets `_pool_key`; the final `else` **fails closed** (raises `InterfaceError`) instead of connecting without an identity.
- "Fail closed" = if we can't build a safe identity, refuse to connect rather than risk a leak.
- New guard: an access token that isn't `bytes` (e.g. a `str`) is rejected up front at **both** the Python and C++ boundaries. A `bytes` token behaves exactly as before.

```mermaid
flowchart TD
    S["Connection.__init__"] --> K["compute_identity_key"]
    K --> Q{"identity without a token?<br/>Managed Identity"}
    Q -->|yes| F1["set key<br/>lazy factory"]
    Q -->|no| E["get token now<br/>fail: InterfaceError"]
    E --> Q2{"identity kind?"}
    Q2 -->|"account id"| F2["set key<br/>lazy factory"]
    Q2 -->|"raw token"| T["key from hash<br/>pass token through"]
    Q2 -->|"none"| FC["InterfaceError"]
    F1 --> RAW{"raw token in attrs,<br/>no key yet?"}
    F2 --> RAW
    T --> RAW
    RAW -->|"bytes"| RN["key = connStr + tok:hash"]
    RAW -->|"str / other"| REJ["InterfaceError<br/>fail closed"]
    RAW -->|"none"| NAT["hand to native Connection"]
    RN --> NAT
```

---

## Pass 3 — Interactive tokens without nagging the user
🔗 [`_acquire_token` silent-first](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/auth.py#L294-L321) · [`disable_automatic_authentication`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/auth.py#L285) · [`_authenticate_interactive`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/auth.py#L186-L214) · tests: [`TestSilentFirstInteractive`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_008_auth.py#L379), [`TestAuthenticateInteractive`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_008_auth.py#L541)

**What it does:** for interactive/device-code logins, try the cached session first; only show a login prompt if that genuinely fails.

**Why it matters:** pooling reconnects often — without this, every reconnect could pop a login prompt. We also cache the account ID so later reconnects stay silent.

**What to check:**
- Credential built with `disable_automatic_authentication=True` (forces the silent attempt first).
- Interactive `authenticate()` runs only on `AuthenticationRequiredError`.
- `home_account_id` is cached and reused.

```mermaid
sequenceDiagram
    participant App as connect()
    participant Auth as acquire_token
    participant Cred as Credential
    participant Cache as account cache
    App->>Auth: need interactive token
    Auth->>Cache: known account id?
    Auth->>Cred: get_token silently
    alt cached session works
        Cred-->>Auth: token, no prompt
    else login required
        Auth->>Cred: authenticate, prompt once
        Cred-->>Auth: account id
        Auth->>Cache: remember it
        Auth->>Cred: get_token
        Cred-->>Auth: token
    end
    Auth-->>App: token + expiry + account id
```

---

## Pass 4 — Native pool: keying + minting tokens only on a miss
🔗 [`ConnectionPool::acquire`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection_pool.cpp#L58-L120) · [`invokeTokenFactory`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection.cpp#L534-L550) · [pybind signatures](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection_pool.h#L71-L74) · [binding, defaulted args](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/ddbc_bindings.cpp#L5914) · test: [`TestNativeTokenFactory`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_009_pooling.py#L971)

**What it does:** the C++ pool files connections under the **pool key** (not the raw connection string), and calls the token factory **only when it must open a new physical connection** (a "pool miss").

**Why it matters:** (1) keying by pool key is what physically separates identities. (2) "mint on miss only" means a **pool hit reuses a connection with zero token work** — the performance win.

**What to check (C++ reviewers):**
- Map keyed by `pool_key`; the real connection string is still used for the ODBC connect.
- Token factory + ODBC connect run **outside both locks**.
- On a pool **hit**, no token is minted.

```mermaid
flowchart TD
    A["acquire()"] --> P1["prune dead/idle"]
    P1 --> LOOP{"connection on shelf?"}
    LOOP -->|"yes"| POP["validate outside lock"]
    LOOP -->|"no, room"| RES["reserve slot"]
    LOOP -->|"no, full"| FULL["pool full error"]
    POP --> NE{"factory set +<br/>near expiry?"}
    NE -->|"no"| V{"alive + resettable?"}
    NE -->|"yes"| EXP["expiry refresh<br/>Pass 5"]
    V -->|"yes"| REUSE["reuse<br/>no token minted"]
    V -->|"no"| DISC["discard"]
    EXP --> DISC
    DISC --> LOOP
    RES --> OPEN["mint token + connect<br/>outside lock"]
    REUSE --> DONE["hand to caller"]
    OPEN --> DONE
```

---

## Pass 5 — Don't hand out a connection whose token is about to expire
🔗 [near-expiry refresh block](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection_pool.cpp#L129-L196) · [`tokenExpirySafelyBeyond`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection_pool.cpp#L24-L33) · [`isTokenNearExpiry`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection.cpp#L556-L565) · tests: [`test_near_expiry_token_refreshed_on_checkout`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_009_pooling.py#L1150), [`..._factory_token_bytes_are_extracted`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_009_pooling.py#L1185)

**What it does:** if a pooled connection's token is within ~5 minutes of expiring, mint a fresh token and compare. Same token → reuse the connection. Different (rotated) token → discard the stale connections and open fresh ones with the new token.

**Why it matters:** without this you could hand out a connection that dies mid-query because its token just expired.

**What to check:**
- Threshold is 5 minutes (`TOKEN_EXPIRY_THRESHOLD_SECS`).
- Comparison is a **byte compare** — same token means the session just refreshed under the hood, so the connection is still valid.
- On a rotated token, sibling connections holding the *old* token are drained too, and the fresh token is reused for the reopen (no double mint).
- **Gotcha to confirm is acceptable:** "unknown expiry" (epoch 0) is treated as *not* near expiry, so it's never force-refreshed. Bring-your-own tokens are explicitly not expiry-managed.

```mermaid
flowchart TD
    NE["token near expiry"] --> MINT["mint fresh token"]
    MINT --> CMP{"same token +<br/>far from expiry?"}
    CMP -->|"yes"| SAFE["update expiry<br/>reuse"]
    CMP -->|"no, rotated"| ROT["drain stale siblings<br/>discard"]
    ROT --> REOPEN["reopen with new token"]
```

---

## Pass 6 — Cleaning up unused pools, and the on/off switch
🔗 [`acquireConnection` entry + eviction](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection_pool.cpp#L362-L441) · [eviction sweep](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection_pool.cpp#L383-L418) · [`closePools`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pybind/connection/connection_pool.cpp#L484-L517) · [`pooling.py enable`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pooling.py#L31-L78) · [`disable`](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/mssql_python/pooling.py#L81-L99) · tests: [idle-evicted](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_009_pooling.py#L410), [in-use-not-evicted](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_009_pooling.py#L520), [re-enable works](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_009_pooling.py#L899), [disable during connects](https://github.com/microsoft/mssql-python/blob/0636c6c90fbb4fb3d8fc35c9c6922b281ac92c34/tests/test_009_pooling.py#L574)

**What it does:** two housekeeping concerns. (1) **Eviction:** every distinct identity now creates its own pool, so a one-off identity would leave an idle pool forever — on later calls we sweep out pools idle too long. (2) **Lifecycle:** `PoolingManager.enable()/disable()` turns pooling on/off cleanly.

**Why it matters:** without eviction, per-identity pools slowly leak memory/handles. And disable→re-enable must actually work (there was a latent bug that kept it "closed").

**What to check:**
- Only evict a pool that's idle **and not currently in use** (`use_count()==1` guard).
- Evicted pools are closed **after** releasing the manager lock.
- An `accepting` gate: while disabling, new requests are refused and fall back to a plain non-pooled connection.
- `enable → disable → enable` re-arms correctly (`_pools_closed` reset).

```mermaid
stateDiagram-v2
    [*] --> Off
    Off --> On: enable()
    On --> Off: disable()
    Off --> On: enable() again
    On --> On: connect, pooled
    Off --> Off: connect, non-pooled
```

---

### Two practical notes for reviewers
- **Native tests (`test_009`) need a live SQL Server and a compiled build** — they won't run in a plain unit-test job. Please confirm the integration CI is green rather than eyeballing the C++ paths.
- **Intentionally deferred (not in this PR):** Managed Identity object-id vs client-id alignment (*that* one is a customer-facing breaking change); DB-API exception-type polish on the eager auth path; an optional size cap on the credential/account caches.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -513,14 +531,90 @@ std::chrono::steady_clock::time_point Connection::lastUsed() const {
...
+    }
+    return result.cast<py::dict>();
+}
+
+void Connection::setTokenExpiry(long long epochSeconds) {
+    _tokenExpiryEpoch = epochSeconds;
+}
+
+bool Connection::isTokenNearExpiry(int thresholdSecs) const {
+    if (_tokenExpiryEpoch == 0) {
```
sumitmsft, line 563: In `isTokenNearExpiry`, a `_tokenExpiryEpoch` of 0 (unknown expiry) is treated as "not expiring," so a factory-backed msi/acct connection whose factory didn't return an expiry just gets reused forever and never comes up for refresh. 
What caught my eye is that the sibling helper `tokenExpirySafelyBeyond` over in `connection_pool.cpp` does the opposite, it treats `<= 0` as not safe and fails closed. So the two functions disagree on what "unknown expiry" should mean, which felt worth a second look.

I don't think this is actually dangerous in practice: real MSI and interactive tokens always come back with `expires_on` populated, and even if a token did slip through and expire, ODBC's connection resiliency would just re-auth on the next use. Worst case is a query-time error, not one identity getting handed another's connection, the pool key still keeps identities separated. So this reads as a reliability edge, not a security hole.

Two small things from me:

1. Can you confirm the real credential paths always give us an expires_on? If so, I'm happy to ship as-is.
2. For the follow-up you already mentioned, it'd be nice to make `isTokenNearExpiry` fail closed like `tokenExpirySafelyBeyond` for factory-backed pools, just so the two are consistent.

sumitmsft (dismissed):
just have one comment which is not a blocker.

tests/test_009_pooling.py
```diff
@@ -668,6 +896,34 @@ def test_pooling_enable_disable_cycle(conn_str):
...
+
+    PoolingManager._reset_for_testing()
+    try:
+        PoolingManager.enable()
+        PoolingManager.disable()
+        # Re-enable: this must re-arm the guard (reset _pools_closed to False).
+        PoolingManager.enable()
+        assert PoolingManager._pools_closed is False
+
+        with patch("mssql_python.pooling.ddbc_bindings.disable_pooling") as mock_disable:
```
bewithgaurav: flakes on the Ubuntu ARM64 leg:

```
ModuleNotFoundError: No module named 'mssql_python.pooling.ddbc_bindings'; 'mssql_python.pooling' is not a package
```

`mssql_python.pooling` is both the public `pooling()` function and the `pooling.py` submodule, so the attribute resolves to whichever imported last. when it is the function, mock on py3.10 can't resolve the target and tries to import it as a package. repros in a 3.10 container, passes on 3.13.

fix, patch the canonical module instead:

```python
with patch("mssql_python.ddbc_bindings.disable_pooling") as mock_disable:
```

same module object (pooling.py does `from mssql_python import ddbc_bindings`), so it still intercepts the call.

generate_codecov.sh
```diff
@@ -38,7 +38,19 @@ echo "==================================="
 
 # Cleanup old coverage
 rm -f .coverage coverage.xml python-coverage.info cpp-coverage.info total.info
-rm -rf htmlcov unified-coverage
+rm -f default.profraw default.profdata
+rm -rf htmlcov unified-coverage profraw
+
+# Capture one raw profile *per process* so subprocess-based tests count too.
```
bewithgaurav, line 44: this is a great enhancement on overall code coverage, thanks for adding

mssql_python/pybind/connection/connection.cpp
```diff
@@ -513,14 +531,90 @@ std::chrono::steady_clock::time_point Connection::lastUsed() const {
...
+            .count());
+    return (now + static_cast<long long>(thresholdSecs)) >= _tokenExpiryEpoch;
+}
+
+std::string Connection::currentAccessToken() const {
+    auto it = _attrBytesBuffers.find(SQL_COPT_SS_ACCESS_TOKEN);
+    return it != _attrBytesBuffers.end() ? it->second : std::string();
+}
+
+bool Connection::accessTokenEquals(const std::string& token) const {
```
bewithgaurav: nit: this is exactly `currentAccessToken() == token` (same empty-map result) and it has one caller. drop it and inline at connection_pool.cpp:199:

```cpp
if (sibling->currentAccessToken() == stale_token) {
```

-11 lines, no behavior change.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -513,14 +531,90 @@ std::chrono::steady_clock::time_point Connection::lastUsed() const {
...
+        // Non-pooled connect still honors the lazy token factory: a
+        // token is materialized only when a physical connection is opened. The
+        // factory may also carry the token expiry, recorded on the
+        // connection for completeness (a non-pooled connection is never reused,
+        // so expiry-aware checkout does not apply here).
+        if (tokenFactory && !tokenFactory.is_none()) {
+            long long expiry = 0;
+            py::dict connect_attrs = Connection::invokeTokenFactory(tokenFactory, expiry);
+            _conn->connect(connect_attrs);
+            _conn->setTokenExpiry(expiry);
```
bewithgaurav: nit: `_tokenExpiryEpoch` is only read on the pooled acquire path, and a non-pooled connection never gets there. this store is dead. can drop it

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -6020,6 +6027,18 @@ PYBIND11_MODULE(ddbc_bindings, m) {
...
+        // Disarm new-pool creation *before* closing so a connect racing this
+        // disable cannot resurrect a pool after the map is cleared: any
+        // acquireConnection serialized after setAccepting(false) declines and
+        // falls back to a non-pooled connection. closePools() then reaps every
+        // pool created before the disarm.
+        auto& manager = ConnectionPoolManager::getInstance();
+        manager.setAccepting(false);
+        manager.closePools();
+    }, "Disable global connection pooling and close all pools");
+    m.def("_pool_count", []() { return ConnectionPoolManager::getInstance().poolCount(); },
```
bewithgaurav: tldr: we shouldn't expose a test-only diagnostic pool count in the ddbc_bindings public API. 
`_pool_count` will ship with ddbc bindings - only the 7 assertions in test_009_pooling.py touch it

I evaluated in my local and turns out we don't need this for tests
what these tests actually want to prove is that an aged-out pool's connection gets reclaimed, and that is observable at the server

ran the equivalent test in actual DB without the hook: capture the pooled connection's immutable `connection_id` from `sys.dm_exec_connections`, then assert it's gone after the idle timeout plus a later acquire on another key.

```python
a = connect(db("master")); ga = guid(a); a.close()       # A pooled + idle
assert alive(ga) == 1                                     # server connection lingers
time.sleep(3)                                             # age past idle_timeout=1
b = connect(db("tempdb")); b.cursor().execute("SELECT 1") # triggers the eviction sweep
assert alive(ga) == 0                                     # A's connection gone -> evicted
```

ran it repeatedly with no flakes, and it matched `_pool_count` every time
this method actually tests the real reclaim & not the map size essentially
this covers the eviction, checked-out, and disabled-manager tests directly.

the only one it can't do cleanly is `test_disable_during_concurrent_connects_leaves_no_pool`, since "no pool escaped" across a 4-thread race isn't observable per-connection
but that test isn't pulling its weight: the guarantee it checks (disable disarms new-pool creation) is already covered deterministically by `test_disabled_manager_creates_no_pool`

the concurrent version only adds a probabilistic smoke test of the timing window, it can't actually prove the fix (a run can just miss the race), and it's one of the sleep(0.2)/sleep(0.3) timing tests most likely to flake. I'd drop it or make it assert through behavior.

so below are two options - you can pick one:
- replace the `_pool_count` assertions with the server-side check above. it tests the real reclaim, not the map size, and ships nothing
- if you find cases which need the exact hook, gate it behind a test-build macro so it doesn't land in the release wheel.

bewithgaurav (changes requested):
some LLD nitpicks, one major suggestion - we need to take care of a public facing API

tests/test_009_pooling.py
jahnvi480: Made the required changes

mssql_python/pybind/connection/connection.cpp
jahnvi480: Made the required changes

mssql_python/pybind/connection/connection.cpp
jahnvi480: Made the required changes

mssql_python/pybind/ddbc_bindings.cpp

mssql_python/pybind/connection/connection.cpp
jahnvi480: 1. Always expires_on? Yes. Every token path goes through azure-identity's credential.get_token(), which returns an AccessToken(token, expires_on) where expires_on is required. The getattr(..., None) on auth.py:315 is just a test-double fallback — a live credential never hits it. So epoch == 0 only happens with non-token/synthetic factories, never a real Entra token.

2. Fail closed — done. A token-bearing connection with unknown expiry now comes up for refresh (consistent with tokenExpirySafelyBeyond); an empty-token connection stays reusable since there's nothing to expire. Sole caller is factory-guarded, so non-factory pools are unaffected.

bewithgaurav (dismissed):
lgtm

bewithgaurav (approved):

sumitmsft (approved):

PR #663, opened 2026-07-07
CHORE: Add standalone mssql-python-odbc package for ODBC driver binaries

jahnvi480:

> [AB#46217](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46217)

### Summary

Splits the ODBC driver binaries out of `mssql-python` into a new standalone, pure-data package `mssql-python-odbc` (import name `mssql_python_odbc`, version 18.6.0). This is the code / packaging half of the split; the ADO release pipelines are in companion PR #664.

#### What is included

- `mssql_python_odbc/__init__.py` - pure-Python package (no native extension) that ships the ODBC driver `libs/` tree and exposes `get_driver_path()` and `get_libs_dir()` with the same per-platform layout the C++ loader expects.
- `setup_odbc.py` - builds the `mssql-python-odbc` wheels. Each wheel is platform-specific but Python-agnostic (`py3-none-<platform>`), so we ship 7 wheels total (Windows amd64/arm64, macOS universal2, manylinux_2_28 x86_64/aarch64, musllinux_1_2 x86_64/aarch64) instead of a per-Python matrix. On Windows the wheel intentionally ships the VC++ runtime (`vcredist/msvcp140.dll` + license) alongside the driver binaries - see "VC++ runtime" below.
- `mssql_python/pybind/ddbc_bindings.cpp` - `LoadDriverOrThrowException()` now resolves the driver / libs base directory via a new `GetOdbcLibsBaseDir()` helper, which imports `mssql_python_odbc` when present and falls back to the bundled `mssql_python` libs when it is not. The fallback is GIL-safe and Alpine/musl-safe.
- `tests/test_024_odbc_package_split.py` - verifies Python vs C++ driver-path parity.
- `.gitignore` - ignores local `mssql_python_odbc/libs/` build copies and egg-info.

#### VC++ runtime (`vcredist`) on Windows

The Windows wheel ships `libs/windows/<arch>/vcredist/msvcp140.dll` (with its license). This is a deliberate, documented choice - **not** "parity" with the main wheel's file layout, and not a reversal of a licensing/size decision:

- The main `mssql-python` wheel also ships the VC++ runtime, just relocated: `build.bat` copies `msvcp140.dll` out of `vcredist/` to the package root next to the compiled `ddbc_bindings` extension, and `setup.py` then excludes the now-duplicate `vcredist/` folder. That exclusion is **de-duplication**, not a removal of the runtime.
- This package has no compiled extension, so there is no package-root `.pyd` to place the runtime beside. Keeping `msvcp140.dll` in its original `vcredist/` folder keeps the driver binaries and the runtime they were built against together, and keeps the wheel self-contained.
- No Phase-2 runtime skew: `msodbcsql18.dll`'s dependency on `msvcp140.dll` is satisfied in-process by the mssql-python extension (built `/MD`, which loads `msvcp140.dll` from beside the `.pyd`), so the external-package and bundled-libs paths behave identically. This copy is completeness + attribution, which is why `GetOdbcLibsBaseDir()`'s completeness check verifies the driver + `mssql-auth.dll` but not `vcredist`.

#### Phasing (no breaking change in this PR)

- Phase 2 (this PR): `mssql-python` prefers the external package but keeps bundling `libs/` as a fallback, so existing installs keep working.
- Phase 3 (future, v2.0.0): remove the bundled `libs/` and depend on `mssql-python-odbc` exclusively.

#### Testing

- `test_024` parity suite: 7 passed, 1 skipped.
- Built `mssql_python_odbc-18.6.0-py3-none-win_amd64.whl` locally: correct wheel tag, `Root-Is-Purelib: false`, driver DLLs present, `vcredist` (`msvcp140.dll` + license) present, `twine check` passes.

#### Notes

Companion PR (release pipelines): #664. Infra registration (ESRP onboarding of `mssql-python-odbc`, ADO build-definition id) is tracked there.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1215,6 +1299,16 @@ std::string GetDriverPathCpp(const std::string& moduleDir) {
         platform = "debian_ubuntu";  // Default to debian_ubuntu for other distros
     }
 
+    // NOTE (version coupling): the Linux driver filename is pinned to the exact
+    // msodbcsql minor version (18.6 -> libmsodbcsql-18.6.so.2.1), matching the
```
bewithgaurav: hardcoding is fine atm, but eventually shouldn't be followed. I would consider these scenarios before planning the next phase, where the pinning strategy might make or break things (adding since `setup.py` has no pin in this phase):

- loose pin (`>=`): pip won't pull the new odbc on a mssql-python upgrade, so new code looks for a driver filename the stale odbc doesn't ship. has to be exact `==`.
- user installs/upgrades mssql-python-odbc on its own, or `--no-deps`: the two drift and it surfaces as a cryptic "driver not found," not "your mssql-python-odbc version is wrong."

two separate fixes would be needed here: exact `==` pin makes upgrades safe (pip pulls the matching odbc), and a single source of truth for the version so the hardcoded filename and the pin can't drift
silently
no change needed to get this going, can just note for the next phase plan.

tests/test_024_odbc_package_split.py
bewithgaurav, line 1: this test file would be dropped if we drop the redundant python resolvers, what's left is trivial (version + import checks), and the dependency/existence coverage already lives in `test_000`, so anything genuinely worth keeping (e.g. "odbc package ships a driver for this platform") can go there instead of a separate file

mssql_python_odbc/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+    if os.path.exists("/etc/alpine-release"):
+        return "alpine"
+    if os.path.exists("/etc/redhat-release") or os.path.exists("/etc/centos-release"):
+        return "rhel"
+    if os.path.exists("/etc/SuSE-release") or os.path.exists("/etc/SUSE-brand"):
+        return "suse"
+    return "debian_ubuntu"  # default for Debian/Ubuntu and other glibc distros
+
+
+def get_driver_path() -> str:
```
bewithgaurav: from what i checked - `get_driver_path` / `_detect_arch` / `_detect_linux_distro_family` / `_DRIVER_FILENAME` in here are a Python resolver that nothing calls at runtime

the C++ loader reads only `__file__` from the package and resolves with its own `GetDriverPathCpp`, and the odbc-absent fallback goes to bundled mssql_python (no get_driver_path there either), so C++ keeps its resolver regardless.

these are just crowding `__init__` and look like redundant second copy of the `arch`/`distro`/`filename` logic that exists only for tests - let C++ stay the single resolver. removes` _DRIVER_FILENAME` (one fewer filename copy to keep in sync)

bewithgaurav (changes requested):
some duplicacies are there & need to be resolved, comments added inline
requesting changes

mssql_python/pybind/ddbc_bindings.cpp
sumitmsft: Agreed with this comment. IMO, let's not plan it in next phase - if we forget it will be detrimental to our release.  C++ and Python should not drift at any moment. Let's fix it and move away from hardcoded versions.

mssql_python_odbc/__init__.py
sumitmsft: Agreed to this as well.. Let's C++ is the real resolver here.

sumitmsft (changes requested):

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: So are we planning to keep a range instead of hard version pinning for mssql-python-odbc?

bewithgaurav: as discussed, this is in reference hardcoding versions in diff places (can find ways to eliminate) vs hard-pinning versions (`==` in reqs.txt which is required)
we're gonna be making sure there's minimal scope of errors when we upgrade odbc binaries
keep one version globally = imports everywhere

sumitmsft (approved):

bewithgaurav (approved):
lgtm, lets do the version pinning once we're published and ready to link

PR #664, opened 2026-07-07
CHORE: Add OneBranch release pipelines for mssql-python-odbc

jahnvi480:

> [AB#46218](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46218)

### Summary

This pull request introduces support for building and releasing two independent Python packages, `mssql-python` and `mssql-python-odbc`, from the same pipeline. The changes allow users to select which package(s) to build and release, optimize pipeline execution by running only the necessary stages, and ensure correct artifact labeling and validation. The release pipeline is also updated to handle both packages, with appropriate validation and symbol publishing logic.

**Build Pipeline Enhancements:**

* Added a `buildPackage` parameter to select which package to build (`mssql-python`, `mssql-python-odbc`, or `both`), with scheduled builds always producing both.
* Introduced separate build matrices and stages for `mssql-python-odbc`, including dedicated Windows, macOS, and Linux configurations and a new `ConsolidateOdbc` stage for artifact collection.
* Pipeline logic now gates `mssql-python` and `mssql-python-odbc` build stages and consolidation based on the selected package, reducing unnecessary work and enabling concurrent builds.
* Ensured SBOM (Software Bill of Materials) is named after the actual package being built to avoid mislabeling artifacts.

**Release Pipeline Updates:**

* Added a `releasePackage` parameter to the dummy release pipeline, allowing selection of which package to test-release and dynamically choosing the correct consolidated artifact. 
* Validation and symbol publishing steps are now conditional on the selected package, including strict checks for the `mssql-python-odbc` dependency pin in `mssql-python` wheels.


OneBranchPipelines/stages/build-odbc-linux-stage.yml
```diff
@@ -0,0 +1,191 @@
+# ODBC Linux Single Configuration Stage Template
+# Builds the platform-specific `mssql-python-odbc` wheel for ONE libc/arch combo
+# (manylinux_2_28 or musllinux_1_2 x x86_64 or aarch64).
+#
+# Data-only package: no native compilation, no per-Python matrix, no pytest.
```
bewithgaurav, line 5: the odbc wheels ship the driver but nothing runs pytest against them as of now
the pr-validation & daily build-release pipelines resolve the driver from bundled `mssql_python/libs`, so `GetOdbcLibsBaseDir` never gets hit in any case actually

`GetOdbcLibsBaseDir` runs on every connect, but CI only ever hits its fallback branch (bundled libs) since nothing installs the odbc wheel. the external-resolution branch never executes, so a break there ships green.

I think we need a separate pytest step here and have a nightly build routine for mssql-python-odbc as well
we should test installing the built package in a clean env with latest mssql-python build & run the same pytest suite
no new tests or smoketest neded, just the existing suite pointed at the built package

bewithgaurav (changes requested):

sumitmsft (dismissed):

OneBranchPipelines/stages/build-odbc-linux-stage.yml
jahnvi480: Added a task to perform pytest every night when both the wheel files are getting generated

sumitmsft (dismissed):

sumitmsft (approved):

bewithgaurav (approved):

Issue #671, opened 2026-07-09
Deadlock under concurrent multithreaded use when DEBUG logging is enabled via setup_logging()


sdebruyn:
### Describe the bug

When driver logging is enabled via `mssql_python.setup_logging(output="file")` (DEBUG level) **and** multiple threads concurrently open connections and execute statements (for example a `ThreadPoolExecutor` with 4 workers), the process permanently deadlocks. It sits at 0% CPU — fully frozen, not merely slow.

The same code with `setup_logging` **not** called does **not** deadlock. It also does **not** deadlock with a single worker thread. The trigger is therefore the combination of **DEBUG logging + concurrency** (`workers > 1`).

### Environment

- **mssql-python version:** 1.10.0
- **Python version:** 3.12.11
- **Operating system:** macOS (arm64)
- **Authentication:** `ActiveDirectoryDefault` (Entra ID) against a Microsoft Fabric Warehouse. The deadlock is in the logging path and is very likely authentication-independent.

### Observed thread stacks

Captured with Python's `faulthandler`. The dumps were **identical across 7 dumps over roughly 7 minutes**, which confirms a permanent deadlock rather than slowness.

Thread A (opening a connection):

```
File ".../mssql_python/connection.py", line 571 in setautocommit
File ".../mssql_python/connection.py", line 410 in __init__
File ".../mssql_python/db_connection.py", line 55 in connect
-> (calls into Python logging) ...
File ".../logging/handlers.py", line 75 in emit
File ".../logging/__init__.py", line 1144 in flush   # blocked here
```

Thread B (executing a statement, a `COPY INTO`):

```
File ".../mssql_python/cursor.py", line 1545 in execute   # blocked in native execute
```

### Server-side check

While the client is frozen, the database server shows **no query running** for this session. This confirms the block is entirely client-side and is not waiting on the server.

### Hypothesized mechanism

mssql-python emits DEBUG log records from within native code paths (for example `setautocommit` during connect, and `cursor.execute`) that hold the GIL / internal locks. Under concurrency this creates a lock-acquisition cycle with Python's `logging` `Handler` locks across threads:

- Thread A holds the logging handler lock and is blocked in `flush`.
- Thread B is blocked in native `execute` and cannot proceed.

This is a classic logging-from-native-under-the-GIL deadlock. It only reproduces with `workers > 1`.

### Steps to reproduce (minimal sketch)

1. Call `mssql_python.setup_logging(output="file")` to enable DEBUG logging.
[...]

PR #678, opened 2026-07-14
FIX: prevent GIL/mutex deadlock when logging under concurrent multithreaded use

bewithgaurav:

> ADO Work Item: Fixed [AB#46334](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46334)
> GitHub Issue: #671

### Summary

This pull request addresses a critical deadlock issue (#671) that could occur when DEBUG logging is enabled and multiple threads interact concurrently with the connection pool and logging system. The main changes ensure that any logging (which acquires Python's GIL) is never performed while holding native mutexes, thus preventing lock-order inversions and deadlocks. It also adds regression tests to guarantee that this deadlock does not return.

**Deadlock Prevention and Logging Safety:**

- Refactored all code paths in `connection.cpp` and `connection_pool.cpp` so that logging calls (`LOG`, `LOG_ERROR`) are made only after releasing any native mutexes, preventing GIL/native-mutex lock-order inversions that caused deadlocks under concurrent DEBUG logging. Comments were added to explain the rationale and reference issue #671. [[1]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecL59-R67) [[2]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR144-R153) [[3]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecL280-R312) [[4]](diffhunk://#diff-35a698bc3405cc30c1bf656066bd0575302239b5e39083a8160a6202026ca360L248-R261) [[5]](diffhunk://#diff-35a698bc3405cc30c1bf656066bd0575302239b5e39083a8160a6202026ca360L432-R454)
- In `logger_bridge.cpp`, removed the use of a native mutex in the logging bridge, relying solely on the GIL for thread safety, which avoids deadlocks when logging from multiple threads.

**Connection Pooling and Handle Management:**

- Updated connection pool logic to construct `Connection` objects and perform ODBC handle allocation outside of pool mutexes, ensuring that any logging or GIL acquisition happens only after releasing native locks. [[1]](diffhunk://#diff-35a698bc3405cc30c1bf656066bd0575302239b5e39083a8160a6202026ca360L110-R114) [[2]](diffhunk://#diff-35a698bc3405cc30c1bf656066bd0575302239b5e39083a8160a6202026ca360L235-R242) [[3]](diffhunk://#diff-35a698bc3405cc30c1bf656066bd0575302239b5e39083a8160a6202026ca360L273-R284) [[4]](diffhunk://#diff-35a698bc3405cc30c1bf656066bd0575302239b5e39083a8160a6202026ca360R392)
- Improved child handle compaction and logging in `Connection::disconnect()` and `allocStatementHandle()` by collecting compacted handle statistics while holding the mutex, but deferring logging until after the mutex is released. [[1]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR116-R134) [[2]](diffhunk://#diff-eca696c13d997f510e5f9b16288ed1deb0ad132768c283eda1518f78edf9b6ecR281-R282)

**Testing and Regression Coverage:**

- Added a new regression test file `test_025_logging_concurrency_deadlock.py` that launches multi-threaded workloads under DEBUG logging in a subprocess, verifying that no deadlocks occur. The test is robust, runs out-of-process, and will fail if the deadlock regresses.

These changes collectively ensure that the driver is robust against concurrency-related deadlocks when logging is enabled, and that future regressions will be caught by automated tests.


mssql_python/pybind/connection/connection.cpp
```diff
@@ -138,6 +141,16 @@ void Connection::disconnect() {
...
 
+        // Log after releasing _childHandlesMutex (#671): LOG()/LOG_ERROR() acquire
+        // the GIL and must not run while a native mutex is held.
+        LOG("Compacted child handles: %zu -> %zu (removed %zu expired)",
+            originalSize, afterCompactSize, originalSize - afterCompactSize);
+        LOG("Marking %zu child statement handles as implicitly freed", afterCompactSize);
+        if (badHandleCount > 0) {
+            LOG_ERROR("CRITICAL: %zu non-STMT handle(s) found in _childStatementHandles. "
+                      "This will cause a handle leak!", badHandleCount);
+        }
```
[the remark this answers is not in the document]
bewithgaurav: fixed. every `LOG()` in `disconnect()` is gated on `hasGil` now, computed once up top, so it lines up with the no-log branch that was already there for the shutdown path. also made `hasGil` hold up after `Py_Finalize()` by checking `Py_IsInitialized()` first, since a plain gil-state check isn't reliable once the interpreter is gone.

tests/test_025_logging_concurrency_deadlock.py
```diff
@@ -0,0 +1,118 @@
...
+    The child imports mssql_python the same way this process did (the parent's
+    sys.path is forwarded via PYTHONPATH), and its configuration is passed via
+    the environment so the connection string never appears in the process list.
+    """
+    env = dict(os.environ)
+    env["DB_CONNECTION_STRING"] = conn_str
+    env["MSSQL671_WORKERS"] = str(workers)
+    env["MSSQL671_ITERS"] = str(iters)
+    env["MSSQL671_LOG_FILE"] = log_file
+    env["PYTHONPATH"] = os.pathsep.join(p for p in sys.path if p)
```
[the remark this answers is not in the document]
bewithgaurav: fixed. forwarding the full `sys.path` to the child now, same as `test_023_ssh_tunnel_gil_release.py`, so the empty cwd entry survives for a script-launched child. added `test_child_pythonpath_forwards_cwd_entry` that asserts it, so the filter can't come back quietly.

mssql_python/pybind/connection/connection.cpp
```diff
@@ -56,7 +56,15 @@ Connection::~Connection() {
...
-    auto _envHandle = getEnvHandle();
+    SqlHandlePtr _envHandle;
+    {
+        // Fetch/initialize the shared env handle without holding the GIL (#671):
+        // its first-time initialization runs under a C++ static-init guard and
+        // emits log records; a thread waiting on that guard while holding the GIL
+        // would deadlock the initializing thread that needs the GIL to log.
+        py::gil_scoped_release gil_release;
+        _envHandle = getEnvHandle();
+    }
```
subrata-ms: Probably a better approach to use lambda's here

auto envHandle = [&] {
        py::gil_scoped_release gil_release;
        return getEnvHandle();
      }();

mssql_python/pybind/connection/connection_pool.cpp
```diff
@@ -232,7 +236,10 @@ std::shared_ptr<Connection> ConnectionPool::acquire(const std::u16string& connSt
         if (have_pending_token) {
             std::lock_guard<std::mutex> lock(_mutex);
             if (_current_size < _max_size) {
-                valid_conn = std::make_shared<Connection>(connStr, true);
+                // Reserve the slot here but construct the Connection outside
+                // _mutex (Phase 3): the constructor emits GIL-acquiring log
+                // records, and holding _mutex across a GIL acquisition
+                // deadlocks a thread that holds the GIL and waits on _mutex (#671).
```
subrata-ms, line 242: Moving this outside the mutex lock will cause an edge-case concern 
**One Edge-Case Concern**

The move extends the existing capacity-accounting race to constructor failures:

1. Thread A reserves capacity and begins construction outside [_mutex].
2. [close()] resets [_current_size] to zero.
3. Thread B reserves the newly visible slot.
4. Thread A’s constructor fails.
5.  A executes:
       if (_current_size > 0) {
                      --_current_size;
            }

That decrement removes B’s reservation rather than A’s. Another acquire can then exceed [_max_size].
Before this PR, construction happened while holding [_mutex] only failures during the later [connect()] phase were exposed to this race. The PR expands exposure to driver initialization, environment setup, memory allocation, and DBC-handle allocation failures.

**Conclusion:** the move is necessary and correct for the main multithreaded path. Its only concrete regression risk is the expanded constructor-failure/accounting race during concurrent pool closure.

bewithgaurav: thanks - this is a valid case
it's a pre-existing `connect()`-failure accounting race, and moving construction out of the lock widens it to constructor failures too 
however, the edge case is a bit narrow - a new connection has to fail at the same moment pooling is being disabled., and the effect is a transient `_current_size` drift, not a crash

I'll open a separate issue to fix the counting properly & comment it here, covering both the constructor and connect paths - let me know if you have any other suggestions

subrata-ms: I am good if you want to track this with different task/issue. You are right that the issue is pre-exist but now with the new change, the edge case becomes more vulnerable (widen).

subrata-ms (approved):
The edge case mentioned in my earlier review comment need to be track with separate task.

sumitmsft (approved):
Reviewed for correctness, concurrency safety, resource handling, regression coverage, repository conventions, and cross-platform implications. No new actionable issues were identified.

**Recommendation:** Approve

mssql_python/pybind/connection/connection_pool.cpp
bewithgaurav: tracked as #746 for a separate fix.

Issue #684, opened 2026-07-16
add_output_converter documents/accepts an integer SQL type, but dispatch keys on the Python type (integer keys silently never fire)


dlevy-msft-sql:
### Summary

`Connection.add_output_converter(sqltype, func)` is documented to take an **integer** ODBC SQL type code (pyodbc-compatible signature), but at fetch time the converter is looked up by the **Python type** stored in `cursor.description[i][1]`. As a result, registering a converter with an integer SQL type constant (for example `SQL_DECIMAL`) silently no-ops, and the value comes back unconverted. Only registering the Python type (for example `decimal.Decimal`) works.

Verified against `mssql-python` 1.11.0.

### Repro

```python
import mssql_python
import decimal

conn = mssql_python.connect(connection_string)

# Documented style: register by integer SQL type constant -> silently ignored
conn.add_output_converter(mssql_python.SQL_DECIMAL, lambda v: None if v is None else float(v))

cur = conn.cursor()
cur.execute("SELECT TOP 1 ListPrice FROM Production.Product WHERE ListPrice > 0")
print(type(cur.fetchone()[0]))   # <class 'decimal.Decimal'>  (converter never ran)

# Works only when keyed by the Python type
conn.add_output_converter(decimal.Decimal, lambda v: None if v is None else float(v))
cur.execute("SELECT TOP 1 ListPrice FROM Production.Product WHERE ListPrice > 0")
print(type(cur.fetchone()[0]))   # <class 'float'>  (fires for money and decimal columns)
```

### Expected

Either:
- the integer SQL type key documented in the API works, or
- the docstring reflects that the key is a Python type.

Ideally the method accepts **both** an integer SQL type code and a Python type (see suggested fix).

### Actual

- Registering an integer SQL type constant is stored but never matched, so no conversion occurs.
- Registering the Python type works, and (because many SQL types collapse to one Python type) a single `decimal.Decimal` converter fires for `decimal`, `numeric`, `money`, and `smallmoney` columns alike.

### Root cause

The description type code is a **Python type**, not the ODBC integer code. `Cursor._map_data_type` maps the ODBC SQL type code to a Python type for `cursor.description[i][1]`:

```python
# cursor.py, _map_data_type
sql_to_python_type = {
    ddbc_sql_const.SQL_DECIMAL.value: decimal.Decimal,
    ddbc_sql_const.SQL_NUMERIC.value: decimal.Decimal,
    ddbc_sql_const.SQL_INTEGER.value: int,
[...]

PR #690, opened 2026-07-28
FIX: dispatch output converters on integer ODBC SQL type codes (#684)

jahnvi480:

> [[AB#46648](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46648)]

> GitHub Issue: #684

### Summary

`Connection.add_output_converter(sqltype, func)` documents (and pyodbc accepts) an
**integer ODBC SQL type code** as the key — e.g. `SQL_DECIMAL`, `SQL_INTEGER`. However,
dispatch in `Cursor._build_converter_map()` keyed on the **Python type** stored in
`cursor.description[i][1]` (via `_map_data_type`), so integer-keyed converters were stored
but **silently never fired**. Only Python-type keys (e.g. `decimal.Decimal`, `str`) worked.
This makes the driver diverge from pyodbc and from its own documentation.

### Root cause

The raw ODBC SQL type code (`col["DataType"]`) is available when the description is built
in `_initialize_description`, but it was discarded. `_build_converter_map` then looked up
converters using `description[i][1]` (a Python type), which can never match an integer key.

### Fix

- `cursor.py`
  - `_initialize_description` now also stores `self._column_sql_types` — the raw ODBC SQL
    type codes, parallel to `description` (reset to `None` when there are no columns).
  - `_build_converter_map` dispatches in pyodbc-compatible order: **(1)** integer SQL type
    code, **(2)** Python type in `description[i][1]`, **(3)** the legacy `WVARCHAR` catch-all.
    This is purely additive — existing Python-type converters keep working.
  - `_prepare_metadata_result_set` now builds the converter map so catalog metadata result
    sets (`columns()`, `tables()`, …) honor converters consistently with normal result sets.
  - Defensive `self._column_sql_types = None` resets alongside the bare `self.description = None`
    assignments (execute/executemany describe-fail paths and `nextset`).
- `connection.py` — `add_output_converter` signature is now `sqltype: Union[int, type]`; the
  docstring is corrected to describe both key styles (integer key takes precedence) and the
  value semantics (already-materialized Python object; `str` passed as UTF‑16LE bytes).
- `mssql_python.pyi` — stub updated to match (`Union[int, type]`).

### Exact-type dispatch

Integer keys use **exact** ODBC type matching, so `SQL_DECIMAL` and `SQL_NUMERIC` are
distinct (they are not collapsed to `decimal.Decimal`), matching pyodbc.

### Tests

Added `tests/test_003_connection.py::test_output_converter_integer_sql_type_key_gh684`
covering: integer key fires; exact-type dispatch (`SQL_INTEGER` does not touch `DECIMAL`,
but fires on `INT`); integer-key precedence over a Python-type key; distinct `DECIMAL` vs
`NUMERIC` converters; NULL stays `None`; the `SQL_WVARCHAR` string path (converter receives
UTF‑16LE bytes); and metadata result sets building a converter map.

### Validation

- `black --check --line-length=100` clean on all changed files.
- Output-converter suite: 14 passed (incl. the new regression test) against a live SQL Server.


mssql_python/connection.py
```diff
@@ -1212,7 +1212,7 @@ def cursor(self) -> Cursor:
         logger.debug("cursor: Cursor created successfully - total_cursors=%d", len(self._cursors))
         return cursor
 
-    def add_output_converter(self, sqltype: int, func: Callable[[Any], Any]) -> None:
+    def add_output_converter(self, sqltype: Union[int, type], func: Callable[[Any], Any]) -> None:
```
subrata-ms, line 1215: Looks like Union[int, type] Is Very Broad. It seems we are now accepting any python type like list ,dict ,tuple ,set, object, ect.
Are all types actually supported?
What happens when user call/implement : add_output_converter(set, func) ?
is it the API is exposing more than it can realistically support?
Please check.

tests/test_003_connection.py
```diff
@@ -1908,6 +1909,88 @@ def test_converter_integration(db_connection):
     db_connection.clear_output_converters()
 
 
+def test_output_converter_integer_sql_type_key_gh684(db_connection):
```
subrata-ms, line 1912: As this is a API change, should we include all possible data type for testing the change?
like object, sets ect?

1. One Python converter applies to multiple SQL types (     CAST(10.5 AS DECIMAL(10,2)),3    CAST(20.5 AS NUMERIC(10,2)),4    CAST(30.5 AS MONEY),5    CAST(40.5 AS SMALLMONEY))
2. Datetime converter registration
3. Bytes converter registration
4. NULL handling
5. Mixed-result-set validation ( int + decimal + str + datetime)
6. API testing for edge cases - 
    a. conn.add_output_converter(set, conv)
    b. conn.add_output_converter(object, converter)
    c. custom instance ( class MyDecimal(decimal.Decimal))
7. Cached converter map behavior.Fetch multiple rows (cursor.fetchall()) and Verify converter selection remains correct after converter-map caching.
8. test coverage for MONEY/SMALLMONEY

mssql_python/connection.py
jahnvi480: The broad annotation is intentional (pyodbc-compatible). Keys are only stored, then dispatched by exact match against a column's integer ODBC SQL code or its materialized Python type. So add_output_converter(set, func) (or list/dict/object/a Decimal subclass) is a harmless no-op - stored but never invoked, no crash. Documented in the docstring and proven by test_output_converter_unsupported_keys_are_noops_gh684.

tests/test_003_connection.py
jahnvi480: Added more test cases

subrata-ms (approved):

subrata-ms (approved):

sumitmsft (approved):
looks good to me. Approved.

Issue #691, opened 2026-07-28
BUG: SQL_WVARCHAR output converter is applied as an unconditional catch-all to non-string columns (INT/DECIMAL/DATE)


jahnvi480:
### Summary

A single output converter registered for `SQL_WVARCHAR` (integer code `-9`) is applied as an **unconditional catch-all to every column** in a result set that lacks a more specific converter — including non-string columns such as `INT`, `DECIMAL`, and `DATE`. Users registering a converter for string decoding do not expect it to be invoked on integer/decimal/date values.

This is independent of #684 (integer-key dispatch) — it reproduces on `main` and was preserved by PR #690.

### Reproduction

```python
import os, mssql_python

conn = mssql_python.connect(os.environ["DB_CONNECTION_STRING"])
cur = conn.cursor()

# Register ONLY a SQL_WVARCHAR converter that tolerates any input type.
conn.add_output_converter(mssql_python.SQL_WVARCHAR, lambda v: f"CONV<{v!r}>")

cur.execute(
    "SELECT CAST(42 AS INT) AS n, "
    "CAST(19.99 AS DECIMAL(10,2)) AS d, "
    "CAST('2024-01-15' AS DATE) AS dt, "
    "CAST(N'hi' AS NVARCHAR(10)) AS s"
)
print(cur.fetchone())
```

### Actual output

```
n   (int)     -> 'CONV<42>'                               # WRONG
d   (Decimal) -> "CONV<Decimal('19.99')>"                 # WRONG
dt  (date)    -> 'CONV<datetime.date(2024, 1, 15)>'       # WRONG
s   (str)     -> "CONV<b'h\x00i\x00'>"                    # expected
```

The `SQL_WVARCHAR` converter fires on the `INT`, `DECIMAL`, and `DATE` columns.

### Expected

A `SQL_WVARCHAR` converter should only apply to string-typed columns (or, more strictly, only to columns whose SQL type is actually `WVARCHAR`). It should not be used as a blanket catch-all across unrelated column types.

### Root cause

`mssql_python/cursor.py` — `_build_converter_map()` uses the `SQL_WVARCHAR` converter as an unconditional fallback for any column that has no other converter, with no check on the column's type:

```python
# 3) last-resort WVARCHAR converter fallback (legacy behavior)
if converter is None:
    converter = self.connection.get_output_converter(ConstantsDDBC.SQL_WVARCHAR.value)
```

[...]

PR #692, opened 2026-07-29
FIX: gate SQL_WVARCHAR output-converter fallback to string/binary columns (#691)

jahnvi480:

> [AB#46685](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46685)

> GitHub Issue: #691

### Summary

A SQL_WVARCHAR output converter was being applied as an unconditional
catch-all to non-string columns. `Cursor._build_converter_map` fell back to the
converter registered for `SQL_WVARCHAR` for **any** column that had no direct
type-keyed converter, so registering a single `SQL_WVARCHAR` converter mangled
`INT` / `DECIMAL` / `DATE` values (e.g. `42` came back converted).

This PR gates that legacy fallback on the column's mapped Python type being
`str`/`bytes`, mirroring the `isinstance(value, (str, bytes))` guard already
present in `Row._apply_output_converters`. The optimized and fallback apply
paths now behave consistently.

### Root cause

`sql_type = desc[1]` is the column's mapped Python type. The fallback ran
whenever no direct converter matched, regardless of the column type:

```python
if converter is None:
    converter = self.connection.get_output_converter(SQL_WVARCHAR)
```

`Row._apply_output_converters` gates the same fallback on
`isinstance(value, (str, bytes))`, so the two paths disagreed.

### Fix

```python
if converter is None and sql_type in (str, bytes):
    converter = self.connection.get_output_converter(ddbc_sql_const.SQL_WVARCHAR.value)
```

`bytes` is intentionally included for parity with `Row._apply_output_converters`
(a `SQL_WVARCHAR` converter still applies to `VARBINARY`, which the new test
asserts). Also removed the now-redundant in-loop `ConstantsDDBC` import.

### Test

`tests/test_003_connection.py::test_output_converter_wvarchar_not_catchall_for_non_string_columns_gh691`
registers only a `SQL_WVARCHAR` converter and selects `INT`, `DECIMAL`, `DATE`,
`NVARCHAR`, and `VARBINARY` columns. It asserts the non-string columns are
untouched, the string/binary columns are converted, and — via a spy — that the
converter is **invoked exactly twice** (only the `str`/`bytes` columns).

The call-count assertion matters: the optimized apply path swallows converter
exceptions, so a plain `value.decode()` converter raises `AttributeError` on
`int`/`Decimal`/`date` and is silently swallowed. A value-only assertion would
therefore pass with **and** without the fix. Verified (via `git stash`) that the
test fails on the unfixed code and passes once the fallback is gated.

### Validation

- `black --check --line-length=100` clean on both files.
- `tests/test_003_connection.py`: 151 passed, 2 skipped.
- `tests/test_004_cursor.py`: 517 passed, 4 skipped.
- New test proven to fail without the fix and pass with it.


tests/test_003_connection.py
```diff
@@ -1908,6 +1908,76 @@ def test_converter_integration(db_connection):
...
+    def wvarchar_spy(value):
+        # Record the invocation first, before anything that could raise, so the
+        # spy captures calls even on non-string values (int/Decimal/date).
+        converter_calls.append(value)
+        if isinstance(value, bytes):
+            return "CONV:" + value.decode("utf-16-le")
+        return "CONV_NON_STRING"
+
+    cursor = db_connection.cursor()
+    db_connection.add_output_converter(sql_wvarchar, wvarchar_spy)
```
subrata-ms, line 1945: Can we add following test case?
1. NULL string values ( SELECT CAST(NULL AS NVARCHAR(20))
2. VARCHAR ( CAST('abc' AS VARCHAR(20))
3. Mixed-result-set scenario ( CAST(1 AS INT),3 CAST('abc' AS NVARCHAR(10)),4 CAST(12.34 AS DECIMAL(10,2))
4. Multiple fetches use cached converter map correctly (Ensuring behavior remains identical on subsequent fetches.)

jahnvi480: Added more tests

subrata-ms (dismissed):

tests/test_003_connection.py
```diff
@@ -1908,6 +1908,203 @@ def test_converter_integration(db_connection):
     db_connection.clear_output_converters()
 
 
+def test_output_converter_wvarchar_not_catchall_for_non_string_columns_gh691(db_connection):
```
sumitmsft, line 1912: There's no test for a converter registered on a **non-string** column type like `INT`. This is worth adding because your fix changes how non-string columns behave, they now skip the WVARCHAR fallback entirely and rely only on their own directly-registered converter.

Two small things a test would confirm:
1. **Positive:** a converter registered for an INT column still runs correctly.
2. **Negative (GH #691 case):** when no converter is registered for INT, the WVARCHAR converter does **not** sneak in and touch it.

Not blocking but just a nice way to lock in the new behavior so it doesn't regress later.

sumitmsft (changes requested):

tests/test_003_connection.py
jahnvi480: Added required testcase

subrata-ms (approved):

sumitmsft (approved):

PR #693, opened 2026-07-31
REFACTOR: remove bundled ODBC driver from mssql-python wheel; require mssql-python-odbc (Phase 2)

jahnvi480:

> [AB#46836](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46836)

### Summary
This pull request makes significant updates to the build and packaging process for the `mssql-python` and `mssql-python-odbc` Python packages, reflecting the transition to Phase 2 where the ODBC driver is fully separated into its own package. The pipeline now always builds both packages together, ensures the external ODBC wheel is installed before testing, and simplifies and streamlines build and test stages. Documentation and prompts are also updated to reflect these structural changes.

**Build and Pipeline Simplification:**

* The pipeline now always builds both `mssql-python` and `mssql-python-odbc` packages together; parameters and logic for building only one package have been removed, simplifying configuration and execution. [[1]](diffhunk://#diff-3db20ac3b805d9b3612357e92c428cda06eaa8a8fa2fafe94ecc247177608996L73-L89) [[2]](diffhunk://#diff-3db20ac3b805d9b3612357e92c428cda06eaa8a8fa2fafe94ecc247177608996L197-L204)
* The ODBC build is consolidated into a single `ODBC_BuildAll` stage that produces all platform wheels from one host, replacing multiple per-platform ODBC build stages.
* All `mssql-python` build stages now depend on the `ConsolidateOdbc` stage and install the external ODBC wheel before running tests, validating the package split end-to-end during the build. The previous post-consolidation "TestBothWheels" stages are removed as redundant.

**Documentation and Prompt Updates:**

* Documentation and prompts are updated to reference the new `mssql_python_odbc/libs` location for ODBC binaries and to clarify that these are now maintained in the standalone `mssql-python-odbc` package. 

**Platform and Security Improvements:**

* The macOS build pool is updated to use the latest available image (`macos-latest`, macOS-15 Sequoia).
* The SBOM (Software Bill of Materials) is now always named after the primary `mssql-python` wheel, reflecting the unified build process.

**Release Pipeline Robustness:**

* The release pipeline now derives the expected ODBC package version directly from `mssql_python_odbc/__init__.py`, ensuring the version pin is always in sync and eliminating the need for manual updates when the ODBC version changes.


setup.py
bewithgaurav: same single-source-of-truth point from #663, it can be followed here as well:

the driver filename derives from `mssql_python_odbc.__version__` now, but the version is still hand-typed in three places: this pin (`setup.py:178`), and `$expectedVersion` in `official-release-pipeline.yml:221` and `dummy-release-pipeline.yml:221`

bump the driver and miss one, they drift silently

`setup_odbc.py` already reads `__version__` via `_read_odbc_version()`, so `setup.py` can too, and the release yaml can awk it like `pr-validation` does. one version, one place.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1182,24 +1157,27 @@ std::string GetOdbcLibsBaseDir() {
...
-            // pybind11 has already fetched and cleared the CPython error
-            // indicator, so re-importing `mssql_python` below is safe.
-            LOG("GetOdbcLibsBaseDir: mssql_python_odbc not installed (%s); "
-                "falling back to bundled libs in mssql_python",
+            // Phase 2: the standalone package is required. Turn the missing
+            // dependency into a clear, actionable error instead of a fallback.
+            LOG("GetOdbcLibsBaseDir: required package mssql_python_odbc is not installed (%s)",
                 e.what());
-            return GetModuleDirectory();
+            ThrowStdException(
```
bewithgaurav, line 1178: as we added last time, the happy path is already tested end to end (build stages install the odbc wheel and run the full suite, pr-validation resolves the driver from the in-repo `mssql_python_odbc` and connects)

there is however gap is in this negative path: nothing installs mssql-python without odbc and checks the new error

need tests for these two exceptions we've now added, `GetOdbcLibsBaseDir` now raises instead of falling back: 

- a missing package points to `pip install mssql-python-odbc` (line 1179 - this one)
- an incomplete one to `--force-reinstall` (line 1166), and nothing asserts either

`test_008_auth` already does exactly this for azure-identity (`test_import_error_when_azure_identity_missing`), so the same shape fits here

OneBranchPipelines/stages/build-odbc-all-stage.yml
```diff
@@ -0,0 +1,217 @@
...
+    type: object
+    default:
+      - { tag: 'win_amd64', arch: 'x64' }
+      - { tag: 'win_arm64', arch: 'arm64' }
+      # macOS: universal2 data-only wheel. macosx_11_0 is the canonical universal2
+      # floor (Big Sur — first arm64 macOS) and the driver dylibs run on macOS 11+
+      # (proven by pre-split connection tests on the macOS-14 agent), so this is
+      # pip-installable on every macOS the mssql-python wheel targets, including the
+      # macOS-14 build/test agent (a macosx_15_0 tag is rejected there).
+      - { tag: 'macosx_11_0_universal2', arch: 'universal2' }
```
bewithgaurav: mssql-python-odbc ships `macosx_11_0_universal2`, mssql-python ships `macosx_15_0_universal2` (`setup.py:37`, `setup_odbc.py:122`). the two are inconsisten on the macOS floor, not an "issue" but a maintainability inconsistency

I traced a bit more on the "why" - although you've already mostly explained in comments :

- the OneBranch macOS build stage is pinned to `macOS-14` (`build-macos-single-stage.yml:48`).
- odbc is pip-installed on that agent before pytest, and a `15_0` wheel won't install on a 14 host, so it was dropped to `11_0`.
- mssql-python is never installed there. it is tested from the source tree, and the `15_0` tag is stamped after pytest (`:253`), so `15_0` is never install-tested in CI.
- PR-validation already runs `macos-latest`, which resolved to `macOS-15` on this run, so the release pipeline is a version behind the gate.

that macOS-14 pin is the only reason odbc is `11_0`. it does not break installs (odbc's lower floor installs anywhere `15` does), so this is consistency, not a runtime bug.

mssql-python stays at `15_0`, so one fix:

- bump the OneBranch macOS stage to `macos-latest`/`macOS-15` like the gate, then both wheels tag `15_0`, and `15_0` finally gets install-tested.

this is not a blocker though - will leave upto you to take it up as a follow up task or fold it in here

bewithgaurav (dismissed):
need to add some failure case tests, and some improvement suggestions

setup.py
jahnvi480: Fixed

OneBranchPipelines/stages/build-odbc-all-stage.yml
jahnvi480: Fixed

mssql_python/pybind/ddbc_bindings.cpp
jahnvi480: Added

bewithgaurav (approved):

sumitmsft (approved):

PR #701, opened 2026-08-05
RELEASE: mssql-python-odbc 18.6.2.1 - correct license metadata + PyPI description

jahnvi480:

> [AB#46836](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/46836)

### Summary

Metadata-only re-release of the standalone **mssql-python-odbc** package (identical driver binaries as the already-published 18.6.2) that corrects the package license and ships a proper PyPI description.

## What changed

- **License: MIT → Other/Proprietary License** — the `license` field and trove classifier in `setup_odbc.py`. The wheel redistributes Microsoft's proprietary ODBC Driver 18 for SQL Server (and the VC++ runtime it links against), so the previous `MIT` label was incorrect. The build script's own source stays MIT.
- **PyPI `long_description`** — replaced the plain one-line description with a rendered Markdown description: project role, "do not install directly" guidance, docs link, and a **License information** section linking both redistributed Microsoft license texts (same style as `azure-cognitiveservices-speech`). `long_description_content_type` switched to `text/markdown`.
- **Committed canonical `mssql_python_odbc/licenses/` dir** — byte-exact copies of the ODBC Driver and Visual Studio license texts. The description links and `license_files` now point here, so the links resolve on PyPI/GitHub and the wheel embeds the license texts deterministically — independent of the build-time-populated `libs/` tree (empty on `main`).
- **Version 18.6.2 → 18.6.2.1** — packaging revision 1 of the same driver. A literal `v1` suffix is not a valid PEP 440 version and `+local` versions are rejected by PyPI's upload API, so the revision is expressed as a fourth numeric component. Ordering holds: `18.6.2 < 18.6.2.1 < 18.6.3`.
- **Relaxed `mssql_python/pybind/CMakeLists.txt` version parser** — now accepts a trailing packaging-revision component / PEP 440 suffix. Only the leading MAJOR and MAJOR.MINOR are used to derive the driver filename, so the extra component has no effect on which driver is loaded.

## Ordering

This lands **before** the Phase 2 wheel-slimming PR (#693). After this release publishes `18.6.2.1` to PyPI, a small follow-up bumps `mssql-python`'s pinned dependency from `==18.6.2` to `==18.6.2.1` so installs pull the correctly-licensed wheel.

## Validation

- `black --check --line-length=100 setup_odbc.py` — passes.
- CMake version-parse of `18.6.2.1` exercised with the real regex engine → `MAJOR=18`, `MAJOR_MINOR=18.6`, no `FATAL_ERROR`.
- Driver binaries unchanged from 18.6.2 (metadata-only re-release).


setup_odbc.py
```diff
@@ -211,22 +211,68 @@ def run(self):
 
 _require_min_setuptools()
 
+# PyPI long description (Markdown). Kept at module scope and UNINDENTED so the
+# Markdown is not accidentally turned into indented code blocks by PyPI's
+# renderer. The "License information" links point at the two redistributed
+# Microsoft license texts that ship inside every wheel.
+_LONG_DESCRIPTION = """\
```
bewithgaurav: better to add a `PyPI_Description.md` instead here (same as mssql-python)

bewithgaurav (dismissed):
approving with a minor suggestion

setup_odbc.py
jahnvi480: Added

bewithgaurav (approved):

sumitmsft (approved):

Issue #670, opened 2026-07-09
executemany() sporadically fails to insert rows, throws no exception


Mapkinasekinase:
### Describe the bug
Short version:
When inserting data from a pandas dataframe through executemany(), the insert will *sporadically* fail (0 rows inserted) and *sporadically* work (86 rows inserted), even if everything else remains the same.
No exceptions are thrown, neither on the SQL side, nor in Python.

Longer version / additional context:
I import an .xpt file using pd.readsas(). The file contains 86 rows and 19 columns. For 2 rows, the data is partially incomplete and information for some of the columns is missing.
The bug is correlated with this particular file – other files import successfully – but, for runs in which all 86 rows are imported successfully, the data is then completely fine and properly stored in the DB.
When I omit the two rows with incomplete data, all rows are always inserted correctly. Likewise, when I omit one particular column, all rows are always inserted correctly.
When I iterate through many execute() commands, rather than using executemany(), all rows are always inserted correctly.
When I convert the datatypes to string for all columns ahead of the executemany() command, all rows are always inserted correctly.

Exception message: None
Stack trace:

SQL Server Profiler:

A successful run will show:
- 1 sp_prepare statement
- 86 exec sp_execute statements
- 1 exec sp_unprepare statement

A non successful run will show:
- 1 sp_prepare statement
- 1 exec sp_unprepare statement

MSSQL Python Logs:
Complete logs attached.

A successful run will end in:

2026-07-09 10:28:51.194, 19736, DEBUG, ddbc_bindings.cpp:2658, DDBC, SQLExecuteMany: PARAMSET_SIZE set to 86
2026-07-09 10:28:51.202, 19736, DEBUG, ddbc_bindings.cpp:2665, DDBC, SQLExecuteMany: SQLExecute completed - rc=0
2026-07-09 10:28:51.203, 19736, DEBUG, ddbc_bindings.cpp:1611, DDBC, SQLGetAllDiagRecords: Retrieving all diagnostic records for handle 0000026954E2BE60, handleType=3
2026-07-09 10:28:51.203, 19736, DEBUG, ddbc_bindings.cpp:5708, DDBC, SQLRowCount_wrap: Get number of rows affected by last execute
2026-07-09 10:28:51.203, 19736, DEBUG, ddbc_bindings.cpp:5721, DDBC, SQLRowCount_wrap: SQLRowCount returned 86
[...]

PR #702, opened 2026-08-05
FIX: Initialize numeric array indicators in executemany

gargsaumya:

> GitHub Issue: #670 

### Summary
- Initialize every fixed-width numeric parameter indicator before ODBC array execution.
- Prevent late NULL values from leaving earlier indicator slots uninitialized.
- Add regression coverage for TINYINT, SMALLINT, INT, and FLOAT batches with late NULLs.

### Root cause
The affected numeric binding paths allocated the ODBC indicator array only after encountering the first NULL. Earlier rows then contained uninitialized indicators, which could be interpreted as data-at-execution markers and cause a batch to insert zero rows without an exception.

### Validation
- Native extension build succeeded on Windows x64 with Python 3.13.
- Reproduction improved from 209/300 anomalous batches to 0/300.
- Focused cursor tests: 2 passed.
- Black check and git diff whitespace check passed.


subrata-ms (approved):
Looks good to me. Only concern is the code coverage showing pretty less. Probably you need to modify the existing test to cover the new  AllocateParamBufferArray allocation call.

sumitmsft (approved):

PR #714, opened 2026-08-11
FIX: forked-pr-coverage.yml env-injection via untrusted artifact

sumitmsft:

[AB#46466](https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/46466)

### Summary

- Treat fork-produced coverage artifacts as untrusted input and validate their schema and values before use.
- Bind coverage comments to the triggering pull request instead of accepting an artifact-supplied PR number.
- Remove untrusted artifact propagation through GITHUB_ENV and reject unexpected files, unsafe URLs, control characters, and oversized values.
- Pin GitHub Actions to immutable commit SHAs and disable persisted checkout credentials.
- Add regression coverage for environment injection, cross-PR targeting, executable artifacts, and Markdown/URL injection.

### Validation

- 12 targeted security regression tests passed.
- Changed workflow YAML parsed successfully.
- Black formatting checks passed.
- Native extension build completed successfully.


tests/test_fork_coverage_security.py
```diff
@@ -0,0 +1,153 @@
+import importlib.util
+import json
+from pathlib import Path
+
+import pytest
+
+SCRIPT_PATH = Path(__file__).parents[1] / ".github" / "scripts" / "prepare_fork_coverage_comment.py"
+SPEC = importlib.util.spec_from_file_location("prepare_fork_coverage_comment", SCRIPT_PATH)
+coverage_comment = importlib.util.module_from_spec(SPEC)
+SPEC.loader.exec_module(coverage_comment)
```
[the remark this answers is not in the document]
sumitmsft: No change here. The import targets a fixed, checked-in repository path, and a missing script already fails test collection immediately. Explicit SPEC/loader assertions would only alter diagnostics and do not affect the runtime or security contract covered by this PR.

.github/scripts/prepare_fork_coverage_comment.py
```diff
@@ -0,0 +1,243 @@
...
+            raise ValidationError("artifact must not contain symbolic links")
+        if path.is_file():
+            files.append(path.relative_to(artifact_directory).as_posix())
+    if files != ["pr-info.json"]:
+        raise ValidationError("artifact must contain only pr-info.json")
+
+    data = _load_json(artifact_directory / "pr-info.json")
+    if not isinstance(data, dict) or set(data) != EXPECTED_FIELDS:
+        raise ValidationError("pr-info.json does not match the expected schema")
+
```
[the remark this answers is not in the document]
sumitmsft: Good catch. Fixed in 5aec312d: the validator now inspects only the top-level entries and requires exactly one regular, non-symlink pr-info.json. I also added a regression test that rejects nested directory trees.

.github/scripts/prepare_fork_coverage_comment.py
```diff
@@ -0,0 +1,243 @@
...
+    matching_pulls = [
+        pull
+        for pull in associated_pulls
+        if pull.get("head", {}).get("sha") == head_sha
+        and pull.get("base", {}).get("ref") == default_branch
+        and pull.get("base", {}).get("repo", {}).get("full_name") == repository_name
+        and isinstance(pull.get("number"), int)
+        and pull["number"] > 0
+    ]
+    if len(matching_pulls) != 1:
```
bewithgaurav, line 188: forked PRs never get a comment. this line always fails for them.

`GET /commits/{sha}/pulls` only sees commits that live in the base repo, and a fork's head commit doesn't, so `matching_pulls` is empty every time. checked all open fork PRs I could find across mssql-python, every one returns 0. the same call returns 1 for a same-repo PR like this one, which is why it looks fine from here.

ran a real artifact plus the live api responses for 3 forked PRs through the script, all three exit 2 on `workflow run must resolve to exactly one pull request`.

`GET /pulls?state=open` matched on `head.sha` resolves all three. the exactly-one-match, base repo and base branch checks all carry over, so no protection is given up

.github/scripts/prepare_fork_coverage_comment.py
```diff
@@ -0,0 +1,243 @@
...
+        raise ValidationError("event is missing the repository name")
+    if not isinstance(default_branch, str) or not default_branch:
+        raise ValidationError("event is missing the default branch")
+    if not isinstance(head_sha, str) or not re.fullmatch(r"[0-9a-f]{40}", head_sha):
+        raise ValidationError("workflow run has an invalid head SHA")
+    if head_repository.get("full_name") == repository_name:
+        raise ValidationError("workflow run did not originate from a fork")
+
+    event_pulls = workflow_run.get("pull_requests") or []
+    if len(event_pulls) == 1:
```
bewithgaurav: this is dead code actually

line 170 already bails out when the run isn't from a fork, and github always sends an empty `pull_requests` on fork runs, 8/8 real fork runs I sampled, so this never fires

the cost is that `test_prepares_safe_comment_for_trusted_event_pr` only exercises this branch. it goes green while the path that actually runs is untested and broken

.github/workflows/forked-pr-coverage.yml
```diff
@@ -27,85 +27,78 @@ jobs:
...
+      github.event.workflow_run.head_repository.full_name != github.repository
     permissions:
+      actions: read
       pull-requests: write
       contents: read
     
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
```
bewithgaurav: #716 that was raised yesterday pins these same SHAs, plus setup-python, github-script, devskim and codeql, and carries the exact patch version in the comment

worth dropping from here so the two don't collide

same for `persist-credentials: false`, that landed with #713

.github/scripts/prepare_fork_coverage_comment.py
```diff
@@ -0,0 +1,243 @@
...
+    if (
+        len(entries) != 1
+        or entries[0].name != "pr-info.json"
+        or not entries[0].is_file()
+        or entries[0].is_symlink()
+    ):
+        raise ValidationError("artifact must contain only pr-info.json")
+
+    data = _load_json(artifact_directory / "pr-info.json")
+    if not isinstance(data, dict) or set(data) != EXPECTED_FIELDS:
```
bewithgaurav: suggestion: check the six fields are there rather than requiring an exact match.

as it stands, anyone who adds a field to the artifact later turns fork comments off without knowing, and nothing on the PR says why - the extra field is harmless anyway, nothing here reads past the six it wants

.github/workflows/forked-pr-coverage.yml
```diff
@@ -27,85 +27,78 @@ jobs:
...
-            exit 0
-          fi
-          
-          # Verify artifact was downloaded
-          if [[ ! -f pr-info.json ]]; then
-            echo "⚠️ Artifact downloaded but pr-info.json not found"
-            echo "This may indicate an issue with artifact upload"
+          set -euo pipefail
+
+          [[ "$RUN_ID" =~ ^[0-9]+$ ]] || {
```
bewithgaurav: suggestion: drop the `RUN_ID`, `HEAD_SHA` and `PR_NUMBER` guards, the script already covers all three.

the first two come straight from github, and the script re-checks the sha anyway
the `PR_NUMBER` one further down is checking this script's own output - same rule is written twice

bewithgaurav (changes requested):
requesting changes - some issues wrt forked pr comments and refactoring suggestions to cut the PR short

.github/scripts/prepare_fork_coverage_comment.py
```diff
@@ -0,0 +1,243 @@
...
+    if len(value.splitlines()) > 10:
+        raise ValidationError("low_coverage_files contains more than 10 lines")
+    if any(
+        character != "\n" and unicodedata.category(character).startswith("C") for character in value
+    ):
+        raise ValidationError("low_coverage_files contains control characters")
+    return value
+
+
+def _validate_ado_url(value) -> str:
```
gargsaumya, line 96: ado_url comes from ADO's _links.web.href, but this only accepts dev.azure.com, an exact path, and a lone buildId param. If ADO returns extra query params or the *.visualstudio.com host, the comment is silently dropped. Please confirm a real build's href passes these rules

.github/workflows/forked-pr-coverage.yml
```diff
@@ -27,85 +27,78 @@ jobs:
...
+          PR_NUMBER_FILE="${RUNNER_TEMP}/coverage-pr-number"
+
+          gh run download "$RUN_ID" \
+            --repo ${{ github.repository }} \
+            --name coverage-comment-data \
+            --dir "$ARTIFACT_DIR"
+
+          gh api \
+            -H "Accept: application/vnd.github+json" \
+            "repos/${GITHUB_REPOSITORY}/commits/${HEAD_SHA}/pulls" > "$PULLS_FILE"
```
gargsaumya: When this moves to /pulls?state=open (per the resolve fix), add --paginate — the default 30-per-page cap can miss the match on a repo with many open PRs.

.github/scripts/prepare_fork_coverage_comment.py
gargsaumya: +1

/commits/{sha}/pulls can't see a fork's head commit, so it's empty every time. /pulls?state=open + head.sha match fixes it with no loss of protection; head_sha still comes from the trusted workflow_run event. Blocking.

.github/scripts/prepare_fork_coverage_comment.py
gargsaumya: Agree- fork workflow_run events always send empty pull_requests, so this never fires, and it's the only branch the green happy-path test covers.

.github/workflows/forked-pr-coverage.yml
gargsaumya: Agree - all three are re-validated in the script, safe to remove.

gargsaumya (changes requested):

.github/scripts/prepare_fork_coverage_comment.py
sumitmsft: Fixed - switched to `/pulls?state=open` matched on the trusted `head_sha`; `/commits/{sha}/pulls` never sees a fork's head commit.

.github/scripts/prepare_fork_coverage_comment.py
sumitmsft: Removed the dead branch and repointed the happy-path test at real resolution, plus a regression that spoofed event pulls are ignored.

.github/workflows/forked-pr-coverage.yml
sumitmsft: Reverted the action SHA pins to tags so #716 owns pinning; `persist-credentials` already matches main from #713.

.github/scripts/prepare_fork_coverage_comment.py
sumitmsft: Relaxed to require the six fields present, not an exact set, so an added field won't silently disable fork comments - values still strictly validated.

.github/workflows/forked-pr-coverage.yml
sumitmsft: Dropped the `RUN_ID`/`HEAD_SHA`/`PR_NUMBER` guards - all three are re-validated in the script or come straight from the trusted event.

.github/scripts/prepare_fork_coverage_comment.py
sumitmsft: Confirmed against 5 live builds - `_links.web.href` returns `dev.azure.com` + project GUID + a lone `buildId`, which passes the validator and normalizes to the canonical URL. No `visualstudio.com` or extra params in practice.

.github/workflows/forked-pr-coverage.yml
sumitmsft: Done - the resolve fix uses `gh api --paginate --slurp`, so the default 30-per-page cap is handled.

.github/workflows/forked-pr-coverage.yml
```diff
@@ -27,85 +27,65 @@ jobs:
...
-            echo "$LOW_COV_FILES"
-            echo "EOF"
-          } >> $GITHUB_ENV
-          
-          {
-            echo "PATCH_COVERAGE_SUMMARY<<EOF"
-            echo "$PATCH_SUMMARY"
-            echo "EOF"
-          } >> $GITHUB_ENV
+          gh api --paginate --slurp \
```
bewithgaurav, line 63: blocker: gh won't run this. `--slurp` and `--jq` can't be used together:

```
$ gh api --paginate --slurp "repos/.../pulls?state=open&per_page=100" --jq 'add'
the `--slurp` option is not supported with `--jq` or `--template`
exit 1
```

fails on every gh version, and the whole step goes with it 
same combination at line 75 (commented there as well - with a better suggestion)

piping instead works:

```
gh api --paginate --slurp "repos/${GITHUB_REPOSITORY}/pulls?state=open&per_page=100" \
  | jq 'add' > "$PULLS_FILE"
```

checked against the live api and all three fork PRs here resolve correctly through it

.github/scripts/prepare_fork_coverage_comment.py
```diff
@@ -0,0 +1,237 @@
...
+        raise ValidationError("workflow run has an invalid head SHA")
+    if head_repository.get("full_name") == repository_name:
+        raise ValidationError("workflow run did not originate from a fork")
+
+    if not isinstance(associated_pulls, list):
+        raise ValidationError("associated pull request response must be a list")
+    matching_pulls = [
+        pull
+        for pull in associated_pulls
+        if pull.get("head", {}).get("sha") == head_sha
```
bewithgaurav, line 181: optional: match the fork name too, not just the commit.

two forks sitting on the same commit would both match today, and then nobody gets a comment. harmless, just noticed it

.github/workflows/forked-pr-coverage.yml
```diff
@@ -27,85 +27,65 @@ jobs:
...
+          python .github/scripts/prepare_fork_coverage_comment.py \
+            --artifact-directory "$ARTIFACT_DIR" \
+            --event "$GITHUB_EVENT_PATH" \
+            --associated-pulls "$PULLS_FILE" \
+            --comment-output "$COMMENT_FILE" \
+            --pr-number-output "$PR_NUMBER_FILE"
+
+          PR_NUMBER="$(cat "$PR_NUMBER_FILE")"
+
+          COMMENT_ID="$(
```
bewithgaurav, line 77: this has the same blocker as the line 60 one, I have a better suggestion which includes fixing that as well: 

`gh pr comment` does find-or-create on its own, so this whole block collapses to one call.

```
gh pr comment "$PR_NUMBER" --repo "$GITHUB_REPOSITORY" \
  --body-file "$COMMENT_FILE" --edit-last --create-if-none
```

18 lines down to 2. the script writes the body as markdown instead of wrapping it in json, same line either way. the argus writeup named this one directly, a plain `gh pr comment` call, so it stays on the no-node-action path you took.

if you do take it, the line 75 call goes with the block, so that one needs no fix.

bewithgaurav (changes requested):
requesting changes - for a different blocker this time, added a couple of suggestions as well

.github/workflows/forked-pr-coverage.yml
sumitmsft: Confirmed and fixed - gh api --slurp can't combine with --jq (exits 1 on gh 2.96), so I pipe --slurp output to jq 'add'; verified end to end against a real fork PR.

.github/workflows/forked-pr-coverage.yml
sumitmsft: Same --slurp/--jq fix applied here (piped to jq -r). I kept the marker-based find-or-create rather than --edit-last, since --edit-last isn't marker-aware and could edit an unrelated github-actions[bot] comment - happy to switch if you'd prefer the shorter form.

.github/scripts/prepare_fork_coverage_comment.py
sumitmsft: Good catch - now also matching the fork's head.repo.full_name from the trusted workflow_run event, so two forks on the same commit resolve to the right PR instead of colliding; added tests.

bewithgaurav (approved):

gargsaumya (approved):

Issue #712, opened 2026-08-10
`return` in a `finally` block at cursor.py:3046 breaks import on Python 3.14 and discards fetch errors


rgant:
### Describe the bug

`mssql_python/cursor.py:3046` puts a bare `return` inside a `finally` block. The block belongs to the `batch_generator` generator in `Cursor.arrow_reader`. The construct arrived in #644 and shipped in 1.13.0.

This has two effects.

**1. Python 3.14 refuses to import the driver under warnings-as-errors.**

PEP 765 makes `return` in a `finally` block a `SyntaxWarning` in Python 3.14. A project that turns warnings into errors cannot import `mssql_python` at all:

```
Exception message:
  File "/path/to/site-packages/mssql_python/cursor.py", line 3046
    return
    ^^^^^^
SyntaxError: 'return' in a 'finally' block

Stack trace:
  mssql_python/__init__.py:54:  from .db_connection import connect, Connection
  mssql_python/db_connection.py:9:  from mssql_python.connection import Connection, TokenProvider
  mssql_python/connection.py:22:  from mssql_python.cursor import Cursor
```

Version 1.12.0 does not carry this line. Version 1.13.0 does. I compiled `cursor.py` from both wheels to confirm that.

**2. The `return` discards an exception raised inside the generator.**

The guard reads:

```python
cur = cursor_ref[0]
cursor_ref[0] = None
if cur is None or cur.closed or cur.hstmt is None:
    return
```

If `arrow_batch()` raises and the guard is true, the `return` replaces the exception with a normal end of iteration. The caller reads a truncated result set and sees no error. I did not confirm that a closed cursor and an in-flight exception occur together in practice. The language semantics hold either way.

The comment above the block states that the teardown runs on "an exception inside the body". The `return` is the one path where that intent does not hold.

### To reproduce

Effect 1:

```python
import warnings

warnings.simplefilter('error')
import mssql_python  # SyntaxError: 'return' in a 'finally' block
```

Effect 2, reduced to the same shape as `batch_generator`:

```python
def make(guard):
    def gen():
        try:
            yield 1
            raise RuntimeError('fetch failed')
        finally:
            if guard:
                return          # same shape as cursor.py:3046
            print('cleanup ran')
    return gen

g = make(True)()
next(g)
try:
    next(g)
except RuntimeError:
    print('propagated')
except StopIteration:
    print('the RuntimeError was discarded')   # this branch runs

g = make(False)()
next(g)
try:
    next(g)
[...]

PR #718, opened 2026-08-13
FIX: Preserve Arrow reader fetch exceptions (#712)

subrata-ms:

> [AB#47281](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47281)

> GitHub Issue: #712 

### Summary   
  
This pull request improves the robustness and test coverage of the cursor cleanup logic in the `mssql_python` package, particularly around error handling during fetch operations and resource cleanup. It also introduces a build check to ensure that the driver source code compiles cleanly with warnings treated as errors in Python 3.14 and above.

**Error handling and cleanup improvements:**

* Refactored the cleanup guard in `batch_generator()` in `cursor.py` to only skip cleanup if the cursor is `None`, closed, or has no `hstmt`, improving clarity and correctness.

**Test coverage enhancements:**

* Added tests in `test_004_cursor_arrow.py` to verify that fetch errors are properly propagated and not masked by defensive cleanup logic, both when cleanup is skipped and after normal cleanup. These tests use fake cursor objects to simulate various error and cleanup scenarios.

**Build and compatibility checks:**

* Introduced a new test in `test_004_cursor.py` to ensure that `cursor.py` compiles successfully with `SyntaxWarning` promoted to an error, as required by Python 3.14+ (PEP 765). This helps future-proof the codebase against upcoming Python changes.
* Added necessary imports (`subprocess`, `sys`, `Path`) in `test_004_cursor.py` to support the new compilation test.


mssql_python/cursor.py
```diff
@@ -3088,54 +3088,52 @@ def batch_generator():
...
-                # 3) Drain diagnostics produced by SQL_CLOSE itself.  This
-                #    runs unconditionally because SQL_CLOSE can return
-                #    SQL_SUCCESS_WITH_INFO (a *success* code) and still leave
-                #    warning records on the HSTMT diag stack; the previous
-                #    "only on failure" path would silently drop those.
-                try:
-                    cur.messages.extend(ddbc_bindings.DDBCSQLGetAllDiagRecords(cur.hstmt))
-                except Exception as e:  # pylint: disable=broad-exception-caught
-                    logger.debug("arrow_reader cleanup: post-close diag drain failed: %s", e)
+                if cur is not None and not cur.closed and cur.hstmt is not None:
```
bewithgaurav: nit: `cur` is never `None` here, it gets read on the line right above and this block runs once per reader
so `if not cur.closed and cur.hstmt is not None:` would do the same, only mentioning it since the line is already changing

tests/test_004_cursor.py
```diff
@@ -108,6 +111,21 @@
 ]
 
 
+@pytest.mark.skipif(sys.version_info < (3, 14), reason="PEP 765 warnings begin in Python 3.14")
+def test_cursor_compiles_with_warnings_as_errors():
+    """The driver source must compile when SyntaxWarning is promoted to an error."""
+    cursor_source = Path(__file__).parents[1] / "mssql_python" / "cursor.py"
```
bewithgaurav: this checks `cursor.py` only, but what broke was importing the package
the same `return` in a `finally` in any other file breaks it the same way, and this test would stay green

looping over the package covers all of it:

```python
package_dir = Path(__file__).parents[1] / "mssql_python"
for source in sorted(package_dir.glob("*.py")):
    with warnings.catch_warnings():
        warnings.simplefilter("error")
        compile(source.read_text(encoding="utf-8"), str(source), "exec")
```

ran it on 3.14 and 3.13, green on both. so the `skipif` can go and it covers every leg instead of just the 3.14 ones. needs `import warnings`, and `subprocess` becomes unused

tests/test_004_cursor_arrow.py
```diff
@@ -688,6 +688,95 @@ def close(self):
...
+    fake_cursor = FakeCursor()
+    reader = mssql_python.Cursor.arrow_reader(fake_cursor, batch_size=1)
+    try:
+        with pytest.raises(RuntimeError, match="fetch failed"):
+            reader.read_next_batch()
+        assert fake_cursor.hstmt.close_calls == 1
+        assert fake_cursor.rownumber_cleared is True
+        assert fake_cursor.rowcount == -1
+    finally:
+        reader.close()
```
bewithgaurav: requesting changes on this one - since these two go green even when the driver is broken.
they never open a connection, so a real regression under the fake object still shows green here

the same bug is testable through the driver: close the cursor part way through a read and assert it raises. fails on the old guard, passes here, adding a suggestion:

```suggestion
_BIG_QUERY = (
    "SELECT TOP (3000) ROW_NUMBER() OVER (ORDER BY (SELECT NULL)) AS n "
    "FROM sys.all_objects a CROSS JOIN sys.all_objects b"
)


def test_arrow_reader_raises_when_cursor_closes_mid_stream(db_connection):
    """A cursor closed mid-stream must raise, not report a short result set."""
    cur = db_connection.cursor()
    cur.execute(_BIG_QUERY)
    reader = cur.arrow_reader(batch_size=500)

    rows = 0
    with pytest.raises(mssql_python.Error):
        for batch in reader:
            rows += batch.num_rows
            if rows >= 1000:
                cur.close()
    assert 0 < rows < 3000


def test_arrow_reader_raises_when_cursor_scope_already_exited(db_connection):
    """A reader outliving its cursor's `with` block must raise, not yield nothing."""
    with db_connection.cursor() as cur:
        cur.execute(_BIG_QUERY)
        reader = cur.arrow_reader(batch_size=500)

    with pytest.raises(mssql_python.Error):
        for _ in reader:
            pass
```

bewithgaurav (changes requested):
requesting changes on fake cursor tests, a suggestion, and a nit comment
the core fix lgtm

sumitmsft (dismissed):

mssql_python/cursor.py
subrata-ms: fixed

tests/test_004_cursor.py
subrata-ms: fixed

tests/test_004_cursor_arrow.py
subrata-ms: made following changes - 

-  Removed both fake-cursor regression tests.

-  Added a real connection/cursor lifecycle test.

-  It reads one batch, closes the cursor, then verifies the next read propagates [ProgrammingError](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

bewithgaurav (approved):
lgtm

sumitmsft (approved):

PR #719, opened 2026-08-13
FIX: Redact parameter values from executemany Decimal-conversion error

sumitmsft:

> [AB#47300](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47300)

### Summary

The `executemany()` Decimal/NUMERIC conversion path embedded the **entire parameter row** into the raised `ValueError`:

```
ValueError(f"Failed to convert parameter at row {row}, column {i} to Decimal: {e}")
```

Here `row` is the full parameter tuple, so every column value in a failing row (potentially PII such as SSNs, emails, names, or balances) was placed into the exception message. That message propagates to caller error handlers, tracebacks, and APM/log shippers (Sentry, Splunk, App Insights) **even when driver DEBUG logging is never enabled**, because uncaught exceptions surface by default.

This change makes the error **metadata-only**: it reports the row index, column index, and the value's type name, and never the value or the row.

**Before**
```
Failed to convert parameter at row ('Jane Doe', '123-45-6789', 'jane@x.com', 'N/A'), column 3 to Decimal: [<class 'decimal.ConversionSyntax'>]
```

**After**
```
Failed to convert parameter to Decimal at row 0, column 3 (value type: str)
```

#### Hardening: value-bearing exception cause

Redacting only the outer message is not enough. `raise ... from e` chains the original exception, which surfaces through `__cause__` and `traceback.format_exc()` -- exactly what APM/log shippers capture. For an ordinary bad string the chained cause is the value-free `decimal.InvalidOperation` (`ConversionSyntax`), but for a value whose `str()` raises with content (or any error echoing the input) the sensitive text would still leak via the cause.

The conversion now splits `str(val)` from the decimal parse and chains **only** `decimal.DecimalException`, which is proven value-free (it never echoes the input). `str(val)` failures and any other unexpected error are re-raised with the chain suppressed (`from None`), so the metadata-only guarantee holds across `__cause__` and formatted tracebacks, not just `str(exc)`.

#### Changes
- `mssql_python/cursor.py`: enumerate rows for an index; raise a value-free, metadata-only `ValueError`; chain only the value-free `decimal.DecimalException` and suppress the cause (`from None`) for `str(val)` / unexpected failures.
- `tests/test_004_cursor.py`: strengthen the unconvertible-value test to assert the sensitive value and raw row are absent from both the message and the fully formatted traceback; add `test_setinputsizes_sql_decimal_str_raises_no_leak` covering a parameter whose `str()` raises a secret (asserts `__cause__ is None` and the secret is absent from the traceback).

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -2647,7 +2647,7 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         # Process parameters into column-wise format with possible type conversions
         # First, convert any Decimal types as needed for NUMERIC/DECIMAL columns
         processed_parameters = []
-        for row in seq_of_parameters:
+        for row_index, row in enumerate(seq_of_parameters):
             processed_row = list(row)
             for i, val in enumerate(processed_row):
                 if val is None:
@@ -2672,8 +2672,15 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
                         try:
                             processed_row[i] = format(decimal.Decimal(str(val)), "f")
                         except Exception as e:  # pylint: disable=broad-exception-caught
+                            # Do not embed the parameter value or the full row in
+                            # the message: rows may contain PII (SSNs, emails,
+                            # balances) that would leak into caller error handlers,
+                            # tracebacks, and log/APM stores. Report metadata only
+                            # (row index, column index, value type); the original
+                            # decimal error is preserved via exception chaining.
                             raise ValueError(
-                                f"Failed to convert parameter at row {row}, column {i} to Decimal: {e}"
+                                f"Failed to convert parameter to Decimal at row {row_index}, "
+                                f"column {i} (value type: {type(val).__name__})"
                             ) from e
             processed_parameters.append(processed_row)
 
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -2703,14 +2703,15 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         # Get encoding settings
         encoding_settings = self._get_encoding_settings()
 
-        # Add debug logging
+        # Debug logging: emit batch metadata only. Never log parameter values or
+        # row representations here -- rows may contain PII (SSNs, emails,
+        # balances) that would leak into log files and APM/log shippers even
+        # though this is a DEBUG-level statement. Metadata (batch size, column
+        # count) is sufficient for diagnostics without exposing user data.
         logger.debug(
-            "Executing batch query with %d parameter sets:\n%s",
+            "Executing batch query with %d parameter sets (%d columns per row)",
             len(seq_of_parameters),
-            "\n".join(
-                f"  {i+1}: {tuple(p) if isinstance(p, (list, tuple)) else p}"
-                for i, p in enumerate(seq_of_parameters[:5])
-            ),  # Limit to first 5 rows for large batches
+            len(parameters_type),
         )
 
         ret = ddbc_bindings.SQLExecuteMany(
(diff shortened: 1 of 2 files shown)


mssql_python/cursor.py
gargsaumya: This block also still logs the first 5 full parameter rows ([seq_of_parameters[:5]])- the same PII this PR redacts from the exception. Redact or gate this too

bewithgaurav: +1

mssql_python/cursor.py
```diff
@@ -2669,12 +2669,32 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
...
-                            processed_row[i] = format(decimal.Decimal(str(val)), "f")
-                        except Exception as e:  # pylint: disable=broad-exception-caught
-                            raise ValueError(
-                                f"Failed to convert parameter at row {row}, column {i} to Decimal: {e}"
-                            ) from e
+                            processed_row[i] = format(decimal.Decimal(val_text), "f")
+                        except decimal.DecimalException as e:
+                            raise ValueError(err_msg) from e
+                        except Exception:  # pylint: disable=broad-exception-caught
+                            raise ValueError(err_msg) from None
```
bewithgaurav, line 2743: test suggestion: this is an imp branch, and adding tests would be highly beneficial here
the coverage bot flags 2696-2697 as the only miss on the diff

asking since it is reachable cheaply and deterministically. 

`format(Decimal("1e999999999999999999"), "f")` raises `MemoryError`, not a `DecimalException`, in 0.1ms. the fixed-point expansion length is rejected up front so nothing is allocated.

suggesting adding the below - will take the diff from 81% to 100%):

```python
def test_setinputsizes_sql_decimal_non_decimal_error_no_leak(db_connection):
    """A non-DecimalException during conversion must not chain a cause (GH-503)."""

    huge_exponent = "1e999999999999999999"

    cursor.execute("DROP TABLE IF EXISTS #test_sis_dec_nondec")
    try:
        cursor.execute("CREATE TABLE #test_sis_dec_nondec (Price DECIMAL(18,2))")

        cursor.setinputsizes([(mssql_python.SQL_DECIMAL, 18, 2)])

        with pytest.raises(ValueError) as exc_info:
            cursor.executemany(
                "INSERT INTO #test_sis_dec_nondec (Price) VALUES (?)",
                [(huge_exponent,)],
            )

        message = str(exc_info.value)
        assert "Failed to convert parameter" in message
        assert "row 0" in message
        assert "column 0" in message
        assert huge_exponent not in message
        assert exc_info.value.__cause__ is None
        formatted = "".join(
            traceback.format_exception(
                type(exc_info.value), exc_info.value, exc_info.value.__traceback__
            )
        )
        assert huge_exponent not in formatted
    finally:
        cursor.execute("DROP TABLE IF EXISTS #test_sis_dec_nondec")
```

bewithgaurav (changes requested):
no blockers as such, added a test suggestion - requesting changes

mssql_python/cursor.py
sumitmsft: Added in 04505548 as `test_setinputsizes_sql_decimal_non_decimal_exception_no_leak`, using your `1e999999999999999999` trigger (verified it raises MemoryError, not a DecimalException, in ~0.002s). The test asserts a metadata-only ValueError (`row 0`, `column 0`), `__cause__ is None`, and that the input appears in neither the message nor the fully formatted traceback. I kept a slightly different name and also assert the metadata fields, but it covers exactly the 2696-2697 branch the coverage bot flagged. Thanks for the deterministic repro.

mssql_python/cursor.py
sumitmsft: Good catch, fixed in 04505548. The batch debug log no longer emits any parameter values or row representations. It now logs metadata only: `"Executing batch query with %d parameter sets (%d columns per row)"` (batch count + column count). Added `test_executemany_debug_log_no_parameter_values`, which enables DEBUG capture, runs a batch of sensitive-looking values, and asserts they (and their row reprs) are absent from the logs while the batch metadata line is present as a positive control that capture is actually working.

bewithgaurav (approved):

gargsaumya (approved):

PR #724, opened 2026-08-19
CHORE: Flaky Setup - Gate benchmark and baseline steps on job success

bewithgaurav:

> ADO Work Item: Fixed [AB#47384](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47384)

### Summary

the benchmark and baseline steps keep running after SQL Server setup has already failed, so a dead job still burns 20 minutes benchmarking against a database that does not exist.

build 167929, `macOS x86_64 SQL2025`:

    Build + start SQL Server (Colima boot & SQL setup...)   failed               15.6 min
    Install mssql_py_core from NuGet wheels                 skipped
    Uninstall system ODBC Driver before running tests       skipped
    Run pytest with coverage                                skipped
    Download and restore AdventureWorks2022 database        succeeded             0.4 min
    Download macOS baseline from main                       succeeded             0.0 min
    Prepare macOS baseline file                             succeeded             0.0 min
    Run performance benchmarks on macOS SQL2025             succeededWithIssues  20.2 min

pytest was correctly skipped because it relies on the default step condition. the four that ran override it:

    condition: or(eq(variables['sqlVersion'], 'SQL2022'), eq(variables['sqlVersion'], 'SQL2025'))

that only asks which SQL version the leg is. writing an explicit condition replaces the implicit `succeeded()` rather than adding to it, so nothing upstream is checked. `Publish macOS benchmark baseline` already had this right with `and(succeeded(), ...)`.

each condition now starts with `succeeded()`. the windows leg has the same four steps with the same pattern, and its SQL setup steps carry no `continueOnError`, so a failure there propagates the same way. eight sites in total across `pytestonwindows` and `PytestOnMacOS`.

the benchmark step also carries `timeoutInMinutes: 20` and `continueOnError: true`, so when it times out the job reports `succeededWithIssues` instead of failing. the macOS SQL2022 leg in that same build finished `succeededWithIssues` for exactly that reason. this does not change that behaviour, it just stops the step running when there is nothing to benchmark against.

related to [AB#47311](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47311). that one raised the macOS job timeout to 90 minutes, this one stops a failed setup from eating 20 minutes of the budget.


jahnvi480 (dismissed):

eng/pipelines/pr-validation-pipeline.yml
gargsaumya: This step has the same missing-succeeded() pattern the PR fixes for Windows/macOS. 

Consider condition: and(succeeded(), eq(variables['distroName'], 'Ubuntu'), eq(variables['useAzureSQL'], 'false'))

gargsaumya: Same as the Linux DB-restore step above: gate on succeeded() so the benchmark doesn't run after an upstream container/SQL failure.

bewithgaurav: done, thanks for suggesting

bewithgaurav: resolved

gargsaumya (approved):

jahnvi480 (approved):

Issue #725, opened 2026-08-20
connect does not honor timeout parameter


AlexandreArpin:
### Describe the bug

The docstring on `db_connection.connect` says: `timeout (int): The timeout for the connection attempt, in seconds.`. However, it gets mixed up down the line with the query timeout.

```
sqlstate = 'HYT00'
ddbc_error = '[Microsoft][ODBC Driver 18 for SQL Server]Query timeout expired'

    def raise_exception(sqlstate: str, ddbc_error: str) -> None:
        """
        Raise a custom exception based on the given SQLSTATE code.
        This function raises a custom exception based on the provided SQLSTATE code.
        If the code is not found in the mapping, a generic DatabaseError is raised.
    
        Args:
            sqlstate (str): The SQLSTATE code to map to a custom exception.
            ddbc_error (str): The DDBC error message.
    
        Raises:
            DatabaseError: If the SQLSTATE code is not found in the mapping.
        """
        exception_class = sqlstate_to_exception(sqlstate, ddbc_error)
        if exception_class:
            logger.error(f"Raising exception: {exception_class}")
>           raise exception_class
E           mssql_python.exceptions.OperationalError: Driver Error: Timeout expired; DDBC Error: [Microsoft]Query timeout expired
```

### To reproduce
I used the following test case to reproduce the issue

```python
async def test_connection_timeout_repurposes_to_query_timeout(connection_string):
    conn = mssql_python.connect(connection_string, timeout=3)
    cursor = conn.cursor()
    cursor.execute("WAITFOR DELAY '00:00:10'; SELECT 1")  # times out after ~3s

    val = cursor.fetchval()
    assert 1 == val
```

### Expected behavior
connect(timeout=3) should make the connection attempt time out after 3s.

### Further technical details
Python version: 3.13
SQL Server version: Azure SQL Server
Operating system: Windows 11

**Additional context**
Discovered by Claude as I was migrating from pyodbc/aiodbc to mssql-python, I did A/B tests and can confirm that I believe it's an "unintended" bug in the library, hopefully I did by due diligence correctly!

Some more context from Claude

```
In connection.py, timeout is documented twice as a login timeout — "Login timeout in seconds. 0 means no timeout." (line 295) and "The timeout for the connection attempt, in seconds" in connect(). Its complete set of uses:

631:  self._timeout = timeout          # stored
928:  return self._timeout             # property documented as "the current query timeout"
950:  self._timeout = value            # setter: "Set the query timeout for all operations"
[...]

PR #728, opened 2026-08-25
FIX: connect(timeout=) sets login timeout, not query timeout (GH-725)

jahnvi480:

> [AB#47540](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47540)

> GitHub Issue: #725 

### Summary

`connect(timeout=N)` is documented (and matches pyodbc) as the login /
connection-attempt timeout, but it was stored into the query-timeout field and
applied to every cursor as `SQL_ATTR_QUERY_TIMEOUT`. The connect path never set
`SQL_ATTR_LOGIN_TIMEOUT`, so the parameter silently became a per-statement query
timeout and `Connection.timeout` returned it instead of bounding the login.

### Fix

- Route the constructor `timeout` to `SQL_ATTR_LOGIN_TIMEOUT` via `attrs_before`
  (`setdefault`, so an explicit `attrs_before[SQL_ATTR_LOGIN_TIMEOUT]` still wins).
- `Connection.timeout` (the per-statement query timeout) is now a separate field
  defaulting to `0`; the property setter behavior is unchanged.
- Add a shared `_validate_timeout` helper used by both the constructor and the
  setter so the two entry points can never drift apart again; `bool` is rejected.

### Behavior

- The supported query-timeout API (`conn.timeout = N`) is unchanged.
- `connect(timeout=N)` now bounds the connection attempt (login timeout), as documented.
- Invalid input (negative, non-int, `bool`) fails fast with `ValueError` / `TypeError`.

### Validation

- `black --check` clean; full `tests/test_003_connection.py` = 167 passed, 2 skipped.
- Live repros: unreachable host now fails in ~2s (was ~15s); a 10s query is no longer aborted.


gargsaumya (dismissed):

tests/test_003_connection.py
```diff
@@ -2570,23 +2570,117 @@ def test_timeout_setter(db_connection):
...
+        with pytest.raises(Exception):
+            connect(unresponsive, timeout=3).close()
+        elapsed = time.time() - start
+        # Lower bound: the TCP connect succeeds instantly, so any elapsed >= 2s
+        # proves the driver actually waited in the login handshake and the login
+        # timeout (not an immediate failure) ended it.
+        assert elapsed >= 2.0, f"Connect failed in {elapsed:.1f}s; login timeout not exercised"
+        # Upper bound: well under the ~15s driver default, proving the 3s login
+        # timeout was actually applied.
+        assert elapsed < 12, f"Login timeout not honored; connect took {elapsed:.1f}s"
```
subrata-ms: **Medium: timeout regression test can false-pass**
[test_003_connection.py:2655-2665] accepts any [Exception] occurring between 2 and 12 seconds. It should assert [OperationalError] with a timeout diagnostic and use [time.monotonic()]. The runtime produces the expected [OperationalError] after 3.018s, so the test can be stricter.

tests/test_003_connection.py
```diff
@@ -2570,23 +2570,117 @@ def test_timeout_setter(db_connection):
...
-        # Execute a quick query to ensure the timeout doesn't interfere
         cursor.execute("SELECT 1")
         result = cursor.fetchone()
-        assert result[0] == 1, "Query execution should succeed with timeout set"
+        assert result[0] == 1, "Query execution should succeed with login timeout set"
+    finally:
+        conn.close()
+
+
+def test_constructor_timeout_does_not_become_query_timeout(conn_str):
```
subrata-ms, line 2598: **Low: redundant integration coverage adds ~5 seconds per CI leg**
[test_003_connection.py:2598] and [test_003_connection.py:4949] both verify that constructor timeout does not abort a five-second query. Consolidating them would reduce matrix runtime without losing coverage.

subrata-ms:
**What Is Good**

- Fixes Issue #725 at the root: [connect(timeout=N)] now configures [SQL_ATTR_LOGIN_TIMEOUT] before SQLDriverConnect.

- Cleanly separates login timeout from [Connection.timeout], which remains the query timeout.

- Works correctly with pooling, pool misses, and deferred Entra token connections.

- Validates invalid timeout values early and documents the behavioral change clearly.

- Adds broad regression coverage.

- Native build, formatting, dependency tests, and the full CI matrix pass.

**What Needs Fixing**

1. Tighten the login-timeout test at [test_003_connection.py:2655-2665]:

- Expect [OperationalError], not any [Exception].
- Assert the error contains “timeout”.
- Use [time.monotonic()] for elapsed-time measurement.

2. Remove duplicate five-second query coverage between [test_003_connection.py:2597] and [test_003_connection.py:4947] to reduce CI runtime.

3. Consider moving the loopback timeout test into a no-database test module. The auto-use fixture at [test_003_connection.py:49] currently requires a live SQL Server even though this test itself is designed to be hermetic.


Overall, the implementation is correct and has no production-code blockers; the remaining improvements concern test precision, isolation, and runtime.

tests/test_003_connection.py
jahnvi480: Fixed this

tests/test_003_connection.py
jahnvi480: Fixed this

gargsaumya (approved):

subrata-ms (approved):

PR #397, opened 2026-01-07
FEAT: [POC] BCP implementation in mssql-python driver -

subrata-ms:
### Work Item / Issue Reference  

> AB#<WORK_ITEM_ID>

> GitHub Issue: #<ISSUE_NUMBER>

-------------------------------------------------------------------
### Summary   
  

This pull request introduces Rust-based Python bindings for SQL Server bulk copy operations, providing an alternative to the existing C++ bindings. It adds a new `mssql_rust_bindings` module using PyO3, integrates a `BulkCopyWrapper` for efficient data transfer, and includes comprehensive documentation and build scripts for both Linux and Windows. The main demo script is updated to showcase usage of the new Rust bindings for bulk copy.

**Key Components:**

**User/Application:** Calls the cursor's bulk_copy method 
**Cursor (cursor.py):**
       - Parses connection string
       - Creates BulkCopyWrapper with parameters
       - Manages the bulk copy lifecycle
**BulkCopyWrapper (Rust/PyO3):**
       - Creates and manages mssql_core_tds connection internally
       - Checks if bulk_copy is implemented
       - Delegates to mssql_core_tds
**DdbcConnection (mssql_core_tds):**
       - Python bindings to TDS protocol
       - Handles actual database communication
       - SQL Server: Processes the bulk insert
**Sequence Diagram:**

image

**Key changes:**

### 1. Rust Bindings for Python (PyO3 Integration)
- Added new Rust-based Python extension `mssql_rust_bindings` using PyO3, featuring a `RustConnection` class, utility functions, and a `BulkCopyWrapper` for bulk copy operations. Includes module initialization and versioning. [[1]](diffhunk://#diff-8c971d17d1fc064bb1931fd9c38d5e01d0acbdfd8a1d5a4637dcadddd5bd6eb4R1-R101) [[2]](diffhunk://#diff-f89f56fcf904e81a451ff713405bfe31c50f1a230cc4832ae2d6d193f522b789R1-R49)
- Created `Cargo.toml` for Rust package configuration, specifying dependencies, build profiles, and metadata.

### 2. Bulk Copy Functionality
- Implemented `BulkCopyWrapper` in Rust, exposing a Python class that wraps the `bulk_copy` method from the underlying TDS connection, enabling high-performance bulk data transfer.
- Updated `main.py` to demonstrate connecting to SQL Server, creating a temp table, generating test data, and performing a bulk copy using the new Rust bindings. Includes verification and cleanup steps.

### 3. Documentation and Build Scripts
[...]

Issue #394, opened 2026-01-05
[FEATURE] Ability to check if a connection has been closed


martinstuder:
### Is your feature request related to a problem? Please describe.
Currently, checking whether a connection has been closed requires a workaround such as issuing a `getinfo(...)` request and intercepting an `InterfaceError` if the connection has been closed.

### Describe the solution you'd like
Ideally, it would be possible to directly check if a connection has been closed, e.g. through a property `closed`.

### Describe alternatives you've considered
Alternatives considered:
* issuing a `getinfo(...)` request and intercepting `InterfaceError`
* attempting to create a cursor and intercepting `InterfaceError`

PR #730, opened 2026-08-26
CHORE: Add Selectable Native ODBC Providers

gargsaumya:
Linked work item: [AB#47445](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47445)

### Summary
This pull request introduces opt-in/opt-out support for selecting which native ODBC provider is loaded by `mssql-python`, allowing users to choose between the default Microsoft ODBC Driver 18 (`msodbcsql18`) and the Rust-based driver (`mssql-odbc`), if the latter is installed. The selection can be made via a module property or environment variable before the first connection, with diagnostics and warnings for precedence and immutability. The implementation includes a new provider manager, updates to documentation, and improvements to driver loading and shutdown safety.

**Native ODBC provider selection and diagnostics:**

* Added support for selecting the native ODBC provider via the new `mssql_python.native_provider` module property or the `MSSQL_PYTHON_NATIVE_PROVIDER` environment variable (the env var takes precedence). The provider selection is resolved and frozen at the first connection, with warnings for conflicting or late assignments. The default remains `"msodbcsql18"`, but opt-in to `"mssql-odbc"` is supported if the `mssql-python-rs` package is installed. [[1]](diffhunk://#diff-d95f3a67986de29f30453416b1b4c34e6a43207e9a33e2b1b80ef0c378b0a538R604-R618) [[2]](diffhunk://#diff-b073381cfee6e39fdb7cd5eee52930c6dd1eafa4135f394b634791d9694b76bbR1-R250) [[3]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR46-R57) [[4]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L21-R26) [[5]](diffhunk://#diff-8b251a7d56f9bb22b686d2b101b0420b1d6cd6934b29edcd408c1f579f7d9e84R33)

* Introduced `get_native_provider_info()` for diagnostics, reporting the selected provider, source, package, version, driver path, and whether the selection is frozen. [[1]](diffhunk://#diff-d95f3a67986de29f30453416b1b4c34e6a43207e9a33e2b1b80ef0c378b0a538R79-R92) [[2]](diffhunk://#diff-8b251a7d56f9bb22b686d2b101b0420b1d6cd6934b29edcd408c1f579f7d9e84R47)

**Driver loading and connection logic:**

* Updated connection logic so the ODBC provider is resolved and frozen only after all Python-side validation succeeds and just before the native driver loads, preventing premature provider locking on failed connection attempts. [[1]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R31) [[2]](diffhunk://#diff-29bb94de45aae51c23a6426d40133c28e4161e68769e08d046059c7186264e90R734-R743)

* Ensured that enabling connection pooling does not prematurely resolve or freeze the ODBC provider, so explicit pooling configuration does not lock in the default provider before selection.

**Documentation updates:**

* Updated `README.md` and `CHANGELOG.md` to document the new provider selection mechanism, its usage, precedence rules, and diagnostic interface. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L21-R26) [[2]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR46-R57)

**Type hints and interface improvements:**

* Updated type hints in `mssql_python.pyi` to reflect the new `native_provider` property and `get_native_provider_info()` function. [[1]](diffhunk://#diff-8b251a7d56f9bb22b686d2b101b0420b1d6cd6934b29edcd408c1f579f7d9e84R33) [[2]](diffhunk://#diff-8b251a7d56f9bb22b686d2b101b0420b1d6cd6934b29edcd408c1f579f7d9e84R47)
* Minor import and typing improvements in `mssql_python/__init__.py`.

**Native extension (C++ backend) safety:**

* Improved Python shutdown/finalization detection in the C++ extension to use thread-safe, GIL-free APIs (`Py_IsFinalizing` or `_Py_IsFinalizing`), preventing crashes during interpreter shutdown from foreign threads. [[1]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R17) [[2]](diffhunk://#diff-dde2297345718ec449a14e7dff91b7bb2342b008ecc071f562233646d71144a1R953-R980)


Vahid-b (commented):
## Summary

Adds a process-wide, resolve-once selector between `msodbcsql18` and the Rust `mssql-odbc` driver, with `msodbcsql18` as the Phase 1 default and the Rust path failing closed until its wheel ships. The design is sound — Python as the sole resolver, the native side purely a receiver, freeze at first connect, no public API change. One finding I'd treat as blocking; the rest are suggestions and nits.

I verified the load ordering the whole design rests on: `loadDriver()` is lazy behind `std::call_once`, and the only entry into it from a fresh process is the `Connection` constructor at `pybind/connection/connection.cpp:25`. `connection.py:375-376` pushes the provider well before `ddbc_bindings.Connection(...)` at line 741. That part is correct.

All of `copilot-pull-request-reviewer`'s findings and the DevSkim alert look addressed in `e9372ba7` / `094dc2f3` / `f854f8dc`; I am not re-filing any of them. The one that is only half-closed is the `Optional[str]` stub, noted inline.

## Blocking

**The pooling hook (`pooling.py:66-71`) — its stated reason is false, and its only real effect is an unwanted early freeze.** Detail inline. `enable_pooling` does not load the native driver, so the hook protects nothing, but it does make `mssql_python.pooling()` freeze the provider and become able to raise `ImportError`.

## Suggestions

Six inline: the `source` field that is always `None` before the freeze, `effective()` raising on a bad env var, the native side's silent coercion plus unguarded `set_odbc_provider`, the `mssql-auth.dll` justification, the Linux path layout versus the libc split, and the `Optional[str]` stub.

One that has no line to sit on: **no test covers the pooling freeze path.** That is where the blocking finding lives, and a test asserting `PoolingManager.enable()` does *not* freeze the selection would have caught it. Worth adding alongside the fix.

## Nits

Three inline: the mid-file `#include`, the `recwarn` assertion, and the resolve-before-validate ordering in `Connection.__init__`. Plus one that spans the change rather than a line: `NormalizeProviderId` (`ddbc_bindings.cpp:993-1003`) strips *interior* whitespace while Python's `_normalize` (`odbc_provider.py:44-54`) only does `.strip()`. Harmless while Python is the sole resolver, but the two should agree if the native one ever becomes authoritative.

## What I ran

`pytest` isn't installed here and the package needs the compiled `ddbc_bindings` extension, so I could not run the suite — CI is the gate for that. The two runtime observations below came from loading `odbc_provider.py` standalone against a stubbed `mssql_python.logging`:

```
A. get_info BEFORE resolve, env=mssql-odbc: {'id': 'mssql-odbc', 'package': 'mssql_python_rust_odbc', 'source': None, 'frozen': False}
C. get_info AFTER  resolve:                 {'id': 'mssql-odbc', 'package': 'mssql_python_rust_odbc', 'source': 'environment', 'frozen': True}
D. get_info    with bad env RAISED: ValueError Unknown ODBC provider 'bogus-value'. ...
E. effective() with bad env RAISED: ValueError Unknown ODBC provider 'bogus-value'. ...
```

The cross-repo claims (`mssql-auth.dll`, artifact filenames, the Linux libc split) were checked against `microsoft/mssql-rs` source rather than from memory; file and line are cited in each comment.

---

*Reviewed with GitHub Copilot on behalf of @Vahid-b, then checked by hand. Not an approval — push back on anything that looks wrong.*

mssql_python/pooling.py
```diff
@@ -62,6 +63,11 @@ def enable(cls, max_size: int = 100, idle_timeout: int = 600) -> None:
                 max_size,
                 idle_timeout,
             )
+            # Enabling pooling loads the native driver; resolve and push the
+            # ODBC provider first so an explicit pooling() before any connect
+            # still honors the selection (mirrors Connection.__init__).
+            _provider = ProviderManager.ensure_available()
+            ddbc_bindings.set_odbc_provider(_provider)
```
Vahid-b: **Blocking:** the comment above is factually wrong, and the code has a side effect that works against the feature this PR adds.

`enable_pooling` does not load the native driver. It is `ddbc_bindings.cpp:6068-6078` → `ConnectionPoolManager::configure()` (`pybind/connection/connection_pool.cpp:486-493`, pure state assignment) plus `setAccepting(true)` (`:525-528`). There is no `loadDriver()` call anywhere in `pybind/connection/*.cpp` except the `Connection` constructor at `connection.cpp:25`.

So the hook protects nothing, but it does two user-visible things:

- **`mssql_python.pooling()` now freezes the provider.** `ms.pooling(max_size=50)` followed by `ms.odbc_provider = "mssql-odbc"` silently becomes a no-op plus a `RuntimeWarning`. That is exactly the opt-in route this PR exists to add, and `pooling()` is a natural first line in a startup block — so the ordering trap is easy to hit and gives no error, just a warning.
- **`pooling()` can now raise `ImportError`.** A configuration call that previously could only raise `ValueError` on bad parameters now fails when the selected provider's wheel is absent.

Deleting lines 69-70 loses nothing. On the auto-enable path, `Connection.__init__` already pushed at `connection.py:376` before reaching `PoolingManager.enable()` at line 737; on the explicit path, the next `connect()` pushes before the driver loads either way.

If you'd rather keep it as defence-in-depth against a future `enable_pooling` that *does* load the driver, use the non-freezing accessor:

```python
ddbc_bindings.set_odbc_provider(ProviderManager.effective())
```

Either way the comment needs correcting — as written it will stop the next reader from removing this.

mssql_python/odbc_provider.py
```diff
@@ -0,0 +1,177 @@
...
+        return cls._resolved is not None
+
+    @classmethod
+    def get_info(cls) -> Dict[str, object]:
+        """Report the selected provider for diagnostics."""
+        provider = cls._resolved if cls._resolved is not None else cls.effective()
+        return {
+            "id": provider,
+            "package": _PACKAGE_BY_PROVIDER[provider],
+            "source": cls._source,
```
Vahid-b: **Suggestion:** `source` is always `None` until the provider freezes, so the diagnostic is blank in exactly the window you'd call it.

`_compute()` returns the source, but `get_info` reads `cls._source`, which only `resolve()` ever assigns. Confirmed by running the module standalone with `MSSQL_PYTHON_ODBC_PROVIDER=mssql-odbc` set:

```
BEFORE resolve: {'id': 'mssql-odbc', 'package': 'mssql_python_rust_odbc', 'source': None,          'frozen': False}
AFTER  resolve: {'id': 'mssql-odbc', 'package': 'mssql_python_rust_odbc', 'source': 'environment', 'frozen': True}
```

Reporting the id but not where it came from is the opposite of useful when someone is trying to work out *why* they're getting a provider they didn't expect — which is the pre-connect case.

```python
if cls._resolved is not None:
    provider, source = cls._resolved, cls._source
else:
    provider, source = cls._compute()
```

`test_get_info_before_and_after_resolve` (`tests/test_026_odbc_provider.py:97-109`) asserts `id`, `package` and `frozen` before resolve but skips `source`, which is why this passes today.

mssql_python/odbc_provider.py
```diff
@@ -0,0 +1,177 @@
...
+                )
+            return cls._resolved
+
+    @classmethod
+    def effective(cls) -> str:
+        """Return the provider that would be used, without freezing it."""
+        with cls._lock:
+            if cls._resolved is not None:
+                return cls._resolved
+            provider, _ = cls._compute()
```
Vahid-b: **Suggestion:** `_compute()` raises `ValueError` on an unrecognized env var, and that propagates out of `effective()` — so the diagnostics can't diagnose the one misconfiguration they exist for. Verified by running the module standalone with `MSSQL_PYTHON_ODBC_PROVIDER=bogus-value`:

```
get_info()    RAISED: ValueError Unknown ODBC provider 'bogus-value'. Valid providers are: msodbcsql18, mssql-odbc.
effective()   RAISED: ValueError Unknown ODBC provider 'bogus-value'. Valid providers are: msodbcsql18, mssql-odbc.
```

Failing closed at `resolve()` / `ensure_available()` is right and I'm not suggesting changing that. The problem is the read-only paths that share `_compute()`:

- `get_odbc_provider_info()` raises instead of reporting the bad value.
- `effective()` also backs the **getter** for `mssql_python.odbc_provider` (`__init__.py:605-613`), so plain attribute access raises. And since `"odbc_provider"` is in `__all__` (`__init__.py:528`), `from mssql_python import *` would raise `ValueError` at import — before any connection is attempted, from a line that has nothing to do with providers.

Consider having the non-freezing paths report the raw invalid value (e.g. an `error` key in `get_info`, and the default from the getter) and keep the hard failure at resolve time, where it is actionable.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -971,6 +971,63 @@ std::string GetLastErrorMessage();
...
+
+void SetSelectedProvider(const std::string& id) {
+    std::lock_guard<std::mutex> lock(g_providerMutex);
+    g_selectedProvider = NormalizeProviderId(id);
+}
+
+// Effective provider id: the value pushed from Python, else the classic default.
+// Python is the authoritative resolver (env var -> module property -> default)
+// and pushes the result via set_odbc_provider() before the driver loads.
+std::string GetSelectedProviderId() {
```
Vahid-b, line 1079: **Suggestion:** the native side silently coerces anything unrecognized to classic, which contradicts the Python layer's fail-closed contract.

`GetSelectedProviderId()` returns `kProviderMsodbcsql18` for any value that isn't exactly `mssql-odbc`, and `SetSelectedProvider` above just overwrites the global with no validation and no check that the driver has already loaded. `ProviderManager` can't produce a bad value today, so this is defence-in-depth rather than a live bug — but `m.def("set_odbc_provider", ...)` at line 6108 is an unprefixed, public-looking native entry point, and anything that calls it directly bypasses both the validation and the freeze invariant the Python layer maintains. Silently loading a *different* driver than the caller asked for is the failure mode the Python side goes out of its way to avoid.

Two cheap options, not mutually exclusive: reject an unknown id in `SetSelectedProvider` (throw, mirroring `_normalize`), and rename the binding to `_set_odbc_provider` so it reads as internal plumbing. Making it a no-op or throw once `DriverLoader` has run would close the ordering hole too.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1215,8 +1294,12 @@ DriverHandle LoadDriverOrThrowException() {
         LOG("LoadDriverOrThrowException: mssql-auth.dll not found at '%s' - "
             "Entra ID authentication will not be available",
             authDllPath.string().c_str());
-        ThrowStdException("mssql-auth.dll not found. If you are using Entra "
-                          "ID, please ensure it is present.");
+        // mssql-auth.dll ships with the classic driver; the Rust provider does
```
Vahid-b: **Suggestion:** the behavior is right but the stated reason isn't — the Rust driver does use `mssql-auth.dll`.

From `microsoft/mssql-rs`: `mssql-odbc/src/auth/msqa.rs:63` is `const MSQA_LIBRARY: &str = "mssql-auth.dll";`, `auth/interactive.rs:10` describes it as "a Windows-only library it loads with", and `msqa.rs:212` documents "Loads `mssql-auth.dll` and resolves the entry points, once per process". It is required for `ActiveDirectoryInteractive`.

What actually differs is *when* and *from where*. `msqa.rs:62` notes it is "The library msodbcsql loads from **System32**" — the Rust driver resolves it lazily through the normal DLL search path, whereas the classic driver needs it co-located at load time. So not treating its absence as fatal for the Rust provider is correct.

Two things worth changing: restate the comment as "loaded lazily from the system search path, so it is not a load-time dependency for this provider", and note the consequence — a user on `mssql-odbc` doing interactive Entra auth will now get a failure from inside the Rust driver rather than this clear load-time message.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1122,6 +1182,25 @@ std::string GetDriverPathCpp(const std::string& moduleDir) {
     throw std::runtime_error("Unsupported architecture");
 #endif
 
+    // Rust provider (mssql-odbc): ships as `mssql-odbc.{so,dylib,dll}` (no `lib`
+    // prefix on Linux/macOS) under an mssql-python-defined libs/ layout.
+    // mssql-python owns the provider wheel, so this layout is authoritative and
+    // finalized alongside that wheel build.
+    if (GetSelectedProviderId() == kProviderMssqlOdbc) {
+#ifdef __linux__
+        return (basePath / "libs" / "linux" / arch / "lib" / "mssql-odbc.so").string();
```
Vahid-b: **Suggestion:** the Rust Linux path drops the distro segment the classic path has, and there's a known reason it will be needed.

Classic Linux resolves to `libs/linux/<platform>/<arch>/lib/...` with `platform` one of `alpine` / `rhel` / `suse` / `debian_ubuntu` (lines 1205-1230). macOS and Windows in this new block correctly mirror their classic shapes; only Linux is flattened to `libs/linux/<arch>/lib/`.

That matters because `microsoft/mssql-rs` is building **two libc flavours per architecture** — see mssql-rs #394, "Build both Linux libc flavours per architecture on one agent". A flat `libs/linux/<arch>/` has nowhere to put both, so the musl and glibc artifacts would collide. Since the completeness check in `GetOdbcLibsBaseDir` does `fs::exists()` on exactly this path, whatever the wheel ships has to match it.

I realise the description calls this layout "finalized alongside that wheel build", so this is a deliberate open question rather than an oversight — but the libc split is a requirement today, not a future one, and it is cheaper to settle before the wheel ships than after.

Separately, the filenames here match mssql-rs #397 ("Ship mssql-odbc driver artifacts with product filenames"), which is still open. Worth naming that dependency in the Deferred list alongside the others.

mssql_python/mssql_python.pyi
```diff
@@ -30,6 +30,7 @@ threadsafety: int  # 1
 # Module Settings - Properties that can be get/set at module level
 lowercase: bool  # Controls column name case behavior
 native_uuid: bool  # Controls UUID type handling
+odbc_provider: str  # Selects the ODBC provider ('msodbcsql18' or 'mssql-odbc')
```
Vahid-b: **Suggestion:** this only half-closes the earlier review point about `Optional`.

`ProviderManager.set_property()` and the module setter at `__init__.py:614-617` both accept `Optional[str]` (documented as "or None to clear"), so `mssql_python.odbc_provider = None` works at runtime but is a type error for anyone consuming the stubs. A module-level `.pyi` variable can't express asymmetric get/set types, so it has to be one or the other: either annotate `Optional[str]` here, or drop `None` support from the public setter.

I'd lean toward dropping it — clearing the selection has no test, and given the freeze semantics there's a narrow window in which it does anything at all.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -971,6 +971,63 @@ std::string GetLastErrorMessage();
...
+//
+// Two providers are supported: the classic Microsoft ODBC Driver 18
+// ("msodbcsql18", shipped by mssql_python_odbc) and the Rust driver
+// ("mssql-odbc", shipped by mssql_python_rust_odbc). Python is the sole
+// resolver (env var -> module property -> default) and pushes the chosen id
+// here via set_odbc_provider() before the driver loads. The native side does
+// not read the environment itself; if the push has not happened yet, it falls
+// back to the hardcoded classic default.
+// -----------------------------------------------------------------------------
+#include <cctype>
```
Vahid-b: **Nit:** `#include <cctype>` sits ~985 lines into the file rather than with the other includes at the top. It works, but it's the only one down here.

tests/test_026_odbc_provider.py
```diff
@@ -0,0 +1,152 @@
...
+    ProviderManager.resolve()  # freezes as default msodbcsql18
+    with pytest.warns(RuntimeWarning):
+        ProviderManager.set_property(PROVIDER_MSSQL_ODBC)
+    assert ProviderManager.effective() == PROVIDER_MSODBCSQL18
+
+
+def test_same_value_after_freeze_does_not_warn(recwarn):
+    ProviderManager.resolve()
+    ProviderManager.set_property(PROVIDER_MSODBCSQL18)
+    assert len(recwarn) == 0
```
Vahid-b: **Nit:** this asserts *no warnings at all*, not "no `RuntimeWarning`". Any unrelated `DeprecationWarning` raised during the call would fail it. `pytest.warns(None)` is deprecated, so the targeted form is:

```python
assert not [w for w in recwarn if issubclass(w.category, RuntimeWarning)]
```

mssql_python/connection.py
```diff
@@ -368,6 +369,12 @@ def __init__(
             >>> # Return native uuid.UUID objects instead of strings
             >>> conn = ms.connect("Server=myserver;Database=mydb", native_uuid=True)
         """
+        # Resolve and freeze the ODBC provider before the native driver loads,
+        # then hand the selection to the native loader so it imports the matching
+        # provider package.
+        _provider = ProviderManager.ensure_available()
```
Vahid-b: **Nit:** this resolves and freezes the provider before the `native_uuid` type validation just below, so a call that is about to raise `TypeError` still freezes the selection as a side effect. Moving the two lines after the argument validation is free and keeps the freeze tied to a connection attempt that actually proceeds.

saurabh500: **Blocking:** this provider resolution happens after the native extension has already attempted to load the classic driver. Importing `mssql_python` loads the extension, whose `PYBIND11_MODULE` calls `DriverLoader::loadDriver()` at `ddbc_bindings.cpp:6302`. That call is protected by `std::call_once`, so it permanently records either the classic driver handle or the classic load error before this line can push `mssql-odbc`.

I reproduced this by selecting `mssql-odbc` in a fresh process with stand-in provider packages; `connect()` still surfaced the `mssql-python-odbc` completeness error. Please remove the import-time `loadDriver()` block and rely on the existing lazy load from the native connection path, so this push really runs first. A subprocess test should select Rust with incomplete stand-in packages and assert that the native error names `mssql-python-rust-odbc`, not the classic distribution.

mssql_python/odbc_provider.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Called before the native driver loads. Fails closed with an actionable
+        error if the selected provider's package is missing, rather than
+        silently loading a different provider.
+        """
+        provider = cls.resolve()
+        package = _PACKAGE_BY_PROVIDER[provider]
+        try:
+            importlib.import_module(package)
+        except ImportError as exc:
```
saurabh500: **Suggestion:** this rewrites every `ImportError` as “the provider package is not installed.” If the package is present but its initialization or a transitive dependency fails, the actionable underlying error is hidden behind an incorrect installation hint.

Catch `ModuleNotFoundError` and translate it only when `exc.name == package`; otherwise re-raise the original exception. The test double should construct `ModuleNotFoundError(..., name=name)` so it matches real import behavior.

```python
except ModuleNotFoundError as exc:
    if exc.name != package:
        raise
    dist = _DIST_BY_PROVIDER[provider]
    ...
```

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1215,8 +1294,12 @@ DriverHandle LoadDriverOrThrowException() {
...
             "Entra ID authentication will not be available",
             authDllPath.string().c_str());
-        ThrowStdException("mssql-auth.dll not found. If you are using Entra "
-                          "ID, please ensure it is present.");
+        // mssql-auth.dll ships with the classic driver; the Rust provider does
+        // not require it, so its absence is only fatal for msodbcsql18.
+        if (GetSelectedProviderId() != kProviderMssqlOdbc) {
+            ThrowStdException("mssql-auth.dll not found. If you are using Entra "
+                              "ID, please ensure it is present.");
+        }
```
saurabh500: **Suggestion:** the Rust provider also requires `mssql-auth.dll` for its Windows interactive-authentication path. This exemption is also unreachable in the missing-file case because `GetOdbcLibsBaseDir()` already rejects every Windows provider whose driver directory lacks the DLL at lines 1064–1068.

If the provider wheel is required to include `mssql-auth.dll`, keep that completeness check, remove this Rust-specific exception, and update the PR description/comment that currently calls the DLL nonfatal.

```suggestion
        ThrowStdException("mssql-auth.dll not found. If you are using Entra "
                          "ID, please ensure it is present.");
```

saurabh500 (commented):
The native load-order issue below blocks the Rust opt-in. I also agree with the existing `PoolingManager.enable()` thread: native `enable_pooling()` only configures pool state, so it should not freeze provider selection when the documented boundary is the first connection. I did not duplicate that thread.

The authentication comment below reflects the clarified requirement that the Rust provider also requires `mssql-auth.dll` for its Windows interactive-authentication path.

mssql_python/connection.py
gargsaumya: Confirmed and fixed. 
Removed the eager load in ddbc_bindings.cpp; the driver now loads lazily at first use, always after the Python push. Added the subprocess regression you suggested. Thanks for the clear repro.

mssql_python/odbc_provider.py
gargsaumya: Fixed. Now catches ModuleNotFoundError and only translates it to the "not installed" hint when exc.name == package; anything else re-raises the original error.

mssql_python/pooling.py
gargsaumya: Correct. Fixed now.

mssql_python/odbc_provider.py
gargsaumya: Fixed.

mssql_python/odbc_provider.py
gargsaumya: Fixed

mssql_python/odbc_provider.py
```diff
@@ -0,0 +1,211 @@
...
+            ) from exc
+        return cls.resolve()
+
+    @classmethod
+    def is_frozen(cls) -> bool:
+        """Whether the provider has been resolved and can no longer change."""
+        return cls._resolved is not None
+
+    @classmethod
+    def get_info(cls) -> Dict[str, object]:
```
bewithgaurav, line 178: the diagnostics are something we'll lean on once there are two providers live

the case I have in mind is a customer encounters a driver-level bug: we ask which provider is live, which version, and which binary actually got loaded, and paste the answer into the issue/ticket

the doc says this call returns `id`, `version`, `abi_profile`, `driver_path`
we return `id`, `package`, `source`, `frozen`

can we add the two cheap ones:

- `version`, from `mssql_python_odbc.__version__` (`18.6.2.1` on my box)
- `driver_path`, from the existing `GetDriverPathCpp` (`.../libs/macos/arm64/lib/libmsodbcsql.18.dylib`)

`driver_path` matters most. `package` is the same string on every machine, so it will not tell the two providers apart once the default flips.

since `abi_profile` needs the rust handshake, we can leave it out of this PR

mssql_python/odbc_provider.py
```diff
@@ -0,0 +1,211 @@
...
+        has been resolved is ignored with a warning; the env var still takes
+        precedence over this value when both are set.
+        """
+        with cls._lock:
+            canonical = _normalize(value) if value is not None else None
+            if cls._resolved is not None:
+                if canonical != cls._resolved:
+                    cls._warn_frozen()
+                return
+            cls._property_value = canonical
```
bewithgaurav, line 96: edge case: you can ask for the rust driver, not get it, and never know.

```
os.environ["MSSQL_PYTHON_NATIVE_PROVIDER"] = "msodbcsql18"   # set by ops or the base image
mssql_python.native_provider = "mssql-odbc"                  # app opts in

mssql_python.native_provider -> msodbcsql18
```

no error, no warning, the app runs on msodbcsql18 the whole time
the only way to catch it is to call `get_native_provider_info()` and read the value back, so someone tries the rust driver, tells us it works fine, and was never on it

we already warn if you set it too late. can we warn here too? 
env var taking a priority is fine I guess

bewithgaurav (commented):
was skimming through, and found a couple of cases worth thinking about - putting as a comment review

gargsaumya:
@copilot resolve the merge conflicts in this pull request

mssql_python/mssql_python.pyi
```diff
@@ -30,6 +30,7 @@ threadsafety: int  # 1
 # Module Settings - Properties that can be get/set at module level
 lowercase: bool  # Controls column name case behavior
 native_uuid: bool  # Controls UUID type handling
+native_provider: Optional[str]  # Selects the native ODBC provider ('msodbcsql18' or 'mssql-odbc')
```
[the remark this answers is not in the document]
gargsaumya: The getter does always return str, but this module property intentionally accepts None in its setter to clear a pending property selection before the provider is resolved. Since a module-level variable annotation cannot express asymmetric getter/setter types, Optional[str] is required to type that supported assignment correctly. Changing it to str would make valid mssql_python.native_provider = None usage fail static checking. We could make it str only by removing None support from the public API, which would be a separate behavioral change.

mssql_python/pybind/ddbc_bindings.h
```diff
@@ -250,13 +251,14 @@ class DriverLoader {
   public:
     static DriverLoader& getInstance();
     void loadDriver();
+    bool isDriverLoaded() const;
```
bewithgaurav: yagni: `isDriverLoaded()` and `m_driverLoaded` are dead now. the race fix in this branch moved the post-load guard to `g_providerSelectionLocked`, so nothing reads them anymore (the getter has no callers, the field is write-only). the `std::atomic` upgrade and `<atomic>` include only exist to back the unused getter, so all three can go

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1023,46 +1118,52 @@ std::string GetOdbcLibsBaseDir() {
...
 #endif
         if (!externalComplete) {
-            LOG("GetOdbcLibsBaseDir: mssql_python_odbc present at '%s' but its ODBC driver "
+            LOG("GetOdbcLibsBaseDir: %s present at '%s' but its ODBC driver "
                 "binaries are missing or incomplete for this platform",
-                parentDir.string().c_str());
+                packageName.c_str(), parentDir.string().c_str());
             ThrowStdException(
-                "The 'mssql-python-odbc' package is installed but its ODBC driver binaries "
+                "The '" + distName + "' package is installed but its ODBC driver binaries "
```
bewithgaurav, line 1146: the error names a package the user may not have installed. `mssql-python` bundles `mssql_py_core` for bulkcopy without the `mssqlodbc.*` binary, so selecting mssql-odbc on a plain `pip install mssql-python` reaches this throw and reports "the 'mssql-python-rs' package is installed but its ODBC driver binaries are missing" even when mssql-python-rs was never installed. 

it fails closed, so no wrong-driver surprise, just a confusing hint. worth a wording pass once the mssql-python-rs packaging is settled.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -999,6 +1003,94 @@ std::string GetLastErrorMessage();
...
+        return kProviderMssqlOdbc;
+    }
+    return kProviderMsodbcsql18;
+}
+
+std::string ProviderPackageForId(const std::string& id) {
+    return (id == kProviderMssqlOdbc) ? "mssql_py_core" : "mssql_python_odbc";
+}
+
+std::string ProviderDistForId(const std::string& id) {
```
bewithgaurav, line 1092: nit: `ProviderPackageForId` and `ProviderDistForId` hardcode the same provider to package and provider to dist mapping that `odbc_provider.py` already owns in `_PACKAGE_BY_PROVIDER` / `_DIST_BY_PROVIDER`. 

two copies means a third provider has to land in both languages or the native error strings drift from the python ones, not worth a shared source with two providers today, but a comment on each side pointing at the other would save the next person from updating only one.

bewithgaurav (commented):
some changes suggested on refactoring, errors etc.

one nit - will leave upto you to include or not

mssql_python/pybind/ddbc_bindings.h
gargsaumya: Done in bc2f70c2 — removed `isDriverLoaded()`, the `m_driverLoaded` field, and the now-unused `<atomic>` include. Confirmed zero remaining references; the post-load immutability guard lives entirely in `g_providerSelectionLocked`.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Confirmed the scenario: `mssql_py_core` is bundled into the `mssql-python` wheel (for bulkcopy) without the `mssqlodbc.*` binary, so selecting `mssql-odbc` on a plain `pip install mssql-python` reaches this throw and can report `mssql-python-rs` as "installed" when only the bundled core is present. Agreed it fails closed, so no wrong-driver risk. Leaving the wording until the `mssql-python-rs` packaging lands, since the right message depends on whether the driver binaries ship in the bundled core or the standalone dist — tracking it for that pass.

mssql_python/pybind/ddbc_bindings.cpp
gargsaumya: Done in bc2f70c2 — added reciprocal "keep in sync" comments on both `ProviderPackageForId`/`ProviderDistForId` (ddbc_bindings.cpp) and `_PACKAGE_BY_PROVIDER`/`_DIST_BY_PROVIDER` (odbc_provider.py) so the duplicated mapping can't drift silently.

bewithgaurav (approved):
lgtm

saurabh500 (approved):

Issue #563, opened 2026-05-11
Publish mssql-python as a conda-forge package


martinstuder:
### Is your feature request related to a problem? Please describe.
Many data science and machine learning workflows rely on conda as the primary package and environment manager. Currently, mssql-python is only available via PyPI, which creates friction for conda-based projects. When a project's dependency tree is managed through conda (e.g. using conda-forge), mixing pip-installed packages can cause environment inconsistencies, dependency conflicts, and reproducibility issues - particularly in regulated or enterprise environments where environment lockfiles and auditable builds are required.

### Describe the solution you'd like
It would be great if mssql-python would be available as a conda package on the conda-forge channel (or an alternative channel). This would allow users to install it via `conda install -c conda-forge mssql-python`. This involves submitting a feedstock recipe to the conda-forge/staged-recipes repository. The recipe would define the package metadata, source (pointing to the PyPI release or source tarball), build requirements, and runtime dependencies - following the standard conda-forge contribution process. Ideally the conda package would be kept in sync with PyPI releases so that both channels remain up to date.

### Describe alternatives you've considered
* Using pip inside a conda environment: this is the current workaround, but it undermines the integrity of conda environment solves and is discouraged in environments where full conda-managed reproducibility is required.
* Self-hosting a conda package: it is possible to build a local conda package from the PyPI source using conda build. However, this places a maintenance burden on end-users and is not a scalable or shareable solution.

### Additional context
The conda-forge contribution process is well-documented and community-supported. The grayskull tool can auto-generate an initial feedstock recipe directly from the PyPI package metadata, which significantly reduces the effort required: grayskull pypi mssql-python. Many comparable database drivers (e.g. pyodbc, psycopg2) are already available on conda-forge.

sumitmsft:
Hello @martinstuder

Thanks for your suggestions. 

Publishing mssql-python on Conda package manager is on our roadmap and the work on it will start this month. Please stay tuned and we will update this space once we are ready to publish the package on Conda. 

Sumit

oscarrobertson:
@sumitmsft  any progress on this? 

A concrete data point for this request: I'm on RHEL 9, managing a conda env using pixi (and installing this lib via pypi-dependencies), some of the changes around splitting out the odbc driver now make this more challenging. The pip wheel fails to load in conda environments that contain conda-forge krb5 (which is common — libcurl, pyarrow etc. pull it in transitively).

[...]

PR #734, opened 2026-08-31
FEAT: conda recipe for mssql-python (self-contained repackage of the signed wheel)

jahnvi480:

> [AB#47315](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47315)

### Summary
First of a series splitting the large conda onboarding branch (#720) into smaller, reviewable PRs.

Adds the conda-build **recipe** that repackages the prebuilt, ESRP-signed `mssql-python` wheel into a **self-contained** conda package (it vendors the ODBC Driver 18 payload — there is no separate companion package). The recipe compiles nothing; it repackages the already-signed wheel, fully offline.

**Files**
- `conda/mssql-python/meta.yaml` — recipe manifest (version from `MSSQL_PYTHON_VERSION`; repackage-safe build flags so conda-build never rewrites/scans the signed binaries).
- `conda/mssql-python/build.sh` — Linux/macOS: install/extract the code + odbc wheels into site-packages, then stamp the relative `$ORIGIN` RUNPATH climb (#563) onto the vendored Linux driver so it resolves the env's own krb5/openssl.
- `conda/mssql-python/bld.bat` — Windows: extract the code + odbc wheels; on **win-arm64** it strips the x64 `mssql_py_core` (no arm64 build exists yet, so bulk copy is a documented limitation on that platform).
- `conda/.gitattributes`

**Safety:** this recipe is **dormant on `main`** — nothing runs it until the standalone conda build pipeline (a follow-up PR) invokes it, so merging is safe and changes no existing behavior.

Follow-up PRs in this split: (2) conda build pipeline + the gates/probes it runs, (3) conda release pipeline, (4) the GitHub conda-audit workflow.

### DIFF
--- a/conda/mssql-python/bld.bat
+++ b/conda/mssql-python/bld.bat
@@ -1,0 +1,57 @@
+@echo on
+REM Repackage the prebuilt, ESRP-signed wheel into a conda package (offline), and
+REM vendor the ODBC Driver 18 payload INTO it (v1.11.0 model: libs ship inside).
+REM PKG_NAME / PKG_VERSION are exported by conda-build; WHEELS_DIR + MSSQL_ODBC_VERSION
+REM by the pipeline.
+setlocal enabledelayedexpansion
+
+REM Site-packages of the (possibly cross-targeted) host env, plus the underscore form
+REM of the package name used in wheel filenames (mssql-python -> mssql_python).
+set "SP=%PREFIX%\Lib\site-packages"
+if not exist "%SP%" mkdir "%SP%"
+set "PKG_UNDERSCORE=%PKG_NAME:-=_%"
+
+REM Install the prebuilt code wheel. On a NATIVE leg the host Python runs and pip
+REM installs it (pip auto-selects the wheel matching this host). On a CROSS win-arm64
+REM build the win-arm64 host Python CANNOT execute on the x64 agent (Windows has no
+REM reverse emulation), so extract the wheel (a zip) WITHOUT Python -- the same
+REM approach build.sh uses for the osx-arm64 cross-build. conda-build exports CONDA_PY
+REM (e.g. 314), which selects the matching cp tag of the win_arm64 slice.
+"%PYTHON%" -c "import sys" >nul 2>&1
+if errorlevel 1 (
+  echo Host Python "%PYTHON%" not executable here ^(non-emulated cross-build^); extracting cp%CONDA_PY% win_arm64 code wheel without Python.
+  set "CODE_WHL="
+  for %%W in ("%WHEELS_DIR%\!PKG_UNDERSCORE!-%PKG_VERSION%-cp%CONDA_PY%-*-win_arm64.whl") do if exist "%%~fW" set "CODE_WHL=%%~fW"
+  if not defined CODE_WHL (
+    echo ERROR: no %PKG_NAME%==%PKG_VERSION% cp%CONDA_PY% win_arm64 wheel in "%WHEELS_DIR%"
+    exit /b 1
+  )
+  echo Extracting "!CODE_WHL!" into "%SP%"
+  tar -xf "!CODE_WHL!" -C "%SP%"
+  if errorlevel 1 exit /b 1
+  REM No arm64 Windows build of mssql_py_core exists yet, so the win-arm64 code wheel
+  REM bundles the x64 (win_amd64) one. An x64 .pyd cannot load on arm64 -- it would
+  REM crash bulk copy AND fails the PE-arch assert -- so REMOVE it from the win-arm64
+  REM package. Bulk copy then raises a clean "not available" error (the import is lazy);
+  REM the rest of the DBAPI is unaffected (ddbc_bindings is native arm64). Restore once
+  REM an arm64 mssql_py_core ships in the feed.
+  echo Removing x64 mssql_py_core from the win-arm64 package; bulk copy is unavailable on win-arm64 until an arm64 build ships.
+  if exist "%SP%\mssql_py_core" rmdir /s /q "%SP%\mssql_py_core"
+  if exist "%SP%\mssql_py_core.libs" rmdir /s /q "%SP%\mssql_py_core.libs"
+) else (
+  "%PYTHON%" -m pip install --no-deps --no-index --find-links "%WHEELS_DIR%" %PKG_NAME%==%PKG_VERSION% -vv
+  if errorlevel 1 exit /b 1
+)
+
+REM Extract the python-agnostic py3-none-win odbc wheel into the SAME site-packages
+REM so mssql_python_odbc\libs\ sits beside mssql_python\ (the loader finds the driver
+REM there). WHEELS_DIR is staged per-target, so a single matching odbc wheel is present.
+set "ODBC_WHL="
+for %%W in ("%WHEELS_DIR%\mssql_python_odbc-%MSSQL_ODBC_VERSION%-py3-none-win_*.whl") do if exist "%%~fW" set "ODBC_WHL=%%~fW"
+if not defined ODBC_WHL (
+  echo ERROR: no mssql_python_odbc==%MSSQL_ODBC_VERSION% py3-none-win wheel in "%WHEELS_DIR%"
+  exit /b 1
+)
+echo Extracting "!ODBC_WHL!" into "%SP%"
+tar -xf "!ODBC_WHL!" -C "%SP%"
+if errorlevel 1 exit /b 1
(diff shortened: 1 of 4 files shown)


conda/mssql-python/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    # Bare `python` for the same reason as host (the floor is the per-build python_abi,
+    # NOT a range on python -- a range breaks the per-Python solve).
+    - python
+    # conda does NOT inherit the wheel's install_requires, so pin azure-identity here.
+    - azure-identity >=1.12.0
+    # ODBC Driver 18 payload deps (the driver ships inside this package, so its
+    # security-serviced deps are declared here). OpenSSL for TLS is dlopen'd, so
+    # overlinking can't see it; pinned <4 (Driver 18 supports the OpenSSL 1.1/3.0 ABI
+    # only). macOS uses Homebrew OpenSSL, Windows SChannel.
+    - openssl >=3,<4          # [linux]
```
bewithgaurav, line 57: just double checked this one - On macOS this package is not self-contained for TLS and still points to homebrew installation instead of conda env

concrete break scenario: end user on a conda-only Mac (no Homebrew) runs `connect("...;Encrypt=yes")`, the default. 
The connect fails with `SSL Provider: [OpenSSL library could not be loaded, make sure OpenSSL 1.0, 1.1, or 3.0 is installed]`.

repro'd on the shipped 1.14.0 osx-arm64 package, one real SQL Server, `Encrypt=yes`:
- Homebrew present: TLS completes (login-failed), and the driver binds `/opt/homebrew/Cellar/openssl@3/3.6.3/lib/libssl.3.dylib`.
- Homebrew hidden (`sandbox-exec` denying reads of `/opt/homebrew`, which simulates a clean Mac without deleting anything): the same connect fails with `OpenSSL library could not be loaded`.

why the recipe misses it: 

`openssl` is declared `[linux]` only, the macOS payload bundles no `libssl`/`libcrypto`, and the `$ORIGIN` climb in `build.sh` has no macOS twin

conda's own openssl does land in `$PREFIX/lib` (pulled in by python), but the driver never falls back to it, so the env copy is unreachable.

Why CI and manual runs look green: 
- both have Homebrew OpenSSL, so the driver silently binds that copy
- the masking-immune audit only covers Linux, so nothing catches this on macOS.

direction: give the macOS driver a way to resolve the env's own openssl (the `@loader_path` twin of the Linux climb pointing at `$PREFIX/lib`, or bundle/symlink it into the payload), and add a macOS self-containment check on a Homebrew-free agent so it cannot be masked again.

bewithgaurav (changes requested):
~~blocker on macOS, will re-review once done~~
edit: discussed and checked - this is a limitation w.r.t. macOS

conda/mssql-python/bld.bat
```diff
@@ -0,0 +1,53 @@
...
+  if exist "%SP%\mssql_py_core.libs" rmdir /s /q "%SP%\mssql_py_core.libs"
+) else (
+  "%PYTHON%" -m pip install --no-deps --no-index --find-links "%WHEELS_DIR%" %PKG_NAME%==%PKG_VERSION% -vv
+  if errorlevel 1 exit /b 1
+)
+
+REM Extract the py3-none-win odbc wheel into the SAME site-packages so
+REM mssql_python_odbc\libs\ sits beside mssql_python\ and the loader finds the driver.
+set "ODBC_WHL="
+for %%W in ("%WHEELS_DIR%\mssql_python_odbc-%MSSQL_ODBC_VERSION%-py3-none-win_*.whl") do if exist "%%~fW" set "ODBC_WHL=%%~fW"
```
sumitmsft: The code wheel is matched arch-specifically (`cp%CONDA_PY%-*-win_arm64.whl`) but the ODBC wheel here uses `py3-none-win_*`, which matches both `win_amd64` and `win_arm64`. The `py3-none` tag only means "no Python bytecode" — the ODBC payload DLLs inside are still architecture-specific.

This currently works only because of the implicit "one wheel per target `WHEELS_DIR`" contract. If both arch wheels are ever staged in the same dir, the `for` loop picks a filesystem-order-dependent match (last one wins, no `break`, no arch assert) and could silently vendor an x64 driver into a win-arm64 package. Unlike the code wheel, there's no post-extract arch guard here, so a wrong-arch driver would import fine and only fail at the user's first `connect()`.

Consider making this glob arch-specific to mirror the code-wheel handling (e.g. select `-win_arm64.whl` on the arm64 leg / `-win_amd64.whl` otherwise), and optionally fail if more than one candidate matches. The same `py3-none-*` asymmetry exists in `build.sh` (lower-risk there since macOS ships universal2 payloads).

conda/mssql-python/meta.yaml
jahnvi480: As we discussed, we've exhausted the in-env options: the driver dlopens OpenSSL only from fixed system paths (Homebrew / MacPorts / /usr/lib) by unversioned name, with no @loader_path/rpath/bare-leaf hook — so DYLD_FALLBACK, the RTLD_GLOBAL preload, and payload bundling were all validated as ineffective on real hardware.

So this will be a known limitation, documented: macOS conda Encrypt=yes needs system OpenSSL (brew install openssl / MacPorts) — the same requirement as the PyPI wheel today. Windows (SChannel) and Linux ($ORIGIN climb) remain self-contained.

conda/mssql-python/bld.bat
jahnvi480: Fixed this

conda/mssql-python/bld.bat
```diff
@@ -0,0 +1,60 @@
...
+    exit /b 1
+  )
+  REM The win-arm64 wheel bundles the x64 mssql_py_core (no arm64 build exists yet); an
+  REM x64 .pyd can't load on arm64 and fails the PE-arch assert, so strip it. Bulk copy
+  REM then raises a clean "not available" error (lazy import); the rest of the DBAPI works.
+  echo Removing x64 mssql_py_core from the win-arm64 package; bulk copy is unavailable on win-arm64 until an arm64 build ships.
+  if exist "%SP%\mssql_py_core" rmdir /s /q "%SP%\mssql_py_core"
+  if exist "%SP%\mssql_py_core.libs" rmdir /s /q "%SP%\mssql_py_core.libs"
+) else (
+  set "ODBC_ARCH=win_amd64"
```
sumitmsft: This branch treats every executable `%PYTHON%` Windows build as `win_amd64`. A native Windows ARM64 build would install the ARM64 code wheel but then extract or require the x64 ODBC wheel, leaving the package without the `libs\windows\arm64` payload the loader expects. Please set `ODBC_ARCH` from the conda target platform (e.g. `target_platform`/`CONDA_SUBDIR`, mapping `win-arm64`→`win_arm64` and `win-64`→`win_amd64`) rather than from whether Python can execute.

sumitmsft (commented):
Automated review (proposal-only workflow), posting one high-confidence packaging reliability finding. Non-blocking comment.

**Native Windows ARM64 selects the x64 ODBC wheel.** On a native Windows ARM64 conda-build host the ARM64 Python is executable, so the `else` branch runs and installs the ARM64 code wheel — but `ODBC_ARCH` is hardcoded to `win_amd64`, so the later ODBC extraction targets the x64 payload. The build then either fails looking for `mssql_python_odbc-...-py3-none-win_amd64.whl` or vendors the wrong driver DLLs, leaving the package without the `libs\windows\arm64` payload the loader expects. Consider deriving `ODBC_ARCH` from the conda target platform (`target_platform`/`CONDA_SUBDIR`) rather than from whether Python can execute.

conda/mssql-python/bld.bat
jahnvi480: Fixed

sumitmsft (approved):
Re-reviewed at `95c62f64`. The earlier finding is resolved: `conda/mssql-python/bld.bat` now derives `ODBC_ARCH` from the conda target platform (`target_platform`/`CONDA_SUBDIR`) and uses the exact arch-specific ODBC wheel tag, so the native Windows ARM64 x64-wheel mis-selection no longer applies.

No new high-confidence correctness, security, reliability, or breaking-change findings. The outstanding macOS OpenSSL concern was already raised by another reviewer and is not duplicated here.

bewithgaurav (approved):

conda/mssql-python/meta.yaml
bewithgaurav: ack, this is a genuine limitation of internal ODBC impl - lets make sure to document this properly
edited my top level comment, resolving this thread

PR #735, opened 2026-09-01
CHORE: load bundled Windows driver and auth DLLs from package-local directories

bewithgaurav:
### Summary

On Windows the vendored ODBC driver (`msodbcsql18.dll`) and Entra auth DLL (`mssql-auth.dll`) are loaded with `LoadLibraryW`, so their dependencies are resolved via the default process search order, which also consults the current working directory and `%PATH%`. This makes dependency resolution dependent on ambient environment.

This change loads both DLLs with `LoadLibraryExW` using `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR`, which keeps the current working directory and `%PATH%` out of the search and confines dependency resolution to the driver's own directory, the application directory, and System32.

**Changes**
- `mssql_python/pybind/ddbc_bindings.cpp`: `LoadLibraryExW` with constrained search flags on the driver and auth-DLL loads (replacing `LoadLibraryW`). The flag constants are `#define`d defensively in case the build SDK gates them behind an older `_WIN32_WINNT`.
- `tests/test_026_windows_dll_search.py`: a source-contract regression test that fails if the loader reintroduces a bare `LoadLibraryW` call or drops the constrained-search flags. It is deliberately a source check, not a runtime one — the search restriction only manifests at DLL-resolution time on Windows and cannot be observed without dropping a file on disk, and a "does it still load" check would pass on the unhardened code too (any host with the dependency already in System32), so it would guard nothing.

**Scope / notes**
- macOS and Linux are unaffected — the change is entirely within `#ifdef _WIN32` blocks, and the shipped Unix libraries already carry `@loader_path` / `$ORIGIN`.
- No global process state is mutated (`SetDefaultDllDirectories` / `AddDllDirectory` are intentionally not used), so a consuming application's own DLL resolution is unchanged. The bundled VC++ runtime already loads from the trusted copy next to the pybind extension (built `/MD`), so no extra search directory is needed.
- Requires Windows 8+ (or Windows 7 + KB2533623), which the supported Python versions already imply.

**Validation**
This is a Windows-only native change and is validated by the ADO Windows build/test legs (x64/arm64/x86); it cannot be exercised on a non-Windows dev host.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -18,6 +18,7 @@
 #include <cstring>  // For std::memcpy
 #include <filesystem>
 #include <iostream>
+#include <mutex>  // std::once_flag / std::call_once
 #include <utility>  // std::forward
 #include <datetime.h>  // CPython datetime API (PyDateTime_IMPORT, PyDateTime_GET_*, etc.)
 
@@ -25,6 +26,18 @@
 //-------------------------------------------------------------------------------------------------
 // Macro definitions
 //-------------------------------------------------------------------------------------------------
+
+#ifdef _WIN32
+// Constrained DLL search flags (Windows 8+ / Win7 + KB2533623). Defined
+// defensively in case the build's SDK headers gate them behind an older
+// _WIN32_WINNT than this project targets.
+#ifndef LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR
+#define LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR 0x00000100
+#endif
+#ifndef LOAD_LIBRARY_SEARCH_DEFAULT_DIRS
+#define LOAD_LIBRARY_SEARCH_DEFAULT_DIRS 0x00001000
+#endif
+#endif  // _WIN32
 
 #ifndef SQL_C_DATE
 #define SQL_C_DATE (9)
@@ -1047,9 +1060,19 @@ DriverHandle LoadDriverLibrary(const std::string& driverPath) {
     // fs::path::c_str() returns wchar_t* on Windows with correct encoding
     namespace fs = std::filesystem;
     fs::path pathObj(driverPath);
-    HMODULE handle = LoadLibraryW(pathObj.c_str());
+    // Resolve the vendored driver and its dependencies from trusted,
+    // package-local directories only. LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR adds the
+    // driver's own folder for its dependency lookups, and
+    // LOAD_LIBRARY_SEARCH_DEFAULT_DIRS restricts the rest of the search to
+    // System32, the application directory, and directories registered via
+    // AddDllDirectory (see LoadDriverOrThrowException) -- excluding the current
+    // working directory and %PATH%, which the legacy LoadLibraryW search order
+    // would otherwise include.
+    HMODULE handle = LoadLibraryExW(
+        pathObj.c_str(), nullptr,
+        LOAD_LIBRARY_SEARCH_DEFAULT_DIRS | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR);
     if (!handle) {
-        LOG("LoadDriverLibrary: LoadLibraryW failed for path='%s' - %s", driverPath.c_str(),
+        LOG("LoadDriverLibrary: LoadLibraryExW failed for path='%s' - %s", driverPath.c_str(),
             GetLastErrorMessage().c_str());
         ThrowStdException("Failed to load library: " + driverPath);
     }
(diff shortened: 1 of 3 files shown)


tests/test_026_windows_dll_search.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+
+def test_driver_and_auth_loads_use_constrained_search():
+    code = _code_without_comments(_LOADER_SRC.read_text(encoding="utf-8"))
+    # Both the driver and the auth DLL are loaded with the hardened API.
+    assert (
+        len(re.findall(r"\bLoadLibraryExW\s*\(", code)) >= 2
+    ), "expected LoadLibraryExW for both the driver and the auth DLL loads"
+    assert "LOAD_LIBRARY_SEARCH_DEFAULT_DIRS" in code
```
gargsaumya: The regression test does not verify flags are used.

These assertions do not associate the flags with either LoadLibraryExW call. The flag names remain present in the #define declarations, so replacing both calls’ third arguments with 0 still passes every assertion here. Please verify that each driver/auth invocation contains both required flags, or add a runtime DLL-resolution test.

bewithgaurav: nice catch, I've pushed a commit to change this thing

gargsaumya (dismissed):

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -1047,9 +1059,17 @@ DriverHandle LoadDriverLibrary(const std::string& driverPath) {
...
-    HMODULE handle = LoadLibraryW(pathObj.c_str());
+    // Resolve the vendored driver's dependencies with a constrained search
+    // path. LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR adds the driver's own folder for
+    // its dependency lookups, and LOAD_LIBRARY_SEARCH_DEFAULT_DIRS restricts the
+    // rest of the search to System32 and the application directory -- excluding
+    // the current working directory and %PATH%, which the legacy LoadLibraryW
+    // search order would otherwise include.
+    HMODULE handle = LoadLibraryExW(
+        pathObj.c_str(), nullptr,
+        LOAD_LIBRARY_SEARCH_DEFAULT_DIRS | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR);
```
sumitmsft: `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` also includes `LOAD_LIBRARY_SEARCH_USER_DIRS` (per the LoadLibraryEx docs it is APPLICATION_DIR + SYSTEM32 + USER_DIRS), so this search is not limited to the driver dir, application dir, and System32 as the comment above states — it also searches directories registered process-wide via `AddDllDirectory`/`SetDllDirectory`. If another in-process module has registered such a directory, a missing/transitive dependency of the driver could resolve from it. Consider using explicit `LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32 | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR` here (and for the `mssql-auth.dll` load at line 1224), or update the comment/tests to note user DLL dirs are intentionally trusted.

sumitmsft (changes requested):
**Constrained DLL search still includes process-global user directories (Medium).** Both `LoadLibraryExW` calls use `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR`. Per the Win32 docs, `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` is a combination of `LOAD_LIBRARY_SEARCH_APPLICATION_DIR`, `LOAD_LIBRARY_SEARCH_SYSTEM32`, **and `LOAD_LIBRARY_SEARCH_USER_DIRS`**. So dependency resolution is not confined to the driver directory, the application directory, and System32 as the code comment (lines 1064-1065) states — it also searches any directory registered process-wide via `AddDllDirectory`/`SetDllDirectory`.

**Impact:** If another module loaded in the same process registers a DLL directory, a missing/transitive dependency of the vendored driver (or `mssql-auth.dll`) can resolve from that ambient location, weakening the intended package-local confinement. This is narrower than a CWD/%PATH% hijack (which this PR correctly excludes) since USER_DIRS requires in-process registration, but the stated intent and the actual behavior do not match.

**Suggested correction:** Replace `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` with explicit `LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32 | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR` on both loads (the fully-qualified path already satisfies `DLL_LOAD_DIR`), or update the comment/tests to acknowledge that user DLL directories are intentionally trusted.

mssql_python/pybind/ddbc_bindings.cpp
bewithgaurav: thanks - this is a valid case, have fixed it
both loads now use `LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32 | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR`, so user DLL dirs added via `AddDllDirectory`/`SetDllDirectory` are out of the search now

the regression test asserts these flags per call and fails if `DEFAULT_DIRS` comes back.

sumitmsft (approved):
Re-reviewed at `5e6268c8`. The earlier finding is resolved: both `LoadLibraryExW` calls in `mssql_python/pybind/ddbc_bindings.cpp` now use explicit `LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32 | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR`, so process-global user DLL directories are no longer searched, and the regression test rejects `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` per call.

No further high-confidence correctness, security, reliability, or breaking-change findings. Existing reviewer comments were cross-checked and not duplicated.

gargsaumya (approved):

Issue #739, opened 2026-09-02
[MSSQL-Django] Binary() rejects memoryview, breaking Django BinaryField writes and serialization


bewithgaurav:
### Describe the bug

`mssql_python.Binary()` accepts only `str`, `bytes`, and `bytearray`. It raises `TypeError` on a `memoryview`, which pyodbc and most DB-API drivers accept. `memoryview` is a standard buffer type, and Django's `BinaryField` hands it straight to the driver, so any `BinaryField` write or lookup fails under mssql-python.

Exception message:
```
TypeError: Cannot convert type memoryview to bytes. Binary() only accepts str, bytes, or bytearray objects.
```

The check is in `mssql_python/type.py` (`Binary()`, line 123): it handles `bytes`, `bytearray`, and `str`, then raises for everything else.

### To reproduce

Standalone, no database needed:

```python
import mssql_python

mv = memoryview(b"\x01\x02\x03")
mssql_python.Binary(mv)
# TypeError: Cannot convert type memoryview to bytes.
# Binary() only accepts str, bytes, or bytearray objects.
```

For contrast, pyodbc accepts it:

```python
import pyodbc
pyodbc.Binary(memoryview(b"\x01\x02\x03"))   # returns bytearray(b'\x01\x02\x03')
```

Through Django on the mssql-django backend, a plain `BinaryField` save hits the same path and fails:

```python
# class Blob(models.Model): data = models.BinaryField()
Blob.objects.create(data=b"\x00\x01\x02")
#   File ".../django/db/models/fields/__init__.py", in get_db_prep_value
#     return connection.Database.Binary(value)
# TypeError: Cannot convert type memoryview to bytes. ...
```

### Expected behavior

`Binary(memoryview(...))` returns the bytes, matching pyodbc and the DB-API convention of accepting buffer-protocol objects. Django `BinaryField` writes, lookups, and fixture load/dump then work under mssql-python.

### Further technical details

Python version: 3.13 (also reproduces on 3.10 through 3.14)
SQL Server version: SQL Server 2022
Operating system: driver-side, OS-independent (seen on macOS and Ubuntu)
mssql-python version: 1.14.0

**Additional context**

Found while validating mssql-python as a driver for mssql-django. This accounts for 7 of the 9 Django integration-suite failures: 2 in `model_fields.test_binaryfield` and 5 in `serializers.test_data` (json/jsonl/python/xml/yaml), since Django's serializers roundtrip `BinaryField` data through `memoryview`.

[...]

PR #741, opened 2026-09-02
FIX: accept memoryview in Binary()

bewithgaurav:

> GitHub Issue: #739

### Summary

`Binary()` rejected `memoryview` with a `TypeError`, so any DB-API caller that
passes a buffer-protocol value failed. Django's `BinaryField` hands the driver a
`memoryview`, which broke `BinaryField` writes and the serializer roundtrips.
Accept `memoryview` via `tobytes()`, matching pyodbc and the DB-API convention.


mssql_python/type.py
```diff
@@ -122,13 +122,15 @@ def TimestampFromTicks(ticks: int) -> datetime.datetime:
 
 def Binary(value) -> bytes:
```
jahnvi480: Since `py.typed` ships, type checkers read the inline signature here, and `value` is currently untyped (`Any`). Consider annotating the runtime function directly so the new accepted-input contract actually reaches callers:


Python
```
def Binary(value: Union[str, bytes, bytearray, memoryview]) -> bytes:
```

bewithgaurav: fixing, good catch - this is a great usecase to have the type checker in CI

jahnvi480 (approved):

sumitmsft (approved):

**Details**
- `Binary()` normalizes `memoryview` to an exact `bytes` via `.tobytes()`, preserving the existing `bytes` return contract. This keeps the downstream binding path unchanged, since the binder already maps `bytes` to `SQL_VARBINARY`/`SQL_C_BINARY` (including large-value streaming).
- `.tobytes()` produces an owned copy in logical C order, so non-contiguous, bytearray-backed, and empty memoryviews all round-trip correctly with no aliasing/lifetime concern.
- Accepting `memoryview` aligns with pyodbc's acceptance of buffer-protocol values (pyodbc materializes to `bytearray`; this driver keeps its `bytes` contract) and resolves GH-739 (Django `BinaryField` hands the driver a `memoryview`).
- Runtime signature and the PEP 561 `.pyi` stub are kept in sync; a focused regression test covers memoryview inputs and the updated error message; Black passes.

**Recommendation:** Approve

Issue #740, opened 2026-09-02
[MSSQL-Django] Decimal in money range bound as string, causing arithmetic overflow on numeric columns


bewithgaurav:
### Describe the bug

`mssql_python` binds a Python `Decimal` whose value falls within the SQL Server money range (about -214,748 to +214,748) as a string (`SQL_VARCHAR` / `SQL_C_CHAR`), based on the value alone, ignoring the target column type. SQL Server then does a server-side varchar to numeric conversion at execution. When the value does not fit the target `numeric`/`decimal` column, that conversion overflows and raises, instead of the value simply not matching. pyodbc binds the same value as `SQL_NUMERIC` client-side, so the comparison just returns no rows.

Exception message:
```
DataError: ... Arithmetic overflow error converting varchar to data type numeric.
```

The heuristic is in `mssql_python/cursor.py` (`_map_sql_type`, line 646). A `Decimal` within `SMALLMONEY_MIN`..`SMALLMONEY_MAX` (line 797) or `MONEY_MIN`..`MONEY_MAX` returns a `SQL_VARCHAR` string bind. Only values outside those ranges reach the `SQL_NUMERIC` path (line 819) that binds with self-derived precision.

### To reproduce

```python
import mssql_python
from decimal import Decimal

conn = mssql_python.connect("Server=localhost;Database=tempdb;UID=sa;PWD=...;Encrypt=no;TrustServerCertificate=yes;")
conn.autocommit = True
cur = conn.cursor()

cur.execute("CREATE TABLE #t (v numeric(5,2))")   # max value 999.99
cur.execute("INSERT INTO #t VALUES (?)", [Decimal("12.34")])

# Out-of-range value that sits inside the money range (~ +/-214,748):
cur.execute("SELECT COUNT(*) FROM #t WHERE v = ?", [Decimal("12345.6789")])
# mssql_python.exceptions.DataError:
#   Arithmetic overflow error converting varchar to data type numeric.

# A value OUTSIDE the money range works (binds as SQL_NUMERIC):
cur.execute("SELECT COUNT(*) FROM #t WHERE v = ?", [Decimal("300000.00")])
print(cur.fetchone()[0])   # 0, no error
```

For contrast, pyodbc returns 0 with no error for the same in-range value.

### Expected behavior

A `Decimal` comparison against a `numeric`/`decimal` column returns the correct result (no match) without raising, matching pyodbc. Binding should not depend on whether the value happens to fall in the money range.

### Further technical details

[...]

Issue #503, opened 2026-04-08
setinputsizes() with SQL_DECIMAL/SQL_NUMERIC crashes with C type mismatch


dlevy-msft-sql:
### Describe the bug

Calling `cursor.setinputsizes()` with `SQL_DECIMAL` or `SQL_NUMERIC` type tuples causes a `RuntimeError` when followed by `executemany()`. The C binding rejects Python `Decimal` objects because `_get_c_type_for_sql_type` maps `SQL_DECIMAL`/`SQL_NUMERIC` to `SQL_C_NUMERIC`, but the binding doesn't convert `Decimal` to the `SQL_NUMERIC_STRUCT` that ODBC expects.

Without `setinputsizes()`, the driver infers the correct type and inserts succeed.

```
Exception message:
RuntimeError: Parameter's object type does not match parameter's C type. paramIndex - 2, C type - SQL_C_NUMERIC
```

### To reproduce

```python
import mssql_python
from decimal import Decimal

conn = mssql_python.connect(
    "Server=myserver;Database=mydb;UID=user;PWD=pass;"
    "TrustServerCertificate=yes;Encrypt=yes"
)
cursor = conn.cursor()

cursor.execute("""
    IF OBJECT_ID('dbo.SetInputSizesTest') IS NOT NULL
        DROP TABLE dbo.SetInputSizesTest;
    CREATE TABLE dbo.SetInputSizesTest (
        Name NVARCHAR(100),
        CategoryID INT,
        Price DECIMAL(18,2)
    )
""")
conn.commit()

# This works fine - no setinputsizes
cursor.executemany(
    "INSERT INTO dbo.SetInputSizesTest (Name, CategoryID, Price) VALUES (?, ?, ?)",
    [("Widget", 1, Decimal("19.99")), ("Gadget", 2, Decimal("29.99"))]
)
conn.commit()
print("Without setinputsizes: OK")

cursor.execute("DELETE FROM dbo.SetInputSizesTest")
conn.commit()

# This crashes
cursor.setinputsizes([
    (mssql_python.SQL_WVARCHAR, 100, 0),   # nvarchar(100) - works
    (mssql_python.SQL_INTEGER, 0, 0),       # int - works
    (mssql_python.SQL_DECIMAL, 18, 2)       # decimal(18,2) - crashes
])

cursor.executemany(
    "INSERT INTO dbo.SetInputSizesTest (Name, CategoryID, Price) VALUES (?, ?, ?)",
[...]

PR #742, opened 2026-09-02
FIX: bind Decimal as SQL_NUMERIC regardless of value

bewithgaurav:

> GitHub Issue: #740

### Summary

The standard execute path chose a `Decimal`'s bind type from its value, sending
anything in the MONEY/SMALLMONEY range as a formatted VARCHAR. Comparing such a
value against a smaller numeric column made SQL Server convert varchar to numeric
and overflow, so `WHERE v = ?` raised an arithmetic overflow instead of simply not
matching. Bind every finite `Decimal` as `SQL_NUMERIC` with its own precision and
scale, matching pyodbc.

Removing the shortcut surfaced a second bug: the numeric parameter's descriptor
record number was hardcoded to 1, so a numeric parameter in any position other than
the first wrote its precision/scale onto the wrong record and the driver raised
"Numeric value out of range". Use the parameter's own 1-based position.

The fix covers both `execute()` routes: the native C++ detection path, and the Python
legacy path reached when `setinputsizes()` covers fewer positions than parameters (an
uncovered money-range `Decimal` there previously fell back to the VARCHAR shortcut and
overflowed). `executemany` is intentionally left on its batch VARCHAR string binding
(GH-503), so a money-range `Decimal` compared through `executemany` can still overflow;
that is a tangential finding filed separately as #745.


mssql_python/pybind/param_detect.hpp
```diff
@@ -511,45 +512,11 @@ inline std::vector<ParamInfo> DetectParamTypes(PyObject* params) {
...
-                    // PyList_SetItem steals (decrefs) the item even on failure,
-                    // so raw is already freed — do NOT Py_DECREF here.
-                    throw py::error_already_set();
-                }
-                continue;
-            }
-
-            // Build SQL_NUMERIC_STRUCT from the Decimal object. Store as a pybind11-castable
-            // object in the param list so BindParameters can extract it as NumericData.
+            // Bind every finite Decimal as SQL_NUMERIC using its own precision and scale,
```
jahnvi480, line 515: The identical money→VARCHAR shortcut still lives in the Python `_map_sql_type`, which this PR doesn't touch, and that path is still reachable from a real `execute()`. If a caller uses `setinputsizes()` for only some positions, the un-sized params fall back to `_map_sql_type` and GH-740 reproduces again. `executemany()` has the same problem for `WHERE numeric_col = ?`.

Can we either mirror this change in `_map_sql_type`  or explicitly scope/track the remaining path?

tests/test_020_money_smallmoney.py
```diff
@@ -4,9 +4,12 @@
...
 through MONEY, SMALLMONEY, and DECIMAL columns with proper precision handling.
 
-Key implementation detail: MONEY-range Decimals use string binding (SQL_VARCHAR)
-because SQL_NUMERIC binding fails with ODBC "Numeric value out of range" error.
-String binding preserves full precision and SQL Server handles conversion.
+Key implementation detail: on the execute() path every finite Decimal binds as
+SQL_NUMERIC using its own precision and scale, regardless of value. Binding no longer
+depends on whether the value falls in the MONEY/SMALLMONEY range, so an in-range value
+compared against a smaller numeric column returns no match instead of a varchar->numeric
+overflow (GH-740). executemany still string-binds Decimals (SQL_VARCHAR) to preserve
```
jahnvi480, line 11: Since native binding is now NUMERIC, `test_023_execute_path_parity.py::test_money_range_decimal_binds_wide` (L515) is misleading — it's a **native-path** test whose docstring says money-range decimals bind as text, but it only asserts a value round-trip, so it still passes even though the C type changed. The file's whole "native == `_map_sql_type` reference" premise is now violated for this case. Worth updating it here to assert the native base type is `numeric` so we keep real parity coverage instead of a green-but-vacuous test.

tests/test_020_money_smallmoney.py
```diff
@@ -640,3 +643,119 @@ def test_both_null(cursor, db_connection):
...
+        assert cursor.fetchone()[0] == 0
+        # The matching value still matches.
+        cursor.execute(f"SELECT COUNT(*) FROM {table_name} WHERE v = ?", [Decimal("12.34")])
+        assert cursor.fetchone()[0] == 1
+    finally:
+        drop_table_if_exists(cursor, table_name)
+        db_connection.commit()
+
+
+def test_gh740_numeric_param_not_in_first_position(cursor, db_connection):
```
jahnvi480, line 680: These use a `NULL` as the earlier param. Since the old bug also clobbered whatever was bound at **record 1**, a stronger guard would put a non-null value first (e.g. `(some_int_or_str, Decimal(...))`) and assert that first value round-trips intact - that pins the *collateral* corruption of position 1, not just the numeric's own misplacement.

mssql_python/pybind/param_detect.hpp
```diff
@@ -511,45 +512,11 @@ inline std::vector<ParamInfo> DetectParamTypes(PyObject* params) {
...
+            // regardless of value. The previous MONEY/SMALLMONEY VARCHAR shortcut chose the
+            // bind type from the value alone, ignoring the target column, so an in-range value
+            // compared against a smaller numeric column triggered a server-side varchar->numeric
+            // overflow instead of simply not matching (GH-740).
             info.paramSQLType = SQL_NUMERIC;
```
sumitmsft, line 520: This scopes the fix to the **native** execute path. When `setinputsizes()` is active but supplies fewer entries than there are parameters (which only raises a warning, not an error), any `Decimal` past the end of the list falls through to the legacy `_map_sql_type` (`cursor.py:1313-1316`), which still binds MONEY/SMALLMONEY-range Decimals as VARCHAR (`cursor.py:796-818`) — so GH-740 can still reproduce there.

Example:
```python
cur.setinputsizes([(SQL_INTEGER, 0, 0)])              # 1 entry
cur.execute("SELECT ... WHERE v = ?", [5, Decimal("12345.6789")])  # 2 params
# the Decimal at index 1 is uncovered -> _map_sql_type -> bound as VARCHAR -> overflow
```

Can we close this so the fix holds for every `execute()` path? e.g. make uncovered execute-time Decimals use numeric binding in the legacy path too (without changing `executemany`'s GH-503 string binding), and add a partial-`setinputsizes` regression test. If it's intentionally out of scope, please add a short note here and in the PR description and file a follow-up so it's explicit.

tests/test_020_money_smallmoney.py
```diff
@@ -4,9 +4,10 @@
 Validates that Python Decimal values are correctly bound and round-tripped
 through MONEY, SMALLMONEY, and DECIMAL columns with proper precision handling.
 
-Key implementation detail: MONEY-range Decimals use string binding (SQL_VARCHAR)
-because SQL_NUMERIC binding fails with ODBC "Numeric value out of range" error.
-String binding preserves full precision and SQL Server handles conversion.
+Key implementation detail: every finite Decimal binds as SQL_NUMERIC using its own
+precision and scale, regardless of value. Binding no longer depends on whether the
+value falls in the MONEY/SMALLMONEY range, so an in-range value compared against a
+smaller numeric column returns no match instead of a varchar->numeric overflow (GH-740).
```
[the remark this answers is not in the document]
bewithgaurav: fixed, the docstring now scopes the always-numeric binding to the `execute()` path and calls out that `executemany` still string-binds Decimals (GH-503)

mssql_python/pybind/param_detect.hpp
bewithgaurav: good call, I mirrored it on the legacy `execute()` path: a `decimal_as_numeric` flag makes the un-sized params bind numeric like the native path, so a partial `setinputsizes()` no longer overflows, with a regression test (`test_setinputsizes_shorter_than_params_money_decimal_binds_numeric`)

`executemany` keeps its batch string binding (GH-503) and is tracked in #745, since binding it numeric needs batch-wide precision/scale, not just dropping the shortcut

tests/test_020_money_smallmoney.py
bewithgaurav: fixed

tests/test_020_money_smallmoney.py
bewithgaurav: done

mssql_python/pybind/param_detect.hpp
bewithgaurav: closed it in the PR rather than scoping it out

`executemany` stays on its GH-503 batch string binding untouched - that path is genuinely heavier (needs batch-wide precision/scale derivation), so I filed it as #745 and noted it in the PR description

jahnvi480:
This is a **behavioral change** on the wire (money-range Decimals now bind NUMERIC, not VARCHAR) - bare `SELECT ?` returns `Decimal` not `str`, `sql_variant` stores `numeric`, and since NUMERIC outranks MONEY/VARCHAR in type precedence, `WHERE money_or_varchar_col = ?` can flip an index seek to a `CONVERT_IMPLICIT` scan. Worth a CHANGELOG entry.

tests/test_020_money_smallmoney.py
```diff
@@ -640,3 +643,125 @@ def test_both_null(cursor, db_connection):
...
+            cursor.execute(f"INSERT INTO {table_name} VALUES (?)", [value])
+            db_connection.commit()
+            cursor.execute(f"SELECT v FROM {table_name}")
+            assert cursor.fetchone()[0] == value
+        finally:
+            drop_table_if_exists(cursor, table_name)
+            db_connection.commit()
+
+
+def test_gh740_same_statement_changing_precision(cursor, db_connection):
```
jahnvi480, line 752: Two additional tests would harden this: (a) an exact `numeric(38,38)` round-trip of `Decimal("1E-38")` — the existing extreme-scale coverage elsewhere uses a tolerance loose enough that a silent `0` would pass; and (b) `Decimal("-0")`/`-0.00`, since NUMERIC normalizes signed zero where the old VARCHAR path preserved the text. Both are cheap and pin behavior the switch actually changed

sumitmsft (dismissed):

The earlier finding on the legacy execute path (an uncovered money-range Decimal binding as VARCHAR when setinputsizes is shorter than the parameter list) is fully addressed here via `decimal_as_numeric=True`, with a dedicated regression test in tests/test_023_execute_path_parity.py. The executemany residual is disclosed in the description and tracked as a separate follow-up (#745).

**Recommendation:** Approve

mssql_python/pybind/param_detect.hpp
sumitmsft: Confirmed. Verified at head 1fcda2ce: the legacy execute path now passes `decimal_as_numeric=True` (cursor.py:1850), so an uncovered money-range Decimal with a partial `setinputsizes` binds as SQL_NUMERIC instead of the VARCHAR shortcut, and the new `test_setinputsizes_shorter_than_params_money_decimal_binds_numeric` in test_023 covers it. Explicit-`setinputsizes` Decimals still string-bind (GH-503 preserved), and executemany is correctly left untouched and tracked as #745. Resolving.

jahnvi480 (approved):

sumitmsft (approved):

PR #736, opened 2026-09-01
PERF: route setinputsizes through the native C++ execute pipeline

bewithgaurav:
### Work Item / Issue Reference

> GitHub Issue: #500

-------------------------------------------------------------------
### Summary

move `setinputsizes` overrides into the native detect, bind, and execute path and remove `DDBCSQLExecuteLegacy`.

-------------------------------------------------------------------
#### macOS arm64, Python 3.13, SQL Server 2022, release build

| Path | Median | Improvement |
|------|--------|-------------|
| `main` legacy path | 2.205s | |
| this PR native path | 1.686s | **24% faster** |

the issue #500 workload runs 100 executions with 2,000 declared parameters each (1,000 `INT`/`VARCHAR` rows), calling `setinputsizes()` before every execute. results are the median of three runs and were confirmed with the base/PR execution order reversed.

#### Cross-platform CI

| Platform | `main` | This PR | Raw improvement |
|----------|--------|---------|-----------------|
| Windows SQL2022 | 4.0770s | 2.8424s | **30.3% faster** |
| Windows SQL2025 | 3.8320s | 3.1110s | **18.8% faster** |
| macOS SQL2022 | 18.0371s | 9.0534s | **49.8% faster** |
| macOS SQL2025 | 9.6811s | 7.1356s | **26.3% faster** |
| Linux SQL2022 | 2.7965s | 2.3523s | **15.9% faster** |

`main` results are from [build 171915](https://dev.azure.com/sqlclientdrivers/public/_build/results?buildId=171915). PR results are from [build 172065](https://dev.azure.com/sqlclientdrivers/public/_build/results?buildId=172065), except macOS SQL2022, which is from the last completed run in [build 171939](https://dev.azure.com/sqlclientdrivers/public/_build/results?buildId=171939); the rerun reached the 20-minute benchmark timeout before that scenario completed.

these are raw medians. the existing pyodbc-normalized score is not used here because the Windows pyodbc main baseline did not reproduce across hosted runners.

PR #743, opened 2026-09-02
FEAT: add setinputsizes performance benchmark

bewithgaurav:

> ADO Work Item: Fixed [AB#47799](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47799)

### Summary

add a dedicated `setinputsizes` scenario to the existing 100K-row insertmanyvalues benchmark. both mssql-python and pyodbc declare all 2,000 parameter types before every execute, so main publishes a normalized baseline for #736 and later binding changes.

the existing insertmanyvalues scenario remains the no-`setinputsizes` control.

### Local Baseline

macOS arm64, Python 3.13, SQL Server 2022, 10 iterations with the first discarded as warmup:

| Driver | Median |
|--------|--------|
| mssql-python | 1.8498s |
| pyodbc | 0.3288s |

normalized score: **5.63x slower than pyodbc**.

### DIFF
--- a/benchmarks/perf-benchmarking.py
+++ b/benchmarks/perf-benchmarking.py
@@ -1,8 +1,9 @@
 """
 Performance Benchmarking: mssql-python vs pyodbc
 
-Runs scenarios (fetch queries + insertmanyvalues), compares mssql-python against pyodbc,
-and optionally against a baseline JSON from the main branch.
+Runs scenarios (fetch queries + insertmanyvalues, with and without setinputsizes),
+compares mssql-python against pyodbc, and optionally against a baseline JSON from
+the main branch.
 
 Usage:
     python benchmarks/perf-benchmarking.py                          # 2-col: PR vs pyodbc
@@ -38,7 +39,7 @@ from typing import List, Optional
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
 
 import pyodbc
-from mssql_python import connect
+from mssql_python import SQL_INTEGER, SQL_VARCHAR, connect
 
 # Configuration
 CONN_STR = os.getenv("DB_CONNECTION_STRING")
@@ -74,7 +75,6 @@ def _init_conn_strings():
         # connection string.  Strip it so both drivers can share one env var.
         parts = [p for p in CONN_STR.split(";") if not p.strip().lower().startswith("driver=")]
         CONN_STR = ";".join(parts)
-
 
 
 class BenchmarkResult:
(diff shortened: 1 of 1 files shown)


jahnvi480 (approved):

benchmarks/perf-benchmarking.py
```diff
@@ -241,6 +249,28 @@ def run_insertmany_mssql(iterations: int) -> BenchmarkResult:
...
+        lambda cs: pyodbc.connect(cs),
+        CONN_STR_PYODBC,
+        "Setinputsizes (100K rows)",
+        iterations,
+        input_sizes * INSERTMANY_BATCH_SIZE,
+    )
+
+
+def run_setinputsizes_mssql(iterations: int) -> BenchmarkResult:
+    input_sizes = [(SQL_INTEGER, 0, 0), (SQL_VARCHAR, 100, 0)]
```
sumitmsft, line 264: Optional (Low): the `input_sizes` template hardcodes the 2-column `(INT, VARCHAR(100))` shape and repeats it `INSERTMANY_BATCH_SIZE` times. It's implicitly coupled to `_build_batch_sql`'s `(?,?)` and the `VARCHAR(100)` column, with nothing enforcing that link — a future change to the batch schema (columns added/resized) would silently desync only this scenario. Consider deriving the per-row type template alongside the table definition, or asserting the expected per-batch param count, so the declarations stay in sync with the table shape. Same applies to the pyodbc runner just above.

sumitmsft (approved):
## Review Summary

**Intent:** Add a dedicated `setinputsizes` scenario to the existing 100K-row insertmanyvalues benchmark, declaring all 2,000 parameter types before every execute for both mssql-python and pyodbc, to publish a normalized baseline for later binding work. The existing no-`setinputsizes` insert scenario remains the control.

**Scope reviewed:** The single changed file `benchmarks/perf-benchmarking.py` and directly related code — the shared `_run_insertmany` runner, `main()` wiring, the `SQL_INTEGER`/`SQL_VARCHAR` top-level exports, and `setinputsizes` semantics in `cursor.py`.

**Assessment:** Clean, self-contained, benchmark-only change. Verified correctness:
- Param alignment: `[INT, VARCHAR] * 1000` = 2,000 declarations, matching the interleaved `(id, val) × 1000` flat params and the `(?,?)×1000` SQL.
- The per-batch `setinputsizes` call is necessary and correct — `execute` resets input sizes afterward (`cursor.py` `_reset_inputsizes()`), so declaring once would only type the first batch. A count mismatch raises `ProgrammingError` rather than silently producing misleading timings.
- pyodbc 3-tuple form `(type, size, scale)` is valid; `size=0` is harmless for fixed-width INT, and VARCHAR is correctly declared as 100.
- Distinct scenario name; no collision with `"Insertmanyvalues (100K rows)"`.

No production, DB-API, native, or packaging surface is affected. All GitHub and ADO cross-platform validation checks are green.

**Overall risk:** Low  
**Design compliance:** Not Applicable (no approved design/spec governs a benchmark script; change is consistent with the file's existing scenario structure)

### Non-blocking notes (benchmark validity)
- **Partial-result robustness:** the shared runner's per-iteration `except Exception` prints and continues, so a run with even one surviving sample still prints `OK` and is serialized. Consider requiring a minimum sample count before treating the scenario as a valid published baseline.
- **Interpretation:** the control-vs-`setinputsizes` delta reflects the end-to-end explicitly-typed binding path, not purely the `setinputsizes()` call overhead, since explicit typing can take a different code path than automatic inference. Fine as a baseline; just not a microbenchmark of the method call alone.

**Recommendation:** Approve with minor (optional) changes.

PR #744, opened 2026-09-03
FEAT: add standalone conda build pipeline + build-validation tooling

jahnvi480:
### Summary

This pull request introduces a new standalone OneBranch pipeline for building, validating, and consolidating self-contained `mssql-python` conda packages across all supported platforms and Python versions. The pipeline is carefully structured to ensure per-platform correctness, best-effort artifact gathering, and robust isolation between build legs. It also includes a template job for consolidating conda artifacts and a `.gitattributes` update for shell scripts.

**New Standalone Conda Build Pipeline**  
*Major features and structure:*

- **Adds `conda-build-pipeline.yml`**: Implements a full multi-stage pipeline that builds and validates self-contained `mssql-python` conda packages for Windows, macOS, and Linux, including cross-platform builds (e.g., win-arm64, osx-arm64, linux-aarch64) using platform-appropriate agents and emulation where needed. Each leg is carefully isolated to prevent failures in one subdir from impacting others, and best-effort logic is used for cross-arch builds.

- **Artifact Consolidation**:  
  - **Adds `consolidate-conda-artifacts-job.yml`**: Provides a job template that collects all per-platform conda packages into a single artifact, preserving the subdir structure. The consolidation is best-effort in the build pipeline, with hard gating deferred to the release pipeline to avoid blocking primary wheel deliverables.

**Build and Release Process Improvements**

- **Best-Effort and Gating Strategy**: The pipeline is designed so that conda packaging failures do not block wheel releases. Completeness and artifact presence are strictly enforced only at release time, not during the build. [[1]](diffhunk://#diff-e258bc0778f9b125cb5214aecba116312126f1a37e3db101606171b05a2b8302R1-R349) [[2]](diffhunk://#diff-6a4150f95dfbb94f92e9ec3f674414deb459b23077ac34a99deb40b0479aee29R1-R129)

**Cross-Platform and Cross-Arch Handling**

- **Cross-Build Logic**: Implements cross-building for win-arm64, osx-arm64, and linux-aarch64 using CONDA_SUBDIR and QEMU emulation as appropriate, with architecture-specific validation and clear blast-radius isolation to prevent a single platform???????????????s failure from impacting others.

**Repository Hygiene**

- **Shell Script Line Endings**:  
  - Adds `OneBranchPipelines/scripts/.gitattributes` enforcing LF line endings for `.sh` files under that folder (the conda recipe scripts are covered separately by `conda/.gitattributes`).

**Build-Validation Tooling + Tests**
*The "build-validation tooling" named in the title -- the exact set of files in this slice:*

- **`OneBranchPipelines/scripts/build_conda_packages.py`**: the single cross-platform orchestrator (gather wheels -> Miniforge/conda-build -> build the self-contained package -> masking-immune audit -> solve a fresh env + import + driver-load + optional reachability gate -> stage). Replaces the former per-OS PowerShell/bash build scripts.
- **`OneBranchPipelines/steps/conda-build-validate-step.yml` + `-posix.yml`**: the Windows / POSIX step templates that invoke the orchestrator.
- **`eng/scripts/audit_bundled_binaries.py`, `assert_pe_machine.py`, `assert_macho_arch.py`, `_conda_pkg.py`**: masking-immune RUNPATH, ELF/PE/Mach-O architecture, required Linux driver-inventory, and package-reader validation.
- **`requirements.txt`**: declares `zstandard` so `.conda` payload tests execute on Python versions without the stdlib `compression.zstd` backend.
- **`conda/driver_load_probe.py`**: the DB-less driver-load proof.
- **`tests/test_029_bundled_binary_audit.py`, `test_030_pe_machine_assert.py`, `test_033_driver_load_probe.py`, `test_034_conda_verify_cwd.py`, `test_035_conda_macho_assert.py`**: unit coverage for Linux payload completeness/RUNPATH, PE and Mach-O architecture checks, the driver-load probe, neutral-cwd verification, and blocking win-arm64 environment installation.

**Not in this slice (deliberately):** the live `Encrypt=yes` TLS gate (`conda/tls_connect_probe.py` and its `test_028`/`test_031`) is a *release*-pipeline concern -- the build pipeline never enables it (no leg sets `CONDA_TLS_PROBE_CONN`), so it ships with the release slice, not here. The masking-immune static RUNPATH audit already guards the OpenSSL layout at build time.


conda/tls_connect_probe.py
```diff
@@ -0,0 +1,286 @@
...
+    """
+    segments = []
+    buf = ""
+    depth = 0
+    for ch in conn.strip():
+        if ch == "{":
+            depth += 1
+            buf += ch
+        elif ch == "}":
+            depth = max(0, depth - 1)
```
sumitmsft: `_split_top_level` decrements brace depth on every `}`, but the shipped parser (`connection_string_parser.py::_parse_braced_value`) treats `}}` as an escaped literal `}` inside a braced value. A value containing `}}` can therefore be mis-split at an internal `;`, which lets `force_tls` corrupt the connection string and lets `_redact` print a fragment of a braced password verbatim (`<<NO-VALUE:...>>`) to the logs on line 263. Please match the production escaped-brace grammar (treat `}}` as a literal, depth unchanged) and add a `}}` test case.

OneBranchPipelines/conda-build-pipeline.yml
```diff
@@ -0,0 +1,343 @@
...
+              type: linux
+              isCustom: true
+              name: Azure Pipelines
+              vmImage: 'ubuntu-latest'
+            variables:
+              ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
+              # H1: activate the runtime gates in build-conda-packages.sh. The TLS probe
+              # runs when a connection string is supplied; the ldd reachability gate runs
+              # (fail-closed) only when explicitly enabled on a minimal base.
+              CONDA_TLS_PROBE_CONN: ${{ parameters.condaTlsProbeConn }}
```
sumitmsft: `condaTlsProbeConn` is a non-secret runtime parameter, so a connection string entered here (potentially with `UID`/`PWD`) is not masked by ADO and can surface in the run UI/logs. Per the repo's secret-handling convention, please source this from a secret variable / variable group mapped into the step `env:` rather than a plain pipeline parameter.

conda/tls_connect_probe.py
```diff
@@ -0,0 +1,286 @@
...
+            conn.close()
+        except Exception:  # noqa: BLE001 - best-effort cleanup only
+            pass
+    except Exception as exc:  # noqa: BLE001 - deliberately classified below
+        outcome = exc
+
+    if tls_completed(outcome):
+        print("TLS_OK (OpenSSL backend reachable; " + describe(outcome) + ")")
+        return
+    sys.exit("TLS/OPENSSL BACKEND UNREACHABLE: " + describe(outcome))
```
sumitmsft: This exit message (and `describe()` at L108) labels the failure "TLS/OPENSSL BACKEND UNREACHABLE", but the same path is reached on cert/handshake failures where OpenSSL is present. Consider neutral wording (e.g. "TLS handshake did not complete") and reserve the "OpenSSL unreachable" label for the `libssl`/`libcrypto`/`cannot open shared object` markers, so the label matches what actually failed.

conda/driver_load_probe.py
```diff
@@ -0,0 +1,139 @@
...
+            conn.close()
+        except Exception:  # noqa: BLE001 - best-effort cleanup only
+            pass
+    except Exception as exc:  # noqa: BLE001 - deliberately classified below
+        outcome = exc
+
+    if driver_loaded(outcome):
+        print("DRIVER_LOADED (" + describe(outcome) + ")")
+        return
+    sys.exit("DRIVER DID NOT LOAD / wrong arch / missing companion: " + describe(outcome))
```
sumitmsft: "missing companion" implies a separate companion library, but the package is self-contained (it vendors the ODBC payload). Consider dropping "missing companion" from this self-contained probe so the failure label isn't misleading during triage.

OneBranchPipelines/scripts/build-conda-packages.sh
```diff
@@ -0,0 +1,422 @@
...
+  # -> celery/boto3/botocore (~9 MB); see conda-forge/azure-core-feedstock#71.
+  # --strict-channel-priority keeps the freshly built local package authoritative.
+  "$conda" create -y -n "$envName" -c "$localChannel" -c microsoft -c conda-forge --strict-channel-priority --override-channels "python=$py" mssql-python
+  # Whether the freshly built package's Python can EXECUTE on this host.
+  target_runnable=1
+  "$conda" run -n "$envName" python -c "import sys" >/dev/null 2>&1 || target_runnable=0
+  if [ "$target_runnable" = "0" ]; then
+    # The ONLY leg allowed to skip the runtime proof is the osx-arm64 cross-build on
+    # an Intel agent (no reverse Rosetta): the arm64 Python genuinely cannot run here,
+    # and the pipeline's static arm64-slice audit (lipo/otool/file on the arm64 Mach-O
```
sumitmsft: This comment states a "static arm64-slice audit (lipo/otool/file on the arm64 Mach-O payload) stands in" for the skipped osx-arm64 import, but no Mach-O audit is implemented in this PR -- only the Windows PE assert (`assert_pe_machine.py`). Please either add the Mach-O/`e_machine` slice check for osx-arm64 or correct this comment (and the matching one in `conda-build-pipeline.yml`) so it doesn't imply a gate that doesn't exist. As-is, osx-arm64 packages are staged with no arch check.

eng/scripts/assert_pe_machine.py
```diff
@@ -0,0 +1,220 @@
...
+                f"{name}: PE machine {_MACHINES.get(machine, hex(machine))} "
+                f"!= expected {_MACHINES[expected]} for subdir '{subdir}'."
+            )
+        else:
+            print(f"  {subdir}/{os.path.basename(name)}: PE machine={_MACHINES[expected]} OK")
+
+    if native_seen == 0:
+        errors.append(
+            f"{base_name}: no .pyd/.dll found in a '{subdir}' package -- the native binding "
+            f"(ddbc_bindings*.pyd) + the vendored ODBC driver DLLs must be present."
```
sumitmsft, line 125: This docstring/error says the native binding "+ the vendored ODBC driver DLLs must be present", but the check passes on >=1 arch-matching binary -- a win-arm64 package with the binding `.pyd` but missing driver DLLs would pass. Since win-arm64 skips the runtime import, this is the main presence gate; consider asserting the required binary categories (binding + driver DLLs) or softening the docstring to match what's enforced.

eng/scripts/audit_bundled_binaries.py
```diff
@@ -0,0 +1,491 @@
...
+    dirs_with_driver: set = set()
+    dirs_with_inst: set = set()
+    vendored: list[str] = []
+
+    for name, data in _iter_payload_members(path):
+        base = posixpath.basename(name)
+        norm = "/" + name
+        member_dir = posixpath.dirname(name)
+        # Track every driver lib dir (mssql_python_odbc/libs/linux/<distro>/<arch>/lib).
+        if "/libs/linux/" in norm and member_dir.endswith("/lib"):
```
sumitmsft, line 324: The RUNPATH audit only validates *discovered* `libs/linux/<distro>/<arch>/lib` dirs, so a package missing an entire distro payload (e.g. no RHEL) passes silently. Consider asserting the expected distro/arch dir set so a dropped payload is caught, or documenting that distro completeness isn't enforced here.

OneBranchPipelines/scripts/build-conda-packages.sh
```diff
@@ -0,0 +1,422 @@
...
+#
+# Use a DEDICATED env instead of `install -n base`: a pre-installed conda whose base
+# is pinned to a too-new python (the GitHub-hosted runner's Miniconda pins python
+# 3.14, which no conda-build<26 supports) makes a base install UNSOLVABLE. A fresh env
+# lets conda pick a python conda-build<26 supports, independent of the base pin.
+# conda-forge only (--override-channels) avoids the defaults-channel ToS; zstandard
+# rides along so the RUNPATH audit reads .conda metadata from this same env.
+condaBuildEnv="conda_builder"
+echo "=== creating dedicated conda-build env ($condaBuildEnv: conda-build<26) ==="
+"$conda" create -y -n "$condaBuildEnv" -c conda-forge --override-channels "conda-build<26" zstandard
```
sumitmsft: `conda create -n conda_builder` has no pre-remove, unlike the verify envs (`env remove ... || true`) and the `.ps1` (`install -n base`). This is harmless on the hosted legs (each installs an isolated per-leg Miniforge), but it can abort a rerun on a reused/self-hosted agent or where a system `conda` is on `PATH`. Can you confirm the target agents never expose a shared base `conda`? If they might, mirroring the `env remove ... || true` idiom would make reruns robust.

eng/scripts/audit_bundled_binaries.py
```diff
@@ -0,0 +1,491 @@
...
+            )
+    # openssl must be RANGE-pinned for Driver 18 (which supports only the OpenSSL
+    # 1.1/3.0 ABI; conda-forge has begun shipping openssl 4), not merely present.
+    if "openssl" in dep_names:
+        spec = next(
+            (str(d) for d in (index.get("depends") or []) if str(d).split()[:1] == ["openssl"]),
+            "openssl",
+        )
+        constraint = spec[len("openssl") :].strip()
+        if ">=3" not in constraint or "<4" not in constraint:
```
sumitmsft: The openssl range-pin check is a substring test (`">=3" not in ... or "<4" not in ...`). A valid alternate spelling could false-pass or false-fail (e.g. an upper bound written only as `<4.0a0`, or a spec with no upper bound). Can you confirm the recipe's actual `openssl` spec spelling always satisfies this heuristic?

sumitmsft (changes requested):
**Summary of review (automated pre-review, pending maintainer judgment)**

Reviewed the added conda build pipeline and validation tooling for correctness, secret handling, validation-gate coverage, and repository conventions. The tooling is well-structured and high-coverage (ran the 5 new test files: 97 passed, 4 skipped for the optional zstd `.conda` path). All specifics are left as inline comments.

**Requesting changes on two Medium items** (inline): the brace-splitter divergence from the shipped connection-string parser in `tls_connect_probe.py` (connect-string corruption + password fragment in logs), and the non-secret `condaTlsProbeConn` pipeline parameter. The remaining inline comments are Low-severity suggestions and two questions.

**Recommendation:** Request Changes

conda/tls_connect_probe.py
jahnvi480: Done. _split_top_level now treats }} as an escaped literal } (depth unchanged, via a lookahead), matching _parse_braced_value, so a braced value with }} + internal ; isn't mis-split. Added a }} case to test_028.

OneBranchPipelines/conda-build-pipeline.yml
jahnvi480: Fixed - removed the plaintext parameter; CONDA_TLS_PROBE_CONN is now sourced from a secret variable ($(condaTlsProbeConn), defined in a variable group / Key Vault) mapped into the step env. Unset ⇒ probe skips.

conda/tls_connect_probe.py
jahnvi480: Done - neutral wording now (TLS HANDSHAKE DID NOT COMPLETE), and describe() just returns the raw diagnostic; the OpenSSL-specific labels are no longer asserted on cert/handshake failures.

conda/driver_load_probe.py
jahnvi480: Dropped "missing companion" from the label (self-contained package vendors the ODBC payload - there's no separate companion).

OneBranchPipelines/scripts/build-conda-packages.sh
jahnvi480: Corrected both comments - the arm64 slice is trusted from the universal2 wheel tag; no Mach-O audit is claimed. The enforced arch guards are the PE assert (Windows) and the ELF e_machine gate (Linux).

eng/scripts/assert_pe_machine.py
jahnvi480: Tightened - it now asserts the binding (ddbc_bindings*.pyd) and the core driver (msodbcsql18*.dll) are both present, independently (not just ≥1 arch-matching binary). +2 tests (missing-binding, support-DLL-only).

eng/scripts/audit_bundled_binaries.py
jahnvi480: Documented - added a note that the audit validates each discovered libs/linux/<distro>/<arch>/lib dir but does not assert a required distro set; the repackaged wheel is the source of truth for which distro payloads exist.

OneBranchPipelines/scripts/build-conda-packages.sh
jahnvi480: Added "$conda" env remove -y -n "$condaBuildEnv" 2>/dev/null || true before the create, so a reused/self-hosted agent (or a system conda on PATH) reruns cleanly. Matches the verify-env / .ps1 idiom.

eng/scripts/audit_bundled_binaries.py
jahnvi480: The recipe pins exactly openssl >=3,<4, so it satisfied the old check - but you're right it's brittle, so I replaced the substring test with a proper operator+version parse (_openssl_range_ok): >=3,<40 now correctly fails, >=3,<4.0a0 passes, and a bare >=3 (no upper) fails.

bewithgaurav (changes requested):
one thought on the two build scripts (build-conda-packages.ps1 + .sh).

they're the same pipeline written twice:
- same 7 steps, same 7 args
- one in powershell, one in bash
- ~440 lines of parallel logic, plus the same comments duplicated on both sides

the reason i'd push on it: they've already drifted.
- step 2 installs conda-build<26 into a dedicated env in the .sh, but into base in the .ps1
- the .sh comment literally explains that base is the thing that goes unsolvable on a 3.14-pinned runner
- so the ps1 is doing what the sh warns against. that's the gap
- two files means every change is a two-file change, and this one already diverged

also a big chunk of the ps1 complexity only exists because the yaml dot-sources it with `&`:
- the ErrorActionPreference flip, the `2>$null`, the `cmd /c "exit 0"` reset at the end
- if the step just ran `python build_conda_packages.py ...` as a normal process, it reads the exit code and all of that goes away

suggestion: one build_conda_packages.py orchestrator.
- conda is python and every agent already has a bootstrap interpreter, so it runs everywhere
- the platform-specific bits (installer invocation, qemu vs ToS env, the linux-only reachability + tls gates) are a handful of if branches, not a second 360-line script
- both step templates collapse to one `python ...` call too

on effort: this is small.
- the scripts are short and well scoped, and the logic already exists
- it's collapsing two files into one, not writing new pipeline
- only real work is one ADO run to confirm the legs stay green

OneBranchPipelines/conda-build-pipeline.yml
```diff
@@ -0,0 +1,350 @@
...
+            variables:
+              ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
+              # H1: activate the runtime gates in build-conda-packages.sh. The TLS probe's
+              # Encrypt=yes connection string carries a password, so it is sourced from a
+              # SECRET variable -- NOT a queue-time parameter, which ADO leaves UNMASKED in
+              # the run UI/logs. Define `condaTlsProbeConn` as a secret in a variable group /
+              # Key Vault linked to this pipeline; unset = the probe SKIPs loudly. The ldd
+              # reachability gate and the MANDATORY-TLS gate (CONDA_TLS_PROBE_REQUIRED) run
+              # (fail-closed) only when explicitly enabled on a minimal base.
+              CONDA_TLS_PROBE_CONN: $(condaTlsProbeConn)
```
sumitmsft: > **Follow-up to my earlier review comment on `OneBranchPipelines/conda-build-pipeline.yml:268`** (now outdated on this revision), which asked to source the TLS-probe connection string from a secret rather than a plain parameter. Thanks for moving it to `$(condaTlsProbeConn)` — this flags the **remaining** wiring gap so the secret actually reaches the step.

**[Medium] TLS-probe secret is mapped at job scope and never reaches the posix step's `bash:` process**

`CONDA_TLS_PROBE_CONN: $(condaTlsProbeConn)` is defined under the job's `variables:` block, but the posix validate step (`steps/conda-build-validate-step-posix.yml`) runs a bare `bash:` script with no step-level `env:`. Azure DevOps does **not** auto-inject *secret* variables into a script's process environment via `variables:` mapping — secrets must be surfaced explicitly through a step-level `env:` block (see this repo's own `steps/symbol-publishing-step.yml`, which does exactly that). As written, `CONDA_TLS_PROBE_CONN` is empty inside the script, so the TLS reachability / mandatory-TLS gate either silently SKIPs (default) or fails with `TLS_PROBE_REQUIRED_BUT_UNSET` when `CONDA_TLS_PROBE_REQUIRED=1` — the gate never actually runs against the intended connection.

**Suggested fix:** add an opt-in `tlsProbeConn` parameter to `steps/conda-build-validate-step-posix.yml`, map it inside a step-level `env:` block (`CONDA_TLS_PROBE_CONN: ${{ parameters.tlsProbeConn }}`), and pass `$(condaTlsProbeConn)` only from the linux-64 call site — dropping the job-level `variables:` mapping of the secret.

conda/tls_connect_probe.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+    segments = []
+    buf = ""
+    depth = 0
+    s = conn.strip()
+    i = 0
+    while i < len(s):
+        ch = s[i]
+        if ch == "{":
+            depth += 1
```
sumitmsft: > **Follow-up to my earlier review comment on `conda/tls_connect_probe.py:127`** (now outdated on this revision), which flagged `_split_top_level` diverging from the production parser on `}}`. Thanks for fixing the `}}` escaped-literal case — this flags the **remaining** divergence, on an inner `{`.

**[Low] `_split_top_level` treats an inner `{` as nesting, diverging from the production parser**

This splitter increments `depth` on every `{`, but the production connection-string parser (`mssql_python/connection_string_parser.py`, `_parse_braced_value`) treats a `{` that appears **inside** an already-open braced value as a literal character — a braced value is single-level and ends only at the next `}`. For a password like `Pwd={a{b};Encrypt=no`, this probe stays at depth 2 after `{a{b`, so `;Encrypt=no` is absorbed *into* the Pwd segment rather than recognized as a separate keyword. `force_tls` then keeps the user's `Encrypt=no` and appends `Encrypt=yes`, producing a duplicate `Encrypt`; when `mssql_python.connect()` parses that, it raises `Duplicate keyword 'encrypt' found` and the gate rejects a legitimate config. (No secret leak — the parser error embeds only the key name.)

**Suggested fix:** model a single-level braced-value state (a boolean "in braced value") instead of a depth counter, matching `_parse_braced_value`.

eng/scripts/audit_bundled_binaries.py
```diff
@@ -0,0 +1,475 @@
...
+    has_upper_4 = False
+    for clause in constraint.split(","):
+        m = re.match(r"\s*(>=|<=|==|=|>|<)\s*([0-9][0-9.]*)", clause)
+        if not m:
+            continue
+        op, ver = m.group(1), m.group(2)
+        major = int(ver.split(".")[0])
+        if op in (">=", ">", "==", "=") and major == 3:
+            has_lower_3 = True
+        elif op in ("<", "<=") and major == 4:
```
sumitmsft: > **Follow-up to my earlier review comment on `eng/scripts/audit_bundled_binaries.py:311`** (now outdated on this revision), which flagged the openssl range-pin substring test. Thanks for rewriting it into clause parsing — this flags a **residual** leniency in the new `_openssl_range_ok`.

**[Low] `_openssl_range_ok` accepts upper bounds that still admit OpenSSL 4.x**

The upper-bound check keys only on `op in ("<", "<=") and major == 4`, so `<=4`, `<4.1`, and `<4.0.1` all satisfy `has_upper_4` — yet each admits some OpenSSL 4.x build, which is outside the Driver-18 ABI range this audit is meant to enforce. The recipe currently pins the canonical `<4.0a0`, so this is latent today, but the checker is looser than its own docstring claims.

**Suggested fix:** accept only an *exclusive* `<` whose version is `4` or a `4.0`-prefixed pre-release (e.g. `<4.0a0` / `<4`); reject `<=` on major 4 and any `<4.<positive>`.

eng/scripts/audit_bundled_binaries.py
```diff
@@ -0,0 +1,475 @@
...
+    except Exception as exc:  # H2: malformed/unreadable must FAIL, never skip.
+        return [f"{base_name}: unreadable/malformed package metadata ({exc})."]
+
+    subdir = str(index.get("subdir", ""))
+    if not subdir.startswith("linux"):
+        print(f"  SKIP (no Linux ELF payload): {base_name} [subdir={subdir or '?'}]")
+        return []
+
+    # The conda subdir is the arch authority; every vendored ELF must match it.
+    expected_machine = _SUBDIR_MACHINE.get(subdir)
```
sumitmsft, line 276: **[Low] Unknown Linux subdir silently skips the ELF machine (arch) gate**

`_SUBDIR_MACHINE.get(subdir)` returns `None` for any `linux-*` subdir not in the map (e.g. a future `linux-ppc64le`). Downstream, a `None` `expected_machine` skips the per-ELF architecture assertion, so a package built for an unrecognized target passes the arch gate. The RUNPATH / deps / lib-dir checks still run, but the architecture check — the one this function exists to enforce — is silently bypassed.

**Suggested fix:** if `subdir.startswith("linux")` but is not in `_SUBDIR_MACHINE`, emit a violation (fail closed) rather than proceeding with `expected_machine = None`.

eng/scripts/_conda_pkg.py
```diff
@@ -0,0 +1,88 @@
...
+def iter_payload_members(path: str):
+    """Yield ``(member_name, data_bytes)`` for the files in a ``.conda`` / ``.tar.bz2`` payload."""
+    if path.endswith(".conda"):
+        with zipfile.ZipFile(path) as zf:
+            pkg_name = next(
+                (n for n in zf.namelist() if n.startswith("pkg-") and n.endswith(".tar.zst")),
+                None,
+            )
+            if pkg_name is None:
+                return
```
sumitmsft: **[Low] `iter_payload_members` silently yields nothing for a `.conda` missing its `pkg-*.tar.zst`**

If a `.conda` archive contains no `pkg-*.tar.zst` member, the function `return`s with zero yields instead of signaling the malformed input. End-to-end this is currently caught by downstream presence assertions, but a silent empty iteration is a fragile contract — a future caller that only checks "did any member match" would treat a corrupt package as clean.

**Suggested fix:** `raise ValueError(f"{path}: no pkg-*.tar.zst payload found in .conda archive")` instead of a bare `return`.

sumitmsft (changes requested):
### Review Summary
- **PR intent:** Add a standalone conda build pipeline plus build-validation tooling (PE/ELF architecture asserts, bundled-binary audit, and TLS / driver-load reachability probes) for the mssql-python conda packages.
- **Scope reviewed:** The full PR diff at head SHA `ae9e4a64` — the OneBranch conda pipeline YAML + step/job templates, the build scripts (`build-conda-packages.ps1` / `.sh`), and the Python validation tooling under `conda/` and `eng/scripts/`.
- **Overall assessment:** The build/validation design is careful and largely fail-closed. One Medium pipeline-wiring defect prevents the TLS gate from ever receiving its secret connection string, plus four Low-severity robustness gaps in the audit/probe tooling.
- **Overall risk:** Medium
- **Proposed decision:** Request Changes
- **Design compliance:** Not Applicable (no formal design/ADR governs this tooling)

**Findings:** 1 Medium, 4 Low — all posted inline.

**Recommendation:** Request Changes. The Medium (job-level secret never reaches the posix `bash:` step) should be fixed so the TLS gate actually runs against the intended connection; the four Low items are robustness hardening of the audit/probe tooling.

OneBranchPipelines/conda-build-pipeline.yml
jahnvi480: Resolved by removing the live Encrypt=yes TLS gate from the build pipeline entirely — it needs a reachable server + secret conn (release infra), so it moves to the release pipeline where the secret is always present and the gate is unconditional. The job-level variables: mapping is gone.

conda/tls_connect_probe.py
jahnvi480: Fixed — rewrote it to match _parse_braced_value exactly: single-level braced values (inner { literal), and a { only opens a brace at a value's start. Added tests for the inner-{, mid-value-{, and force_tls duplicate-Encrypt cases.

eng/scripts/audit_bundled_binaries.py
jahnvi480: Tightened — the upper bound is now accepted only as an exclusive < at 4.0(.0…); <=4, <4.1, <4.0.1 are rejected, <4.0a0 still passes. Tests added.

eng/scripts/audit_bundled_binaries.py
jahnvi480: Now fails closed — an unrecognized linux-* subdir with no _SUBDIR_MACHINE entry emits a violation instead of skipping the arch gate.

eng/scripts/_conda_pkg.py
jahnvi480: Now raises ValueError on a .conda missing its pkg-*.tar.zst; both audit scripts turn that into a clean violation.

OneBranchPipelines/scripts/build_conda_packages.py
```diff
@@ -0,0 +1,772 @@
...
+        _die(f"RUNPATH audit script not found at {audit}")
+    _log("=== RUNPATH self-containment audit (eng/scripts/audit_bundled_binaries.py) ===")
+    run(
+        [conda, "run", "-n", builder, "python", audit, "--root", bld],
+        env=env,
+        what="RUNPATH self-containment audit",
+    )
+    # win-arm64 is cross-built on x64 where its runtime import is skipped, so its arch is
+    # trusted from the wheel filename UNLESS the PE machine assert reads it out of the payload.
+    if target_subdir == "win-arm64":
```
sumitmsft, line 338: `audit_packages` runs the PE-machine assert only for `win-arm64`. `osx-arm64` is cross-built on the Intel agent so its runtime import auto-skips (verify lines 504-511), and — as the pipeline comment at conda-build-pipeline.yml:216-218 notes — its arch is trusted purely from the `universal2` filename tag with no Mach-O check. A payload missing its arm64 slice would build, solve, skip execution, and stage. This is Low (documented, `continueOnError`, build-only, slices validated upstream at wheel build), but consider closing the asymmetry with `win-arm64` by adding a static Mach-O slice assert (parse the fat header / `lipo -archs`) for the osx legs.

eng/scripts/_conda_pkg.py
```diff
@@ -0,0 +1,88 @@
...
+                raise ValueError(f"{path}: no pkg-*.tar.zst payload found in .conda archive")
+            blob = zstd_decompress(zf.read(pkg_name))
+        with tarfile.open(fileobj=io.BytesIO(blob)) as tf:
+            for m in tf.getmembers():
+                if not m.isfile():
+                    continue
+                f = tf.extractfile(m)
+                if f is not None:
+                    yield m.name, f.read()
+    elif path.endswith(".tar.bz2"):
```
sumitmsft, line 52: `iter_payload_members` has no final `else`, so a path that is neither `.conda` nor `.tar.bz2` yields an empty iterator silently (fail-open), whereas `read_index` (line 88) raises `ValueError("unrecognized conda package extension")` for the same case. It's currently unreachable (the audit globs only `*.conda`/`*.tar.bz2`), so this is latent — but for fail-closed consistency and future callers, consider adding `else: raise ValueError(f"{path}: unrecognized conda package extension")`.

sumitmsft (commented):
## Scheduled Copilot review — conda build pipeline + build-validation tooling

**Recommendation: Approve with Minor Changes** (2 Low, non-blocking). Overall risk: Low. Design compliance: Not Applicable (no formal ADR governs this CI/packaging tooling).

Reviewed the true PR scope (three-dot `origin/main...55e56c43`, 14 conda/build-tooling files) for correctness, security/trust boundaries, supply chain, cross-platform/arch coverage, fail-closed behavior, tests, and repository conventions. All prior-round findings are resolved. The orchestrator is injection-safe (list-form subprocess), SHA256-verifies Miniforge before execution, keeps fail-closed static audits, and pins the conda version to the ESRP-signed wheel filename. Two small non-blocking gaps are noted inline.

### Questions / Needs Verification (non-blocking)
- **win-arm64 real `conda create` failure is treated as infrastructure** (`build_conda_packages.py` ~467-477): any nonzero exit from the real create is logged and skipped after the blocking `--dry-run` solve. Arch is still enforced by the PE-machine assert and solvability by the dry-run, and nothing ships (build-only) — please confirm a genuinely-broken-but-solvable package can't stage undetected on this leg.
- **`gather_wheels` uses `odbc_matches[0]`** (`build_conda_packages.py` ~97) while copying all mssql matches. Per-leg `odbcWheelFilter` values are arch-specific and pip's platform-tag enforcement would reject a wrong-arch pick, so `[0]` is safe in practice — please confirm the odbc drop can only ever contain one wheel matching each filter, else consider asserting `len(odbc_matches) == 1`.
- **Linux audit does not assert a required distro SET** (`audit_bundled_binaries.py` 438-443, documented): a wholesale-missing distro payload passes the audit, and on the emulated `linux-aarch64` leg the driver-probe failure is swallowed — please confirm the wheel-build validation is the intended source of truth for distro-set completeness.

OneBranchPipelines/scripts/build_conda_packages.py
jahnvi480: Added `assert_macho_arch.py` and wired it into both macOS legs before staging.

eng/scripts/_conda_pkg.py
jahnvi480: Added the final `else` raising `ValueError` for unsupported extensions, matching `read_index`’s fail-closed behavior.

jahnvi480:
> ## Scheduled Copilot review — conda build pipeline + build-validation tooling
> **Recommendation: Approve with Minor Changes** (2 Low, non-blocking). Overall risk: Low. Design compliance: Not Applicable (no formal ADR governs this CI/packaging tooling).
> 
> Reviewed the true PR scope (three-dot `origin/main...55e56c43`, 14 conda/build-tooling files) for correctness, security/trust boundaries, supply chain, cross-platform/arch coverage, fail-closed behavior, tests, and repository conventions. All prior-round findings are resolved. The orchestrator is injection-safe (list-form subprocess), SHA256-verifies Miniforge before execution, keeps fail-closed static audits, and pins the conda version to the ESRP-signed wheel filename. Two small non-blocking gaps are noted inline.
> 
> ### Questions / Needs Verification (non-blocking)
> * **win-arm64 real `conda create` failure is treated as infrastructure** (`build_conda_packages.py` ~467-477): any nonzero exit from the real create is logged and skipped after the blocking `--dry-run` solve. Arch is still enforced by the PE-machine assert and solvability by the dry-run, and nothing ships (build-only) — please confirm a genuinely-broken-but-solvable package can't stage undetected on this leg.
> * **`gather_wheels` uses `odbc_matches[0]`** (`build_conda_packages.py` ~97) while copying all mssql matches. Per-leg `odbcWheelFilter` values are arch-specific and pip's platform-tag enforcement would reject a wrong-arch pick, so `[0]` is safe in practice — please confirm the odbc drop can only ever contain one wheel matching each filter, else consider asserting `len(odbc_matches) == 1`.
> * **Linux audit does not assert a required distro SET** (`audit_bundled_binaries.py` 438-443, documented): a wholesale-missing distro payload passes the audit, and on the emulated `linux-aarch64` leg the driver-probe failure is swallowed — please confirm the wheel-build validation is the intended source of truth for distro-set completeness.

I've addressed these rather than relying on the upstream assumptions:

1. **Windows ARM64 installation failure:** The real `conda create` now fails the leg on any nonzero exit, even after a successful dry-run solve. It no longer classifies every installation failure as infrastructure, and failed installation cannot reach staging. Added a regression test. Runtime execution can still be skipped when the target Python cannot run on the x64 host.

2. **ODBC wheel selection:** `gather_wheels` now requires exactly one matching ODBC wheel. Zero or multiple matches fail with a diagnostic instead of silently selecting the first. Added coverage for ambiguous matches.

3. **Linux distro completeness:** The Conda audit now independently requires the shipped driver inventory: `alpine/debian_ubuntu/rhel/suse` for x64 and `alpine/debian_ubuntu/rhel` for ARM64. There is no shipped SUSE ARM64 tree. Missing an entire required tree now fails, in addition to the existing per-directory driver/manager checks. Added a missing-tree regression test.

tests/test_029_bundled_binary_audit.py
```diff
@@ -0,0 +1,627 @@
...
+def test_zstd_decompress_preserves_decompression_errors(monkeypatch):
+    conda_pkg = sys.modules.get("_conda_pkg")
+    assert conda_pkg is not None, "loading the audit module should have imported _conda_pkg"
+
+    class BrokenPayload:
+        @staticmethod
+        def decompress(_raw):
+            raise ValueError("invalid zstd frame")
+
+    fake_backend = type("FakeZstandard", (), {"ZstdDecompressor": BrokenPayload})
```
bewithgaurav, line 638: every python 3.14 validation leg is red on this test, so the pr cannot merge in its current state. the fallback fake is ignored on 3.14 because `compression.zstd` imports first:

```text
3.12: ValueError: invalid zstd frame
3.13: ValueError: invalid zstd frame
3.14: compression.zstd.ZstdError: Unknown frame descriptor
```

force the `compression` import to fail in this test before installing the fake `zstandard` module. that exercises the fallback exception path this test is trying to cover.

bewithgaurav (changes requested):
went through the conda build pipeline and validation tooling. the overall approach looks good and the earlier review feedback is addressed. one test issue remains inline, so requesting changes on that.

eng/scripts/audit_bundled_binaries.py
```diff
@@ -0,0 +1,529 @@
...
+
+def audit_package(path: str) -> list[str]:
+    """Return a list of violation strings for one package (empty == clean)."""
+    base_name = os.path.basename(path)
+    try:
+        index = read_index(path)
+    except Exception as exc:  # H2: malformed/unreadable must FAIL, never skip.
+        return [f"{base_name}: unreadable/malformed package metadata ({exc})."]
+
+    subdir = str(index.get("subdir", ""))
```
bewithgaurav: a package whose `info/index.json` exists but omits `subdir` bypasses the entire linux audit.

I reproduced this with `read_index = lambda _: {}`: `audit_package()` prints `SKIP ... [subdir=?]` and returns `[]`. this is the only platform audit invoked without `--subdir`, so no later check catches the missing metadata.

please fail when `subdir` is missing, or pass the expected target subdir into this audit and enforce it.

OneBranchPipelines/conda-build-pipeline.yml
```diff
@@ -0,0 +1,360 @@
...
+                  pythonVersions: '${{ parameters.pythonVersions }}'
+              # osx-64: NATIVE on the Intel agent (BLOCKING -- real import + driver-load proof).
+              - template: /OneBranchPipelines/steps/conda-build-validate-step-posix.yml@self
+                parameters:
+                  condaSubdir: 'osx-64'
+                  condaTargetSubdir: 'osx-64'
+                  odbcWheelDir: '$(Pipeline.Workspace)/odbc_wheels'
+                  odbcWheelFilter: 'mssql_python_odbc-*macosx*universal2.whl'
+                  pythonVersions: '${{ parameters.pythonVersions }}'
+              - task: PublishPipelineArtifact@1
```
bewithgaurav, line 281: partial macOS artifacts can still be lost here.

if osx-arm64 stages successfully and the later blocking osx-64 build fails, this publish task keeps its default `succeeded()` condition and is skipped. `ConsolidateConda` runs under `succeededOrFailed()`, but there is then no macOS artifact for it to download.

please publish the staged output under `succeededOrFailed()` while still excluding cancellation.

bewithgaurav (changes requested):
on a 2nd review, found 2 more suggestions. both are inline.

the earlier feedback is addressed: the build scripts are consolidated, the linux reachability check remains intact, and osx-arm64 now has a static Mach-O check.

the zstd fallback test remains the merge-blocking issue. these two are smaller fail-closed and partial-diagnostics gaps.

tests/test_029_bundled_binary_audit.py
jahnvi480: **Fixed.** The test explicitly blocks `compression` import to exercise the fallback.

eng/scripts/audit_bundled_binaries.py
jahnvi480: **Fixed.** Missing, empty, `None`, and malformed values fail, with regression coverage.

OneBranchPipelines/conda-build-pipeline.yml
jahnvi480: **Fixed.** The macOS publish task now uses `succeededOrFailed()`, excluding cancellation.

bewithgaurav (dismissed):
re-reviewed the latest changes. the zstd fallback test, fail-closed subdir validation, and partial artifact publishing gaps are fixed.

the additional helper error handling is also in place, and all checks are green. approving.

bewithgaurav:
non-blocking I guess - one last thing I noticed before this merges.

this pipeline can pick wheels from any successful branch build, not just `main`. scheduled runs are marked `Official`, so if the newest successful wheel build came from a feature branch, those wheels could be packaged by mistake.

we already have successful non-main runs in pipeline 2199, so this can happen.

can we make `main` the default, and block `Official` runs when the selected wheel build is not from `main`? manual `NonOfficial` runs can still use branch builds for testing.

`official-release-pipeline.yml` has the same gap, so we should fix that separately too.

conda/driver_load_probe.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Deferred so this module can be imported (and ``driver_loaded`` unit-tested)
+    # WITHOUT triggering the native ``mssql_python`` import, which needs the
+    # compiled extension + driver payload.
+    import mssql_python
+
+    # Unreachable endpoint (nothing listens on TCP port 1) -> the driver loads,
+    # attempts the socket, and fails fast at the network stage. The loopback:1 is a
+    # dummy DB-less probe target, never a live endpoint.
+    conn_str = "Server=127.0.0.1,1;Database=x;Uid=x;Pwd=x;Encrypt=no;TrustServerCertificate=yes;"  # DevSkim: ignore DS162092
```
sumitmsft: This probe never reaches the native driver load, so the validation gate is broken. The connection string has a bare `******` token with no `=`. `mssql_python.connect()` -> `Connection.__init__` -> `_construct_connection_string` runs `_ConnectionStringParser(validate_keywords=True)._parse()` before the native `ddbc_bindings.Connection(...)` load, and `_parse` raises `Incomplete specification: keyword '******' has no value (missing '=')` for any token without `=` (this syntax check is not gated by `validate_keywords`). `driver_loaded()` is fail-closed and does not recognize that parser error as a loaded-driver marker, so the probe exits non-zero every run regardless of whether the driver payload is healthy. Please use a valid dummy password attribute (a proper `keyword=value` token, e.g. `PWD=x`) so the probe can actually reach and test driver loading.

eng/scripts/assert_macho_arch.py
```diff
@@ -0,0 +1,270 @@
...
+
+    # Presence gate (mirror the PE assert): osx-arm64 skips the runtime import, so this static
+    # pass IS its arch+presence check. A package with the binding but no driver (or vice versa)
+    # must fail here.
+    if binding_seen == 0:
+        errors.append(
+            f"{base_name}: no native binding (mssql_python/ddbc_bindings*.so) found in a "
+            f"'{subdir}' package."
+        )
+    if target_driver_seen == 0:
```
sumitmsft: This presence gate only fails when no `libmsodbcsql*.dylib` is found: `target_driver_seen` is incremented solely for `base_low.startswith("libmsodbcsql")`, and this check fails only when that counter is zero. Missing load-bearing helper dylibs -- `libodbcinst.2.dylib`, `libodbc.2.dylib`, `libltdl.7.dylib` -- do not fail the audit. Since osx-arm64 skips the runtime import on the Intel build agent, a package with the driver present but helpers missing passes this static gate and can be staged, then fail at `dlopen` on Apple Silicon. Please require the complete expected dylib set under `mssql_python_odbc/libs/macos/<arch>/lib/` and add regression tests that delete each helper while keeping `libmsodbcsql` present.

sumitmsft (commented):
Two validation-gate correctness issues on this revision. Details inline; reasons summarized below.

1. **Driver-load probe never reaches the native driver load (High).** In `conda/driver_load_probe.py` the probe connection string contains a bare `******` token with no `=`. `mssql_python.connect()` runs `Connection.__init__` -> `_construct_connection_string` -> `_ConnectionStringParser(validate_keywords=True)._parse()` *before* the native `ddbc_bindings.Connection(...)` load. `_parse` raises `Incomplete specification: keyword '******' has no value (missing '=')` for any token lacking `=` (that syntax check is not gated behind `validate_keywords`). The probe's own `driver_loaded()` classifier is fail-closed and only returns `True` on a clean connect or a recognized driver-loaded marker, so the parser `ValueError` is classified as not-loaded and the gate exits non-zero on every run regardless of payload health.

2. **macOS presence gate does not require load-bearing helper dylibs (Medium).** In `eng/scripts/assert_macho_arch.py` the presence counter is incremented only for `libmsodbcsql*.dylib` and the gate fails only when that counter is zero, so a package missing `libodbcinst.2.dylib` / `libodbc.2.dylib` / `libltdl.7.dylib` still passes. Because osx-arm64 skips the runtime import on the Intel agent, such a package passes static validation but can fail at `dlopen` on Apple Silicon.

sumitmsft (approved):
Prior findings are resolved. Targeted conda tests pass.

bewithgaurav (approved):

Issue #682, opened 2026-07-15
FEAT: Built-in configurable connection and transient-fault retry logic


dlevy-msft-sql:
### Summary

Add first-class, configurable retry logic to mssql-python so applications get transient-fault resiliency without hand-rolling their own retry loops. Today the only retry surface is the ODBC-level `ConnectRetryCount`/`ConnectRetryInterval` keywords, which only silently reconnect a *dropped idle* connection. They do **not** retry a `connect()` that fails transiently, nor a *query* that fails with a recoverable error (deadlock victim, lock/query timeout, or Azure SQL throttling such as 40197/40501/49918). Every app has to reimplement this, and most get the backoff, jitter, and "which errors are retriable" classification subtly wrong.

### Motivation

- Parity with the .NET Microsoft.Data.SqlClient configurable retry providers (`SqlRetryLogicBaseProvider`, `SqlConnection.RetryLogicProvider` / `SqlCommand.RetryLogicProvider`) and the Azure SDK retry-policy conventions.
- Azure SQL Database and SQL Managed Instance routinely surface transient errors during failover, scaling, and throttling; robust handling is effectively required for production.
- Reduces copy-pasted, error-prone retry loops and centralizes the transient-error taxonomy in the driver where it can be maintained authoritatively.

### Proposed API (for discussion)

A retry policy object, attachable at the connection level and overridable per cursor/execute:

```python
from mssql_python import RetryPolicy, connect

policy = RetryPolicy(
    max_attempts=3,           # total tries, not extra retries
    backoff="exponential",    # "exponential" | "fixed" | "none"
    base_delay=1.0,           # seconds
    max_delay=30.0,           # cap per wait
    jitter=True,              # full jitter to avoid thundering herd
    # Retriable classification, defaulted by the driver but overridable:
    # connect-scope errors -> fresh connection; query-scope -> same connection
)

conn = connect(conn_str, retry_policy=policy)      # applies to connect() + queries
cursor = conn.cursor(retry_policy=policy)          # optional override
cursor.execute(sql, *params, retry_policy=policy)  # optional per-call override
```

Key behaviors:

- **Two scopes**, because recovery differs: connection-establishment/connection-loss errors need a *fresh* connection; connection-surviving errors (deadlock, query timeout, throttling) retry on the *same* connection.
- **Driver-maintained default retriable set** (transient SQLSTATEs / native error numbers), overridable by the caller.
[...]

Issue #581, opened 2026-05-14
FEAT: Expose SQLSTATE (and native error number) as attributes on exception objects


dlevy-msft-sql:
## Summary

`mssql_python` discards the SQLSTATE code after using it to choose an exception subclass, so callers cannot programmatically classify errors by SQLSTATE. The only stable hook left is substring-matching the English `driver_error` text, which is brittle and forces every consumer to duplicate the driver's internal SQLSTATE → text table.

Please expose SQLSTATE as a public attribute on the exception (and ideally also the native SQL Server error number).

## Why this matters

SQLSTATE is the one stable, standardized, locale-independent error identifier in ODBC. It's exactly what application code needs to decide things like:

- Is this transient? Retry it. (e.g. `HYT00`, `08S01`, `40001`)
- Is this a deadlock specifically vs. some other operational error? (`40001`)
- Is this an Azure SQL throttling response? (engine 40501 under `42000`)
- Is this an auth failure that means "refresh the access token"? (`28000`)

Today none of that is possible without one of two bad options:

1. **Substring-match `driver_error`** against a frozenset of English phrases copied out of [`exceptions.py`](https://github.com/microsoft/mssql-python/blob/main/mssql_python/exceptions.py). This pins user code to internal driver constants, breaks if the strings are ever reworded or localized, and forces every consumer to maintain their own copy of the SQLSTATE-to-phrase mapping.
2. **Regex-scrape `ddbc_error`** for `Error NNNN` to recover the SQL Server engine number. Fragile and explicitly not a stable contract.

Compare to peer drivers:

| Driver | Stable error identifier |
| --- | --- |
| pyodbc | `exc.args[0]` is the 5-char SQLSTATE |
| pymssql | `exc.args[0]` is the SQL Server engine error number |
| mssql-python | Neither; SQLSTATE is consumed and discarded in `sqlstate_to_exception()` |

PEP 249 doesn't standardize this, but every mature driver exposes *something*. `mssql-python` is currently the odd one out.

## Proposed change

In [`mssql_python/exceptions.py`](https://github.com/microsoft/mssql-python/blob/main/mssql_python/exceptions.py):

1. Add `sqlstate` (and optionally `native_error`) as keyword args to `Exception.__init__` and store them on `self`.
2. Update `sqlstate_to_exception(sqlstate, ddbc_error)` to pass `sqlstate=sqlstate` into every constructor call (it already has the value in hand).
[...]

PR #751, opened 2026-09-03
FEAT: add optional RetryPolicy for transient failures on connect() (GH-682)

Om-singhaI:

> GitHub Issue: #682

### Summary

First of two PRs for #682, `connect()` scope only; cursor and `execute()` retry follow separately. Adds `mssql_python.RetryPolicy` and `retry_policy=` on `connect()` / `Connection()`, with the constructor shape from the issue. The issue's `backoff="none"` is spelled `backoff="fixed", base_delay=0` here. Without a policy nothing changes.

### How it works

- The loop wraps only the native connect in `Connection.__init__`, below connection string parsing and any token acquired on the Python side, so every attempt reuses the same inputs.
- The SQLSTATE is read from the `SQLSTATE:XXXXX:message` the C++ layer already throws. `exceptions.py` is untouched, so this does not preempt #581.
- Retriable code with attempts left: a warning line (attempt, SQLSTATE, delay), sleep, retry. Otherwise `_raise_connection_error` runs exactly as today, same exception type, nothing rewrapped. If at least one retry happened, one more warning says which attempt failed last and with what SQLSTATE, so the give up shows in the logs too.
- Default set is the seven transient SQLSTATEs from the [Learn retry page](https://learn.microsoft.com/sql/connect/python/mssql-python/retry-logic): `HYT00 HYT01 08001 08S01 08007 40001 40003`. `08004` stays out, the page lists it under never retry. `retriable_sqlstates=` replaces the set.
- `max_attempts` is total tries including the first, as in the issue. The Learn sample counts retries instead.
- Every invalid setting raises `ValueError`. `base_delay` and `max_delay` top out at 86400 seconds, so a policy that validates can't fail inside `time.sleep` halfway through a retry.
- The stub now declares `token_provider` ahead of `retry_policy` on `connect()` and `Connection()`. It was missing upstream, which put `retry_policy` in its positional slot.

### Out of scope

Azure SQL throttling. Those are engine error numbers, and the native number is dropped in `SQLCheckError_Wrap`. Plumbing it out is #581's territory, so it goes with the second PR.

### Validation

- `tests/test_027_retry_policy.py`: 74 passed, no server. The native constructor is faked as in `test_006_exceptions.py`, and `retry._sleep` / `_random` are patched so delay sequences are asserted exactly. No policy is one attempt and the same `OperationalError`; two transient failures then success is three calls with sleeps `[1.0, 2.0]`; exhaustion keeps the mapped type; `28000`, `08004`, `42000` and a message with no SQLSTATE fail once; a transient failure followed by `28000` or no SQLSTATE logs one retry line and one give up line; an error of another type after a retry, like one from a deferred token factory, keeps its own type and still logs the give up; a `token_provider` token is acquired once across three attempts; bad settings, including huge ints, delays over a day and SQLSTATEs that aren't five ASCII letters or digits, raise `ValueError`; log lines never contain the connection string.
- `tests/test_028_stub_signature_parity.py`: 3 passed. It parses the stub and the runtime with `ast`, needs no native module, and fails if `connect()` or `Connection.__init__` drift in name, order, kind or default.
- `test_006_exceptions.py` server free tests: 20 passed. `black` and `flake8` with the CI flags clean.
- Not run against a live server. The failure path is the unchanged `_raise_connection_error`.


Om-singhaI:
@sumitmsft this is the `connect()` half of #682. Nothing has run on it beyond the CLA check, so I think it needs someone to kick off the pipelines.

The statement scope half is built on top of this branch and I've been holding it back rather than stacking two open PRs on the same issue. Happy to open it as soon as this one lands, or sooner if you'd rather review them together.

One thing I'd flag while you're in here: `max_attempts` counts total tries, so 1 means no retry. The docs page counts retries instead. I went with the issue, but say if you'd rather match the docs and I'll change it.

mssql_python/retry.py
```diff
@@ -0,0 +1,220 @@
...
+        if self.backoff == "exponential":
+            # Double once per failed attempt and stop as soon as the cap is reached, so a large
+            # attempt number can never overflow the way a direct power of two would.
+            doublings = attempt - 1
+            while doublings > 0 and 0.0 < delay < self.max_delay:
+                delay *= 2.0
+                doublings -= 1
+        delay = min(delay, self.max_delay)
+        if self.jitter:
+            delay = min(delay * (0.5 + _random()), self.max_delay)
```
bewithgaurav: suggestion: clients retrying the same outage lose part of the intended spread once the delay reaches `max_delay`. every random draw at or above `0.5` becomes exactly the same wait because of the second cap.

in a controlled sweep of 10,000 draws at a 30-second cap, 5,000 returned exactly 30 seconds. this is a delay calculation result, not a concurrent load measurement.

can we use full jitter over the already-capped delay instead?

```suggestion
            delay *= _random()
```

this deliberately changes the documented behavior and allows shorter waits, including zero. please update the jitter docstrings and assertions together, including a case where the backoff has reached the cap.

mssql_python/mssql_python.pyi
```diff
@@ -273,6 +274,40 @@ class _ArrowReader:
         use_internal_transaction: bool = False,
     ) -> Dict[str, Any]: ...
 
+# Retry Policy for transient failures at connect() time
+class RetryPolicy:
```
bewithgaurav: suggestion: can we re-export the annotated `RetryPolicy` instead of maintaining a second copy of its constructor, properties and methods here?

```python
from .retry import RetryPolicy as RetryPolicy
```

this keeps the public type and read-only properties without duplicating the policy API. the `FrozenSet` import can go once the copied class block is removed.

tests/test_027_retry_policy.py
```diff
@@ -0,0 +1,406 @@
...
+    stdlib_logger = logging.getLogger("mssql_python")
+    previous_level = stdlib_logger.level
+    stdlib_logger.setLevel(logging.WARNING)
+    handler = RecordingHandler()
+    mssql_python.logging.logger.addHandler(handler)
+    try:
+        yield handler
+    finally:
+        mssql_python.logging.logger.removeHandler(handler)
+        stdlib_logger.setLevel(previous_level)
```
bewithgaurav: optional: can we reuse `caplog` here and drop `RecordingHandler` plus the manual log-level restoration?

the driver logger doesn't propagate, so attach `caplog.handler` directly, use `caplog.at_level(logging.WARNING, logger="mssql_python")`, and remove the handler in `finally`. the assertions can use `caplog.records`.

bewithgaurav (changes requested):
this addresses the connection-retry scope. I'd like capped retries to stay spread out before this lands; the other comments are cleanup suggestions. requesting changes.

Om-singhaI:
All three done.

Jitter's full now, `delay *= _random()`. I ran your case before touching it and got 49.7% landing on exactly the cap over 100k draws at 30 seconds, so your 5000 in 10000 holds. Docstrings say the wait can be shorter than `base_delay` and can be zero, and there's a new test that a capped delay never comes back as `max_delay` and doesn't bunch up in any tenth of the range.

The stub imports `RetryPolicy` from `retry` now instead of keeping a copy. `FrozenSet` went with it.

Took the caplog one too. `RecordingHandler` and the manual level restore are gone.

sumitmsft (dismissed):
PR #751: No actionable findings in the reviewed changes. Retries remain opt-in and limited to connection establishment, with bounded attempts and capped jitter. Statements are not replayed.

mssql_python/mssql_python.pyi
```diff
@@ -312,6 +316,7 @@ class Connection:
         attrs_before: Optional[Dict[int, Union[int, str, bytes]]] = None,
         timeout: int = 0,
         native_uuid: Optional[bool] = None,
+        retry_policy: Optional[RetryPolicy] = None,
         **kwargs: Any,
```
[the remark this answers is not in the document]
Om-singhaI: Fixed, added `token_provider` to the stub ahead of `retry_policy` in both `Connection.__init__` and `connect`, and added `test_028_stub_signature_parity.py` so the stub can't drift from the runtime signatures again.

Om-singhaI:
@bewithgaurav @sumitmsft I've addressed everything from the earlier rounds, including the stub order Copilot caught. Could you take another look when you get a chance? And could you approve the workflow runs too? They're stuck on approval since this comes from a fork.

sumitmsft (dismissed):

bewithgaurav (dismissed):
lgtm

Om-singhaI:
Pushed fixes for both of today's comments.

`test_028`: `_signature` now tags `posonlyargs` and `args` separately, so a stub that drops the `/` no longer matches the runtime. Defaults still line up across both groups. The new case compares `def f(a, /, b=1)` with `def f(a, b=1)`: the old helper reported them as matching, the new one doesn't.

`test_027`: the spread test fed `_random` from a seeded `random.Random`, which is what devskim flagged. It now gets an evenly spaced sweep over `[0, 1)`, so the test has no random source and doesn't depend on a seed. The old clamping jitter still fails both assertions under the sweep.

The Debian ARM64 failure is `test_concurrent_connections_with_same_token_provider` in `test_008_auth.py`. It fails inside `conn.close()` with the native connect mocked, and without a `retry_policy` the connect loop makes a single attempt. This branch doesn't touch `close()` or that file, so I don't think it comes from this change, but a rerun would show whether it repeats.

@sumitmsft @bewithgaurav the workflow runs on the new commits are waiting for approval again. Could one of you approve them and `/azp run`?

sumitmsft (dismissed):

bewithgaurav (dismissed):

Om-singhaI:
Merged main again to clear the conflict from #613. `RowMapping` and `RetryPolicy` both go into `__all__`, and nothing else changed. @bewithgaurav the runs on the new head will need approving again. The coverage report on the last head timed out waiting for the Azure build, so it'll need `/azp run` too.

PR #752, opened 2026-09-04
FIX: bind executemany money-range Decimals as SQL_NUMERIC

### LINKED ISSUE #745: Decimal in money range still string-binds on the executemany path
### Describe the bug

#740 fixed money-range `Decimal` binding on both `execute()` routes (the native C++ path and the Python legacy path used when `setinputsizes()` covers fewer positions than parameters). `executemany()` is the one remaining path that still carries the money-range VARCHAR shortcut, so a money-range `Decimal` compared through `executemany` can still overflow:

```python
cur.executemany("UPDATE t SET x = 1 WHERE v = ?", [(Decimal("12345.6789"),)])
# against v numeric(5,2): Arithmetic overflow error converting varchar to data type numeric
```

This is a non-shape in practice (`executemany` runs the statement per row and discards results, so nobody runs a comparison-SELECT through it), and it is not reachable from Django, which is why it is lower priority than #740.

### Why it was scoped out of #740

`executemany` declares one parameter type for the whole batch and lets the server coerce each row's formatted string (GH-503). Switching money-range Decimals to `SQL_NUMERIC` there needs a single `NUMERIC(precision, scale)` that fits every row in the batch, which the code does not compute today — it tracks max string length, not max precision and scale. A naive switch regresses a mixed-sign batch (the GH-557 shape) with a string-truncation error.

### Suggested fix

Add batch-wide precision/scale derivation in `executemany` so it can bind money-range Decimals as `SQL_NUMERIC` (like both `execute()` paths) without breaking the GH-503 batch string binding or the GH-557 mixed-sign column sizing.

### Further technical details

mssql-python version: reproduces on the `main` branch alongside #740.
SQL Server version: SQL Server 2022
Operating system: driver-side, OS-independent

vyrnsynx:

> GitHub Issue: #745

### Summary   
`executemany` auto-detect still used the MONEY/SMALLMONEY-range VARCHAR shortcut after #742 fixed `execute()`. A money-range `Decimal` compared against a smaller numeric column could still overflow on that path.

This change binds auto-detected Decimal columns as `SQL_NUMERIC` with a batch-wide precision/scale (values still go through the existing `SQL_C_CHAR` string conversion). The `setinputsizes` DECIMAL/NUMERIC string path (GH-503) and mixed-sign VARCHAR sizing (GH-557) are left alone.

Fixes #745


mssql_python/cursor.py
```diff
@@ -2663,7 +2716,15 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
...
-                    # Ensure columnSize accommodates the longest string representation
+                    # One NUMERIC(precision, scale) must fit every Decimal in the
+                    # batch (GH-745). Sample-only precision/scale is not enough.
+                    batch_precision, batch_scale = self._batch_decimal_precision_scale(column)
+                    if batch_precision > paraminfo.columnSize:
+                        paraminfo.columnSize = batch_precision
+                    if batch_scale > paraminfo.decimalDigits:
+                        paraminfo.decimalDigits = batch_scale
+                    # Ensure columnSize also accommodates the longest string form
+                    # (mixed-sign batches, GH-557).
```
sumitmsft: For a `SQL_NUMERIC` param, `columnSize` is the numeric **precision** passed to `SQLBindParameter` (ddbc_bindings.cpp, 6th arg), not a string length. The `max_decimal_len > columnSize` override just below counts the decimal point and sign, so it sets precision to `batch_precision + 1/2`. Now that money-range Decimals flow into this NUMERIC block, a full-scale value like `Decimal("1E-38")` or `Decimal("0."+"1"*38)` becomes `NUMERIC(40,38)` and the driver rejects it (max precision 38), a regression from the prior VARCHAR path and inconsistent with the `execute()` fix (GH-740). `batch_precision` already sizes precision correctly; the string-length need is handled separately in the `SQL_VARCHAR` block below. Please drop this override in the NUMERIC branch and add an `executemany` test for `Decimal("1E-38")` into `NUMERIC(38,38)`.

mssql_python/cursor.py
```diff
@@ -2663,7 +2716,15 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
                     ddbc_sql_const.SQL_NUMERIC.value,
                 ):
                     paraminfo.paramCType = ddbc_sql_const.SQL_C_CHAR.value
-                    # Ensure columnSize accommodates the longest string representation
+                    # One NUMERIC(precision, scale) must fit every Decimal in the
+                    # batch (GH-745). Sample-only precision/scale is not enough.
+                    batch_precision, batch_scale = self._batch_decimal_precision_scale(column)
```
sumitmsft, line 2729: `batch_precision` is applied to `columnSize` without a <=38 check. The sample-level guard in `_map_sql_type` doesn't cover the whole batch, so a mixed batch (many integer digits in one row, many fractional in another) can produce `batch_precision > 38` even when each value is individually storable, and the driver then fails with an opaque message. Consider raising the existing "precision too high" `ValueError` when `batch_precision > 38`. Don't clamp, since that would silently truncate.

mssql_python/cursor.py
```diff
@@ -2638,6 +2683,13 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
                 )
                 sample_value, min_val, max_val, max_decimal_len = self._compute_column_type(column)
 
+                # GH-745: auto-detected Decimal columns bind as SQL_NUMERIC (skipping
+                # the money-range VARCHAR shortcut) so a money-range value compared
+                # against a smaller numeric column does not overflow. executemany still
+                # string-binds via SQL_C_CHAR below; setinputsizes DECIMAL stays on the
+                # GH-503 string path above.
+                decimal_as_numeric = isinstance(sample_value, decimal.Decimal)
```
sumitmsft: Deriving `decimal_as_numeric` from just the sample type forces the whole column onto the NUMERIC/string-decimal path. A heterogeneous column (Decimal sample plus a stray non-numeric value) that previously bound as VARCHAR will now raise in the Decimal conversion loop below. Consider gating this on "all non-NULL values are Decimal," or documenting heterogeneous columns as unsupported on this path.

sumitmsft (changes requested):
Reviewed the change that routes `executemany` auto-detected `Decimal` columns to `SQL_NUMERIC` (GH-745). The core approach is sound and the new `_decimal_sql_precision_scale` / `_batch_decimal_precision_scale` helpers correctly derive a batch-wide precision/scale (and fix a latent sample-only sizing gap). Verified the precision/scale math matches `_get_numeric_data`, and `black --check` passes.

One blocking issue: the retained `max_decimal_len` override inside the NUMERIC block sets the NUMERIC **precision** from a formatted string length, which can push precision past 38 and break near-max-precision money-range Decimals that previously worked. Two smaller reliability points below.

**Recommendation:** Request changes.

vyrnsynx:
@sumitmsft Thanks for the review — addressed in the latest commit:

1. Dropped the `max_decimal_len` override inside the NUMERIC branch so `columnSize` stays as numeric precision (not string length).
2. Raise `ValueError` when batch precision exceeds 38 (no clamp).
3. Force `decimal_as_numeric` only when every non-NULL value in the column is a `Decimal`.

Also added unit coverage for near-max precision (`1E-38`), batch precision >38, heterogeneous columns, and non-finite Decimals.

vyrnsynx:
@microsoft-github-policy-service agree

bewithgaurav:
@vyrnsynx - requesting you to please look into the failing tests, let us know if you need any help

mssql_python/cursor.py
```diff
@@ -2576,9 +2637,21 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
                     ddbc_sql_const.SQL_NUMERIC.value,
                 ):
                     paraminfo.paramCType = ddbc_sql_const.SQL_C_CHAR.value
-                    # Ensure columnSize accommodates the longest string representation
-                    if max_decimal_len > paraminfo.columnSize:
-                        paraminfo.columnSize = max_decimal_len
+                    # One NUMERIC(precision, scale) must fit every Decimal in the
+                    # batch (GH-745). Sample-only precision/scale is not enough.
+                    # columnSize is NUMERIC precision for SQLBindParameter, not a
+                    # string buffer length — do not widen it with max_decimal_len.
```
bewithgaurav: this uses sql numeric precision as the array's character-buffer width too. precision only counts digits, but the fixed-point string also needs space for the sign, decimal point, leading zero, and terminator. `Decimal("1E-38")` gets `(38, 38)` here but formats to 40 characters, so `PyStringArrayDataBuffer` rejects it before sql server sees it.

I ran the same cases on this head and the exact merge base. `Decimal("1E-38")` and `[-12.34, 56.78]` both complete on the base; this head raises `RuntimeError: Input string exceeds allowed column size at parameter index 0`. the linux and windows jobs show the same 13 decimal failures too.

can we keep sql precision and scale separate from the `SQL_C_CHAR` buffer stride, sizing the buffer from the longest encoded fixed-point value plus its terminator?

bewithgaurav (changes requested):
requesting changes. the issue comparison still errors, while normal decimal batches that work on the merge base now fail before reaching sql server. I reproduced this locally, and the same regression appears in the linux and windows CI jobs.

vyrnsynx:
@bewithgaurav thanks for catching this — you were right that we were treating SQL numeric precision as the CHAR array width.

Pushed a follow-up that keeps `columnSize` / `decimalDigits` as SQL precision/scale for `SQLBindParameter`, and sizes the `SQL_C_CHAR` array stride separately (`bufferSize`) from the longest fixed-point encoding in the batch (with a precision+3 fallback in the binder). That covers `Decimal("1E-38")` and the mixed-sign short-precision case like `[-12.34, 56.78]`.

Added/updated unit coverage for those, plus an executemany `numeric(38,38)` round-trip test.

mssql_python/cursor.py
```diff
@@ -2500,12 +2548,30 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
...
+                    max_encoded = 0
+                    for row in seq_of_parameters:
+                        value = row[col_index]
+                        if value is None:
+                            continue
+                        try:
+                            if isinstance(value, decimal.Decimal):
+                                encoded = format(value, "f")
+                            else:
+                                encoded = format(decimal.Decimal(str(value)), "f")
```
bewithgaurav: with explicit numeric input sizes, failed conversions can now escape as raw exceptions and expose value-bearing error messages in the traceback, instead of the existing `ValueError` with row/column details.

this sizing pass converts non-`Decimal` inputs before the protected conversion loop. its catch misses `MemoryError` and `RuntimeError`.

the existing `test_setinputsizes_sql_decimal_non_decimal_exception_no_leak` fails here with raw `MemoryError` for `"1e999999999999999999"`. it passes on the exact merge base.

I also reproduced the message exposure with a synthetic marker:

```python
class ExplodingStr:
    def __str__(self):
        raise RuntimeError("synthetic-private-parameter")

cursor.execute("CREATE TABLE #pr752_error_boundary (v DECIMAL(18,2))")
cursor.setinputsizes([(mssql_python.SQL_DECIMAL, 18, 2)])
cursor.executemany(
    "INSERT INTO #pr752_error_boundary VALUES (?)",
    [(ExplodingStr(),)],
)
```

head `977654f` raises `RuntimeError: synthetic-private-parameter`. base `cb0bd6c` raises the sanitized `ValueError` identifying `row 0, column 0`, with the marker absent from the formatted traceback.

please derive the buffer width from text produced through the existing protected conversion path, rather than converting independently here, and cover both error cases.

bewithgaurav (changes requested):
the original decimal buffer issue is fixed. requesting changes for a new regression in how conversion errors are handled, detailed inline.

mssql_python/cursor.py
vyrnsynx: @bewithgaurav thanks — buffer sizing for the setinputsizes DECIMAL/NUMERIC path now goes through the protected conversion path.

The provisional sizing pass only measures values that are already `Decimal` (no independent `str`/`Decimal` conversion there). After the existing protected loop produces fixed-point text, `bufferSize` is widened from that text, so `MemoryError` / `RuntimeError` still become a sanitized `ValueError` with row/column details and no value leakage. `Decimal("1E-38")` buffer sizing is covered as well.

Please re-check when you get a chance.

mssql_python/cursor.py
```diff
@@ -2287,6 +2288,53 @@ def _transpose_rowwise_to_columnwise(
...
+        max_int_digits = 0
+        found = False
+        for value in column:
+            if not isinstance(value, decimal.Decimal):
+                continue
+            # Propagate non-finite errors; do not silently skip them.
+            precision, scale = self._decimal_sql_precision_scale(value)
+            found = True
+            max_scale = max(max_scale, scale)
+            max_int_digits = max(max_int_digits, precision - scale)
```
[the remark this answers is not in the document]
bewithgaurav: confirmed. with auto-detection, a batch containing `Decimal("1000000000000000.00")` and `"20000000000000000"` fails with `DataError` / `Numeric value out of range`, although both fit the destination `DECIMAL(38,14)`.

the same batch succeeds on the merge base. the buffer fits both values now, but `NUMERIC(18,2)` allows only 16 integer digits and the string needs 17. please add this case as regression coverage when addressing this.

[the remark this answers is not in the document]
bewithgaurav: @vyrnsynx - requesting you to please check this one

sumitmsft (commented):
**Remaining item (Low, non-blocking): non-finite handling on the executemany path**

The added `ValueError` guards (and `test_gh745_batch_decimal_rejects_non_finite`, which calls `_batch_decimal_precision_scale` directly) do not cover the real `executemany` entry. `_compute_column_type` runs first (line 2527) and does `v_tuple.exponent < 0` / `sample_tuple.exponent < 0` (lines 2391/2397) on the Decimal, which raises `TypeError: '<' not supported between instances of 'str' and 'int'` for a NaN/Infinity (string exponent) before your `ValueError`. So a NaN in a multi-value Decimal batch surfaces a confusing `TypeError` (order-dependent; a single-value column does reach the `ValueError`).

Consider a finiteness guard in `_compute_column_type` (`not v.is_finite()` raising the same `ValueError`) plus an end-to-end executemany NaN test for both row orders. The underlying comparison is pre-existing.

mssql_python/cursor.py
```diff
@@ -2287,6 +2288,53 @@ def _transpose_rowwise_to_columnwise(
...
+        """Return SQL NUMERIC (precision, scale) for a finite Decimal.
+
+        Matches the precision/scale rules used by _map_sql_type / _get_numeric_data.
+        """
+        decimal_as_tuple = value.as_tuple()
+        digits_tuple = decimal_as_tuple.digits
+        num_digits = len(digits_tuple)
+        exponent = decimal_as_tuple.exponent
+        if isinstance(exponent, str):
+            raise ValueError("Cannot bind non-finite Decimal (NaN/Infinity) as SQL NUMERIC")
```
sumitmsft, line 2304: Non-blocking (Low): this non-finite guard is reached too late on the `executemany` path. `_compute_column_type` runs first (called at line 2527) and does `v_tuple.exponent < 0` / `sample_tuple.exponent < 0` (lines 2391/2397) on the Decimal, which raises `TypeError: '<' not supported between instances of 'str' and 'int'` for a NaN/Infinity (string exponent) before this `ValueError` is hit. So a NaN in a multi-value Decimal batch surfaces a confusing `TypeError` (order-dependent; a single-value column does reach this `ValueError`). `test_gh745_batch_decimal_rejects_non_finite` calls `_batch_decimal_precision_scale` directly, so it does not catch this. Consider a finiteness guard in `_compute_column_type` (`not v.is_finite()` raising the same `ValueError`) plus an end-to-end executemany NaN test for both row orders. The underlying comparison is pre-existing.

mssql_python/cursor.py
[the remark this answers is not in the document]
vyrnsynx: Addressed in 763a8dd.

After the protected DECIMAL/NUMERIC conversion, `columnSize` / `decimalDigits` are now widened from every fixed-point string in the batch (via `Decimal(text)` + `_decimal_sql_precision_scale`), not from Decimals alone. Your mixed case (`Decimal("1000000000000000.00")` + `"20000000000000000"`) becomes `NUMERIC(19,2)` instead of under-sizing to `(18,2)`.

Added unit + e2e coverage: `test_gh745_executemany_mixed_decimal_string_widens_precision` and `test_gh745_executemany_mixed_decimal_string_precision`.

mssql_python/cursor.py
vyrnsynx: Addressed in 763a8dd.

`_compute_column_type` now rejects non-finite Decimals with `is_finite()` before any `exponent < 0` comparison, raising the same `ValueError` as `_decimal_sql_precision_scale`. Added executemany coverage for both row orders (`test_gh745_executemany_rejects_non_finite_both_orders` / `test_gh745_executemany_rejects_nan_both_orders`).

mssql_python/cursor.py
```diff
@@ -2678,6 +2776,54 @@ def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-s
...
+                ptype.bufferSize = max(prior, max_encoded, 1)
+            if found_numeric_text:
+                batch_precision = max(max_int_digits + max_scale, 1)
+                if batch_precision > 38:
+                    raise ValueError(
+                        "Precision of the numeric value is too high. "
+                        "The maximum precision supported by SQL Server is 38, "
+                        f"but got {batch_precision}."
+                    )
+                if batch_precision > ptype.columnSize:
```
sumitmsft: This widening pass also runs for parameters configured through `setinputsizes()`, so it can silently replace the caller's declared precision and scale. For example:

```python
cursor.setinputsizes([(SQL_NUMERIC, 10, 2)])
cursor.executemany("SELECT ?", [(Decimal("1.234"),)])
```

The explicit branch initially creates `NUMERIC(10,2)`, but this code widens `decimalDigits` to `3`, so `SQLBindParameter` receives `NUMERIC(10,3)`. That changes observable server-side conversion and metadata behavior instead of honoring the explicit declaration.

Please track whether each parameter was explicitly configured. Always widen `bufferSize` so the encoded value fits, but only widen `columnSize` and `decimalDigits` for auto-detected parameters. Please also add a regression test confirming that explicit `(10,2)` metadata remains unchanged.

vyrnsynx: Addressed in 107a78c (also merges current `main` and resolves the `cursor.py` conflicts).

The post-conversion widen pass now:
- **always** grows `bufferSize` so the encoded fixed-point text fits
- widens `columnSize` / `decimalDigits` **only for auto-detected** columns

Explicit `setinputsizes` metadata is tracked via `explicit_inputsize_cols` and left unchanged. Added `test_setinputsizes_sql_numeric_keeps_declared_precision_scale` for your `NUMERIC(10,2)` + `Decimal("1.234")` case.

PR #770, opened 2026-09-10
GH-745: bind money-range Decimal batches as SQL_NUMERIC in executemany

### LINKED ISSUE #745: Decimal in money range still string-binds on the executemany path

IMGillusion:
## What
Fixes #745. A `Decimal` batch in the MONEY / SMALLMONEY range passed to `executemany()` still took the sample row's `VARCHAR` money shortcut, so comparing it against a smaller `numeric`/`decimal` column (`WHERE v = ?`) made SQL Server convert each row's `varchar` -> `numeric` and raise an arithmetic overflow — the `executemany` remnant of what #740 fixed on both `execute()` paths.

## How
`Cursor.executemany` now calls a new `_apply_money_batch_declared_type()`: when the auto-detected declaration for a column is the `VARCHAR` money shortcut **and** every non-NULL value in the column is a `Decimal` inside the money range, it re-declares the parameter as `SQL_NUMERIC` with `decimalDigits` = the **batch-wide** maximum scale.

Design notes (per the issue):
- The batch still string-binds (`SQL_C_CHAR`, GH-503) — the existing override below the call keeps that, so no C-binding or per-row conversion changes.
- `decimalDigits` uses the max scale across the *whole batch*, not just the sample. A naive sample-only switch would regress a mixed-scale batch (the GH-557 shape) with a truncation/overflow on a row whose scale exceeds the sample's.
- Only pure-Decimal columns fully inside the money range are affected. Wider values (already `SQL_NUMERIC`), mixed-type columns, all-NULL columns, and `setinputsizes()`-typed columns are untouched.

## Test
New `tests/test_020_money_smallmoney.py::test_money_batch_redeclared_as_numeric` covers the declaration switch: in-range mixed-scale batch -> `SQL_NUMERIC` + max scale; integer-valued -> scale 0; above `MONEY_MAX`, mixed Decimal/str, all-NULL, and already-non-VARCHAR columns all left unchanged.

Changelog entry added under Unreleased / Fixed.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -2371,6 +2371,38 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
 
         return sample_value, None, None, max_decimal_formatted_len
 
+    def _apply_money_batch_declared_type(self, paraminfo, column):
+        """GH-745: re-declare a money-range Decimal column as SQL_NUMERIC.
+
+        The auto-detection sample path takes the VARCHAR money shortcut
+        (see _map_sql_type), but a money-range Decimal compared against a
+        smaller numeric column overflows on the server side when bound as
+        VARCHAR - the same bug GH-740 fixed on the execute() paths. Declare
+        SQL_NUMERIC instead; the existing override below keeps the
+        SQL_C_CHAR string binding (GH-503), and decimalDigits uses the
+        batch-wide max scale so every row fits the declared type (a naive
+        sample-only switch regresses mixed-scale batches, GH-557 shape).
+
+        Returns True if the declaration was changed.
+        """
+        if paraminfo.paramSQLType != ddbc_sql_const.SQL_VARCHAR.value:
+            return False
+        non_nulls = [v for v in column if v is not None]
+        if not non_nulls or not all(isinstance(v, decimal.Decimal) for v in non_nulls):
+            return False
+        if not all(SMALLMONEY_MIN <= v <= MONEY_MAX for v in non_nulls):
+            return False
+        paraminfo.paramSQLType = ddbc_sql_const.SQL_NUMERIC.value
+        paraminfo.decimalDigits = max(
+            (-v.as_tuple().exponent) if v.as_tuple().exponent < 0 else 0 for v in non_nulls
+        )
+        logger.debug(
+            "executemany: money-range batch re-declared as SQL_NUMERIC, "
+            "decimalDigits=%d (GH-745)",
+            paraminfo.decimalDigits,
+        )
+        return True
+
     def executemany(  # pylint: disable=too-many-locals,too-many-branches,too-many-statements
         self,
         operation: str,
@@ -2560,6 +2592,13 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
                     min_val=min_val,
                     max_val=max_val,
                 )
+
+                # GH-745: a money-range Decimal batch declared as SQL_VARCHAR
+                # overflows on the server when compared against a smaller
+                # numeric column (the executemany remnant of GH-740). Re-declare
+                # as SQL_NUMERIC with batch-wide scale; string binding (GH-503)
+                # is preserved by the override below.
+                self._apply_money_batch_declared_type(paraminfo, column)
 
                 # GH-610: all-NULL columns now pass SQL_UNKNOWN_TYPE to C++,
                 # where BindParameterArray resolves the correct type via the
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -2390,7 +2390,13 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         non_nulls = [v for v in column if v is not None]
         if not non_nulls or not all(isinstance(v, decimal.Decimal) for v in non_nulls):
             return False
-        if not all(SMALLMONEY_MIN <= v <= MONEY_MAX for v in non_nulls):
+        # The money shortcut (see _map_sql_type) string-binds any Decimal in
+        # the full MONEY range; SMALLMONEY is a subset of that range. The
+        # lower bound must be MONEY_MIN, not SMALLMONEY_MIN: a value like
+        # Decimal("-300000") is in the MONEY range but below SMALLMONEY_MIN,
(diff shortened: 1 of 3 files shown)


IMGillusion:
Worked through the review.

The range floor is the main one: I had it at SMALLMONEY_MIN, but the money shortcut covers the full MONEY range, so a value like -300000 (in range, just below the smallmoney floor) was being left as VARCHAR and the overflow came back for negative money. It's MONEY_MIN now, with a test for that edge.

On the columnSize precision point: for the money shortcut the values have at most 4 decimal places, so the string length that carries over is never above 38, and the declared precision is never less than the scale. That branch isn't actually reachable for money-range values — happy to set an explicit precision if you'd rather not rely on that.

Also cleaned up the test docstring that was implying the helper avoided the native extension (it doesn't, the Cursor import does).

bewithgaurav:
@IMGillusion - closing this PR since this issue is already been taken care of in #752 - marking this one as duplicate
thanks for your contribution!

mssql_python/cursor.py
```diff
@@ -2371,6 +2371,44 @@ def _compute_column_type(self, column):
...
+            return False
+        # The money shortcut (see _map_sql_type) string-binds any Decimal in
+        # the full MONEY range; SMALLMONEY is a subset of that range. The
+        # lower bound must be MONEY_MIN, not SMALLMONEY_MIN: a value like
+        # Decimal("-300000") is in the MONEY range but below SMALLMONEY_MIN,
+        # and using the smaller floor left such batches as SQL_VARCHAR,
+        # reintroducing the GH-745 overflow on negative money-range values.
+        if not all(MONEY_MIN <= v <= MONEY_MAX for v in non_nulls):
+            return False
+        paraminfo.paramSQLType = ddbc_sql_const.SQL_NUMERIC.value
```
sumitmsft, line 2401: **High — derive numeric precision separately from SQL_C_CHAR buffer capacity.** This changes the declared SQL type to `SQL_NUMERIC`, but leaves `columnSize` as the sample's formatted-string length. `BindParameterArray` passes that field as `SQLBindParameter`'s `ColumnSize`, which ODBC defines as precision for `SQL_NUMERIC`.

For example, `[Decimal("9999"), Decimal("0.0001")]` selects `0.0001` and becomes `NUMERIC(6,4)`, so `9999` no longer fits. `Decimal("1E-38")` passes the existing precision-38 check but becomes invalid `NUMERIC(40,38)` because its formatted text is 40 characters. This is established from the Python inference path and native ODBC arguments; I could not run a live DB reproduction because the cp313-amd64 extension is not built in this worktree.

Please derive a common batch precision as maximum integer digits plus maximum scale, enforce precision <= 38, and carry SQL_C_CHAR text-buffer capacity separately in the native binder. Changing `columnSize` to precision alone would truncate strings containing a sign, decimal point, or leading zero. Add focused live regressions for the two examples above.

tests/test_020_money_smallmoney.py
```diff
@@ -528,6 +529,92 @@ def test_executemany_money_smallmoney(cursor, db_connection):
...
+
+    cur = _bare_cursor()
+    varchar = ddbc.SQL_VARCHAR.value
+    numeric = ddbc.SQL_NUMERIC.value
+
+    # In-range batch, mixed scales -> redeclared NUMERIC with max scale 4
+    p = _ParamInfo(varchar)
+    assert cur._apply_money_batch_declared_type(
+        p, [Decimal("12345.6789"), Decimal("-0.0001"), None]
+    ) is True
```
sumitmsft, line 571: **Low — apply the blocking Black format.** The repository's `black --check --line-length=100 mssql_python/ tests/` gate reformats this assertion and the analogous below-`MONEY_MIN` assertion around line 591.

```suggestion
    assert (
        cur._apply_money_batch_declared_type(p, [Decimal("12345.6789"), Decimal("-0.0001"), None])
        is True
    )
```

sumitmsft (changes requested):
## Review summary

This correctly fixes the `MONEY_MIN` range floor and preserves the `SQL_C_CHAR` batch path, but the new declaration derives only batch scale and reuses character length as ODBC numeric precision. That creates undersized or invalid `NUMERIC(p,s)` declarations and can regress ordinary `executemany()` inserts.

- **Risk:** High
- **Decision:** Request Changes
- **Design compliance:** Partially Compliant with #745; batch-wide precision remains unimplemented.
- **Findings:** 1 blocking correctness defect, 1 formatting suggestion.

The independent rubber-duck pass substantiated the precision/buffer-capacity defect and strengthened it to a plain-insert regression. The proposed non-sample-NaN concern was excluded as pre-existing and unreachable through the new helper, and the earlier `MONEY_MIN` issue is fixed at this HEAD. A live SQL Server test was not run because the required cp313-amd64 native extension is unavailable locally; the blocking Black failure was reproduced locally.

PR #773, opened 2026-09-10
FIX: Correct GetInfoConstants IDs and ODBC result decoding

jahnvi480:

> [AB#48074](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/48074)

### Summary

Correct all 11 misassigned SQLGetInfo information-type IDs and add the standard `SQL_TIMEDATE_FUNCTIONS` spelling at 52, preserving `SQL_DATETIME_FUNCTIONS` as the canonical enum name.

- Retain all 11 misplaced enum attributes and their original values as deprecated compatibility members. Preserve module-level imports and `get_info_constants()` lookups, expose the appropriate `ConstantsDDBC` groups, and add the standard SQL conformance result flags.
- Use one immutable, module-level ODBC return-type registry with exact numeric widths, including pointer-sized handles. Decode unsigned values without guessing; reject wrong-width or truncated buffers. Preserve non-byte floats/bools unchanged and safely convert decimal strings.
- Correct the five known unlisted standard IDs: `SQL_DBMS_NAME` (17), `SQL_DBMS_VER` (18), `SQL_XOPEN_CLI_YEAR` (10000), `SQL_ASYNC_MODE` (10021), and `SQL_CREATE_ASSERTION` (127).
- Add independent reference coverage for 102 non-deprecated enum names/aliases, 11 compatibility names, and five additional raw IDs. Cover exact forwarding, widths, unsigned boundaries, Unicode, legacy payloads, imports/stubs, enum pickles, connection-local results, and live native-payload comparisons. Replace blanket `None` skips with strict assertions and explicit Driver Manager-only exclusions.

**Deprecation-first compatibility policy**

The misplaced enum members remain available throughout **1.x**, with removal deferred to **2.0 or later**, only after maintainer approval and an explicit migration notice. No removal date is scheduled. This revision implements documented deprecation without custom enum-access hooks or runtime warnings.

Existing calls such as `GetInfoConstants.SQL_ROW_NUMBER.value` continue to work. The deprecated `SQL_SQL92_*_SQL` names retain **127/128/129**; new code should use `SQL_SC_SQL92_ENTRY=1`, `SQL_SC_FIPS127_2_TRANSITIONAL=2`, `SQL_SC_SQL92_INTERMEDIATE=4`, and `SQL_SC_SQL92_FULL=8` to interpret conformance results.

Deprecated names remain in `get_info_constants()` for dictionary-lookup compatibility, but are not valid information-type names. Their numeric values still request unrelated information; rejecting those integers would also reject legitimate colliding information types. The explicit decoding registry is independent of these compatibility names.

Correcting the actual information-type IDs intentionally changes incorrect results. Persisted raw IDs and value-based enum pickles for corrected members remain ambiguous and must be rebuilt from the intended names. Name-based enum pickles resolve retained names to their corrected values. Both formats are covered explicitly because the default reducer differs across Python versions. The existing `SQL_DATETIME_FUNCTIONS.name` is preserved.

**Validation and limitations**

- Windows x64, Python 3.13, Microsoft ODBC 18.6.2.1, SQL Server LocalDB: **1,235 focused tests passed**, with three explicit Driver Manager-only skips. All 137 pickle cases also passed with the name-based default emulated locally. Previous full-suite run: **3,629 passed**, 137 skipped, 42 deselected, and the same two previously reproduced baseline environment failures (password-mask expectation under integrated authentication; Windows long-path limit).
- `Connection.getinfo` focused coverage: **50/52 statements (96.2%) and 26/26 branch exits (100%)**, using repository exclusions. The two uncovered statements are the retained defensive exception handler in the unregistered binary fallback; native C++ coverage was not measured.
- Earlier workload checks exercised independent catalog changes, Unicode rows, unchanged transaction state, and 3,200 metadata calls across eight separate connections. The registry revision also passed live raw-ID, catalog-isolation, transaction-state, and Unicode-row scenarios.
- Cross-platform execution and the optional Rust ODBC provider have not been validated locally.

**Scope**

No native-code or dependency changes from this fix, no additional native calls, and no cross-connection metadata cache. Other unregistered raw IDs retain their legacy handling; this is not a universal SQLGetInfo overhaul. Correct IDs do not add support for Driver Manager-only handle queries. Standard Y/N character responses are retained, but nonstandard lowercase/binary responses no longer receive the old normalization.

The native-error contract is deliberately unchanged: unsupported-type, timeout, and connection-loss errors are logged and return `None`. Changing that contract is deferred to separate work. Providers can return cached metadata after connection loss, so **`getinfo()` is not a connection-health check**.


mssql_python/constants.py
```diff
@@ -190,19 +191,37 @@ class ConstantsDDBC(Enum):
...
+    SQL_SQL92_ENTRY_SQL = SQL_SC_SQL92_ENTRY
+    SQL_SQL92_INTERMEDIATE_SQL = SQL_SC_SQL92_INTERMEDIATE
+    SQL_SQL92_FULL_SQL = SQL_SC_SQL92_FULL
+
 
 class GetInfoConstants(Enum):
-    """
-    These constants are used with various methods like getinfo().
-    """
+    """ODBC information-type IDs accepted by Connection.getinfo()."""
```
bewithgaurav, line 221: existing callers using `GetInfoConstants.SQL_ROW_NUMBER.value` now get `AttributeError`. keeping the module-level import does not preserve that usage.

the PR description still says the compatibility decision is pending. please either retain the old enum attributes through a deprecation period or explicitly document the decision and release plan for removing them before merge.

bewithgaurav (changes requested):
the metadata fixes address the reported issue, but existing enum callers still break without an agreed migration approach. requesting changes until that compatibility decision is settled.

mssql_python/constants.py
jahnvi480: Agreed. I’ve chosen a **deprecation-first approach** for this fix. In `5f68257f`, I restored all 11 legacy enum attributes and their original values, preserving existing imports and `get_info_constants()` lookups.

We’ll retain these compatibility members throughout **1.x**. Any removal will be a separately announced breaking change in **2.0 or later**; this PR does not schedule removal.

The ODBC ID and decoding fixes remain. I’ve updated the PR description and changelog to make this decision explicit. Please re-review.

gargsaumya (approved):

bewithgaurav (approved):
lgtm

PR #780, opened 2026-09-11
FIX: Make SQL Server CI setup retry-safe

jahnvi480:

> [AB#47797](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/47797)

### Summary

Replace duplicated Unix SQL-container setup with one shared, ownership-aware helper and one bounded SQL-only retry, including macOS SQL2025, Linux matrices, and coverage. Capture redacted failure evidence before removing the failed container, recreate only job-owned containers, bound readiness/commands/cleanup, and propagate cancellation and permanent failure.

Keep dependency installation, native builds, and pytest outside the retry boundary. Preserve macOS build/setup overlap, optional AzureSQL behavior, existing ARM build retry settings, amd64 SQL containers, and current database/image selections. Isolate final SQL cleanup from test-container cleanup and retain safe image/state/log evidence without publishing dumps or credentials.

The PR changes only the pipeline YAML and shared SQL setup helper. No test files are added or modified. Verification uses temporary development checks kept outside this PR, normal PR validation, and separate hosted Linux/macOS SQL2025 recovery and permanent-failure experiments. The local Windows x64/Python 3.13 native build and Black gate passed. Hosted results and review feedback are being assessed; the PR remains a draft.

This is recovery mitigation for observed SQL setup failures, not a fix for the underlying SQL-engine crash or a guarantee that hosted failures cannot recur.


eng/scripts/setup_sql_container.py
```diff
@@ -0,0 +1,595 @@
...
+            except subprocess.TimeoutExpired:
+                timed_out = True
+            if not timed_out:
+                drain_timed_out = not output_done.wait(
+                    timeout=max(0, work_deadline - time.monotonic())
+                )
+        finally:
+            teardown_deadline = min(deadline, time.monotonic() + 4)
+            if process.poll() is None or not output_done.is_set() or read_errors:
+                reaped = self.stop(process, teardown_deadline)
```
bewithgaurav: macOS SQL2022 is failing before SQL setup in [build 174400](https://dev.azure.com/sqlclientdrivers/public/_build/results?buildId=174400&view=logs&j=e3c5db7b-2dae-59db-aa6e-413d28319b7b). the Colima startup path reaches `stop()` and raises:

```text
os.killpg(process.pid, signal.SIGKILL)
PermissionError: [Errno 1] Operation not permitted
```

I reproduced the same exception locally on macOS with a synthetic daemon-style launcher, without mocking subprocesses or signals:

```python
import sys
from eng.scripts.setup_sql_container import Commands

launcher = (
    "import subprocess,sys;"
    "p=subprocess.Popen([sys.executable,'-c','import time; time.sleep(5)']);"
    "print(p.pid,flush=True)"
)
Commands("").run([sys.executable, "-c", launcher], 2)
```

the launcher exits successfully, but its child keeps stdout open. waiting for EOF sends this through group teardown anyway. the exception then bypasses the captured startup diagnostics.

please separate Colima launcher completion from its background processes and preserve diagnostics when teardown fails. catching `PermissionError` alone still leaves the successful-launcher case rejected. this is new behavior introduced by the helper.

bewithgaurav (dismissed):
this addresses the SQL-only retry scope, but the new startup handling breaks macOS before SQL setup begins. requesting changes to fix that path before merging.

eng/scripts/setup_sql_container.py
```diff
@@ -0,0 +1,595 @@
...
+        drain_timed_out = False
+        reaped = True
+        try:
+            try:
+                reader.start()
+                reader_started = True
+            except RuntimeError:
+                raise SetupFailure("Cannot start setup output reader", retryable=False) from None
+            try:
+                process.wait(timeout=max(0, work_deadline - time.monotonic()))
```
gargsaumya, line 361: macOS setup fails after a successful Colima launcher exit.
The latest macOS SQL2022 validation also fails because process-group teardown raises `PermissionError: [Errno 1] Operation not permitted`. This exception escapes the helper and suppresses the intended bounded diagnostics.

Please handle daemon-style launchers correctly: once the launcher exits successfully, descendants retaining stdout should not cause the command to be treated as failed. Catching PermissionError alone would hide the exception but would not fix that lifecycle issue.

gargsaumya (changes requested):

eng/scripts/setup_sql_container.py
jahnvi480: Fixed this

eng/scripts/setup_sql_container.py
jahnvi480: Resolved this

gargsaumya (dismissed):

gargsaumya (approved):

sumitmsft (approved):

PR #783, opened 2026-09-11
FIX: Prevent late Windows ODBC ENV cleanup during shutdown

jahnvi480:

> [AB#48085](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/48085)

### Summary

Prevent Windows process-exit access violations caused by the static ODBC environment handle calling `SQLFreeHandle` after Python finalization during DLL teardown.

Extend the existing finalization guard to `SQL_HANDLE_ENV` on Windows only. Preserve normal-lifetime handle cleanup, existing STMT/DBC protection, implicit-free checks, and GIL-release behavior. Unix ENV cleanup remains unchanged.

Strengthen the existing shutdown tests to verify query results, pooled and nonpooled explicit cleanup, completion of Python cleanup, and actual subprocess exit codes. No new test files, production pipeline changes, or diagnostic infrastructure are included.

The frozen candidate was validated in hosted diagnostic [174475](https://dev.azure.com/SqlClientDrivers/public/_build/results?buildId=174475): 700 fresh subprocesses on SQL2022/Python3.13.15 and 700 on SQL2025/Python3.14.7, all exiting successfully without access violations or timeouts on Windows Server2025 x64. Those results apply to the candidate based on `89c22515`; normal PR validation of the current-main integration remains pending.

This addresses the demonstrated late-ENV shutdown defect, not every possible Windows access violation.

### DIFF
--- a/mssql_python/pybind/ddbc_bindings.cpp
+++ b/mssql_python/pybind/ddbc_bindings.cpp
@@ -1596,7 +1596,13 @@ void SqlHandle::free() {
         // 2. OS-level cleanup at process termination recovers any remaining resources
         // 3. This tradeoff prioritizes crash prevention over resource cleanup, which
         //    is appropriate since we're already in shutdown sequence
-        if (pythonShuttingDown && (_type == SQL_HANDLE_STMT || _type == SQL_HANDLE_DBC)) {
+        bool skipDuringShutdown = _type == SQL_HANDLE_STMT || _type == SQL_HANDLE_DBC;
+#ifdef _WIN32
+        // The static ENV is destroyed during DLL_PROCESS_DETACH, after Python
+        // finalization. Calling ODBC then can access already-torn-down SSPI state.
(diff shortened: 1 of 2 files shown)


gargsaumya (approved):

subrata-ms (approved):
Below test coverage looks like require - 
1. Medium: Regression test is not a deterministic revert detector
test_013_SqlHandle_free_shutdown.py:313-377 verifies the subprocess exit code, which correctly catches a crash occurring after the PASS marker. However, it runs each pooled/nonpooled scenario only once and cannot prove that the new ENV skip branch executed.

Because the original failure was intermittent, reverting the C++ fix could still leave this test green.

Recommendation: Run the Windows shutdown workload repeatedly, preferably 10–30 times in one subprocess-launch loop, or add test-only native instrumentation that confirms SQLFreeHandle(SQL_HANDLE_ENV) was skipped during finalization.

Issue #784, opened 2026-09-12
bug: replace `strptime()` with ISO parsers for date/time parsing


nos1dot618:
The `_parse_date`, `_parse_datetime`, and `_parse_time` helpers use `datetime.strptime()`, triggering Ruff `DTZ007` due to naive datetime construction.

Replace the manual format loops with the appropriate `fromisoformat()` APIs to:

* Remove the `DTZ007` warnings.
* Simplify the parsing logic.
* Preserve the existing supported date, datetime, and time formats.

Add/update regression tests to verify the existing parsing behavior remains unchanged.

nos1dot618:
After some digging, I found that `date.fromisoformat()` is stricter than the previous `strptime(..., "%Y-%m-%d")` behavior. The previous implementation accepted dates with single-digit month/day components, such as `2024-5-2`, whereas `fromisoformat()` rejects them and returns `None`.

Since this changes previously supported behavior and the goal is to maintain compatibility, I think it would be better to retain the existing functionality by normalizing the components before calling `fromisoformat()` (or retaining a compatibility fallback), along with a regression test for this case.

PR #785, opened 2026-09-12
FIX: replace strptime with fromisoformat for date parsing

nos1dot618:

> GitHub Issue: #784 

### Summary

Replace `datetime.strptime()` with the corresponding `fromisoformat()` APIs for date, datetime, and time parsing in `Cursor`. This removes the Ruff `DTZ007` warnings caused by constructing naive datetimes with `strptime()` while preserving the existing parsing behavior. The existing cursor tests already cover date, datetime, and time parsing, so no new tests are required.

### DIFF
--- a/mssql_python/cursor.py
+++ b/mssql_python/cursor.py
@@ -447,13 +447,10 @@ class Cursor:  # pylint: disable=too-many-instance-attributes,too-many-public-me
         Returns:
             A datetime.date object if parsing is successful, else None.
         """
-        formats = ["%Y-%m-%d"]
-        for fmt in formats:
-            try:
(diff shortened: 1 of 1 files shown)


nos1dot618:
After some digging, I found that `date.fromisoformat()` is stricter than the previous `strptime(..., "%Y-%m-%d")` behavior. The previous implementation accepted dates with single-digit month/day components, such as `2024-5-2`, whereas `fromisoformat()` rejects them and returns `None`.

Since this changes previously supported behavior and the goal is to maintain compatibility, I think it would be better to retain the existing functionality by normalizing the components before calling `fromisoformat()` (or retaining a compatibility fallback), along with a regression test for this case.

PR #791, opened 2026-09-17
FIX: prevent native log format-string injection

sumitmsft:

> [AB#48241](https://sqlclientdrivers.visualstudio.com/c6d89619-62de-46a0-8b46-70b92a84d85e/_workitems/edit/48241)

### Summary

Prevents server-controlled Arrow column metadata from being interpreted as a native printf-style format string.

Converts all dynamic native LOG calls to literal formats, fixes existing format/argument mismatches exposed by compiler checking, and enables compile-time printf validation for GCC, Clang, and AppleClang.

Adds a source-contract regression test that rejects nonliteral native LOG formats across first-party C++ sources and headers.

### Validation

- Windows x64 native extension build completed with 0 errors.
- 37 no-database dependency and logging-security tests passed; 3 platform-specific tests skipped.
- Rebuilt package import passed.
- Python formatting and diff checks passed.
- Independent rubber-duck review completed with no remaining findings.


gargsaumya (approved):

tests/test_039_native_logging_format_security.py
```diff
@@ -0,0 +1,62 @@
...
+        text,
+        flags=re.DOTALL,
+    )
+    return "\n".join(
+        "" if line.lstrip().startswith("#define LOG") else re.sub(r"//.*", "", line)
+        for line in text.splitlines()
+    )
+
+
+def test_native_log_calls_use_literal_format_strings():
```
bewithgaurav, line 30: suggestion: add a subprocess regression with DEBUG logging enabled and `SELECT CAST(1 AS sql_variant) AS [%s%n%p%%]`. assert that it raises the unsupported-type error without crashing and preserves the alias verbatim in the native log.

the source guard is useful, but it doesn't exercise the original failure path.

bewithgaurav (approved):
this addresses the reported format-string issue and keeps the existing error behavior intact. one optional suggestion to strengthen regression coverage. approving.

PR #802, opened 2026-09-21
FIX: Validate caller-controlled native buffer sizes

gargsaumya:

> ADO Work Item: [AB#47843](https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47843)

### Summary

- Reject invalid Python row counts before crossing native boundaries.
- Check parameter allocation arithmetic and metadata cardinality.
- Tighten binary and wide-character input validation.

No tracking issue exists for this security hardening work.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -5841,10 +6130,16 @@ SQLRETURN FetchAll_wrap(SqlHandlePtr StatementHandle, py::list& rows,
...
+    ReserveNativeFetchBytes(
+        reservedBytes,
+        CheckedMultiplySize(static_cast<size_t>(numCols), static_cast<size_t>(fetchSize),
+                            "Native fetch indicator buffer is too large"),
+        sizeof(SQLLEN));
     ColumnBuffers buffers(numCols, fetchSize);
 
     // Bind columns
-    ret = SQLBindColums(hStmt, buffers, columnNames, numCols, fetchSize, charCtype);
+    ret = SQLBindColums(hStmt, buffers, columnNames, numCols, fetchSize, reservedBytes, charCtype);
```
sumitmsft, line 6142: `fetchall()` still chooses its internal batch using the old 1-GiB heuristic. Seventeen `VARCHAR(8000)` columns select 1,000 rows and require 272,170,000 bytes with default wide binding and indicators—above the new 256-MiB limit. Even an empty or one-row result therefore fails before fetching.

Arrow has the same problem with its fixed 64-row native allocation, including schema-only requests. Please derive internal batch sizes from actual native storage and the remaining budget, and skip data-rowset allocation for schema-only requests.

mssql_python/cursor.py
```diff
@@ -1188,20 +1208,34 @@ def setinputsizes(self, sizes: List[Union[int, tuple]]) -> None:
...
 
                     sql_type = ODBC3_TEMPORAL_SQL_TYPES.get(sql_type, sql_type)
 
                     # Validate size and precision
-                    if not isinstance(column_size, int) or column_size < 0:
+                    if (
+                        isinstance(column_size, bool)
+                        or not isinstance(column_size, int)
+                        or column_size < 0
+                        or column_size > MAX_NATIVE_PARAMETER_SIZE
```
sumitmsft, line 1227: This applies a byte-allocation limit to numeric precision metadata before the existing normalization runs. For example, `setinputsizes([(SQL_DECIMAL, 10**100, 10**100)])` previously normalized to supported precision/scale; it now raises immediately, breaking the existing numeric-clamping regression.

Please distinguish numeric precision from buffer length and preserve numeric normalization. Enforce allocation limits against the actual storage required.

tests/test_004_cursor.py
```diff
@@ -1730,6 +1730,59 @@ def test_arraysize(cursor):
...
+def test_setinputsizes_rejects_excessive_column_size():
+    cursor = mssql_python.Cursor.__new__(mssql_python.Cursor)
+    with pytest.raises(ValueError, match="column size"):
+        cursor.setinputsizes([(mssql_python.SQL_VARCHAR, 1 << 40, 0)])
+
+
+def test_executemany_rejects_excessive_cumulative_parameter_buffers(cursor):
+    cursor.setinputsizes([(mssql_python.SQL_WVARCHAR, 50_000_000, 0)] * 3)
+    with pytest.raises(RuntimeError, match="Parameter buffers exceed the 256 MiB allocation limit"):
+        cursor.executemany("SELECT ?, ?, ?", [("", "", "")])
```
sumitmsft, line 1770: With these oversized string hints, the empty-string values select data-at-execution (DAE), not the array-allocation path this test intends to cover. The call raises `Parameter count does not match parameter metadata count` instead of the asserted allocation-limit error.

Use `[(None, None, None)]` with the same metadata to select non-DAE binding. That produces a 300,000,030-byte allocation plan, which the preflight rejects before allocating the large buffers. Keep the specific budget-error assertion.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -322,12 +323,161 @@ ParamType* AllocateParamBuffer(std::vector<std::shared_ptr<void>>& paramBuffers,
...
+        case SQL_C_SS_TIMESTAMPOFFSET:
+            return sizeof(DateTimeOffset);
+        case SQL_C_NUMERIC:
+            return sizeof(SQL_NUMERIC_STRUCT);
+        case SQL_C_GUID:
+            return sizeof(SQLGUID);
+        case SQL_C_DEFAULT:
+            return sizeof(char);
+        default:
+            ThrowStdException("Unsupported C type for parameter array allocation");
```
bewithgaurav, line 416: blocker: this blocks every linux build. the merge CI compiles with warnings as errors and stops here with:

`control reaches end of non-void function [-Werror=return-type]`

`ThrowStdException` is not declared `[[noreturn]]`, so this function still needs a return after the switch, or the throw helper needs that annotation.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -322,12 +323,161 @@ ParamType* AllocateParamBuffer(std::vector<std::shared_ptr<void>>& paramBuffers,
...
+
+size_t CheckedMultiplySize(size_t left, size_t right, const char* errorMessage) {
+    if (left != 0 && right > std::numeric_limits<size_t>::max() / left) {
+        ThrowStdException(errorMessage);
+    }
+    return left * right;
+}
+
+constexpr int MAX_NATIVE_ROW_COUNT = 1000000;
+constexpr size_t MAX_NATIVE_FETCH_BYTES = 256ULL * 1024 * 1024;
```
bewithgaurav, line 350: blocker: valid wide result sets now fail before the first row is fetched. 33 `NVARCHAR(4000)` columns pass, while 34 throw `Native fetch buffers exceed the 256 MiB allocation limit`; both pass on `main`.

`FetchAll_wrap` still chooses 1,000 rows against its 1 GiB heuristic, which reserves 272,340,000 bytes for this shape and collides with the new 256 MiB cap. derive the batch size from this same budget, or shrink and retry before binding.

mssql_python/cursor.py
bewithgaurav, line 1227: blocker: this breaks the existing numeric precision contract. `setinputsizes([(SQL_DECIMAL, 10**100, 10**100)])` is intentionally clamped before narrowing and round-trips on `main`; this branch now raises before that clamp, and the existing regression test is red in every completed job.

apply the allocation cap only after numeric precision and scale normalization, or exempt numeric metadata from this generic size check.

tests/test_004_cursor.py
bewithgaurav, line 1770: blocker: this test never reaches the allocation guard it claims to cover, and it keeps every completed test job red. the 50,000,000-character inputs are marked DAE, so the call takes the row-by-row path and fails with `Parameter count does not match parameter metadata count`.

use `None` values, or another non-DAE payload, so the native array-binding budget is actually exercised.

bewithgaurav (changes requested):
the hardening covers most of the reported unsafe size paths, but four blockers remain across compilation, valid fetch behavior, parameter compatibility, and the new regression test. requesting changes.

PR #803, opened 2026-09-21
FIX: Validate driver-provided fetch sizes

gargsaumya:

> ADO Work Item: [AB#47843](https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47843)

### Summary

- Validate ODBC row counts, indicators, and column-buffer layout before access.
- Check streamed fetch growth and Arrow source/destination bounds.
- Correct fixed-width buffer lengths and reject malformed wide-character data.

Stacked on #802 so this review contains only driver-provided fetch validation.


mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -6145,9 +6250,8 @@ SQLRETURN FetchAll_wrap(SqlHandlePtr StatementHandle, py::list& rows,
         return ret;
     }
 
-    SQLULEN numRowsFetched;
-    SQLSetStmtAttr_ptr(hStmt, SQL_ATTR_ROW_ARRAY_SIZE, (SQLPOINTER)(intptr_t)fetchSize, 0);
-    SQLSetStmtAttr_ptr(hStmt, SQL_ATTR_ROWS_FETCHED_PTR, &numRowsFetched, 0);
+    SQLULEN numRowsFetched = 0;
+    FetchStateGuard fetchStateGuard(hStmt, &numRowsFetched, fetchSize);
```
sumitmsft, line 6254: On the non-LOB `fetchall()` error path, this guard runs statement cleanup before Python calls `check_error()`. Those ODBC calls replace the failed fetch’s diagnostics, losing its SQLSTATE and driver message. A specific `DataError` or `OperationalError` can consequently become `DatabaseError: Unknown DDBC error`.

Please capture the diagnostics before cleanup and preserve the existing DB-API exception mapping, while retaining cleanup before the buffers are destroyed.

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -4389,6 +4425,11 @@ SQLRETURN FetchBatchData(SQLHSTMT hStmt, ColumnBuffers& buffers, py::list& colum
             ret);
         return ret;
     }
+    for (SQLUSMALLINT col = 0; col < numCols; ++col) {
```
bewithgaurav, line 4428: suggestion: these checks are the only safety net when a driver returns malformed row counts, indicators, or buffer lengths, but this diff adds no tests and this stacked branch has no coverage run. the normal fetch and arrow tests pass without exercising any rejection path.

can we add a fake ODBC callback test seam for oversized row counts, odd `SQLWCHAR` lengths, out-of-capacity indicators, and `SQL_NO_TOTAL` with no progress?

mssql_python/pybind/ddbc_bindings.cpp
```diff
@@ -346,6 +346,46 @@ size_t CheckedMultiplySize(size_t left, size_t right, const char* errorMessage)
     return left * right;
 }
 
+size_t CheckedFetchAdd(size_t left, size_t right, const char* errorMessage) {
```
bewithgaurav, line 349: optional: `CheckedFetchAdd` and `CheckedFetchMultiply` only forward to the existing checked-size helpers without adding fetch-specific behavior. using `CheckedAddSize` and `CheckedMultiplySize` directly keeps one vocabulary for the same overflow contract.

bewithgaurav (changes requested):
the driver-provided size validation is well scoped and the normal fetch and arrow paths hold up. I found one suggestion worth incorporating before merge, so requesting changes.

